Issue #2, opened 2022-06-15
Think through our schema


slifty:
Similar to #1 this issue is intended as a place to discuss and explore our schema; after a little bit of iteration here we'll want to actually document the schema in the form of a PR, and then of course eventually through migrations.

slifty:
```mermaid
erDiagram
  Test {
    int foo
  }
  Another {
    string bar
  }

  Test ||--o{ Another : blorp
```

slifty:
Yes! This is spot on.

As I mentioned on Zulip I'll now take a stab at moving this from a class diagram to an ERD (and think through the fields).

bickelj:
Looking at it again, I guess OrganizationSpecificFields would have been better named OrganizationProposalFields (or the Proposal being somewhere in there). But it's neither here nor there.

slifty:
```mermaid
erDiagram
  CanonicalField {
    int id
    string label
    string type
    datetime dateCreated
  }
  Applicant {
    int id
    string externalId 
    bool hasGivenPermission
  }
  Proposal {
    int id
    int applicantId
    int opportunityId
    datetime dateCreated
  }
  Opportunity {
    int id
    string title
    datetime dateCreated
  }
  ApplicationSchema {
    int id
    int opportunityId
    int version
    datetime dateCreated
  }
  ApplicationSchemaField {
    int id
    int applicationSchemaId
    int canonicalFieldId
    int order
    string label
    datetime dateCreated
  }
  ProposalVersion {
    int id
    int proposalId
    int version
    datetime dateCreated
  }
  ProposalFieldValue {
    int id
    int proposalVersionId
    int proposalSchemaId
    int index
    string value
    datetime dateCreated
  }
  ApplicantFieldValue {
    int id
    int canonicalFieldId
    int sequence
    string value
    datetime dateCreated
  }
  ActivityLog {
    int id
    string actionType
    string log
    datetime dateCreated
  }
  
  Applicant ||--o{ Proposal : submits
  Proposal }|--|| Opportunity : "responds to"
  Opportunity ||--|{ ApplicationSchema : establishes
  ApplicationSchema ||--|{ ApplicationSchemaField : has
  ApplicationSchemaField }o--|| CanonicalField : represents
  Proposal ||--|{ ProposalVersion : has
  ProposalVersion ||--|{ ProposalFieldValue : contains
  ProposalFieldValue }o--|| ApplicationSchemaField : populates
[...]

PR #6, opened 2022-06-20
Add example sequences

slifty:
This PR adds some example data flows to help explain more about how this schema might work in the real world.

Related to #2

### DIFF
--- a/docs/ENTITY_RELATIONSHIP_DIAGRAM.md
+++ b/docs/ENTITY_RELATIONSHIP_DIAGRAM.md
@@ -114,3 +114,87 @@ PDC would then:
 * Collect the most recent ExternalFieldValues for each CanonicalField for that Applicant.
 
 It would use the ApplicationFieldValue set as the primary source, and the ExternalFieldValue set as a secondary source.
+
+## Examples
+### Registering an Application Schema
+
+```mermaid
+sequenceDiagram
+  actor Admin
+  participant API
+  participant Database
+  Admin ->>+ API : Here is a new application schema
+  API ->>+ Database : Register any new canonical fields
+  Database ->>- API : OK!
+  API ->>+ Database : Register the new application schema
+  Database ->>- API : OK!
+  API ->>- Admin :  OK!
+```
+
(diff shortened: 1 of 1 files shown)


docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -114,3 +114,87 @@ PDC would then:
...
+      "type": "boolean",
+      "canonicalField": "haveSeenMonaLisa",
+    }
+  ]
+}
+```
+
+The PDC API would then ingest that new schema document.  It would first register any `Canonical Fields` that did not already exist.  It would then register the `Application Schema` and `Application Schema Fields`, with field-level associations to the `Canonical Fields`.
+
+The system does not differentiate between "core" and "custom" fields.  Rather, there will be a set of `Canonical Fields` that are used by varying numbers of `Application Schemas`.  We will likely see that some `Canonical Fields` are used more often than others, and some are only used by a single `Application Schema`.  We might choose a subset of the `Canonical Fields` to highlight in our documentation and might call those "core" fields; that decision is not directly relevant to the schema.
```
bickelj, line 157: I would say "The database does not differentiate..." because it is possible that the APIs or the UI could potentially have a concept of "core" fields. It might also be helpful to add something like "One can think of 'core' fields as those canonical fields we would expect to see in most or all application schemas, across organizations, whereas a 'custom' field can be thought of as those fields that might only appear for one application schema or one organization." I don't know if that helps relate the idea to the DB completely effectively.

bickelj (changes requested):

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: Great call re the db not differentiating

Re the second suggestion I'd like going to avoid providing a heuristic for core fields at this stage; I'm worried it might cause folks to start down a path that isn't relevant to this convo (I don't want folks to start deliberating on what a core field is yet, since the idea is we can have that convo later when we know more)

slifty:
Thank you all!  Hitting the go button :D

PR #7, opened 2022-06-21
Set up TypeScript

slifty:
This PR sets up TypeScript along with all of the other development environment tooling.

The README goes over the new commands available in the project.

To test this try the following:

```
npm ci
npm run lint
npm run test
npm run build
```

If you want to get fancy you can try writing some invalid typescript in `src/index.ts` and see what happens when you do those steps.


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
```
bickelj, line 9: Appears to work fine in (vs)codium.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 16
+    - run: npm ci
+    - run: npm run lint
```
bickelj, line 18: Thanks for adding this too, nice!

.node-version
```diff
@@ -0,0 +1 @@
+16
```
bickelj, line 1: I have a preference for exact versions elsewhere but this is fine here, I assume, and package-lock should pin down a version, right?

bickelj (approved):
Thank you!

.node-version
slifty: Very happy to be exact here -- package-lock won't pin down a version of node.  I'll update.

slifty: Just to expand a bit more -- package-lock is only for locking down npm dependencies.

This file is used by nothing out of the box; I personally like to run `ln -s .node-version .nvmrc` in my projects since I use nvm, and then nvm will install whatever version is specified in this file.

Some Node projects commit `.nvmrc` directly but I personally think of that file as a local development dependency (not everybody uses nvm / there are other ways to manage local node versions)

`.node-version` is a clear enough file name!

Hope that helps explain better than "it won't do that" 😅

.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [ main ]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 16
```
bickelj, line 16: I was about to say that https://github.com/PhilanthropyDataCommons/pdc/runs/6990785047?check_suite_focus=true#step:3:3 appeared to use the .nodeversion but no, it is what is specified here, I guess.

.node-version
bickelj: If used by nothing, maybe it's superfluous. But I guess it makes sense to be specific. And yes, it makes sense that package-lock will only be used by npm.

slifty: Well it *is* used by humans in the sense that it documents what version of node we are targeting for development so I do think it serves a purpose (otherwise someone setting up would need to ask us what version of node to use)

.github/workflows/lint.yml
slifty: Indeed -- if we had more time I'd spend a cycle checking to see if there's a way to auto-use .node-version but FWIW I haven't seen that pattern in the past (and it actually might be worthwhile to decouple CI from dev env, since in theory we might want to use other versions in CI)

slifty:
Fixed up the node version to be more specific -- merging!

PR #10, opened 2022-06-22
Add a bare bones server component

### LINKED ISSUE #9: Add initial server component(s)
Get a simple server running, e.g. add express, a "hello world" API

bickelj:
This PR adds Express to the project and sets up a hello world endpoint

Fixes #9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,8 +8,12 @@
       "name": "@pdc/service",
       "version": "0.0.0",
       "license": "AGPL-3.0",
+      "dependencies": {
+        "express": "^4.18.1"
+      },
       "devDependencies": {
         "@tsconfig/node16": "^1.0.3",
+        "@types/express": "^4.17.13",
         "@types/jest": "^28.1.2",
         "@types/node": "^18.0.0",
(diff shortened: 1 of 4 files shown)


slifty (changes requested):
This looks good!

My only requested change is just a commit style one -- we should follow the guidelines for commit messages laid out in https://code.librehq.com/ots/meta/-/blob/main/CONTRIBUTING.md#writing-commit-messages

Those seven rules are nice!

So your commit might look something like:

```
Add bare bones server component

This "hello world" server is going to be useful for testing our
CI / CD.  We chose `express` because it is very robust and practically
ubiquitous.

Issue #9 Add initial server component(s)
```

^ Note that including the issue title in addition to the number is particularly useful in case we move away from GitHub.

The "fixes #9" would then go in the PR / PR description (which is what lets github do its magic / auto-closing etc)

bickelj:
It looks like I needed to set esModuleInterop to true to satisfy all of IDE, lint, and compiler. If that's not right, let me know.

bickelj:
> My only requested change is just a commit style one

Better?

Issue #8, opened 2022-06-22
Set up schema


slifty:
Now that #2 is resolved it's time to implement our schema.

One nice pattern I've seen used is to use [tinypg](https://www.npmjs.com/package/tinypg) as an alternative to a full blown ORM, along with very simple SQL based migrations.  I think that suits our needs for this project well; we can always get more complex later if we want to, but this approach is very explicit and still has the benefits of sanitization and structured migration.

Here's an example project that uses this pattern: https://gitlab.com/dial/principles/sustainability-toolkit-api/-/tree/development/src/database

slifty:
This is complete -- tinypg and postgres-migrations are part of the project!

bickelj:
My interpretation of the issue being open and there being no schema in the codebase was that the schema is still needed, but I agree with the direction of implementing the parts of the schema needed when it is needed, which I guess we might have discussed elsewhere and agreed to. Regardless, PR #29 is focused on the two most important endpoints `applications` and `opportunities`.

slifty:
Right!  I think that implementing the full schema was probably a poorly scoped issue (my bad; I opened it) -- but *supporting* the ability to develop the schema as we add features is a good way to go.

So really this issue should have been "set up database and schema definition patterns"

PR #11, opened 2022-06-22
[DRAFT] DB support

slifty:
This PR adds database and migration support and then utilizes that support to migrate the schema defined in #2 

I'm keeping it a draft for now because I think I'd like to split up the content here into multiple PRs that also utilize the resulting schema / add queries and API endpoints.

Related to #8


src/database/migrations/0004_create_opportunities.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE opportunities (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  title TEXT,
+	created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
```
bickelj, line 4: ?

src/database/migrations/0006_create_application_schema_fields.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE application_schema_fields (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  application_schema_id INTEGER,
+  canonical_field_id INTEGER,
```
bickelj, line 4: Do we want a unique constraint across `application_schema_id` and `canonical_field_id` as well?

src/database/migrations/0002_create_applicants.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE applicants (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  external_id TEXT,
```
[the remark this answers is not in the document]
slifty: Oh snap yes -- not null is needed in tons of places.  I'll go through and add that.

src/database/migrations/0001_create_activity_logs.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE activity_logs (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  activity_type TEXT,
```
bickelj, line 3: Most or all of the TEXT fields could be VARCHAR. In postgres, they should work the same (assuming no length limit is specified). The latter should port better if ever needed.

src/database/migrations/0007_create_applications.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE applications (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  applicant_id INTEGER,
+  opportunity_id INTEGER,
```
bickelj, line 4: I wonder if this should rather be the `application_schema_id`

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -6,89 +6,94 @@ erDiagram
...
+    string website
+    timestamp created_at
   }
   Application {
     int id
-    int applicantId
-    int opportunityId
-    datetime dateCreated
+    int applicant_id
+    int opportunity_id
```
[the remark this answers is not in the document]
bickelj: Then again, I suppose we don't have a natural (external to the db) identity for each application schema, so that would require the caller to get that surrogate (internal to the db) identity first. Hmm.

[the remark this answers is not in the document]
bickelj: I still think something is not quite right. It looks like you could mix fields from multiple application schemas within a single application version. To resolve, either application needs to be for a certain application schema or application version needs to be for a certain application schema.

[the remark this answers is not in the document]
slifty: I figured this would be enforced easily by business logic rather than at the database level (the restriction you describe is a business logic restriction, not necessarily a function of normalization)

Which is to say: we can add it, but adding it does not prevent the normalization risk.  In EITHER CASE the business logic will need to ensure that the proper application schema is used for all fields of a given version.

The application should not be linked to a schema though -- it is a response to an opportunity.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -6,89 +6,94 @@ erDiagram
...
-  ApplicationSchemaField }o--|| CanonicalField : represents
+  Opportunity ||--|{ ApplicationSchema: establishes
+  Application ||--o{ ApplicationOutcome : "has"
+  ApplicationSchema||--|{ ApplicationSchemaField: has
+  ApplicationSchemaField}o--|| canonical_field : represents
   Application ||--|{ ApplicationVersion : has
   ApplicationVersion ||--|{ ApplicationFieldValue : contains
-  ApplicationFieldValue }o--|| ApplicationSchemaField : populates
+  ApplicationFieldValue }o--|| ApplicationSchemaField: populates
   Applicant ||--o{ ExternalFieldValue : "is described by"
```
bickelj, line 95: I think I see the idea of having applicants described by the same kinds of fields that are in an application (or application schema), that is canonical fields. I'm not sure of the advantage of requiring them to be external fields, though. For every external field, there is a canonical field it refers to, so it might be as accurate to say `Applicant ||--o{ CanonicalField : "is described by"`. But neither of the descriptions of the relationship is explicit in the schema, it appears to be indirect via an application and/or application schema.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -6,89 +6,94 @@ erDiagram
...
   ApplicationFieldValue {
     int id
-    int applicationVersionId
-    int applicationSchemaFieldId
-    int index
-    string value
-    datetime dateCreated
+    int application_version_id
+    int application_schema_field_id
+    int position
```
bickelj, line 75: I am pretty sure this is redundant with the order/position in the application schema so I am omitting it in another draft.

bickelj: Oh, nice, that makes sense. Thanks!

bickelj: Then again, could it not be implicit for the N rows created, just create the rows in order?

slifty: Agreed on all fronts -- I do prefer explicit to implicit as a rule of thumb and think it's probably maybe hopefully worth the pain and suffering here...

slifty:
I think we can finally close this sucker!

PR #12, opened 2022-06-22
Add Database Support

slifty:
This PR adds the tooling needed for database interaction and migrations.

It also sneaks in a fix for a small mistake that I accidentally introduced with the current CI files, where they use yarn (I can split that into a separate PR if so desired).

Issue #8


.github/workflows/migrate.yml
```diff
@@ -0,0 +1,38 @@
+name: Migrations
```
slifty, line 1: It looks like this workflow isn't running properly just yet -- I'll fix it!

bickelj (approved):
It works for me:
```
$ npm run migrate

> @pdc/service@0.0.0 migrate
> ts-node -r dotenv/config src/scripts/migrate.ts

Starting migrations...
Migrations complete.
```
and
```
psql (14.4)
Type "help" for help.

pdc=> \d+
                                      List of relations
 Schema |    Name    | Type  | Owner | Persistence | Access method |    Size    | Description 
--------+------------+-------+-------+-------------+---------------+------------+-------------
 public | migrations | table | pdc   | permanent   | heap          | 8192 bytes | 
 public | test       | table | pdc   | permanent   | heap          | 8192 bytes | 
(2 rows)

pdc=> select * from test;
 id | something 
----+-----------
  1 |         2
(1 row)

pdc=> select * from migrations;
 id |          name           |                   hash                   |        executed_at        
----+-------------------------+------------------------------------------+---------------------------
  0 | create-migrations-table | e18db593bcde2aca2a408c4d1100f6abba2195df | 2022-06-23 16:42:09.99462
(1 row)

pdc=> \q
```

bickelj:
I would prefer to have one set of tests that works with the usual npm/software dependencies and a separate set of tests that require external dependencies such as a running postgres instance. Something like `npm run test` for the former and `npm run integration-tests` for the latter. Tests requiring web servers that can run under node perhaps could go in either or could be separate still.

bickelj:
I reproduced the error from https://github.com/PhilanthropyDataCommons/service/runs/7012914109?check_suite_focus=true#step:6:15 locally:
```
$ PGHOST=postgres PGPORT=5432 PGPASSWORD=postgres npm run migrate

> @pdc/service@0.0.0 migrate
> ts-node -r dotenv/config src/scripts/migrate.ts

Starting migrations...
Migrations failed!
Error: getaddrinfo EAI_AGAIN postgres
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:71:26) {
  errno: -3001,
  code: 'EAI_AGAIN',
  syscall: 'getaddrinfo',
  hostname: 'postgres'
}
```

In other words, hostname lookup for 'postgres' failed. The within-docker-network name is 'postgres' (see https://github.com/PhilanthropyDataCommons/service/runs/7012914109?check_suite_focus=true#step:2:86) but I guess the `npm run migrate` (see https://github.com/PhilanthropyDataCommons/service/runs/7012914109?check_suite_focus=true#step:6:2) happens outside that docker network. So maybe `localhost` could be used, along with the container having ports mapped to host ports, e.g. `-p 5432:5432` on the postgres run, or maybe the `npm run migrate` could happen in a container next to the `postgres` container on the same docker network.

slifty:
Thank you for digging in!  I'll fix up this PR (based on the work you did in the other PR) and then ping you for a re-review.

One note -- my intention was that the "migrate" command was *JUST* testing migrations.  As in, it would run the migrations to make sure they work / don't contain errors as part of the CI.

Fully agree that our unit tests should not rely on a working database, and that we could consider some form of separate integration test for that (I am fine with just unit tests for now though)

slifty:
Merge conflicts in package-lock are a special form of rebase 😂 

OK this looks like it's working!  cc @bickelj -- This is really just your other branch with the two patches squashed into the previously-broken migration commit, so hopefully it doesn't demand a deep review.

README.md
```diff
@@ -10,6 +12,19 @@
...
+
+  ```bash
+  cp .env.example .env
+  edit .env
+  ```
+
+3. Run migrations
+
+  ```bash
+  npm run migrations
```
bickelj, line 25: `npm run migrate`?

bickelj:
Works, LGTM.

Issue #3, opened 2022-06-15
Set up CI / CD


slifty:
We need to set up deployment for this repo (and CI / tests / etc.)

Part of doing this will be deciding on Docker vs server deployment, etc.

bickelj:
With a combination of PR #10 and PR #13 and a .conf file change, with the exact image id used, the proxy server reaches the express server just fine.
```
$ docker compose -f compose.yml up -d
[+] Running 4/4
 ⠿ Network deploy_default            Created                                                                                                                                               0.1s
 ⠿ Container deploy-reverse-proxy-1  Started                                                                                                                                               1.2s
 ⠿ Container deploy-web-1            Started                                                                                                                                               1.4s
 ⠿ Container deploy-database-1       Started
$ docker ps
CONTAINER ID   IMAGE                                    COMMAND                  CREATED         STATUS        PORTS                                                                            NAMES
d709d7461497   bitnami/postgresql:14.4.0-debian-11-r1   "/opt/bitnami/script…"   3 seconds ago   Up 1 second   5432/tcp                                                                         deploy-database-1
1a7c073c402c   07254019df1f                             "npm start"              3 seconds ago   Up 1 second   3000/tcp                                                                         deploy-web-1
dc888a8fb244   bitnami/nginx:1.22.0-debian-11-r8        "/opt/bitnami/script…"   3 seconds ago   Up 1 second   0.0.0.0:80->8080/tcp, :::80->8080/tcp, 0.0.0.0:443->8443/tcp, :::443->8443/tcp   deploy-reverse-proxy-1
$ docker logs deploy-reverse-proxy-1
nginx 21:27:15.35 
nginx 21:27:15.35 Welcome to the Bitnami nginx container
nginx 21:27:15.35 Subscribe to project updates by watching https://github.com/bitnami/bitnami-docker-nginx
nginx 21:27:15.35 Submit issues and feature requests at https://github.com/bitnami/bitnami-docker-nginx/issues
nginx 21:27:15.35 
nginx 21:27:15.36 INFO  ==> ** Starting NGINX setup **
nginx 21:27:15.37 INFO  ==> Validating settings in NGINX_* env vars
nginx 21:27:15.38 INFO  ==> No custom scripts in /docker-entrypoint-initdb.d
nginx 21:27:15.38 INFO  ==> Initializing NGINX
realpath: /bitnami/nginx/conf/vhosts: No such file or directory

[...]

PR #13, opened 2022-06-23
Add a Dockerfile

bickelj:
This is a draft of a Dockerfile for the back-end server component.
The goal is for the CI server to build a deployable image using the
already-compiled software components and libraries.

Issue #3 Set up CI / CD


bickelj:
At the moment, it looks like all the node modules will go into the image, assuming `test` was recently run. It might be nice to trim it down to just the prod dependencies.

Issue #15, opened 2022-06-27
Add swagger/OpenAPI spec for API


bickelj:
Draft a specification for the API in the OpenAPI format.
The goal is to be able to generate express routes from this spec.

bickelj:
d5e9b52 has some progress on an OpenAPI spec.

bickelj:
Perhaps the activity log can contain those changes to external field values.

bickelj:
cb21c1e includes calls for external field as well as outcomes. It also tries to clarify the three patterns present for sending data:
1. Put: idempotent immutable add
2. Put: idempotent add
3. Post: log

The `applications` portion is the most uncertain or needing the most work. Currently I have it as accepting an `application_schema_id` but considering the comment by @slifty at https://github.com/PhilanthropyDataCommons/service/pull/11/files#r907891192 and also thinking about the different ways an application may wish to be accessed or modified, it is probably not the best. But `opportunity_title` hardly identifies it, posting the `version_id` seems weird, etc. It is OK to have a dozen ways to get lists of applications, but the question is more about the correct/canonical way to add a single application.

bickelj:
The updated draft at 365fbd6 has the pre-filing an application call and makes some of the fields consistently named.

None of the drafts so far have any JSON Schema types beyond `array` and `object`. I assume there are tools to generate either the JSON Schema from TypeScript or vice versa and that we do not have to hand-synchronize these between the API specification and the software.

bickelj:
The updated draft at a9f4364 uses parameter descriptions.

PR #19, opened 2022-06-29
Add an OpenAPI specification

bickelj:
These are also known as Swagger specs.

Issue #15


src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+  "info": {
+    "title": "Philanthropy Data Commons API",
+    "description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
+    "version": "0.0.3"
+  },
+  "components": {
+    "parameters": {
+      "canonical_field_label": {
+        "name": "canonical_field_label",
+        "description": "The unique name of a canonical PDC field.",
```
slifty, line 12: I'm wondering if the ERD overloaded this field.

Label elsewhere is meant to be the human readable "label" (I tend to agree it should be unique but I also wonder if it must) -- we might also want a canonical field *code* which is intended to be more meaningful than an int ID but also used in semantic contexts (e.g. when defining a mapping between an application schema field and a canonical field).

Thoughts?

(genuine side note that you can ignore: this kind of question might expose a flaw in our plan of "define the whole api spec, and all migrations, at once and then implement the endpoint one by one" since it asks us to think about the implementation for ALL endpoints at once and potentially answer a whole bunch of these kinds of nitty gritty questions outside the context of that implementation.  Whoops!)

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+        "description": "The unique name of an opportunity, RFC, potential award or grant, etc.",
+        "in": "path",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      },
+      "applicant_external_id": {
+        "name": "applicant_external_id",
+        "description": "The unique name of an applicant, e.g. email address.",
```
slifty, line 30: External ID here would be more like Tax ID Number (we have to talk to @jim-mcgowan to figure out how we handle organizations that have multiple "branches" under the same TIN).

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+              "application/json": {
+                "schema": {
+                  "type": "object"
+                }
+              }
+            }
+          }
+        }
+      },
+      "put": {
```
slifty, line 190: Same as canonical fields -- `POST` is the REST verb for create (and we don't want `PUT` to be implemented since we aren't allowing updates -- so it would 404 appropriately.)

slifty (changes requested):
Heya -- I didn't go through every line because I have some larger feedback first.  The TL;DR is:

1. We should use `camelCase` for everything (endpoints and parameter documentation), rather than `snake_case`

2. We should use `POST` for create (and not implement `PUT` except in cases where we want to support updating).

3. We should use primary keys as the URL parameter for specifying a single entity in a REST call, and we should use querystring for looking up an entity in terms of a non-primary key (regardless of whether or not it is unique).

4.

bickelj:
In our conversation today, I think we came to needing to focus on the Application portion of the API, so my next change may cut down on the lower-level API calls and start with the higher-level Application calls.

bickelj:
bb6472c has a simplified API focused on applications and opportunities, still at a relatively high level (no data exchange structures yet). @slifty is this more along the pattern you'd expect?

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+              "application/json": {
+                "schema": {
+                  "type": "array"
+                }
+              }
+            }
+          }
+        }
+      },
+      "put": {
```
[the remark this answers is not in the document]
bickelj: POST and PUT both create resources but PUT is idempotent.

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+                "schema": {
+                  "type": "object"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/applicants/{applicant_external_id}": {
```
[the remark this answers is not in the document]
bickelj: The question was which thing is a primary key: an arbitrary number generated by an implementation detail (the database) or is the primary key a natural key such as an email address, ssn, etc. I simplified and started with just opportunities and applications in bb6472c below.

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+                "schema": {
+                  "type": "object"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/applications/{application_schema_id}": {
```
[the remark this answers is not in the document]
bickelj: Agreed and done in bb6472c.

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+                "schema": {
+                  "type": "array"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/applications/{application_schema_id}/{applicant_external_id}": {
```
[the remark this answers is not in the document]
bickelj: Agreed and done in bb6472c.

src/openapi.json
bickelj: I like your side note, basically we can defer or delay this question.

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+        "description": "The unique name of a canonical PDC field.",
+        "in": "path",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      },
+      "opportunity_title": {
+        "name": "opportunity_title",
+        "description": "The unique name of an opportunity, RFC, potential award or grant, etc.",
```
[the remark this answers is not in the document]
bickelj: Yes, this helps clarify that we will only use auto-generated IDs.

src/openapi.json
```diff
@@ -0,0 +1,566 @@
...
+                "schema": {
+                  "type": "array"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/canonical_fields/{canonical_field_label}": {
```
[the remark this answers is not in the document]
bickelj: The question was whether the label was the natural key and I think we are going in the auto-generated synthetic key direction.

src/openapi.json
```diff
@@ -0,0 +1,191 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Philanthropy Data Commons API",
+    "description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
+    "version": "0.0.4"
+  },
+  "components": {
+    "parameters": {
```
slifty, line 9: Looking at the OpenAPI spec ma few thoughts:

1. The example given in the OpenAPI documentation actually explicitly flags components/parameters as being *operation* parameters -- e.g. for things that truly are "meta" parameters that are used to provide decorative functionality across multiple endpoints (e.g. `limit` or `offset`).

2. That said, I appreciate there is DRY value in specifying things that are simply reused, and that was the original intent.

So that brings me to a question / suggestion!

I think having `opportunityId` defined in three ways is not ideal (e.g. opportunityId, opportunityIdQueryOptional, opportunityIdQueryRequired); it looks like MAYBE you could define it without specifying required in the `components` and then in the actual API you could say:

```
        "parameters": [
          {
            "$ref": "#/components/parameters/opportunityId",
            "required": false,
          }
        ],
```

I haven't tested this but if it works that seems way nicer (and less repetitive)

bickelj: This is a great suggestion and makes perfect sense to me, I am trying it out and my IDE plugin is complaining when I remove either `in` or `path` but let me see if the swagger.io website validator complains.

bickelj: Darn. Both the "OpenAPI (Swagger) Editor" codium extension and also https://editor.swagger.io/ say "no dice" to removing either of `in` or `path`.

bickelj: Just to make sure that's not a bug in the two validators, I see those fields marked "required" in the spec too, at https://swagger.io/specification/#parameterObject

At the moment, some of these are only used once, so it does not make sense to define them up top until there are two uses. I guess I anticipate multiple uses of these but that could be deferred to later. Or perhaps it is simpler/clearer to always define parameters up top in the common parameters area rather than some defined here while others are defined there.

bickelj: It is OK to omit the `required` property when the parameter location is `query`, but then it causes the value to be interpreted `false` even when an override to `true` is attempted in the context of the specific path.

slifty (dismissed):
I think I posted a comment outside of this review but nevertheless:

1. Let's use `components/schema` for our return type schemas, (but let's NOT try to define those schemas in full just yeat)

2. Let's not use `components/parameters` for parameters that aren't meta/operational, I think it ends up being more verbose than just defining the parameters inline

3. If you agree, I'd suggest NOT documenting the standard HTTP headers in responses (though we should totally include them in the implementation).

src/openapi.json
slifty: So I think our best bet may be to just define these things inline rather than in a reusable place -- even when a parameter is "reused" it will generally not be *identical* (just similar in several ways) -- e.g. in one case the opportunityId would be what is being accessed, in another it would be the opportunityId for what is being deleted, in another it would be the opportunityId for what is being updated, etc.

Would that solve the problem?

I do think this piece of the swagger spec is intended for operational parameters that are truly reused (e.g. parameters like `orderBy` or `limit` or `offset` or `auth`)

bickelj: In the narrowest sense, I think there will probably be re-use of some optional query parameters, although I agree it's not there now so not needed now. In the general sense, the parameters regarding opportunityId (whether required or not, whether in this spot or that spot) really are all referring to the same opportunityId. An opportunityId by any other name would smell as sweet. So the thing referred to, the opportunityId or the opportunity itself, is truly identical, but the operation attempted may require it here or there or not at all. Those aren't properties of opportunities but properties of the API spec.

But yes, for the moment, I think they should be inline, because they are not really re-used yet.

src/openapi.json
```diff
@@ -0,0 +1,191 @@
...
+        "parameters": [
+          { "$ref": "#/components/parameters/opportunityIdQueryMandatory" }
+        ],
+        "responses": {
+          "201": {
+            "description": "The new application that was created.",
+            "headers": {
+              "Location": {
+                "description": "The URL including the applicationId for the new application.",
+                "schema": {
```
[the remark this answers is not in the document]
bickelj: For now, I'm ditching it, but we can re-add later if it proves helpful.

src/openapi.json
```diff
@@ -0,0 +1,177 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Philanthropy Data Commons API",
+    "description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
+    "version": "0.0.5"
```
slifty, line 6: Since the other iterations weren't merged idk if they count as taking up a version number -- fine merging this as is but if you wanna make this 0.0.1 that'd be reasonable IMHO (and in future / once this is "released" I don't think we should jump multiple versions just from branch iteration)

slifty (approved):
This looks *GREAT* thank you for all the time and iteration.

One small note about the version which you are welcome to take or ignore (no review needed if you do decide to make the change)

src/openapi.json
bickelj: So versions 0.0.1 and 0.0.2 were force-pushed out of the git history, but versions 0.0.3 and 0.0.4 are still there in the branch (and therefore will be in the eventual history). There really were five versions published (if you count this PR as publication) which was why I erred on the safe side.

Is this something that should be squashed, and then if so maybe we could call it 0.1.0?

Issue #17, opened 2022-06-29
Add a couple of data types/schemas


bickelj:
We need some types or schemas or schemas generated from types. Ideally we can integrate these with the swagger/openapi spec, whether manually or as a build step.

bickelj:
Overcome by events. We will be adding these as needed. The pattern to follows was set up in issue #30 and pull request #55.

PR #21, opened 2022-06-30
Add sample TypeScript types and schema generator

bickelj:
One can run the generator to produce a types.json:

    npm run generate-json-schema

Issue #17 Add a couple of data types/schemas

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,11 @@ To use a development docker image from GitHub Container Registry:
 docker pull ghcr.io/philanthropydatacommons/service:latest
 ```
 
+To generate a json schema from the included types:
+```bash
+npm run generate-json-schema
+```
+
 ### EditorConfig
 
 We use [EditorConfig](https://editorconfig.org/) to help developers maintain proper whitespace habits in the project.  Most IDEs have [an official EditorConfig plugin](https://editorconfig.org/#download) you can install.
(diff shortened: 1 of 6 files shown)


bickelj:
The above types in 7c748e9 are exploratory and so is the JSON schema generator. It would be nice to integrate these with the swagger/OpenAPI definitions somehow. I don't know if we would want full types with all optional fields present or read-only types (with full data) and write types (with just enough to write a new record). There also might be a completely different way of going about it. This is an exploration of one avenue.

slifty:
Oh this is interesting -- I haven't seen it before.

I see how this would be immediately useful for populating the openapi spec, and that's cool to think about having automatically-updating documentation, though I think as you start to explore there might be downsides to automating that since we might lose the ability to be explicit in our documentation since the types might not always be 1:1 to what our API spec is intended to use.

Are there other things you are thinking the json schema will be used for outside of that?

bickelj:
@slifty, sure we can pause. I am not entirely stuck on the idea of generating one to the other, but it seems like we ought to if it can work. I am also OK with deferring the definition of data exchange schemas, deferring definition of TypeScript types, etc., to another time or also to another PR/issue.

slifty:
Yeah that sounds good; we should be defining types as we implement endpoints anyway (that way we avoid the thrash of accidentally over-implementing against imaginary use cases)

bickelj:
I think AJV promises much regarding JSON schema and can probably be integrated with the OpenAPI somehow. This experiment or proposal served its purpose.

PR #22, opened 2022-07-01
Notify a workflow in deploy project on image push

bickelj:
Use curl to call a manual-run workflow in the 'deploy' project to tell
it the image version (docker image tag) that was just pushed to GHCR.
The workflow in 'deploy' in turn updates the deployment scripts.

Issue #3 Set up CI / CD


bickelj:
Though the checks passed and an image was pushed, the curl step did not have the intended outcome:

> ```
> {
>  "message": "Resource not accessible by integration",
>  "documentation_url": "https://docs.github.com/rest/reference/actions#create-a-workflow-dispatch-event"
> }
> ```

bickelj:
It sounds like I need a PAT to do this and one with pretty wide permissions.

bickelj:
Or maybe it can be done with the regular generated token and setting up permissions in the workflow.

.github/workflows/build.yml
```diff
@@ -43,3 +43,8 @@ jobs:
           tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.VERSION }}, ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
           cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
           cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest,mode=max
+      # Notify the top-level deploy project that an image was pushed.
+      - run: "curl -X POST -H \"Accept: application/vnd.github+json\" \
+             -H \"Authorization: token ${{ secrets.GITHUB_TOKEN }}\" \
+             https://api.github.com/repos/PhilanthropyDataCommons/deploy/actions/workflows/update-service-image.yml/dispatches \
```
slifty, line 49: I hear your concern in the PR that this is somewhat dog wagging but I also appreciate that it may simply be the best we can do when using github actions for CD.

That said, rather than running a curl directly what do you think of using an existing github marketplace action.

This one seems to be the most starred / popular: https://github.com/marketplace/actions/repository-dispatch

.github/workflows/build.yml
```diff
@@ -8,7 +8,7 @@ on:
   push:
     branches:
       - 'main'
-      - '3-build-docker-image-on-push'
+      - '3-build-docker-image-on-push-2'
```
slifty, line 11: This line should be removed before merging (calling it out just to make sure it isn't lost in the scuffle)

slifty (changes requested):
I think this approach is fine for now -- only request is that we use a gihub marketplace action in the spirit of cleaner syntax and not having to keep up with the ins and outs of github dispatches in future.

.github/workflows/build.yml
bickelj: Fair enough. It means that exact commit (the one following removal) cannot really be tested but pushed blind, but there is a good enough chance of success when it gets merged.

.github/workflows/build.yml
bickelj: I haven't gotten this curl part to work just yet, and it seems ugly, so I will try the repository-dispatch approach, thanks!

bickelj: I also was hoping to avoid creating a PAT but that appears inevitable either way.

bickelj:
I see that this action appeared to run successfully but the deploy project action did not run. Perhaps a webhook is missing.

bickelj:
A webhook seems incorrect. Perhaps it is this part needed in the other repository action:
> repository_dispatch

bickelj:
Notification succeeded, see https://github.com/PhilanthropyDataCommons/deploy/commit/6a086256f9d56c850519126b9b00c1a60b8e05ac, however the payload is not right yet.

bickelj:
a2bfca2 included less in the json, we'll see if that works.

bickelj:
71e5f91 sets ref to main.

bickelj:
The workflow in the deploy project worked: https://github.com/PhilanthropyDataCommons/deploy/commit/a7219aea874c713a82a03b9baad37825ce830939

The workflow here may need the token to grant read access to workflows in order for the "wait-for-completion" feature to work.

bickelj:
It appears to have worked fine with current perms. We can expect this to break in 30 days, at which time we will need a new personal access token.

slifty:
Oh yikes I didn't realize that this approach would bake in a manual 30 day tech debt cycle! (I didn't fully grok what a PAC was).  I also don't think it's sustainable to have infrastructure depending on a personal token.

all that is to say, I'm not sure this approach can work long term (I don't understand how ANYBODY could use repo dispatches given all of this...); to the point that if this is required we might need to re-think our CD approach (I'm a few cycles behind on knowing what's going on in the `deploy` repo).

I don't want to upset a bunch of progress here... Do you think we should merge for the short term and plan to change it later?  Maybe we should talk through it by voice.

I'll do some digging to see if maybe there's some harder-to-find way to use github action dispatches without a PAC.

bickelj:
> Oh yikes I didn't realize that this approach would bake in a manual 30 day tech debt cycle! (I didn't fully grok what a PAC was). I also don't think it's sustainable to have infrastructure depending on a personal token.

The alternative is to manually run the workflow at https://github.com/PhilanthropyDataCommons/deploy/actions/workflows/update-service-image.yml with an image name or manually update the compose.yml with the version. So I don't think it is not a huge loss when the PAT expires. Also, it is possible to generate an expiration-less PAT. You can also replace it with one of your PATs instead of mine because it is a repository secret named `REPO_ACCESS_TOKEN`.

> all that is to say, I'm not sure this approach can work long term (I don't understand how ANYBODY could use repo dispatches given all of this...); to the point that if this is required we might need to re-think our CD approach (I'm a few cycles behind on knowing what's going on in the `deploy` repo).

> I don't want to upset a bunch of progress here... Do you think we should merge for the short term and plan to change it later? Maybe we should talk through it by voice.

This part is nice to have, so I think it's worth merging. We can update the `REPO_ACCESS_TOKEN` variable now if we want to or find an alternative now too.

> I'll do some digging to see if maybe there's some harder-to-find way to use github action dispatches without a PAC.

:+1:

bickelj:
I see this nugget in the packages area:
> Pick the repositories that can access this package using GitHub Actions.

That gives me hope that the higher-level deploy project could see the package and hopefully see events such as new package available, etc. I could not find a clear example of this in the docs but it is probably doable with some more digging.

slifty (dismissed):
Just to close my part of the loop on this -- we had a long convo about this by voice to just talk about the limitations of workflow dispatches on github (they basically seem to require a PAC)

Where we left was that this approach is the best bang for the buck right now in terms of time.  It's possible there is a more ideal solution that doesn't use personal tokens, but we simply don't have the time / resources to find them right now.

Given that, this is just about ready for merge!  The only two changes I'd ask before hitting full on approve is:

1. Remove the ad-hoc branch trigger (I know that means there is a minor risk but I think that's needed to not clutter up main branch with short term code)
2. Add in a note so someone in the future will know how to generate a proper PAC.

Thank you for all the hard work on this and the rest of CD @bickelj -- it really is cool and great to see a system that is both explicit AND automated!

bickelj:
@slifty  556b99c should close my end as well. Thank you for the careful review and for flagging the awkward Personal Access Token use. There really must be a better way. I created #23 so that we don't completely forget about it and also added a README.md to the GitHub actions directory here. I also removed this branch from the action, so it is now untested code, ready for merge to main :stuck_out_tongue_winking_eye: (I'm pretty sure it will be fine, and agree it will be cleaner).

PR #24, opened 2022-07-06
Update base node image

bickelj:
This is a small change to make a new service image.

Issue #3 Set up CI / CD


bickelj:
I'm assuming simple stuff like this needs not much review besides the automatic checks and local dev checks.

PR #25, opened 2022-07-07
Create a route pattern

### LINKED ISSUE #16: Add a couple of routes
Add some express routes

slifty:
This PR creates a model for defining new routes which includes tests, handlers, and route definitions.  (It also sneaks in a fix for test coverage reports that weren't being generated any longer)

The defined route is functionally just a `hello world` -- the return value is invented, so that part isn't tested.

That said, to test this you can run:

```
npm start
```

and if you navigate to http://localhost:3000/canonicalFields you should see:

image

Resolves #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,15 +58,14 @@ docker build .
 To use a development docker image from GitHub Container Registry:
 ```bash
 docker pull ghcr.io/philanthropydatacommons/service:latest
+```
 
 To start the server:
-
 ```bash
 npm start
 ```
 
 To start the server in a development environment:
-
 ```bash
 npm start:dev
 ```


slifty:
@bickelj fixed I believe!

(for future reference, the steps were:

1. interactive rebase; flag that commit for edit
2. Run `git commit --amend --author="Dan Schultz <slifty@gmail.com>" --no-edit`

Now it's showing me as the author in my history and in the patch view: https://github.com/PhilanthropyDataCommons/service/pull/25/commits/063e4cc6eec5e8f256f0b35148650cbe29da8242.patch

package.json
```diff
@@ -4,13 +4,15 @@
...
   "main": "dist/index.js",
   "scripts": {
-    "build": "npm run cleanup && npm run build:node",
+    "build": "npm run cleanup && npm run build:node && npm run build:assets",
     "build:node": "tsc -p tsconfig.json",
+    "build:assets": "copyfiles -u 1 \"src/**/*.sql\" \"src/**/.keep\" dist",
     "cleanup": "rimraf dist",
     "lint": "eslint ./src --ext .ts",
-    "migrate": "ts-node -r dotenv/config src/scripts/migrate.ts",
+    "migrate": "npm run build && node -r dotenv/config dist/scripts/migrate.js",
```
bickelj, line 12: As annoying as it is to try to remember to build first and migrate second, I think having the `npm run build` here still means the `migrate` task depends on dev dependencies. So I could not run this `migrate` after `npm prune --omit=dev` because `tsc` wouldn't be available.

package.json
```diff
@@ -4,13 +4,15 @@
...
     "test": "jest",
-    "start": "node ."
+    "start": "npm run build && node ./dist/index.js",
```
bickelj, line 14: Same as above. But neither of these prevent merge because we are no worse off than before, we're still better off.

src/__tests__/canonicalFields.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+describe('/canonicalFields', () => {
+  describe('/', () => {
+    it('should return HTTP Status Code 200 OK', async () => {
+      await agent
+        .get('/canonicalFields')
+        .expect(200);
+    });
+  });
+});
```
bickelj, line 14: This is very nice!

src/app.ts
```diff
@@ -0,0 +1,9 @@
+import express from 'express';
+import { rootRouter } from './routers';
+
+const app = express();
+app.use(express.urlencoded({ extended: true }));
+app.use(express.json());
+app.use('/', rootRouter);
+
+export { app };
```
bickelj, line 9: Simple enough, clear enough, I see it now. Thanks!

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -0,0 +1,14 @@
+import type {
+  Request,
+  Response,
+} from 'express';
```
bickelj, line 4: I guess this style allows us to see more easily in history when a new type comes in or is removed. I am surprised that the trailing comma is OK, but that will help with the same, right?

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+function getCanonicalFields(req: Request, res: Response): void {
+  res.status(200)
+    .contentType('application/json')
+    .send('{ "hello": "world"}');
+}
+
+export const canonicalFieldsHandlers = {
+  getCanonicalFields,
+};
```
bickelj, line 14: Also simple enough, clear enough, thanks!

package.json
slifty: Oh snap of course, silly me -- thanks.

I'll patch that up.

I'm gonna have to figure out if it's more common to have a `migrate:dev` or a `migrate:ci`, my initial instinct is it would be better to have migrate run using ts-node and migrate:ci run via node, that way nobody has to remember to build first.  Will spend a quick cycle thinking and throw out a proposal.

slifty: Landed on migrate:dev for those who wanna skip having to run build first.

`migrate` and `start` are both assuming no dev dependencies.

Issue #39, opened 2022-07-12
`no matching manifest for linux/amd64` on `docker pull ...:latest` 


bickelj:
The first attempt to do a docker pull of our image from GHCR resulted in this:
> $ docker pull ghcr.io/philanthropydatacommons/service:latest
> latest: Pulling from philanthropydatacommons/service
> no matching manifest for linux/amd64 in the manifest list entries

Perhaps there is something missing from the Dockerfile, etc.

bickelj:
It may be an issue with attempting to use `latest` as a build cache and also an image because the issue only occurs when pulling latest. Another tag seems to work OK.

bickelj:
That change seems to have done it:

```
$ docker pull --platform linux/amd64 ghcr.io/philanthropydatacommons/service:latest
latest: Pulling from philanthropydatacommons/service
edcd5f1da6cc: Already exists 
0d83af23440a: Already exists 
4677471f91d8: Already exists 
02eda02759b7: Already exists 
af48e8de5664: Already exists 
607df786a4cb: Already exists 
85dd1a03fac5: Already exists 
c6c4f9fbf9ca: Already exists 
7f2f8ee56627: Already exists 
1733068eda3c: Pull complete 
883eee60a4b5: Pull complete 
4f4fb700ef54: Pull complete 
73388c258f27: Pull complete 
97f26975e0e8: Pull complete 
34f4f6598bbd: Pull complete 
Digest: sha256:fc4aeb51446d83bfe933ab9a617c3ee0552b54b79a8bec541ce7ce7ccc43d2d9
Status: Downloaded newer image for ghcr.io/philanthropydatacommons/service:latest
ghcr.io/philanthropydatacommons/service:latest
```

But we are also left with an annoying `buildcache` tag as the top tag under https://github.com/PhilanthropyDataCommons/service/pkgs/container/service and I don't know if there's a good way to hide that.

bickelj:
OK, weird, now the buildcache tag is lower on the list, so perhaps this fixed that too. But we also have a hopefully temporary error in the last build at https://github.com/PhilanthropyDataCommons/service/runs/7308910563?check_suite_focus=true:
```
#16 pushing manifest for ghcr.io/philanthropydatacommons/service:20220712-a0b1c52@sha256:0d84fc54b3e334f55d4b46208588a72ea30c64bd330fea6d7c7b83fac011743c 0.0s done
#16 ERROR: unexpected status: 499 status code 499
------
[...]

PR #40, opened 2022-07-12
Set platform on docker build to linux/amd64

bickelj:


bickelj:
See #39 it was not just setting the platform but perhaps something to do with tags.

.github/workflows/build.yml
```diff
@@ -38,10 +38,11 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
+          platforms: linux/amd64
           push: true
           tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.VERSION }}, ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
-          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
-          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest,mode=max
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache,mode=max
```
bickelj, line 45: The `cache-from` and `cache-to` which we now use with label/tag `buildcache` as I understand it just makes it faster to do a build instead of having to pull the base image every stinking time. In other words, I think GitHub's Container Registry will have our previous image layers locally, and whenever the first few layers are identical (majority case), it will be much faster to re-use them.

bickelj:
I'm not sure why the buildcache has to be a tag that shows up in our service image list, honestly. I was hoping it would be limited to the cache functionality. That happened on my first attempts too, so I was like 'why not re-purpose latest to be the build cache image?' but that changes the nature of the image, I guess, such that you cannot pull `latest` when you use `latest` as the build cache.

bickelj:
`docker pull ghcr.io/philanthropydatacommons/service:buildcache` fails (was not available before this change)
`docker pull ghcr.io/philanthropydatacommons/service:latest` succeeds (failed before this change)
`docker pull ghcr.io/philanthropydatacommons/service:20220712-ee0be60` succeeds (succeeded both before and after)

slifty (approved):
If you could update the commit message to explain what `buildcache` is / where the tag comes from / why we're choosing to use it then this is good to go!

bickelj:
> If you could update the commit message to explain what buildcache is / where the tag comes from / why we're choosing to use it then this is good to go!

Done.

PR #44, opened 2022-07-12
Update to node 16.16.0 and use it consistently

bickelj:
This is a minor change to test that our workflows are still OK after
changes in issue #39 (no matching manifest for linux/amd64 on docker
pull ...:latest)


bickelj:
It looks like I missed `.node-version`

slifty:
Argh sorry for missing this review -- I probably would have missed that anyway 😂

bickelj:
No problem, it was for testing something else anyway. On the next node update we can remember there are five spots not four.

Issue #41, opened 2022-07-12
Add a logger


slifty:
We're using `console.log` in a few places (which our linting rules are about to yell at; fun fact I misconfigured our linting whoops!).

We should set up a real logger soon so we can start logging with meaningful log levels.

I'm happy to take this on since I've set it up a few times; [winston](https://www.npmjs.com/package/winston) seems to still be an oft-used tool for this.

bickelj:
It looks like winston is still number 1 and actively maintained, pino is number 2 and actively maintained, and bunyan is number 3 and stale? Something like that. Whatever is easiest to set up, which it sounds like winston is if you are setting it up. Since I've set up neither I would lean toward pino but that is based on nothing much.

slifty:
@bickelj if you want to run with it and you wanna run with pino that's fine by me!

If you're up for Winston, here's some existing code we can use from another OTS project: https://github.com/PermanentOrg/notification-service/blob/main/src/log.ts

slifty:
Edit: but pino sure does look light weight and easy ;)

I defer to you completely!

bickelj:
Attempt at finding objective reasons to use pino instead: winston has 2644KiB dependencies on `npm install` vs pino which has 2276KiB dependencies. OK, the real reason for the preference is I don't see much/any setup required for pino. I will try it and see if it "just works." If not, winston for the win.

bickelj:
Relatively painless to get request logging working:
```
$ git diff src/app.ts
diff --git a/src/app.ts b/src/app.ts
index 5b849dc..21c7670 100644
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,7 +1,10 @@
 import express from 'express';
+import PinoHttp from 'pino-http';
 import { rootRouter } from './routers';
 
 const app = express();
+const requestLogger = PinoHttp();
+app.use(requestLogger);
 app.use(express.urlencoded({ extended: true }));
 app.use(express.json());
 app.use('/', rootRouter);
```

But that is not all the logging, we would also want to replace console logs, have reasonable levels, etc.

bickelj:
Likewise for plain logging, it was relatively painless:
```
$ git diff src/index.ts
diff --git a/src/index.ts b/src/index.ts
index ff96239..454b8bf 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,10 @@
 import dotenv from 'dotenv';
+import pino from 'pino';
 import { app } from './app';
 
[...]

PR #45, opened 2022-07-12
Add pino logging libraries

bickelj:
Because pino was easy to get started with and claims speed, try it out.

Pino has a request logging library called pino-http. The logs are in
JSON format by default and unix times since epoch as well, but there is
also the pino-pretty library that will make these friendly in terminals.

We can pipe to pino-pretty on our ":dev" npm tasks such as start:dev and
migrate:dev. When doing so, use RFC3339 datetime format when prettyfying
because it is a good, useful standard. Production logs can be sent to it
later, I assume.

Issue #41 Add a logger

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,10 @@
     "cleanup": "rimraf dist",
     "lint": "eslint ./src --ext .ts",
     "migrate": "node -r dotenv/config dist/scripts/migrate.js",
-    "migrate:dev": "ts-node -r dotenv/config src/scripts/migrate.ts",
+    "migrate:dev": "ts-node -r dotenv/config src/scripts/migrate.ts | pino-pretty --translateTime 'UTC:yyyy-mm-dd\"T\"HH:MM:ss.l\"Z\"'",
     "test": "jest",
     "start": "node dist/index.js",
-    "start:dev": "ts-node src/index.ts"
+    "start:dev": "ts-node src/index.ts | pino-pretty --translateTime 'UTC:yyyy-mm-dd\"T\"HH:MM:ss.l\"Z\"'"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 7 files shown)


src/logger.ts
```diff
@@ -0,0 +1,5 @@
+import pino from 'pino';
+
+const logger = pino();
+
+export { logger };
```
bickelj, line 5: The justification for such a seemingly useless file here is if we wished to further configure the logger, that configuration would go here.

slifty: 100% -- this is a good solid file, the node way!!

package.json
```diff
@@ -10,10 +10,10 @@
     "cleanup": "rimraf dist",
     "lint": "eslint ./src --ext .ts",
     "migrate": "node -r dotenv/config dist/scripts/migrate.js",
-    "migrate:dev": "ts-node -r dotenv/config src/scripts/migrate.ts",
+    "migrate:dev": "ts-node -r dotenv/config src/scripts/migrate.ts | pino-pretty --translateTime 'UTC:yyyy-mm-dd\"T\"HH:MM:ss.l\"Z\"'",
```
slifty, line 13: It looks like pino-pretty supports the idea of a config file at `pino-prettyrc` -- I think that would be nicer than sending config commands in here (both because of DRY but also this is now a much more verbose command than it was before!)

slifty (changes requested):
Small change to offload the config to a file rather than CLI argument (I saw the note in Zulip and will make it myself; hope you agree with the idea!).

package.json
slifty: Fixed up in [06139f2](https://github.com/PhilanthropyDataCommons/service/pull/45/commits/06139f2eebaa527dc1a163db055628c2936f3206)

slifty:
OK! I should have said this in the first review but this is *so much nicer than winston* and looks really good.  Love it.

PR #46, opened 2022-07-13
Add integration test support

slifty:
This PR adds support for unit and integration tests -- test files need to be named `.unit.test.ts` and `.int.test.ts` going forward.

Note: I was tempted to have the filename be `integration.test.ts` to avoid shorthand / be explicit but it seems like people prefer `int` out there.

This PR currently has integration tests run in sequence until an upstream bug is patched.  I'll open an issue to track that improvement.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,16 +36,16 @@ jobs:
       - name: Run build
         run: npm run build
 
-      - name: Run migrations
-        run: npm run migrate
+      - name: Run unit tests
+        run: npm run test:unit -- --coverage
+
+      - name: Run integration tests
+        run: npm run test:integration -- --coverage
         env:
           PGHOST: localhost
           PGPORT: 5432
           PGUSER: postgres
           PGPASSWORD: postgres
-
-      - name: Run tests
-        run: npm run test -- --coverage
 
       - name: Upload coverage
         uses: codecov/codecov-action@v2
--- a/jest.config.int.js
+++ b/jest.config.int.js
@@ -1,0 +1,4 @@
+var config = require('./jest.config.base.js');
+config.testMatch = ['**/?(*.)+(int).(spec|test).[jt]s?(x)'];
+config.setupFilesAfterEnv = ["<rootDir>/src/test/integrationSuiteSetup.ts"];
+module.exports = config;
--- a/src/__tests__/canonicalFields.int.test.ts
+++ b/src/__tests__/canonicalFields.int.test.ts
@@ -1,14 +1,20 @@
 import request from 'supertest';
 import { app } from '../app';
+import {
+  prepareDatabaseForCurrentWorker,
+  cleanupDatabaseForCurrentWorker,
+} from '../test/harnessFunctions';
 
 const agent = request.agent(app);
 
 describe('/canonicalFields', () => {
   describe('/', () => {
     it('should return HTTP Status Code 200 OK', async () => {
+      await prepareDatabaseForCurrentWorker();
       await agent
         .get('/canonicalFields')
         .expect(200);
+      await cleanupDatabaseForCurrentWorker();
     });
   });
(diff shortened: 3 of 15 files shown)


src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,14 +1,20 @@
...
+  prepareDatabaseForCurrentWorker,
+  cleanupDatabaseForCurrentWorker,
+} from '../test/utils';
 
 const agent = request.agent(app);
 
 describe('/canonicalFields', () => {
   describe('/', () => {
     it('should return HTTP Status Code 200 OK', async () => {
+      await prepareDatabaseForCurrentWorker();
```
slifty, line 13: This should probably be automatic / handled by a `beforeAll` hook for any `integration.test.ts` file.

I was hoping for a way to be more explicit about which tests require a database connection -- e.g. there might be an integration test that does NOT use the database and so spinning up a schema / migrations would be a total waste of time.

.github/workflows/test.yml
```diff
@@ -36,16 +36,16 @@ jobs:
       - name: Run build
         run: npm run build
 
-      - name: Run migrations
-        run: npm run migrate
```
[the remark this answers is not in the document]
bickelj: Oh my mistake, this is the workflow, not the `package.json`. Disregard.

src/scripts/migrate.ts
```diff
@@ -6,8 +6,7 @@ const logger = root_logger.child({ source: 'migrate' });
 logger.info('Starting migrations...');
 migrate()
   .then(() => {
-    logger.info('Migrations complete.');
-    process.exit();
+    process.stdout.write('Migrations complete.\n');
```
bickelj, line 9: Should it be `logger.info`?

slifty: Oh yes -- I kept the wrong line when resolving the merge conflict

src/test/utils.ts
```diff
@@ -0,0 +1,40 @@
+import {
```
bickelj, line 1: Is there a better name than `utils.ts`? Perhaps `setupAndTeardown.ts`, `harnessFunctions.ts`, `databaseSetup.ts`? Something like that.

src/test/integrationSuiteSetup.ts
```diff
@@ -0,0 +1,5 @@
+import { db } from '../database';
+
+afterAll(async () => {
+  await db.close();
+});
```
bickelj, line 5: I am still trying to find where this fits in, but frankly I have not followed closely how any of the other pieces fit together, so I don't intend to spend too long looking.

jest.config.int.js
```diff
@@ -0,0 +1,4 @@
+var config = require('./jest.config.base.js');
+config.testMatch = ['**/?(*.)+(int).(spec|test).[jt]s?(x)'];
+config.setupFilesAfterEnv = ["<rootDir>/src/test/integrationSuiteSetup.ts"];
```
bickelj, line 3: Ah. Here is where that below file is referenced.

src/test/integrationSuiteSetup.ts
bickelj: Found it above.

slifty: I can add a comment explaining what is going on for the casual observer (as you saw below, this is Jest Magic and it's not explicit).

This file is also where I'd add the automatic "db up / down before each integration test" logic.

src/__tests__/canonicalFields.int.test.ts
slifty: While I think this is still true -- I think we can do that as an easy refactor later in the name of enabling the baseline ability to write tests.

bickelj (approved):
👌

Issue #31, opened 2022-07-12
Implement `GET /opportunities` endpoint


slifty:
This endpoint should return all registered opportunities in the system.

bickelj:
It seems like #46 would be helpful prior to adding this, maybe not strictly required.

slifty:
Yes for sure -- I'll get that merged in ASAP

bickelj:
Wow, this worked:
```
async function getOpportunities(req: Request, res: Response):Promise<void> {
  const opportunities: Result<Opportunity[]> =
      await db.sql<Opportunity[]>('selectAllOpportunities');
  res.status(200)
    .contentType('application/json')
    .send(opportunities.rows);
}
```

bickelj:
I get the same exact result when using this instead, so there is no magic in the additional type information:
```
async function getOpportunities(req: Request, res: Response):Promise<void> {
  const opportunities = await db.sql('selectAllOpportunities');
  res.status(200)
    .contentType('application/json')
    .send(opportunities.rows);
}
```

bickelj:
The type info does seem useful though. As far as not getting an integer when it is a bigint, it looks like that's some weirdness with the node pg-types library and/or JavaScript itself. I see this nugget at https://github.com/brianc/node-pg-types/

> Let's do something I commonly like to do on projects: return 64-bit integers (int8) as JavaScript integers. Because JavaScript doesn't have support for 64-bit integers node-postgres cannot confidently parse int8 data type results as numbers because if you have a huge number it will overflow and the result you'd get back from node-postgres would not be the result in the database. That would be a very bad thing so node-postgres just returns int8 results as strings and leaves the parsing up to you.

Oh yeah. JavaScript `number` covers both ints and floats. I guess I had forgotten how weird JavaScript can be.

bickelj:
So treating it like a string in JSON is appropriate, I suppose.

bickelj:
There are some interesting possibilities with https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-CREATION-TABLE but first we should get the straighter-forward thing working.

bickelj:
Done in PR #66

PR #49, opened 2022-07-18
Implement GET /opportunities (first draft)

bickelj:
A first draft of an implementation of HTTP GET on /opportunities. It
includes SQL for the schema, removal of cruft, SQL for the query, types,
and the boilerplate/glue code needed to wire it to the express server.
What it still lacks is tests, passing lint, and probably node-ish-ness.

Issue #31 Implement GET /opportunities endpoint

### DIFF
--- a/src/types/opportunitiesTypes.ts
+++ b/src/types/opportunitiesTypes.ts
@@ -1,0 +1,10 @@
+export interface OpportunityMinimal {
+  title: string;
+}
+
+export interface Opportunity extends OpportunityMinimal {
+  id: bigint;
+  created_at: Date;
+}
+
+// export type Opportunity = OpportunityMinimal | OpportunityMaximal;
--- a/src/database/migrations/0003-create-opportunities.sql
+++ b/src/database/migrations/0003-create-opportunities.sql
@@ -1,0 +1,10 @@
+create table opportunities (
+  id bigint primary key generated always as identity,
+  created_at timestamp with time zone not null default now(),
+  title varchar not null
+);
+
+comment on table opportunities is
+  'Funding opportunities such as available grants or awards.';
+comment on column opportunities.title is
+  'The external identifier; not necessarily unique across organizations.';
--- a/src/database/queries/selectAllOpportunities.sql
+++ b/src/database/queries/selectAllOpportunities.sql
@@ -1,0 +1,7 @@
+select id,
+     created_at,
+     title
+from opportunities
+order by id
+offset 0
+fetch next 10000 rows only;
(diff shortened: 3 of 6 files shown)


bickelj:
I know linting fails, but something in here is probably a simple misunderstanding due to my node/typescript noobness. Regardless, is this close to the shape of what you would expect for an implementation of this subset of the API?

bickelj:
I see the integration tests taking more than two minutes to run as well, canceling those.

src/types/opportunitiesTypes.ts
```diff
@@ -0,0 +1,10 @@
+export interface OpportunityMinimal {
+  title: string;
+}
+
+export interface Opportunity extends OpportunityMinimal {
```
slifty, line 5: In TypeScript the way to signal this is to define `Opportunity` and specify optional fields where appropriate using `?`.

e.g.

```
export interface Opportunity {
  id?: number;
  title: string;
  createdAt?: Date;
}
```

For situations where we only need a smaller piece of an opportunity there are a series of [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) that let us specify a *contextual* modifications of a given interface.

For instance if we had an API endpoint that ONLY needed a smaller subset of `Opportunity` fields we could use `Pick<Opportunity, 'title' | 'id'>` and that would require both title and id for that particular parameter.

slifty (changes requested):
ALL RIGHT!  Sorry it took so long to get you any feedback, I had a WIP branch that contains some of the patterns I wanted to propose which I wanted to get out there to be able to point to!

Some things I think are worth doing as part of our endpoint implementation are:

1. fixture-based tests
2. type checking of the database response (and 500 erroring with a useful error message if the type check fails)
3. Defining defaults and type guards along side types (and trusting that we can use utility types to define subsets of interfaces where needed, rather than defining a bunch of versions of the same type)

I think we probably want to avoid bigints for the reasons you recognized in another thread; it'd be one thing if it was free but I think the added complexity isn't worth it for the moment.

src/types/opportunitiesTypes.ts
bickelj: Thanks for the explanation here and in https://github.com/PhilanthropyDataCommons/service/pull/55#discussion_r927940853. It looks like I clicked the link you gave here but did not actually go through it and understand it until your other comment. I see the approach and it makes sense!

bickelj:
Closing without merge because this was before a good pattern (and the dependent endpoint implementation) was established by @slifty in #55. Second draft was in #63 which bit off too much and assumed too much.

PR #50, opened 2022-07-19
Add a link to the PDC site

bickelj:
Philanthropy Data Commons site is https://philanthropydatacommons.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
 
 # Philanthropy Data Commons Service
 
+This project is related to [Philanthropy Data Commons](https://philanthropydatacommons.org) (PDC). Please visit https://philanthropydatacommons.org for an overview of PDC.
+
 ## Development
 
 In order to run this software you need to set up a [Postgres 14](https://www.postgresql.org/) database.


README.md
```diff
@@ -5,6 +5,10 @@
 
 # Philanthropy Data Commons Service
 
+This project is related to
```
slifty, line 8: (This is a TINY nit pick but we should decide)

This .md file right now isn't wrapping mid-paragraph -- e.g. we just rely on word wrap for the formatting.

I did it that way just because (A) I find it a little awkward to have to reformat potentailly X lines because I added a word and it cascaded, (B) my tools don't manually do that reformatting <-- I should solve this

All this is to say: should we change that and break on line? If so let's make that change in this PR with a separate commit.  If not, let's smoosh this into one line.

slifty (changes requested):
I am sorry for being pedantic X(

We should probably be consistent with line breaks in our markdown files.

Right now we have this file and the files in `/docs` with NO manual breaks (e.g. paragraphs run on a single line and line breaks are used when double breaking).  The simplest approach would be to maintain this pattern.

README.md
bickelj: I'm smooshing it into one line for the purposes of this PR. I hadn't looked closely at the other lines in this file going way over 80 chars. I probably saw the links at the top which were greater than 80 chars and assumed it was just the links. I agree if we want to change to another pattern we can do that in a separate commit or separate PR.

slifty:
Argh sorry I meant to mark it as approved (with the optional change!) anyway ty!

slifty (approved):

PR #52, opened 2022-07-20
Use consistent logger configuration

### LINKED ISSUE #51: pino-http is generating a lot of logger noise in tests
It looks like pinoHttp is generating logs during tests regardless of how our logger is configured.

I believe we can solve this by passing our pino logger (which is configured) to pinoHttp.  This will also mean that when we add the ability to set log level it will be honored by pinoHttp.

slifty:
This PR passes our configured `pino` loger to `pinoHttp` so that it uses our configurations.

Resolves #51

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,9 +1,12 @@
 import express from 'express';
-import PinoHttp from 'pino-http';
+import pinoHttp from 'pino-http';
 import { rootRouter } from './routers';
+import { logger } from './logger';
 
 const app = express();
-app.use(PinoHttp());
+app.use(pinoHttp({
+  logger,
+}));
 app.use(express.urlencoded({ extended: true }));
 app.use(express.json());
 app.use('/', rootRouter);


src/app.ts
```diff
@@ -1,9 +1,12 @@
 import express from 'express';
-import PinoHttp from 'pino-http';
+import pinoHttp from 'pino-http';
```
bickelj, line 2: Thanks for correcting this nuance. I'm not sure which example I followed or if it was a guess.

I see this in the source at https://github.com/pinojs/pino-http/blob/c9ce2ac15cb1ad3bb72598c4034b46cce5333ea2/index.d.ts#L66
```
export default PinoHttp;
export { PinoHttp as pinoHttp };
```
But then above at https://github.com/pinojs/pino-http/blob/c9ce2ac15cb1ad3bb72598c4034b46cce5333ea2/index.d.ts#L15, PinoHttp is a function, not a type nor interface nor class:
```
declare function PinoHttp(opts?: Options, stream?: pino.DestinationStream): HttpLogger;

declare function PinoHttp(stream?: pino.DestinationStream): HttpLogger;
```
But wait, it's a function of type `HttpLogger` which is an `interface`:
```
export interface HttpLogger {
...
```
I'll assume this makes sense to you and accept the change in capitalization in good faith.

PR #54, opened 2022-07-21
Invoke test setup via suite hooks

### LINKED ISSUE #43: Support integration tests
We're about to need to run integration tests (I imagine many of our tests will rely on DB access in fact!)

This will mean three test commands:

`npm run test:unit` -- runs unit tests (no database)
`npm run test:integration` -- runs untegration tests
`npm run test` -- runs both sets of tests

Part of this issue will be figuring out the right way to do integration tests so that they're run in parallel but don't step on each other.

Potential resources I've found:

https://dev.to/walrusai/testing-database-interactions-with-jest-519n
https://jestjs.io/docs/setup-teardown
https://jestjs.io/docs/configuration#testmatch-arraystring

slifty:
This PR moves the burden of running database setup / teardown within a test over to a "suite" approach.

This also solves a bug where teardown would not happen if a test failed.

Once this is merged I think we "support" integration tests (which isn't to say there isn't more room for improvement)

Resolves #43

### DIFF
--- a/src/test/integrationSuiteSetup.ts
+++ b/src/test/integrationSuiteSetup.ts
@@ -3,7 +3,19 @@
  * via `setupFilesAfterEnv`.
  */
 import { db } from '../database';
+import {
+  prepareDatabaseForCurrentWorker,
+  cleanupDatabaseForCurrentWorker,
+} from './harnessFunctions';
 
 afterAll(async () => {
   await db.close();
+});
+
+beforeEach(async () => {
+  await prepareDatabaseForCurrentWorker();
+});
+
+afterEach(async () => {
+  await cleanupDatabaseForCurrentWorker();
 });
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "migrate": "node -r dotenv/config dist/scripts/migrate.js",
     "migrate:dev": "ts-node -r dotenv/config src/scripts/migrate.ts | pino-pretty",
     "test": "npm run test:unit && npm run test:integration",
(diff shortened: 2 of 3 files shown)


src/test/integrationSuiteSetup.ts
```diff
@@ -3,7 +3,19 @@
  * via `setupFilesAfterEnv`.
  */
 import { db } from '../database';
+import {
+  prepareDatabaseForCurrentWorker,
+  cleanupDatabaseForCurrentWorker,
+} from './harnessFunctions';
```
bickelj, line 9: It looks good. I am trying to imagine if we will have "int" tests that want to opt-out from database setup and teardown. In other words, despite the repetition, would it be worthwhile to include a `beforeEach` declaration for each `....int.test.ts` file in the `describe` block? It is clearer, I think, to have each integration test file have its own beforeEach/afterEach, but two counts against that approach are:
1. We have a suite/all-tests level `afterAll` call in `integrationSuiteSetup`, and
2. I am pretty sure all these tests will need the database (if not, it's a 50-500ms no-op).

bickelj (approved):
Yes, this is good!

src/test/integrationSuiteSetup.ts
slifty: I agree with the concerns for what it's worth.

My ideal solution would be Jest just having a more robust way of defining categories of tests (e.g. having a `@usesDatabase` tag for the test, which isn't a thing but wouldn't it be nice).

Maybe we refactor to that down the line when we know more about our test suite?

PR #55, opened 2022-07-21
Add GET /canonicalFields endpoint

### LINKED ISSUE #30: Implement `GET /canonicalFields` API endpoint
This endpoint will allow a third party application to list all canonical fields.  For this initial implementation we don't need to create filters, offsets, or pagination -- just returning the full list of elements.

slifty:
This PR adds a real implementation for the `GET /canonicalFields` endpoint.

In doing so it also introduces the concept of data validation as well as test fixtures for integration tests.

Resolves #30 

(This includes the code in #54 so it should be rebased against main once that is merged)

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,8 @@
     "typescript": "^4.7.3"
   },
   "dependencies": {
+    "ajv": "^8.11.0",
+    "ajv-keywords": "^5.1.0",
     "dotenv": "^16.0.1",
     "express": "^4.18.1",
     "pino": "^8.1.0",
--- a/src/__tests__/canonicalFields.int.test.ts
+++ b/src/__tests__/canonicalFields.int.test.ts
@@ -1,14 +1,65 @@
 import request from 'supertest';
 import { app } from '../app';
+import { db } from '../database';
+import type { Result } from 'tinypg';
 
 const agent = request.agent(app);
 
 describe('/canonicalFields', () => {
   describe('/', () => {
-    it('should return HTTP Status Code 200 OK', async () => {
+    it('returns an empty array when no data is present', async () => {
       await agent
         .get('/canonicalFields')
-        .expect(200);
+        .expect(200, []);
+    });
+
+    it('returns all canonical fields present in the database', async () => {
+      await db.query(`
+        INSERT INTO canonical_fields (
+          label,
+          short_code,
+          data_type,
+          created_at
+        )
+        VALUES
+          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+      `);
+      await agent
+        .get('/canonicalFields')
+        .expect(
+          200,
+          [
+            {
+              createdAt: '2022-07-20T12:00:00.000Z',
+              dataType: 'string',
+              id: 1,
+              label: 'First Name',
+              shortCode: 'firstName',
+            },
+            {
+              createdAt: '2022-07-20T12:00:00.000Z',
+              dataType: 'string',
+              id: 2,
+              label: 'Last Name',
+              shortCode: 'lastName',
+            },
+          ],
+        );
+    });
+
+    it('should error if the database returns an unexpected data structure', async () => {
+      jest.spyOn(db, 'sql')
+        .mockImplementationOnce(async () => ({
+          rows: [{ foo: 'not a valid result' }],
+        }) as Result<object>);
+      const result = await agent
+        .get('/canonicalFields')
+        .expect(500);
+      expect(result.body).toMatchObject({
+        name: 'ValidationError',
+        errors: expect.any(Array) as unknown[],
+      });
     });
   });
 });
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,8 +1,9 @@
 import express from 'express';
 import pinoHttp from 'pino-http';
 import { rootRouter } from './routers';
-import { logger } from './logger';
+import { getLogger } from './logger';
 
+const logger = getLogger(__filename);
 const app = express();
 app.use(pinoHttp({
   logger,
--- a/src/database/migrations/0001_create_canonical_fields.sql
+++ b/src/database/migrations/0001_create_canonical_fields.sql
@@ -1,0 +1,15 @@
+CREATE TABLE canonical_fields (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  label VARCHAR NOT NULL,
+  short_code VARCHAR NOT NULL UNIQUE,
+  data_type VARCHAR NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+
+
+comment on table canonical_fields is
+  'Canonical fields are those fields that are unique across organizations.';
+comment on column canonical_fields.label is 'A word or phrase that communicates the meaning of the field for end users';
+comment on column canonical_fields.short_code is 'An externally meaningful alphanumeric code tht can be used to reference this field without knowing the internal id';
+comment on column canonical_fields.data_type is 'The type of data that can be associated with this field';
--- a/src/database/queries/canonicalFields/fetchAll.sql
+++ b/src/database/queries/canonicalFields/fetchAll.sql
@@ -1,0 +1,7 @@
+SELECT
+  cf.id as "id",
+  cf.label as "label",
+  cf.short_code as "shortCode",
+  cf.data_type as "dataType",
+  cf.created_at as "createdAt"
+FROM canonical_fields as cf;
--- a/src/errors/ValidationError.ts
+++ b/src/errors/ValidationError.ts
@@ -1,0 +1,16 @@
+// This error was donated by the TV Kitchen project / Bad Idea Factory
+// https://github.com/tvkitchen/countertop/blob/main/src/errors/ValidationError.ts
+import type { ErrorObject } from 'ajv';
+
+export class ValidationError extends Error {
+  public errors?: ErrorObject[] | null;
+
+  public constructor(
+    message: string,
+    errors: ErrorObject[],
+  ) {
+    super(message);
+    this.name = this.constructor.name;
+    this.errors = errors;
+  }
+}
--- a/src/handlers/canonicalFieldsHandlers.ts
+++ b/src/handlers/canonicalFieldsHandlers.ts
@@ -1,13 +1,35 @@
+import { logger } from '../logger';
+import { db } from '../database';
+import { isCanonicalFieldArray } from '../types';
+import { ValidationError } from '../errors';
 import type {
   Request,
   Response,
 } from 'express';
 
-function getCanonicalFields(req: Request, res: Response): void {
-  res.status(200)
-    .contentType('application/json')
-    .send('{ "hello": "world"}');
-}
+const getCanonicalFields = (req: Request, res: Response): void => {
+  db.sql('canonicalFields.fetchAll')
+    .then((canonicalFields) => {
+      logger.debug(canonicalFields, 'canonicalFields.fetchAll');
+      const { rows } = canonicalFields;
+      if (isCanonicalFieldArray(rows)) {
+        res.status(200)
+          .contentType('application/json')
+          .send(rows);
+      } else {
+        throw new ValidationError(
+          'The database responded with an unexpected format.',
+          isCanonicalFieldArray.errors ?? [],
+        );
+      }
+    })
+    .catch((error: unknown) => {
+      logger.error(error);
+      res.status(500)
+        .contentType('application/json')
+        .send(error);
+    });
+};
 
 export const canonicalFieldsHandlers = {
   getCanonicalFields,
--- a/src/handlers/canonicalFieldsHandlers.ts
+++ b/src/handlers/canonicalFieldsHandlers.ts
@@ -1,4 +1,4 @@
-import { logger } from '../logger';
+import { getLogger } from '../logger';
 import { db } from '../database';
 import { isCanonicalFieldArray } from '../types';
 import { ValidationError } from '../errors';
@@ -6,6 +6,8 @@ import type {
   Request,
   Response,
 } from 'express';
+
+const logger = getLogger(__filename);
 
 const getCanonicalFields = (req: Request, res: Response): void => {
   db.sql('canonicalFields.fetchAll')
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,5 +1,7 @@
 import pino from 'pino';
 
-const logger = pino();
+const logger = pino({
+  level: process.env.LOG_LEVEL ?? 'silent',
+});
 
 export { logger };
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,7 +1,10 @@
 import pino from 'pino';
+import type {
+  Logger,
+} from 'pino';
 
 const logger = pino({
   level: process.env.LOG_LEVEL ?? 'silent',
 });
 
-export { logger };
+export const getLogger = (source: string): Logger => logger.child({ source });
--- a/src/test/integrationSuiteSetup.ts
+++ b/src/test/integrationSuiteSetup.ts
@@ -18,4 +18,5 @@ beforeEach(async () => {
 
 afterEach(async () => {
   await cleanupDatabaseForCurrentWorker();
+  jest.restoreAllMocks();
 });
--- a/src/types/CanonicalField.ts
+++ b/src/types/CanonicalField.ts
@@ -1,0 +1,49 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+export interface CanonicalField {
+  id: number;
+  label: string;
+  shortCode: string;
+  dataType: string;
+  createdAt: Date;
+}
+
+export const canonicalFieldSchema: JSONSchemaType<CanonicalField> = {
+  type: 'object',
+  properties: {
+    id: {
+      type: 'integer',
+    },
+    label: {
+      type: 'string',
+    },
+    shortCode: {
+      type: 'string',
+    },
+    dataType: {
+      type: 'string',
+    },
+    createdAt: {
+      type: 'object',
+      required: [],
+      instanceof: 'Date',
+    },
+  },
+  required: [
+    'id',
+    'label',
+    'shortCode',
+    'dataType',
+    'createdAt',
+  ],
+};
+
+export const isCanonicalField = ajv.compile(canonicalFieldSchema);
+
+const canonicalFieldArraySchema: JSONSchemaType<CanonicalField[]> = {
+  type: 'array',
+  items: canonicalFieldSchema,
+};
+
+export const isCanonicalFieldArray = ajv.compile(canonicalFieldArraySchema);
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,8 @@
       "version": "0.0.0",
       "license": "AGPL-3.0",
       "dependencies": {
+        "ajv": "^8.11.0",
+        "ajv-keywords": "^5.1.0",
         "dotenv": "^16.0.1",
         "express": "^4.18.1",
         "pino": "^8.1.0",
@@ -661,6 +663,28 @@
       "engines": {
         "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
       }
+    },
+    "node_modules/@eslint/eslintrc/node_modules/ajv": {
+      "version": "6.12.6",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
+      "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
+      "dev": true,
+      "dependencies": {
+        "fast-deep-equal": "^3.1.1",
+        "fast-json-stable-stringify": "^2.0.0",
+        "json-schema-traverse": "^0.4.1",
+        "uri-js": "^4.2.2"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/epoberezkin"
+      }
+    },
+    "node_modules/@eslint/eslintrc/node_modules/json-schema-traverse": {
+      "version": "0.4.1",
+      "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-0.4.1.tgz",
+      "integrity": "sha512-xbbCH5dCYU5T8LcEhhuh7HJ88HXuW3qsI3Y0zOZFKfZEHcpWiHU/Jxzk629Brsab/mMiHQti9wMP+845RPe3Vg==",
+      "dev": true
     },
     "node_modules/@humanwhocodes/config-array": {
       "version": "0.9.5",
@@ -1686,19 +1710,29 @@
       }
     },
     "node_modules/ajv": {
-      "version": "6.12.6",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
-      "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
-      "dev": true,
+      "version": "8.11.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.11.0.tgz",
+      "integrity": "sha512-wGgprdCvMalC0BztXvitD2hC04YffAvtsUn93JbGXYLAtCUO4xd17mCCZQxUOItiBwZvJScWo8NIvQMQ71rdpg==",
       "dependencies": {
         "fast-deep-equal": "^3.1.1",
-        "fast-json-stable-stringify": "^2.0.0",
-        "json-schema-traverse": "^0.4.1",
+        "json-schema-traverse": "^1.0.0",
+        "require-from-string": "^2.0.2",
         "uri-js": "^4.2.2"
       },
       "funding": {
         "type": "github",
         "url": "https://github.com/sponsors/epoberezkin"
+      }
+    },
+    "node_modules/ajv-keywords": {
+      "version": "5.1.0",
+      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-5.1.0.tgz",
+      "integrity": "sha512-YCS/JNFAUyr5vAuhk1DWm1CBxRHW9LbJ2ozWeemrIqpbsqKjHVxYPyi5GC0rjZIT5JxJ3virVTS8wk4i/Z+krw==",
+      "dependencies": {
+        "fast-deep-equal": "^3.1.3"
+      },
+      "peerDependencies": {
+        "ajv": "^8.8.2"
       }
     },
     "node_modules/ansi-escapes": {
@@ -2882,6 +2916,22 @@
         "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
       }
     },
+    "node_modules/eslint/node_modules/ajv": {
+      "version": "6.12.6",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
+      "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
+      "dev": true,
+      "dependencies": {
+        "fast-deep-equal": "^3.1.1",
+        "fast-json-stable-stringify": "^2.0.0",
+        "json-schema-traverse": "^0.4.1",
+        "uri-js": "^4.2.2"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/epoberezkin"
+      }
+    },
     "node_modules/eslint/node_modules/eslint-scope": {
       "version": "7.1.1",
       "resolved": "https://registry.npmjs.org/eslint-scope/-/eslint-scope-7.1.1.tgz",
@@ -2903,6 +2953,12 @@
       "engines": {
         "node": ">=4.0"
       }
+    },
+    "node_modules/eslint/node_modules/json-schema-traverse": {
+      "version": "0.4.1",
+      "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-0.4.1.tgz",
+      "integrity": "sha512-xbbCH5dCYU5T8LcEhhuh7HJ88HXuW3qsI3Y0zOZFKfZEHcpWiHU/Jxzk629Brsab/mMiHQti9wMP+845RPe3Vg==",
+      "dev": true
     },
     "node_modules/espree": {
       "version": "9.3.2",
@@ -3118,8 +3174,7 @@
     "node_modules/fast-deep-equal": {
       "version": "3.1.3",
       "resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
-      "integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q==",
-      "dev": true
+      "integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q=="
     },
     "node_modules/fast-glob": {
       "version": "3.2.11",
@@ -4754,10 +4809,9 @@
       "dev": true
     },
     "node_modules/json-schema-traverse": {
-      "version": "0.4.1",
-      "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-0.4.1.tgz",
-      "integrity": "sha512-xbbCH5dCYU5T8LcEhhuh7HJ88HXuW3qsI3Y0zOZFKfZEHcpWiHU/Jxzk629Brsab/mMiHQti9wMP+845RPe3Vg==",
-      "dev": true
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-1.0.0.tgz",
+      "integrity": "sha512-NM8/P9n3XjXhIZn1lLhkFaACTOURQXjWhV4BA/RnOv8xvgqtqpAX9IO4mRQxSx1Rlo4tqzeqb0sOlruaOy3dug=="
     },
     "node_modules/json-stable-stringify-without-jsonify": {
       "version": "1.0.1",
@@ -5884,6 +5938,14 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/require-from-string": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/require-from-string/-/require-from-string-2.0.2.tgz",
+      "integrity": "sha512-Xf0nWe6RseziFMu+Ap9biiUbmplq6S9/p+7w7YXP/JBHhrUDDUhwa+vANyubuqfZWTveU//DYVGsDG7RKL/vEw==",
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
     "node_modules/resolve": {
       "version": "1.22.1",
       "resolved": "https://registry.npmjs.org/resolve/-/resolve-1.22.1.tgz",
@@ -6853,7 +6915,6 @@
       "version": "4.4.1",
       "resolved": "https://registry.npmjs.org/uri-js/-/uri-js-4.4.1.tgz",
       "integrity": "sha512-7rKUyy33Q1yc98pQ1DAmLtwX109F7TIfWlW1Ydo8Wl1ii1SeHieeh0HHfPeL2fMXK6z0s8ecKs9frCuLJvndBg==",
-      "dev": true,
       "dependencies": {
         "punycode": "^2.1.0"
       }
@@ -6862,7 +6923,6 @@
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/punycode/-/punycode-2.1.1.tgz",
       "integrity": "sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A==",
-      "dev": true,
       "engines": {
         "node": ">=6"
       }
@@ -7567,6 +7627,26 @@
         "js-yaml": "^4.1.0",
         "minimatch": "^3.1.2",
         "strip-json-comments": "^3.1.1"
+      },
+      "dependencies": {
+        "ajv": {
+          "version": "6.12.6",
+          "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
+          "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
+          "dev": true,
+          "requires": {
+            "fast-deep-equal": "^3.1.1",
+            "fast-json-stable-stringify": "^2.0.0",
+            "json-schema-traverse": "^0.4.1",
+            "uri-js": "^4.2.2"
+          }
+        },
+        "json-schema-traverse": {
+          "version": "0.4.1",
+          "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-0.4.1.tgz",
+          "integrity": "sha512-xbbCH5dCYU5T8LcEhhuh7HJ88HXuW3qsI3Y0zOZFKfZEHcpWiHU/Jxzk629Brsab/mMiHQti9wMP+845RPe3Vg==",
+          "dev": true
+        }
       }
     },
     "@humanwhocodes/config-array": {
@@ -8380,15 +8460,22 @@
       "dev": true
     },
     "ajv": {
-      "version": "6.12.6",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
-      "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
-      "dev": true,
+      "version": "8.11.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.11.0.tgz",
+      "integrity": "sha512-wGgprdCvMalC0BztXvitD2hC04YffAvtsUn93JbGXYLAtCUO4xd17mCCZQxUOItiBwZvJScWo8NIvQMQ71rdpg==",
       "requires": {
         "fast-deep-equal": "^3.1.1",
-        "fast-json-stable-stringify": "^2.0.0",
-        "json-schema-traverse": "^0.4.1",
+        "json-schema-traverse": "^1.0.0",
+        "require-from-string": "^2.0.2",
         "uri-js": "^4.2.2"
+      }
+    },
+    "ajv-keywords": {
+      "version": "5.1.0",
+      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-5.1.0.tgz",
+      "integrity": "sha512-YCS/JNFAUyr5vAuhk1DWm1CBxRHW9LbJ2ozWeemrIqpbsqKjHVxYPyi5GC0rjZIT5JxJ3virVTS8wk4i/Z+krw==",
+      "requires": {
+        "fast-deep-equal": "^3.1.3"
       }
     },
     "ansi-escapes": {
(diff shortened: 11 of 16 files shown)


src/database/migrations/0001_create_canonical_fields.sql
```diff
@@ -0,0 +1,13 @@
...
+  short_code TEXT,
+  field_type TEXT,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+comment on table canonical_fields is
+  'Canonical fields are those fields that are unique across organizations.';
+comment on column canonical_fields.label is 'The description of the field';
+comment on column canonical_fields.short_code is 'The short name of the field';
+comment on column canonical_fields.field_type is 'The data type of the field';
```
slifty, line 13: Should we actually just name this `data_type`?  (I used `field_type` since I think `type` was a reserved word, but I didn't like it.  data_type seems more descriptive and also removes the need for the comment).

src/database/migrations/0001_create_canonical_fields.sql
```diff
@@ -0,0 +1,13 @@
...
+  label TEXT,
+  short_code TEXT,
+  field_type TEXT,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+comment on table canonical_fields is
+  'Canonical fields are those fields that are unique across organizations.';
+comment on column canonical_fields.label is 'The description of the field';
+comment on column canonical_fields.short_code is 'The short name of the field';
```
slifty, line 12: Actually, this applies to all of these column comments: should we just use the more intuitive names for the field names?

e.g. `label` becomes `description and `short_code` becomes `short_name`

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -1,13 +1,35 @@
...
 
-function getCanonicalFields(req: Request, res: Response): void {
-  res.status(200)
-    .contentType('application/json')
-    .send('{ "hello": "world"}');
-}
+const getCanonicalFields = (req: Request, res: Response): void => {
+  db.sql('canonicalFields.fetchAll')
+    .then((canonicalFields) => {
+      logger.debug(canonicalFields, 'canonicalFields.fetchAll');
```
slifty, line 13: @bickelj I'd love your thoughts on when it's good to have built in logger lines in code... This is something I've always been torn about.

src/logger.ts
```diff
@@ -1,5 +1,7 @@
 import pino from 'pino';
 
-const logger = pino();
+const logger = pino({
+  level: process.env.LOG_LEVEL ?? 'silent',
```
slifty, line 4: I changed the default log level to silent in an epic battle against noisy tests, but forgot to change it back to info.

Also we might consider detecting if Jest `--silent` is enabled and having that override log level when set.  Here is an example of how to do that: https://github.com/chriszs/tv-kitchen/blob/ff61ea8611d092740e38a020b06d7aff4cd979f1/lib/utils/jest.setup.js#L9

src/types/CanonicalField.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export interface CanonicalField {
+  id: number;
+  label: string;
+  shortCode: string;
+  fieldType: string;
+  createdAt: Date;
+}
+
+export const defaultCanonicalField = {
```
slifty, line 12: I've found it to be a good / useful habit of providing a "default" structure for various interfaces.  That way a dev who wants to create a canonical field can do something like:

```
import { defaultCanonicalField } from './types'

const myCanonicalField = {
  ...defaultCanonicalField,
  shortCode: 'whatever',
}
```

tsconfig.dev.json
```diff
@@ -2,7 +2,8 @@
 {
   "extends": "@tsconfig/node16/tsconfig.json",
   "compilerOptions": {
-    "noEmit": true
+    "noEmit": true,
+    "resolveJsonModule": true
```
slifty, line 6: This is a byproduct of an experiment and should be taken out.

tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "module": "commonjs",
     "outDir": "dist",
     "declaration": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "resolveJsonModule": true
```
slifty, line 8: Same here -- not used.

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -1,13 +1,35 @@
+import { logger } from '../logger';
```
slifty, line 1: I saw in #49 you used `.child` -- that's super interesting and I think speaks to the broader question I ask in another thread of "how do we want to do logging".

Whatever we decide, let's document it.

src/database/migrations/0001_create_canonical_fields.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE canonical_fields (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  label TEXT,
```
slifty, line 3: all of these should be marked NOT NULL

package.json
```diff
@@ -59,6 +59,8 @@
     "typescript": "^4.7.3"
   },
   "dependencies": {
+    "ajv": "^8.11.0",
+    "ajv-keywords": "^5.1.0",
```
bickelj, line 63: I was about to ask if this needed to be a dev dependency, but no, if we are using it to parse and/or validate, it needs to be included in prod dist.

src/__tests__/__snapshots__/canonicalFields.int.test.ts.snap
```diff
@@ -0,0 +1,20 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
bickelj, line 1: I am still deciding whether I like this approach. Maybe after poking at it for a bit locally I will have more useful feedback.

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,20 +1,23 @@
...
+        .expect(200, []);
+    });
+
+    it('returns all canonical fields present in the database', async () => {
+      await db.sql('fixtures.canonicalFields.returnsAllCanonicalFields.seed');
+      const result = await agent
         .get('/canonicalFields')
         .expect(200);
-      await cleanupDatabaseForCurrentWorker();
+      expect(result.body).toMatchSnapshot();
```
bickelj, line 20: On first glance, the snapshot doesn't look like literal JSON to me, but I am not sure if that's a downside yet.

src/database/migrations/0001_create_canonical_fields.sql
bickelj: Yes, and text can be varchar.

src/database/migrations/0001_create_canonical_fields.sql
bickelj: Perhaps, yes. I think the main thing to document is what purpose the `short_code` serves as distinct from the `label` or `description`. Is there a "no spaces allowed in short_code" rule? Is the short_code going to be unique within an organization? Is it unique across organizations? Is it limited to 15 characters (and if so, why not char(15) or varchar(15))?

src/database/queries/canonicalFields/fetchAll.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+  canonical_fields.id as "id",
```
bickelj, line 2: Is it needed to qualify `canonical_fields.` here or following?

src/database/queries/canonicalFields/fetchAll.sql
```diff
@@ -0,0 +1,7 @@
...
+  canonical_fields.label as "label",
+  canonical_fields.short_code as "shortCode",
+  canonical_fields.field_type as "fieldType",
+  canonical_fields.created_at as "createdAt"
+FROM canonical_fields
```
bickelj, line 7: Maybe `from canonical_fields cf` or `from canonical_fields as cf` so we can say `cf.label` instead. Or again, just omit the qualification entirely since there are no other tables nor joins.

src/database/queries/fixtures/canonicalFields/returnsAllCanonicalFields/seed.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO canonical_fields (
+  label,
+  short_code,
+  field_type,
+  created_at
```
bickelj, line 5: I was about to object to setting this but then again for the test purposes we need it fixed. Good call.

src/handlers/canonicalFieldsHandlers.ts
bickelj: Based on my familiarity with log4j/logback/slf4j etc., I was looking in pino for a way to qualify log messages by file. The child logger was the first, most obvious solution that I saw so I ran with it. If there is a more natural or better way to be able to filter by file, I am up for that too, so long as it doesn't require me to remember with each log message to add something there. If there is a more automated way to find the filename rather than a string literal, that would be cool, too, because then we can avoid the inevitable lies deriving from a copy/paste/fail-to-edit.

src/handlers/canonicalFieldsHandlers.ts
bickelj: Whenever the developer working on that code finds it useful. I think if it was useful before, it is likely to be useful again. Until it's obviously in the way and clear enough, at which point we can remove it.

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -1,13 +1,35 @@
...
-  res.status(200)
-    .contentType('application/json')
-    .send('{ "hello": "world"}');
-}
+const getCanonicalFields = (req: Request, res: Response): void => {
+  db.sql('canonicalFields.fetchAll')
+    .then((canonicalFields) => {
+      logger.debug(canonicalFields, 'canonicalFields.fetchAll');
+      const { rows } = canonicalFields;
+      if (isCanonicalFieldArray(rows)) {
```
bickelj, line 15: Super nice.
I was about to say "maybe we don't need to validate from our own database" but if we have the function available why not?

src/logger.ts
bickelj: What was the noise during tests? From the pino-http request logger? Or did pino somehow get auto-changed to `debug` level during the tests? I am unfortunately a fan of verbosity. Perhaps it is the Java that is in me. I would rather grep a by-default-verbose log then have to go and enable verboser logging and re-run. But sometimes you have to do that anyway (increase verbosity to debug or to trace).

Once we can get the tests to fail properly I would be more comfortable silencing them more. But I'm OK with it now too, I can always set my local level to debug or trace.

src/types/CanonicalField.ts
```diff
@@ -0,0 +1,57 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+export interface CanonicalField {
+  id: number;
```
bickelj, line 5: So this is great for `CanonicalField` but I anticipate with most types we are going to need (at least) two options: one for reading and one for writing. When posting data, we cannot post an ID, we want that auto-generated, and probably the same for created datetime, although as noted above, for test purposes it seems like we want to specify it.

package.json
slifty: Correct!  They are runtime dependencies (companions to the typescript dev dependency).

src/types/CanonicalField.ts
bickelj: This is probably the Java in me again, but if we go that route, why not provide a formal `class` with default values, constructors, etc.? That way whatever rules or conveniences we want to communicate, such as "specify this, not that", are all baked into that class?

src/database/queries/canonicalFields/fetchAll.sql
slifty: Probably not; I've gotten in the habit of it just from a "explicit is nice and consistency regardless of when you JOIN is nice too" perspective but I don't feel strongly and totally fine being more terse.

src/database/queries/canonicalFields/fetchAll.sql
slifty: Totally fine doing either -- which would you prefer?

src/types/CanonicalField.ts
```diff
@@ -0,0 +1,57 @@
...
+      type: 'string',
+    },
+    createdAt: {
+      type: 'object',
+      required: [],
+      instanceof: 'Date',
+    },
+  },
+  required: [
+    'id',
```
bickelj, line 42: This will be required in read-only mode but not required when creating one. Then again we might be able to defer that to another endpoint that accepts POST, but I would like to think about how to best resolve the two worlds in advance.

src/types/CanonicalField.ts
```diff
@@ -0,0 +1,57 @@
...
+    'label',
+    'shortCode',
+    'fieldType',
+    'createdAt',
+  ],
+};
+
+export const isCanonicalField = ajv.compile(canonicalFieldSchema);
```
bickelj, line 50: Again this is super nice.

src/handlers/canonicalFieldsHandlers.ts
slifty: I hear ya -- and ... it may well be overkill, but one other super nice thing that this is doing is typing `rows` for static analysis purposes.  After this line TypeScript knows that `rows` is a `CanonicalField[]`.  That doesn't mean much now, but it's possible some day we'll want to be able to validate that this handler does, in fact, retrun a `CanonicalField[]` outside of runtime.

src/types/CanonicalField.ts
slifty: Yes!  So in those cases I think we're going to want to use the TypeScript [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) which lets us define a full interface and then modify certain fields to be required / optional based on the context.

So we define the full type in one place, and then we define the subsets directly in the various handlers.

All of that said, I don't know *exactly* how the ajv / validation side of that pattern will look, but I do think that path at least means we don't have to define a bunch of sub-types in advance / before they're needed.

src/types/CanonicalField.ts
slifty: Yeah, I think in that sense it probably just shouldn't be listed as required in the "default" object.

src/__tests__/__snapshots__/canonicalFields.int.test.ts.snap
slifty: My initial reply was gonna be "this is the lazy option so you're probably right that I should do better and put it in the test directly"

BUT it also might be valuable to say "this is the actual document generated by the call"...

BUT it's probably more valuable to say "the call returned valid JSON that, when parsed, resulted in this expected object".

Which is a long way of saying that yeah I should probably do better and put it in the test directly ;)

src/database/migrations/0001_create_canonical_fields.sql
slifty: Ah right yes (and we did land on using INTEGER in the end, correct?)

src/database/queries/canonicalFields/fetchAll.sql
bickelj: I see that perspective too and it appeals to me. Not a big deal either way. But I am thinking if I write some SQL, though, I will find myself doing the `from canonical_fields cf`, so I suppose that's my preference, maybe more out of habit and/or laziness and/or trying to keep lines under 80 chars (though here there is no problem).

src/logger.ts
slifty: Yes it was the http logger

So we have our test runner --silence output by default (see the jest.config.js -- `silent: true`)

This can be disabled by passing `--no-silent` to the test call you want `npm run test:unit --no-silent`

The current behavior, though, breaks jest's `silent` option entirely because it bypasses the console output.  Either way we should fix that / make sure our logger plays well with our test runner.

As far as whether --silent should be default that's fair -- I guess the default setting of silent was predicated on the idea that noise on test success means it's harder to notice when tests are failing.  I don't want to see a bunch of HTTP request logs (which are NOT pino prettied btw!) covering up the test progress / stats / etc.

src/database/queries/canonicalFields/fetchAll.sql
bickelj: I guess `canonical_fields cf` but it's fine as-is too.

src/handlers/canonicalFieldsHandlers.ts
bickelj: Oh, yeah, that wasn't clear just from reading this snippet but that is useful and gives me warm fuzzies.

src/handlers/canonicalFieldsHandlers.ts
slifty: Brilliant -- I'll implement that pattern in this PR for sure (and take a stab at making it more copy-paste friendly)

src/database/queries/canonicalFields/fetchAll.sql
slifty: `from canonical_fields cf` it shall be :D

src/types/CanonicalField.ts
bickelj: Right, well in some contexts we want to require the id (reading) and in other contexts we want to require the id to not be set (writing), so it's not exactly optional, either. My approach of the following (which I see your similar comments on) was my first attempt but probably fails to achieve what I'm describing anyway:
```TypeScript
export interface OpportunityMinimal {
  title: string;
}

export interface Opportunity extends OpportunityMinimal {
  id: bigint;
  created_at: Date;
}

// export type Opportunity = OpportunityMinimal | OpportunityMaximal;
```
Actually, my first attempt had that second thing being `OpportunityMaximal extends OpportunityMinimal` and then the third thing being the union. That might have actually described what I wanted but the linter said "no dice, bad style."

src/database/migrations/0001_create_canonical_fields.sql
bickelj: Sigh. Yes, I suppose, considering the strangeness/hackiness required for 64-bit integers in JavaScript. Packing/type-alignment would be nice too, but with the 8-byte timestamp and a 4-byte id it seems weird to have the created_at precede the id when looking at the tables.

src/types/CanonicalField.ts
slifty: I hear ya.

Basically I think the root answer is "because of json".

More directly: generic object interface shape can be serialized and deserialized, and many of our tools are going to be working in terms of generic objects (e.g. express will provide our handlers generic objects parsed from json, the database will be providing generic objects parsed from sql results, etc.)

That means an Interface is not just more versatile, but is actually necessary regardless of whether or not we are going to also define a class.  Which is to say, we'd need the interface *regardless* in order to validate unknown data, but now we would also need to pass that type-validated data into a class constructor (and maintain a class in addition to the interface and JSONSchema).

More generally: TypeScript has moved away from classes and into more "functions with typed objects" and there are lots of other reasons for that (tree shaking is one -- if you provide an SDK and someone only needs the `createCanonicalField` method then a tree shaker can completely ignore all the other "canonical field" functions if those are defined as individual functions; if it's a class with a `delete` `update` etc method then the bundler can't selectively ditch those class methods just because they aren't used.

src/__tests__/__snapshots__/canonicalFields.int.test.ts.snap
bickelj: I guess if I wanted literal JSON in the test document, it would have to be multi-line (do backticks work for this?), and then in order to assert equality, we would not want actual equality but to parse the expected doc, parse the actual doc, and compare those things for value equality. If that's too awkward, this is good. Regardless, not a show-stopper, you actually have a test!!

src/types/CanonicalField.ts
slifty: I think that the right solution here is probably to only require things in the interface that are truly required, that way there's no need for a default object at all, since a default would never make sense (either the field needs to be set, or the field can be omitted).

I admit I am in the process of growing out of a bias against nulls (e.g. I have tended to prefer empty strings to nulls in almost all cases, since you can avoid having to worry about checking for null before use).

---------

Bottom line -- I think I'll take this out!

src/logger.ts
bickelj: I completely agree that we will need to better integrate these and make them play well together and agree with one another, etc. It is kind of annoying that I need to lose my jest coloration or avoid my pino-pretty printing or have one silent and another loud. In other words, yes, it needs to be better.

src/types/CanonicalField.ts
bickelj: Right, my suggestion was to keep the interface but then have a class implement that interface as a convenience rather than the default object as a convenience.

We are still left with trying to communicate the rules of optional/required/must-not-have via the interfaces, so it's probably six one way and half a dozen the other.

I had to look up tree shaking. At this moment my best summary is "it seems better to have relatively dumb data holders and independent functions that can operate on dumb data than to mash a bunch of data and behavior together into one all-or-nothing-package", e.g. a more functional style, less object-oriented?

bickelj:
Overall this is tremendous.

bickelj:
I am able to run it with `npm run start:dev` fine (separately the tests are giving me grief) and get a correct 500 error with well-formatted message from `GET /canonicalFields` (because I have a different db structure at the moment).

src/handlers/canonicalFieldsHandlers.ts
bickelj: My example in another ticket of logging each method entry is far too much to be pushed to main, though, that is something I would only do locally.

src/types/CanonicalField.ts
slifty: > in some contexts we want to require the id (reading) and in other contexts we want to require the id to not be set

Yes!  So what I'm saying is that *in those contexts* we would use the utility types to create context specific types.

It might be that we want to create our own utility class For example: https://bobbyhadz.com/blog/typescript-make-single-property-optional

Basically instead of defining several versions of the type types outside of the specific context we would do something like this *when defining the parameter*

for instance....

```
import { MakeRequired } from './types
interface CanonicalFieldWithRequiredId = MakeRequired<CanonicalField, 'id'>

function updateCanonicalField(
  canonicalField: CanonicalFieldWithRequiredId
)
```

I think that probably the core types should be based on the schema -- so if something is nullable in DB it is nullable in the type, but anything schema-required is required in type (and in create contexts we make it nullable via a utility type).

So: if a given endpoint doesn't require a field then we would define a new type just for that particular endpoint.

src/__tests__/__snapshots__/canonicalFields.int.test.ts.snap
slifty: Ahh got ya; so snapshots are NOT manually written -- they're generated by jest the first time you run the test (we don't control the format).  It's then on us to look at what was written and make sure it reflects what we're expecting.

> we would not want actual equality

Yeah that's what I intended FWIW -- jest equality checks are value equality not memory equality (if you want literal "is" you would say `expect(blah).toBe(blah)`).

---

I'm gonna try moving away from snapshot and see how it feels.

src/types/CanonicalField.ts
bickelj: OK very nice. Thanks for your patience and the link, and the explanation over in https://github.com/PhilanthropyDataCommons/service/pull/49#discussion_r927007792 too. This helped, and the mention of `Omit` specifically resolves the thing I was stuck on (but was too distracted to look up in the docs that this was possible).

So tying it back to my example, the suggestion is to maximally define `Opportunity` (kind of like my `OpportunityMaximal` and then subtract as-needed in the context needed using utility types. My attempt at understanding it more clearly:

```TypeScript
interface Opportunity {
    id: number;
    title: string;
    createdAt: Date;
}

const oppFull: Opportunity = {
    id: 1,
    title: 'chicken',
    createdAt: new Date(),
}

console.log(`${oppFull}`);

function printOpportunity(minOpp: Omit<Opportunity, 'createdAt' | 'id'>) {
    console.log(`minOpp: ${minOpp}`);

    // Property does not exist, cool:
    // console.log(`minOpp.id: ${minOpp.id}`);
    console.log(`minOpp.title: ${minOpp.title}`);

    // Property createdAt does not exist, cool:
    //console.log(`minOpp.createdAt: ${minOpp.createdAt}`);
}

printOpportunity(oppFull);
```

With utility types it looks like we could do it in a subtractive way (kind of like milling) or additive way (kind of like 3D printing) but I think I see your point now. We pick one definition, minimal or maximal as "the thing in general", and then we use the utility types in context. I think I got it, yes?

bickelj: We might even be able to genericize it somewhat with like `Omit<T, 'createdAt' | 'id'>` or something where T is generic type. Idk.

bickelj: Yes, that seems to work fine, actually:
```TypeScript

interface CanonicalField {
    id: number;
    description: string;
    createdAt: Date;
}

type Minimal<T> = Omit<T, 'createdAt' | 'id'>;

function printMinimalCanonicalField(minField: Minimal<CanonicalField>) {
    console.log(`minField.description: ${minField.description}`);
}

function printMinimalOpportunity(minOpp: Minimal<Opportunity>) {
    console.log(`minField.description: ${minOpp.title}`);
}

```

slifty: I *think* that at this point this can be marked as resolved -- id is required at the schema level and we can `Omit` it when it's not needed in a given context.  If that's not an accurate takeaway please un-resolve!

src/database/migrations/0001_create_canonical_fields.sql
slifty: If it makes you feel any better I think future versions of EMCAScript have proposals for bigint being built in

src/__tests__/__snapshots__/canonicalFields.int.test.ts.snap
bickelj: It is useful to have both approaches, I think, both snapshot and manual. When looking at failed tests that were from snapshots/benchmarks I will be more suspicious of those (than the manually-created expectations) because it is more likely (than the manual ones) to have a bad original test. I suppose what I'm saying is manual > snapshot > no tests.

src/logger.ts
slifty: I know you have dug into this on another issue #59 so I'm gonna mark this thread as resolved

src/handlers/canonicalFieldsHandlers.ts
slifty: Attempt has been made -- looking forward to any feedback! (marking this as resolved in the mean time)

slifty:
@bickelj this is ready for re-review.  I believe I addressed all concerns; the most key changes are:

1. Tests no longer use snapshot.

2. I added a change to the way we access the logger, so that a file MUST pass a `name` for the logger (and then we spawn a child with that name specified).  I think that will help us be consistent with our logger file access (I wish there was a way to auto-detect the name of the calling file... but [it looks like that is not really supported](https://stackoverflow.com/questions/13227489/how-can-one-get-the-file-path-of-the-caller-function-in-node-js)).

3. I took out the `default` object.  I agree with your take about that pattern probably not being the right / expected one, and it isn't even needed at this stage.

src/database/migrations/0001_create_canonical_fields.sql
bickelj: I guess ECMAScript has bigint but not necessarily JSON, does that sound right?

slifty:
I noticed I had an untested path in the handler; I've added an extra test to handle that case.  Our global `afterEach` now resets all mocks between tests (which I believe is a good practice anyway?).

bickelj:
To start with, I was again able to migrate and run e4c36fd8a3d6d88b2e0f70fa118fba0c93cf5230 locally, tests are still giving me grief though. The use of `ajv` and the new `getLogger` function are awesome! I'll add more in-band and specific comments too.

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,14 +1,55 @@
...
 const agent = request.agent(app);
 
 describe('/canonicalFields', () => {
   describe('/', () => {
-    it('should return HTTP Status Code 200 OK', async () => {
+    it('returns an empty array when no data is present', async () => {
       await agent
         .get('/canonicalFields')
-        .expect(200);
+        .expect(200, []);
```
bickelj, line 13: Out of curiosity, I did not notice a difference between chaining the `expect` call like `.expect(...` versus having it as a separate statement `expect(...`. Is there any benefit or drawback to either?

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,14 +1,55 @@
...
+    });
+
+    it('returns all canonical fields present in the database', async () => {
+      await db.sql('fixtures.canonicalFields.returnsAllCanonicalFields.seed');
```
bickelj, line 17: It looks like this path is in the production code area, not the test code area:
```shell
$ find . -name "*seed*"
./src/database/queries/fixtures/canonicalFields/returnsAllCanonicalFields/seed.sql
./dist/database/queries/fixtures/canonicalFields/returnsAllCanonicalFields/seed.sql
```
I wonder if it is as easy to go into `src/test/database/fixtures` or something like that. I think `src/test/database/queries/fixtures/canonicalFields/returnsAllCanonicalFields` is too deep and long, but this comment is more a matter of style than substance.

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,14 +1,55 @@
...
+            {
+              createdAt: '2022-07-20T12:00:00.000Z',
+              dataType: 'string',
+              id: 2,
+              label: 'Last Name',
+              shortCode: 'lastName',
+            },
+          ],
+        );
+    });
```
bickelj, line 39: I like this more than the snap thing. It is so clear and simple! There is still the one level of indirection above to find the SQL (as opposed to seeing the test setup SQL directly in the test) but I think that is fine because it is consistent with the way we use the SQL files in the production code, etc., plus I don't think there are multi-line string literals in JavaScript yet.

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,14 +1,55 @@
...
+      jest.spyOn(db, 'sql')
+        .mockImplementationOnce(async () => ({
+          rows: [{ foo: 'not a valid result' }],
+        }) as Result<object>);
+      const result = await agent
+        .get('/canonicalFields')
+        .expect(500);
+      expect(result.body).toMatchObject({
+        name: 'ValidationError',
+        errors: expect.any(Array) as unknown[],
```
bickelj, line 51: This is excellent!

src/app.ts
```diff
@@ -1,8 +1,9 @@
 import express from 'express';
 import pinoHttp from 'pino-http';
 import { rootRouter } from './routers';
-import { logger } from './logger';
+import { getLogger } from './logger';
 
+const logger = getLogger(__filename);
```
bickelj, line 6: The `getLogger` function and its use is also excellent and so much better than having to copy/paste and then remember to modify!
* `getLogger` is a more targeted import (one thing instead of two).
* The logic for which key to use is in one place (`logger.ts`) instead of spread throughout many files.
* The use of `__filename` is safely copied and pasted.
* The `__filename` tells us exactly what we want to know: "where did this message come from!?"
* The choice of whether to use a child logger or recycle another logger is also in one place (`logger.ts`) instead of spread throughout many files.

I like this so much more.

src/database/migrations/0001_create_canonical_fields.sql
```diff
@@ -0,0 +1,13 @@
...
+  label VARCHAR NOT NULL,
+  short_code VARCHAR NOT NULL,
+  data_type VARCHAR NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+comment on table canonical_fields is
+  'Canonical fields are those fields that are unique across organizations.';
+comment on column canonical_fields.label is 'A word or phrase that communicates the meaning of the field for end users';
+comment on column canonical_fields.short_code is 'An externally meaningful alphanumeric code tht can be used to reference this field without knowing the internal id';
```
bickelj, line 12: The comment on `short_code` helps me a lot. But then it raises a question: can we make `short_code` unique? Why or why not? You probably answered this somewhere already but I am still not sure I grasped it.

src/database/migrations/0001_create_canonical_fields.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE canonical_fields (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  label VARCHAR NOT NULL,
```
bickelj, line 3: Thanks for humoring me on `varchar` : )

src/database/queries/canonicalFields/fetchAll.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+  cf.id as "id",
+  cf.label as "label",
+  cf.short_code as "shortCode",
+  cf.data_type as "dataType",
+  cf.created_at as "createdAt"
+FROM canonical_fields as cf;
```
bickelj, line 7: Thanks for humoring me here too with the terser table alias. It looks nicer to me, at least!

src/errors/ValidationError.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  public constructor(
+    message?: string,
+    errors?: ErrorObject[] | null,
+  ) {
+    super(message);
+    this.name = this.constructor.name;
+    this.errors = errors;
+  }
+}
```
bickelj, line 16: This `ValidationError` class is great. For a second I wondered if `ErrorObject` also extends `Error` but no, it does not. My tendency would be to make as many fields mandatory and immutable only as possible for this commit and then relax them to optional as needed in future commits. Perhaps that already happened in the course of TV Kitchen, though, and so making them optional would be inevitable here anyway.

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -1,13 +1,37 @@
...
+          .contentType('application/json')
+          .send(rows);
+      } else {
+        throw new ValidationError(
+          'The database responded with an unexpected format.',
+          isCanonicalFieldArray.errors,
+        );
+      }
+    })
+    .catch((error: unknown) => {
```
bickelj, line 28: I wonder what the default behavior of express is here and if we can define this behavior once in some other, higher-level code that we hook into express.

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -1,13 +1,37 @@
...
+      logger.error(error);
```
bickelj, line 29: With the five levels provided in `slf4j` (the ones I'm most familiar with), `error` is the most severe, and so I would usually reserve that for situations where the process is either dying or about to die, and use `warn` in this kind of situation. But I think we have another more severe level `fatal` in `pino` so I guess we could use that one in that case. In other words: it's OK but I had a moment of nitpickiness that I thought I'd share.

src/logger.ts
```diff
@@ -1,5 +1,10 @@
...
+  Logger,
+} from 'pino';
 
-const logger = pino();
+const logger = pino({
+  level: process.env.LOG_LEVEL ?? 'silent',
+});
 
-export { logger };
+export const getLogger = (name: string): Logger => logger.child({ name });
```
bickelj, line 10: I am not sure if there will be unintended effects of using `name` rather than another key (e.g. does it break apart or flatten the hierarchy of loggers completely?) but the best part of this change is that this kind of choice is made right here in the single `src/logger.ts` so we can easily change the implementation if new information comes to light.

The logging implementations in the Java world typically have some kind of logger hierarchy so that you can amplify this module and silence that module and so forth in a targeted way. I am not sure how to do that with pino yet but I think I read something in their issues or docs about the `name` being special or related to that capability.

src/test/integrationSuiteSetup.ts
```diff
@@ -18,4 +18,5 @@ beforeEach(async () => {
 
 afterEach(async () => {
   await cleanupDatabaseForCurrentWorker();
+  jest.restoreAllMocks();
```
bickelj, line 21: I agree that mocks should be isolated to the test, but I would have thought jest should do it by default instead of needing the explicit call. Hmm.

src/types/CanonicalField.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const canonicalFieldArraySchema: JSONSchemaType<CanonicalField[]> = {
+  type: 'array',
+  items: canonicalFieldSchema,
+};
+
+export const isCanonicalFieldArray = ajv.compile(canonicalFieldArraySchema);
```
bickelj, line 49: This looks great and I am comfortable with the more targeted utility types being used in the context where they are needed. I guess I still have a feeling that there will be a relatively common shape for exactly one other type per entity. We can cross that bridge when we come to it, though, if it comes.

bickelj:
I think it can be merged as-is, but it would be nice to rebase onto main. I see you used `LOG_LEVEL` and I used `PINO_LOG_LEVEL` in another PR. I think `LOG_LEVEL` as you have it here makes sense instead of naming the particular implementation. My only other hesitation is the test code being in the prod code and whether we can add a unique constraint to the short code.

src/database/migrations/0001_create_canonical_fields.sql
slifty: Oh snap I actually meant to add that -- yes, we should.  I'll include that in this migration

src/database/migrations/0001_create_canonical_fields.sql
slifty: Ha not at all!  Thank you for pointing out the benefits of avoiding TEXT

src/database/queries/canonicalFields/fetchAll.sql
slifty: I agree!

src/errors/ValidationError.ts
slifty: That's a good thought and I agree (I forget if the TVK had cases where it needed to be optional but it's so easy to add it later if we find one)

src/handlers/canonicalFieldsHandlers.ts
slifty: Word (appreciate the thinking / sharing).

Argh, I used to have a really good writeup of log levels but I can't find it any more.  The closest I could find was this documentation at [Winston](https://github.com/winstonjs/winston#logging-levels) which is in turn driven by this document: https://datatracker.ietf.org/doc/html/rfc5424

```
              0       Emergency: system is unusable
              1       Alert: action must be taken immediately
              2       Critical: critical conditions
              3       Error: error conditions
              4       Warning: warning conditions
              5       Notice: normal but significant condition
              6       Informational: informational messages
              7       Debug: debug-level messages
```

So basically I figured this is an "error condition" since it's something that actually did go wrong.

I think the difference between `fatal` and `error` is whether the system is able to recover from it.

That said... those are 8 levels and we don't have all 8 in pino, and `error` is peak of npm...

Dang it I really wanna find the writeup since this would be good to be consistent on so it's worth spending a moment on alignment.

slifty: Ah this is the one!

https://stackoverflow.com/a/2031209/159522

```
    Trace - Only when I would be "tracing" the code and trying to find one part of a function specifically.
    Debug - Information that is diagnostically helpful to people more than just developers (IT, sysadmins, etc.).
    Info - Generally useful information to log (service start/stop, configuration assumptions, etc). Info I want to always have available but usually don't care about under normal circumstances. This is my out-of-the-box config level.
    Warn - Anything that can potentially cause application oddities, but for which I am automatically recovering. (Such as switching from a primary to backup server, retrying an operation, missing secondary data, etc.)
    Error - Any error which is fatal to the operation, but not the service or application (can't open a required file, missing data, etc.). These errors will force user (administrator, or direct user) intervention. These are usually reserved (in my apps) for incorrect connection strings, missing services, etc.
    Fatal - Any error that is forcing a shutdown of the service or application to prevent data loss (or further data loss). I reserve these only for the most heinous errors and situations where there is guaranteed to have been data corruption or loss.
```

I think if we subscribe to this approach then this would be fatal to the immediate operation (e.g. the fields are not returned so the GET operation has failed).

src/logger.ts
slifty: That's a great point; `name` is a special option in pino but I don't know if it does anything in the context of children (`ChildOptions` doesn't have a name field for instance).  If our intention is that this is the source file then I think your original `source` field is a fine term.

I'll switch to `source`

slifty:
Yes -- in fact, I'm going to update our main branch protections to require PRs to be up to date before allowing merge.

slifty:
@bickelj the local tests may be giving you trouble if you're running them against a database that has an existing schema where migrations have been run ( this is due to https://github.com/ThomWright/postgres-migrations/pull/93 not being merged yet )

If you set up `.env.test` and point to a DB that has nothing inside does it work?

src/test/integrationSuiteSetup.ts
slifty: Hmm indeed :(

src/handlers/canonicalFieldsHandlers.ts
slifty: (I'll mark this as resolved for now but we can pick this up in an issue for sure, and should probably plan to document the thinking)

src/handlers/canonicalFieldsHandlers.ts
slifty: I was curious about this question!

So it looks like there is default middleware that will handle and process errors: https://expressjs.com/en/guide/error-handling.html

Given that I think I should ditch this try / catch.

We might consider writing our own middleware some day, but at the very least I think it's clear that it's better to let errors bubble up to the middleware.

slifty: Oops, I'm sort of wrong -- so that is all true for errors thrown, but failed promises are a different situation (promises that fail do so outside of the stack and need to be handled directly).

This means that we DO need a `.catch` but what might be better is to pass the error to `next` so the error is given to that earlier mentioned middleware.

I'm gonna switch over to the `pass error to next` approach

slifty: Final update... sorry for the noise.

It turns out the express middleware, unfortunately, does not send the error that was thrown.  Given that I think maybe the current approach is "right" for this PR, but I'm going to make an issue to think more deeply about how we want our error handling to work.

slifty: marking this as closed now that we have https://github.com/PhilanthropyDataCommons/service/issues/61

src/__tests__/canonicalFields.int.test.ts
slifty: You called out the one thing that was still bugging me about this...

Side note: you can have multiline string literals in JS using backticks:

```
export const test = `
asdf
asdf
asdf
`;
```

I also don't like at all that we're muddying up our `queries` directory with test sql.  I would somewhat expect there to be fixtures for this kind of thing (especially if we consider the idea that a test's migrations might get long and complex), but... I'm gonna see what inline SQL looks like for this just for kicks.

src/__tests__/canonicalFields.int.test.ts
slifty: So actually it's just running the assertion via two different libraries!

`.expect()` is using the functionality of `supertest` -- which has a pattern of `expect(status code, body)`

`expect()` is using the functionality of `jest` -- which is a whole thing that can't be summed up in one line.

Basically my approach was that whenever I can verify the result within supertest I stuck with that, but when dealing with more complex expectations (e.g. I want to match against a shape of an object rather than a literal value) I had to move to jest.

src/__tests__/canonicalFields.int.test.ts
slifty: ok you've convinced me ;) -- I'm bailing on the tinypg-stored-procedure-whatever-you-wanna-call-it for test fixtures.  Too cheugy.

src/__tests__/canonicalFields.int.test.ts
slifty: Moved it to inline SQL; it's the way to go.  Also if in future we ever feel the need to move a particularly long / complex DB setup to a `fixture` file, that fixture should be written alongside the test file rather than halfway across the world in the production `database/queries` directory.

slifty:
@bickelj another round!  (sorry I know you keep approving and then I keep making all the optional changes -- please feel free to flag more improvements!)

src/__tests__/canonicalFields.int.test.ts
bickelj: Nice, this makes sense.

src/__tests__/canonicalFields.int.test.ts
bickelj: Very, very nice!!

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -1,14 +1,65 @@
...
+        INSERT INTO canonical_fields (
+          label,
+          short_code,
+          data_type,
+          created_at
+        )
+        VALUES
+          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+      `);
```
bickelj, line 27: :heart:

src/database/migrations/0001_create_canonical_fields.sql
bickelj: :+1:

src/test/integrationSuiteSetup.ts
bickelj: Good enough for now.

bickelj (approved):
I assumed the `.keep` files were for the purpose of keeping directories present in git (no such thing as a directory in git), but I also see `    "build:assets": "copyfiles -u 1 \"src/**/*.sql\" \"src/**/.keep\" dist",`, but I think it is OK because those will not be copied over if they do not exist.

Thanks for iterating on this one with me a few times, I think this is huge. :ok_hand:

slifty:
> I assumed the .keep files were for the purpose of keeping directories present in git (no such thing as a directory in git), 

It served both purposes!  We needed the `migrations` and `queries` directories in prod to prevent the db code from erroring, so copying `.keep` helped there too.  We can take that out in future (I'll merge this as is though)

src/handlers/canonicalFieldsHandlers.ts
bickelj: In general, I am OK with those descriptions. In particular, I most agree with the `info` and `warn` levels there: info is normal operation messages while `warn` is odd but recoverable. As for the descriptions of debug, trace, error, and fatal, there's wiggle room and a lot of subjectivity (and I am not sure what the poster means by "data loss" either) but again I think it's good overall, especially if we collapse the descriptions of trace/debug and error/fatal. In this situation, I think `error` sounds fine to me and if express is about to keel over and die we can use `fatal` in that situation.

bickelj:
> @bickelj the local tests may be giving you trouble if you're running them against a database that has an existing schema where migrations have been run ( this is due to [ThomWright/postgres-migrations#93](https://github.com/ThomWright/postgres-migrations/pull/93) not being merged yet )
> 
> If you set up `.env.test` and point to a DB that has nothing inside does it work?

That works.

src/handlers/canonicalFieldsHandlers.ts
bickelj: Nice, yes. I think the `json` response even on 500 is nice and it doesn't look like that would happen with the default handler anyway. I also think logging the error is nice too. So yes, #61 is the perfect place to pursue something like this!

Issue #56, opened 2022-07-22
Do DB migration on startup


bickelj:
We probably want the database migration to happen on startup rather than manually.

PR #57, opened 2022-07-22
Migrate the database schema on container startup

bickelj:
When doing two or more commands on startup, it is straightforward to use
a small script. The downside is the command is not visible directly in
the Dockerfile, there is one level of indirection. But we do want to
do `npm run migrate` then `npm run start`. The migrate task should be
idempotent and therefore harmless to run on repeated startup even with
the same version of the software (e.g. no new schema changes).

Issue #56 Do DB migration on startup

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -e
+
+npm run migrate
+exec npm run start
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,5 +10,6 @@ WORKDIR /opt/philanthropy-data-commons/server
 COPY --chown=web:web package.json .
 COPY --chown=web:web node_modules ./node_modules
 COPY --chown=web:web dist ./dist
+COPY docker-entrypoint.sh .
 
-CMD ["npm", "start"]
+CMD ["./docker-entrypoint.sh"]


docker-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+
+npm run migrate
+exec npm run start
```
slifty, line 5: I imagine the use of `exec` here is serves an important purpose but figured it couldn't hurt to ask!

slifty (approved):
Looks good -- had a question about `exec` but figure that's there for a reason (if it isn't needed then maybe take it out before merging!  if it is then carry forth!)

docker-entrypoint.sh
bickelj: From what I have read it helps with signal handling in docker. I guess it drops the shell and lets npm be pid 1 within the container. So then whatever signals docker sends on `stop`, `kill`, etc., will go not to the parent bash process but to the intended `npm run start` process. Something like that. I haven't experimented with it to know more deeply but this is probably close.

bickelj: I had hoped I could ctrl-c to send sigint to the process when running with plain old `docker run ... ` but that doesn't appear to work yet either way. So maybe there is more improvement possible here. It seems to be OK under compose but I also have not looked very closely.

Issue #53, opened 2022-07-21
Cause jest integration tests to play nicely with tinypg/pg.pool on connection errors


bickelj:
Expected behavior:

An error (such as failure to connect) during an integration test causes that test to fail shortly after the error, whether that occurs in a `before` section of the test or the test itself.

Actual behavior:

Timeouts occur as well as "open handles potentially keeping Jest form exiting."

Example debug logs from a draft `opportunities.int.test.ts`:
```
$ npm run test:integration | node node_modules/pino-pretty/bin.js

> @pdc/service@0.0.0 test:integration
> jest --config=jest.config.int.js --detectOpenHandles --runInBand

[2022-07-21T17:23:59.145Z] DEBUG (88928 on local): About to set/call afterAll with a function that will close db.
    source: "test/integrationSuiteSetup"
[2022-07-21T17:23:59.152Z] DEBUG (88928 on local): Finished set/call afterAll with a function that will close db.
    source: "test/integrationSuiteSetup"
[2022-07-21T17:23:59.672Z] DEBUG (88928 on local): About to define function getCanonicalFields
    source: "handlers/canonicalFieldsHandlers"
[2022-07-21T17:23:59.675Z] DEBUG (88928 on local): About to define function getOpportunities
    source: "handlers/opportunitiesHandlers"
[2022-07-21T17:23:59.680Z] DEBUG (88928 on local): About to define the createSchema function
    source: "test/harnessFunctions"
[2022-07-21T17:23:59.680Z] DEBUG (88928 on local): About to define the setSchema function
    source: "test/harnessFunctions"
[2022-07-21T17:23:59.680Z] DEBUG (88928 on local): About to define the dropSchema function
    source: "test/harnessFunctions"
[2022-07-21T17:23:59.680Z] DEBUG (88928 on local): About to define the prepareDatabaseForCurrentWorker function
    source: "test/harnessFunctions"
[2022-07-21T17:23:59.680Z] DEBUG (88928 on local): About to define the cleanupDatabaseForCurrentWorker function
    source: "test/harnessFunctions"
[2022-07-21T17:23:59.681Z] DEBUG (88928 on local): About to describe /opportunities
    source: ".__tests__/opportunities.int.test"
[2022-07-21T17:23:59.681Z] DEBUG (88928 on local): About to call/set beforeEach in /opportunities
    source: ".__tests__/opportunities.int.test"
[...]

PR #58, opened 2022-07-25
Release the db client on migrate every time

bickelj:
This seems to help integration tests finish properly when there are
failures related to database setup or tear-down.

Issue #53  Cause jest integration tests to play nicely with tinypg


slifty (approved):
Oh great catch -- thank you for figuring this out.

Issue #59, opened 2022-07-26
Get pino and jest to agree on console verbosity


bickelj:
See comment in PR #55 at https://github.com/PhilanthropyDataCommons/service/pull/55#discussion_r927000538

An example possibility is at https://github.com/chriszs/tv-kitchen/commit/ff61ea8611d092740e38a020b06d7aff4cd979f1

bickelj:
It seems backwards to have the production code look to see if it's in a test environment first and then change the level based on that. Perhaps there is a way to have the test environment set the level instead.

bickelj:
Here's an approach:
1. Drive the log level with env var `PINO_LOG_LEVEL` or `info` if it is not set.
2. In the jest configuration, when --silent is set and `PINO_LOG_LEVEL` is not set, set `PINO_LOG_LEVEL` to `silent`.

I think the above will mean:
* When `PINO_LOG_LEVEL` is set in prod or test, that level takes effect.
* When `PINO_LOG_LEVEL` is not set in prod, `info` takes effect.
* When `PINO_LOG_LEVEL` is not set in test, and also --silent is set, then the test harness/framework will set `PINO_LOG_LEVEL` to `silent` and it will therefore take effect.

bickelj:
The problem is finding where in jest we can hook in to have access to the resolved/merged configuration object. I have yet to find it in the docs.

bickelj:
The closest thing I can find is our existing `setupEnv.ts` which is akin to the above linked file listed in `setupFiles`, but then there is the problem of importing the logger after detecting whether jest has set the console to silent. I think the root issue again is figuring out how to programmatically read the `jest` configuration after it has been resolved. I cannot see any way to do it despite other folks wondering the same. No answers at https://stackoverflow.com/questions/65698821/how-to-read-jest-config-values-within-test and I cannot find anything in the docs or source code indicating they would be exposed. That doesn't mean it isn't there, though.

bickelj:
Perhaps it is the `globalSetup` that I seek: https://jestjs.io/docs/configuration#globalsetup-string

bickelj:
The use of `globalSetup` seems to work.

bickelj:
Unless I am missing something obvious (I probably am), I can choose between the linter complaining and jest complaining.

With `module.exports = function ...` I get this from the linter (but jest complaineth not):

[...]

PR #60, opened 2022-07-26
Use env var `LOG_LEVEL` to set pino log level

bickelj:
The environment variable `LOG_LEVEL` will set the level on the root
logger when it is set. When running tests under jest, the globalSetup
examines the (jest) global and project configurations as well as the env
var `LOG_LEVEL` to decide whether to silence pino using the same env var
`LOG_LEVEL`. If you set `LOG_LEVEL`, that level takes precedence in all
environments. If you do not set `LOG_LEVEL` and the jest config sets
`silent: true`, the globalSetup for jest sets `LOG_LEVEL` to `silent` on
your behalf, essentially passing jest's `silent: true` setting through.

There are multiple ways to set `LOG_LEVEL` such as on the command line
or in a `.env` file.

Issue #59 Get pino and jest to agree on console verbosity

### DIFF
--- a/src/test/globalSetup.ts
+++ b/src/test/globalSetup.ts
@@ -1,0 +1,22 @@
+// Jest expects a single default function to be exported from this file.
+/* eslint-disable import/no-default-export */
+import dotenv from 'dotenv';
+import type { Config } from 'jest';
+
+// TODO: reconsider use of '.env.test' if or when
+// https://github.com/ThomWright/postgres-migrations/pull/93 gets merged/fixed.
+dotenv.config({ path: '.env.test' });
+
+// Production code will run dotenv.config() anyway, so make clear that it
+// happens by running the same explicitly here. If '.env.test' has any variables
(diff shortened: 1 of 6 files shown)


src/test/globalSetup.ts
```diff
@@ -0,0 +1,11 @@
+// Jest expects a single default function to be exported from this file.
+/* eslint-disable import/no-default-export */
+import type { Config } from 'jest';
+
+export default (globalConfig: Config, projectConfig: Config): void => {
```
slifty, line 5: This is a brilliant solution 👨‍🍳💋 (<-- that's supposed to mean "chefkiss")

Also I don't know if you use StackOverflow but if you do and you want internet points you have found a great answer to: https://stackoverflow.com/questions/73058368/is-it-possible-to-check-if-jest-silent-is-enabled-in-runtime

slifty (approved):
So good, thank you for solving this

PR #63, opened 2022-08-01
Draft of `GET /opportunities` implementation

bickelj:
This is a non-linting mostly-working draft of GET /opportunities. Everything about it is a draft including the assumptions it makes about the schema, its use, whether we will include application schemas with opportunities, etc.


src/handlers/opportunitiesHandlers.ts
```diff
@@ -0,0 +1,134 @@
...
+import type { ApplicationSchemaField } from '../types/ApplicationSchemaField';
+
+const logger = getLogger(__filename);
+
+const getOpportunities = (req: Request, res: Response): void => {
+  db.sql('opportunities.fetchAll')
+    .then((opportunitiesRaw) => {
+      logger.debug(opportunitiesRaw, 'opportunities.fetchAll');
+      type ShortOpportunity = Omit<Opportunity, 'applicationSchema'>;
+      type ShortApplicationSchema = Omit<ApplicationSchema, 'fields'>;
```
bickelj, line 23: Is there a better way than to make the type alias here? Maybe just repeat `Omit<ApplicationSchema, 'fields'>` several times?

src/types/ApplicationSchema.ts
```diff
@@ -0,0 +1,54 @@
...
+  // will be the one retrieved for a given opportunity while the schema used by
+  // a particular application will be related to the schema used at the time.
+  // Nevertheless, we will declare the maximal fields here as a rule and where
+  // it is not needed a utility type can be defined instead, subtracting it.
+  version: number;
+  fields: ApplicationSchemaField[];
+  createdAt: Date;
+}
+
+export const applicationSchemaSchema: JSONSchemaType<ApplicationSchema> = {
```
slifty, line 23: Oh no how awkward -- maybe we should call `applicationSchema` something different at the db level...  (We don't have to feel locked into the ERD; I think it makes sense to avoid developer terms of art in our table names)

src/handlers/opportunitiesHandlers.ts
slifty: If you aren't re-using these names I don't think I'd use named types here (as opposed to just using the `Omit` in the relevant line).  Basically I think the name `Short___` is actually less meaningful than the raw `Omit` stuff.

That said I'm still in the process of grokking what's going on overall so maybe I'll have a better answer in a moment!

src/database/queries/opportunities/fetchAll.sql
```diff
@@ -0,0 +1,34 @@
+SELECT o.id AS "opportunityId",
```
slifty, line 1: I wonder if we might actually NOT want to select JOINs -- and instead leave it to the client.

At least in our first pass.

The client wanting a list of opportunities doesn't mean they necessarily care about the applications or canonical fields or schemas associated with those opportunities.  This is a lot of complexity (and DB work) that isn't needed for that use case.

With this approach we would start with a vanilla `GET /opportunity` that returns the flat opportunity list and a `GET /application?opportunityId=...` for the use case of "give me all applications for a given opportunity"

The downside of this is MAYBE more HTTP calls depending on the use case -- but at this point i'm not even usre we'll have use cases that benefit from a fully populated relationship map for Opportunity!

I think this approach might obviate some of the other concerns you were looking for feedback and dramatically simplify this endpoint.

What do you think?

src/types/ApplicationSchema.ts
bickelj: Application Form? Application Fields? Application List? Application Items?

slifty: I think Application Form seems like a good idea

src/database/queries/opportunities/fetchAll.sql
slifty: (Happy to talk through this by the way) -- just to underscore after re-reading this; I think this just-the-object-being-requested approach will simplify the other API endpoint too / we'll have a better shot at hitting our demo goal.

I also think it's expected in a RESTful api that I'm only getting the object I'm asking for (unless again the API supports asking for children as well but I'm not surprised when that kind of behavior requires a flag of some kind).

bickelj: Yes, in my first pass I had it as a pretty dumb endpoint, basically a pass-through to the `opportunity` entity/relation. It is certainly not an optimization to add the complexity of bringing along the application schema. The reason the application schema comes along here with opportunity is to satisfy the minimum-viable product having only these endpoints:

* Implement POST /applicants endpoint
* Implement GET /applicants endpoint
* Implement POST /applications endpoint
* Implement GET /applications endpoint
* Implement POST /opportunities endpoint
* Implement GET /opportunities/{opportunityId} endpoint
* Implement GET /opportunities endpoint
* Implement GET /canonicalFields endpoint

Given these endpoints, and given the DB schema, in order to post applications we need the application schema, therefore the application schema must be added in or inferred from one of three places:
* POST /applicants (doesn't make much sense here)
* POST /applications (probably makes some sense here)
* POST /opportunities (probably makes the most sense here)

Until we develop explicit `POST` and `GET /applicationSchemas` or unless we push the complexity into the `/applications` endpoint, we need to let the application schema be posted as part of the opportunity.

Note that applications do not come back with the current implementation you are looking at, only the latest application schema for the opportunity. In other words, if I wanted to submit an application for an `opportunity`, the call required is first `GET /opportunities` and then `POST /applications` (assuming the applicant has also already been posted).

The API presented to the caller does not necessarily have to match 1-1 with the underlying database schema (although it might be more straightforward to follow the API if, for example, you have some familiarity with the database schema for whatever reason).

The implementation here leaves the door open for the future possibility of either posting an opportunity with its application schema in one post or posting first a simple opportunity and separately posting an application schema, but for a minimum viable product, the application schema has to come from somewhere, so for now it would be sent with `POST /opportunities` and read with `GET /opportunities`.

Does this help?

slifty: > having only these endpoints:

I'm not sure that's a requirement of the MVP right?  I think it'd be fine (maybe preferred / simpler?) to have GET and POST endpoints for `applicationForms`

slifty: I guess what I'm getting at is that I think it would be a lot easier to say "we have a baseline API is dumb and very fast to implement" and then "let's add this complexity in a future PR"

Just from an MVP perspective I think we'd hit the finish line that way, whereas right now I don't think this PR is ready to merge and it introduces a ton of questions about how we want to handle joins in our schema queries, etc.

bickelj: I don't recall a reason, I think I made the guess from the list of issues here in GitHub, but we didn't make any determination on which ones would be included or excluded. I fully support simplifying the API and including `applicationForms` because it appears the extra data mashing is pretty darn slow (at least my feeble first-pass implementation was).

src/handlers/opportunitiesHandlers.ts
slifty: Following up and seeing the lint error which is saying the same thing -- basically the best practice is this kind of type should not be named and should only be defined in context.

Here's a page that explains the rationale for the rule: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-type-alias.md

src/database/queries/opportunities/fetchAll.sql
bickelj: > I guess what I'm getting at is that I think it would be a lot easier to say "we have a baseline API is dumb and very fast to implement and users CAN access all the data they want" and then "let's add some convenience / richer data responses in a future PR"

Yes, fully on board.

> Just from an MVP perspective I think we'd hit the finish line that way, whereas right now I don't think this PR is ready to merge and it introduces a ton of questions about how we want to handle joins in our schema queries, etc.

Yes, you highlighted another downside is that the data validation is no longer as simple as "take a response from the db and validate it" but is "take a response from the db, do a bunch of mashing, then validate it." So yeah, if we do the simpler pattern, I think we will be better off in the long run, let the complexity be transparent to the caller (by transparent I mean visible here, not invisible as we CS people sometimes backwardsly say).

src/handlers/opportunitiesHandlers.ts
```diff
@@ -0,0 +1,134 @@
...
+      type ShortApplicationSchemaField = Omit<ApplicationSchemaField, 'canonicalField'>;
+
+      // Here is how each raw, de-normalized row fetched from the DB looks:
+      interface RowPerField {
+        opportunity: ShortOpportunity;
```
bickelj, line 28: Use #1

src/handlers/opportunitiesHandlers.ts
bickelj: I had to go back and look if I had used it more than once, and yes, I had used it twice below in the context, which was why I had created the alias. I read through https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-type-alias.md and maybe I didn't read it carefully enough but it did not seem to address the local definition of a utility type that would be repeatedly used within that local scope. Regardless, this question becomes moot (for now) when eliminating all the considerations of `ApplicationSchema` or `ApplicationForm` from `/opportunities`.

src/handlers/opportunitiesHandlers.ts
```diff
@@ -0,0 +1,134 @@
...
+        applicationSchema: ShortApplicationSchema;
+        applicationSchemaField: ShortApplicationSchemaField;
+        canonicalField: CanonicalField;
+      };
+
+      const { rows } = opportunitiesRaw;
+      const formalizeRow = (it: any): RowPerField => {
+        const opportunity: ShortOpportunity = {
```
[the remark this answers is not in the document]
slifty: I hear ya -- I *suspect* that this will be moot after refactor (which is not to say it's not a good discussion but maybe we can punt it for later / when it's definitely relevant.)

bickelj:
Closing without merge because this one bites off too much and assumes too much. See first draft from #49, the next draft will be more similar to #55 and a combination of some of the stuff from #49 and this one.

PR #66, opened 2022-08-02
Implement GET /opportunities endpoint

bickelj:
Opportunities simply have a title. To GET or POST the fields associated
with a given opportunity, those are separate calls to /applicationForms.

Issue #31 Implement GET /opportunities endpoint


kfogel:
I see I'm requested as a reviewer here, @bickelj.  I wish I could!  But I'm not hands-on enough in the technical work in this project, compared to you and @slifty, to be a useful reviewer on PRs (and, for other reasons, I can't be a timely one anyway).

kfogel (commented):
I see I'm requested as a reviewer here, @bickelj. I wish I could! But I'm not hands-on enough in the technical work in this project, compared to you and @slifty, to be a useful reviewer on PRs (and, for other reasons, I can't be a timely one anyway).

slifty (approved):
This took a while to review ONLY because I was on vacation!

It looks great.

Steps I took to test:

1. Ran migrations
2. Ran tests
3. Visited `localhost:3000/opportunities` and saw `[]`

slifty (changes requested):
Oh one path wasn't covered by tests -- I'm going to add in that test before marking approve (I think there's a pattern for it already in the opportunities tests so should be quick)

slifty (approved):
OK!  100% diff coverage now so I think we're good.

Issue #33, opened 2022-07-12
Implement `POST /opportunities` endpoint


slifty:
This will allow a third party to create a new opportunity in the system.  Note that authentication / authorization will be implemented separately from this issue.

bickelj:
I expected TypeScript and AJV to narrow the type of `req.body` here such that it knows it has one property `title`, but no dice yet:
```TypeScript
  if (!isOpportunity(req.body)) {
    res.status(400)
      .contentType('application/json')
      .send(isOpportunity.errors ?? { message: 'Expected JSON body having title.' });
    return;
  }

  const opportunityPosted: Omit<Opportunity, 'createdAt' | 'id'> = { title: req.body.title };
```

I am probably missing something obvious, though.

slifty:
Oh fascinating!  I hadn't thought about the "how does typescript handle type guards when run against attributes" question.  Yeah at a first glance I would have thought that to.

That being said I haven't crossed this path because destructuring objects that you plan on access ends up being the recommended approach (it's possible the linter will demand you do so, in fact)

Try this:

```
  const { body } = req;
  if (!isOpportunity(body)) {
    res.status(400)
      .contentType('application/json')
      .send(isOpportunity.errors ?? { message: 'Expected JSON body having title.' });
    return;
  }
  
  const opportunityPosted: Omit<Opportunity, 'createdAt' | 'id'> = { title: body.title };
```

Another fun fact -- you don't have to type on assignment (I don't know if the linter will tell you not to, but I suggest NOT doing that).  This means you can simply write:

```
  const opportunityPosted = { title: body.title };
```

bickelj:
All of this is at compile/lint time, not run time, right. I suppose I can lean on `isOpportunity` rather than try to formally declare type expectations. In this case there is only one property of interest (`title`) but in other cases (other entities) there will be several. Also I expect `Opportunity` (or others) to evolve, so the goal was to have the type checker flag those kinds of code changes needed when a change occurs. Perhaps it can be caught elsewhere, too, though, or in a better way.

With `const { body } = req;` and later `  const opportunityPosted: Omit<Opportunity, 'createdAt' | 'id'> = { title: body.title };`:
> Unsafe array destructuring of a tuple element with an `any` value  @typescript-eslint/no-unsafe-assignment

I think I would rather say something like `const opportunityPosted: Omit<Opportunity, 'createdAt' | 'id'> = body;` but perhaps I am asking too much of TypeScript and AJV.

[...]

PR #70, opened 2022-08-03
Implement POST /opportunities endpoint

bickelj:
Creates a single opportunity when it is valid.

Issue #33 Implement POST /opportunities endpoint

### DIFF
--- a/src/__tests__/opportunities.int.test.ts
+++ b/src/__tests__/opportunities.int.test.ts
@@ -2,13 +2,17 @@ import request from 'supertest';
 import { app } from '../app';
 import { db } from '../database';
 import { getLogger } from '../logger';
+import {
+  getTableMetrics,
+  isoTimestampPattern,
+} from '../test/utils';
 import type { Result } from 'tinypg';
 
 const logger = getLogger(__filename);
 const agent = request.agent(app);
 
 describe('/opportunities', () => {
-  describe('/', () => {
+  describe('GET /', () => {
     logger.debug('Now running an opportunities test');
     it('returns an empty array when no data is present', async () => {
       await agent
@@ -60,7 +64,7 @@ describe('/opportunities', () => {
     });
   });
 
-  describe('/id', () => {
+  describe('GET /id', () => {
     it('returns exactly one opportunity selected by id', async () => {
       const opportunityIds: Result<{ id: number; title: string }> = await db.query(`
         INSERT INTO opportunities (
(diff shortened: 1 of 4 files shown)


bickelj:
I see that it shows the full diff from main to this branch even though the PR is based on a previous branch. This is awkward. I suppose review #66 and #68 first and we can get those merged before looking here.

bickelj:
This PR is intended only to cover 5350f92.

bickelj:
Ah yes, and missing tests for the post as well.

bickelj:
eddebd3 includes the needful tests.

slifty:
Now that the other endpoints are merged  it would be good to rebase this -- I can do it but if you're back then it'd be cleaner for you to!

slifty:
I used some of the patterns set in here in #76 but also added a few others -- for instance the test suite in the new PR might be worth checking out, as well as the validation mechanism.

bickelj:
914ff96bca03e2ba20bd5dccc3d5ee5c4213b6d4 is the rebase without any other changes yet.

bickelj:
I don't see the Codecov comment updated for commit e6d9397 but when looking at the coverage in https://github.com/PhilanthropyDataCommons/service/actions/runs/3130147251/jobs/5080065775 I see it should be OK now.

src/__tests__/opportunities.int.test.ts
```diff
@@ -133,4 +137,111 @@ describe('/opportunities', () => {
       });
     });
   });
+
+  describe('GET /id', () => {
```
slifty, line 141: This test is just to add coverage to a feature added before, right?  Let's put it in a separate commit if that's easy (not a big deal if not)

slifty (approved):
w00t looks great.

One tiny commit organization request which can be ignored or taken as desired.

src/__tests__/opportunities.int.test.ts
bickelj: Correct, good call. Doing it.

bickelj:
https://github.com/PhilanthropyDataCommons/service/pull/88 has just the follow-up to issue #31 with tests. I don't recall how that fell through the cracks but oh well. Let's merge #88 first, then I will rebase here, then merge this.

bickelj:
It turned out to be a bad rebase or bad merge where some (slightly different but largely the same) code got duplicated. The solution was to remove it here rather than add more in #88. This should be very close to ready.

Issue #67, opened 2022-08-02
Add documentation to the root page


bickelj:
A description and/or the OpenAPI docs might be nice to have as static content on the root endpoint.

bickelj:
Resolved with merge of #71

PR #71, opened 2022-08-04
Add OpenAPI documentation to the root page

bickelj:
Now the OpenAPI spec is served as the root page of the service.

Issue #67 Add documentation to the root page

### DIFF
--- a/src/types/JsonObject.ts
+++ b/src/types/JsonObject.ts
@@ -1,0 +1,5 @@
+import type { JsonObject } from 'swagger-ui-express';
+
+export const isJsonObject = (obj: unknown): obj is JsonObject => (
+  typeof obj === 'object'
+);
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "pino": "^8.1.0",
         "pino-http": "^8.1.1",
         "postgres-migrations": "^5.3.0",
+        "swagger-ui-express": "^4.5.0",
         "tinypg": "^7.0.0"
       },
       "devDependencies": {
@@ -25,6 +26,7 @@
         "@types/node": "^18.0.0",
         "@types/pg": "^8.6.5",
         "@types/supertest": "^2.0.12",
+        "@types/swagger-ui-express": "^4.1.3",
         "@typescript-eslint/eslint-plugin": "^5.29.0",
         "@typescript-eslint/parser": "^5.29.0",
         "copyfiles": "^2.4.1",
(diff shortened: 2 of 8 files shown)


bickelj:
This won't pass lint. But there is probably some simple change to fix that.

src/routers/documentation.ts
```diff
@@ -0,0 +1,14 @@
+import express from 'express';
+import swaggerUi from 'swagger-ui-express';
+import documentation from '../openapi.json';
+import type { JsonObject } from 'swagger-ui-express';
+
+const documentationRouter = express.Router();
+
+// Attempting to satisfy this not being any, but this only pushes back the issue.
+const docs: JsonObject = documentation;
```
slifty, line 9: The lint error is because we can't just cast objects in this way since documentation COULD be anything.

Unfortunately we'll need to write a type guard (maybe swagger-ui-express provides one, but if not, it falls on us unfortunately.  I'll take a look.

slifty: (taking on this fix now since we def want this as part of the demo)

slifty:
(Note that I think the codecov report is just completely busted -- this PR was opened before I added the app integration so we can just ignore it.)

I added a type guard for the `JsonObject` (which is... defined as basically just being a Record).  @bickelj would you be up for taking a glance at the lines I added to make sure it's 👍 -- also, sorry for merging rather than having two commits, it was out of habit / I wanted to avoid having a broken commit state (I can pull them out if helpful though)

slifty:
Oh wait, this breaks tests -- and yet it says test pass.  That's ... weird... why...

slifty:
Well, that was amazing -- I opened a bug report in Jest: https://github.com/facebook/jest/issues/13233

I won't recap the bug here, but basically it's possible that Jest fails to produce a failing exit code when a transformer + coverage report + running in band is specified.

bickelj:
I saw "extends" in the `tsconfig.dev.json` and assumed it was referencing the production config. After looking more closely, both the `tsconfig.dev.json` and `tsconfig.json` extend the same base in-tsconfig-package configuration file, so they are completely independent.

slifty (approved):
Hooray for documentation!

src/types/JsonObject.ts
```diff
@@ -0,0 +1,5 @@
+import type { JsonObject } from 'swagger-ui-express';
+
+export const isJsonObject = (obj: unknown): obj is JsonObject => (
+  typeof obj === 'object'
```
bickelj, line 4: This is curious. Would we not want to conform to the actual `JsonObject` rather than our own definition of what it is? Or am I misunderstanding something here?

slifty: (Ack sorry I thought you had approved)

This is basically the definition of JsonObject (insofar as an object in javsascript is defined as being something with `string` keys and `any` value)

Here's the definition of JsonObject:

```
export interface JsonObject {
    [key: string]: any;
}
```

slifty: Oh there's a bug though (and I should write a unit test for it) -- `null` is the one case of something of type object that doesn't conform to the spec!  So this should actually be `typeof obj === 'object' && obj !== null`.

I'll open a fix PR to add that + some tests.

slifty: (Sorry for the triple post)

Something I didn't explain here, but maybe part of the confusion: types / interfaces do not exist in runtime, so we can't use `JsonObject` directly.  The library does not provide type guards, unfortunately (they never do for some reason?? I don't understand why that isn't the norm.).  This means it is our responsibility to create the type guard that conforms with the interface.

This is a bit of a mind bend when moving from other typed languages where typing can often be enforced at runtime, and so a type and a type guard is the same thing.

PR #74, opened 2022-08-30
Rename query result variables

### LINKED ISSUE #73: Fix DB result variable names
I made a mistake when naming the variable that contains a DB result (I named it as though it contained the resulting *object(s)* which is actually one level deeper.

I'd like to see this fixed in the two endpoints we have so far in order to prevent further repetition of the error.

slifty:
This PR renames our DB result variables to be more accurate, it also adds some type safety to the results.

Resolves #73


bickelj (commented):
See https://github.com/PhilanthropyDataCommons/service/issues/73#issuecomment-1232124104

(Aside: GitHub is kind of weird with regard to issues. Is the issue the issue or is the PR the issue?)

bickelj (approved):
Any of the options we discussed is fine, including your original suggestion of `result`. It is not a big deal.

PR #76, opened 2022-09-06
Add `POST /canonicalFields` endpoint

### LINKED ISSUE #75: Implement POST `/canonicalFields` endpoint
This will allow the creation of new canonical fields.

slifty:
This PR adds the ability to create canonical fields.

To test:

```
curl --request POST \
  --url http://localhost:3000/canonicalFields \
  --header 'content-type: application/json' \
  --data '{
  "label": "First name",
  "shortCode": "firstName",
  "dataType": "string"
}'
```

Resolves #75

### DIFF
--- a/src/__tests__/canonicalFields.int.test.ts
+++ b/src/__tests__/canonicalFields.int.test.ts
@@ -1,12 +1,15 @@
 import request from 'supertest';
 import { app } from '../app';
 import { db } from '../database';
+import { getLogger } from '../logger';
+import { isoTimestampPattern } from '../test/utils';
 import type { Result } from 'tinypg';
 
+const logger = getLogger(__filename);
 const agent = request.agent(app);
 
 describe('/canonicalFields', () => {
-  describe('/', () => {
+  describe('GET /', () => {
     it('returns an empty array when no data is present', async () => {
       await agent
         .get('/canonicalFields')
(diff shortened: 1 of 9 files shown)


src/handlers/opportunitiesHandlers.ts
```diff
@@ -42,7 +43,7 @@ const getOpportunities = (req: Request, res: Response): void => {
 interface GetOpportunityParams {
   id: number;
 }
-const isGetOpportunityParams = ajv.compile<GetOpportunityParams>({
+const getOpportunityParamsSchema: JSONSchemaType<GetOpportunityParams> = {
```
slifty, line 46: These changes should be in a separate commit with an explanation

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -33,6 +45,75 @@ const getCanonicalFields = (req: Request, res: Response): void => {
     });
 };
 
+const postCanonicalFieldBodySchema: JSONSchemaType<Omit<CanonicalField, 'createdAt' | 'id'>> = {
```
slifty, line 48: This is redundant -- I've opened [an issue in the ajv repository](https://github.com/ajv-validator/ajv/issues/2085) to explore the idea of creating corresponding "utility" functions in the world of ajv to mirror the TS utility types.

In the mean time I think it will be valuable to create those utility methods directly in our repo and use them, so I'll go add that in.

slifty: I spent a time boxed two hours trying to create a utility that would accomplish this goal, but it's... incredibly complex (maybe impossible) to do in a type safe way.  Here's where I got (this doesn't work but since I'm not gonna commit it I'd like to post it SOMEWHERE for posterity)

```

const removeAttributesFromObject = <T, K extends string | number | symbol>(
  obj: T,
  keysToRemove: (keyof T)[],
): Omit<T, K> => {
  const modifiedObj: Partial<T> = { ...obj };
  keysToRemove.forEach((key) => {
    modifiedObj[key] = undefined;
  });
  return modifiedObj;
};

export function omitFieldsFromSchema<T, K extends string | number | symbol>(
  schema: JSONSchemaType<T>,
  fieldsToOmit: (keyof T)[],
): JSONSchemaType<Omit<T, K>> {
  const filteredProperties = removeAttributesFromObject<T, K>(
    schema,
    fieldsToOmit,
  )
  const filteredRequired = schema.required.filter(
    (item: keyof T) => !fieldsToOmit.includes(item),
  )
  return {
    type: 'schema.type',
    properties: filteredProperties,
    required: filteredRequired,
  }
};
```

slifty: All this is to say, redundant is ugly but I believe it's the best we can do.

The silver lining is that it's reasonable to me that we define the schema explicitly for each API handler anyway, since at the end of the day API signatures should be explicit and hard to accidentally change.

slifty:
I need some of this code in order to make progress on other POST endpoints (particularly in order to detect unique collisions) so I'm going to invoke the "I'm the only dev who isn't out of office!" clause and bypass branch protections to merge.

If there are comments on this PR we can make any fixes as a separate PR!

bickelj:
Without looking at the code, only looking at the concept of a `POST /canonicalFields`, this is interesting. I guessed we had intentionally left this endpoint out of the API in the interest of letting these be created in one way or another by the administrator of the service. For example, I would have populated canonical fields using the database migrations or a script that populated them at or after startup. If we all use the same version of the software, then, we would all have the same IDs for the same canonical fields across instances. Not that we plan to have several instances, but it might be good to know the ID in advance and be very careful about adding canonicalFields. On the other hand, the point is to have a single instance of this service and to let software both programmatically read and write all kinds of data, perhaps with different levels of privilege, so it makes sense to include this endpoint.

With regard to process, I agree that since no reviewer picked it up or indicated a review by the same time the next business day, self-merge is appropriate.

src/__tests__/canonicalFields.int.test.ts
```diff
@@ -62,4 +65,111 @@ describe('/canonicalFields', () => {
...
+        FROM canonical_fields;
+      `);
+      logger.debug('after: %o', after);
+      expect(before.rows[0].count).toEqual('0');
+      expect(result.body).toMatchObject({
+        id: expect.any(Number) as number,
+        label: '🏷️',
+        shortCode: '🩳',
+        dataType: '📊',
+        createdAt: expect.stringMatching(isoTimestampPattern) as string,
```
bickelj, line 100: This and the above `id` expectation are nice and flexible. If we expect 0 rows before, like on line 94, then we should be able to expect that the first row inserted should be 1, right? Unless the db has not been created from scratch. Anyway, this is probably fine, good enough. But we can insert the exact datetime and then verify that we get the same on the other side.

Minor point regarding the timestamp pattern: it's ISO but more specifically an IETF RFC 3339 timestamp pattern.

src/database/db.ts
```diff
@@ -1,6 +1,13 @@
 import path from 'path';
 import { TinyPg } from 'tinypg';
+import { isQueryContextWithError } from '../types';
+import type { TinyPgError } from 'tinypg';
 
 export const db = new TinyPg({
   root_dir: [path.resolve(__dirname, 'queries')],
 });
+
+export const detectUniqueConstraintViolation = (error: TinyPgError): boolean => (
```
bickelj, line 10: Nice function!

src/handlers/canonicalFieldsHandlers.ts
bickelj: I am trying to think if there is ever a second place (outside of the `POST /blah` code) where we would want to validate a minimal or not-yet-inserted entity. I suppose when the situation actually arises, then I would want to put this into a common code area next to the other Ajv types rather than one-off it here. But I agree if it's the only situation then the one-off should be in the context of the one-off.

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -33,6 +45,75 @@ const getCanonicalFields = (req: Request, res: Response): void => {
...
+      type: 'string',
+    },
+  },
+  required: [
+    'label',
+    'shortCode',
+    'dataType',
+  ],
+};
+const isPostCanonicalFieldBody = ajv.compile(postCanonicalFieldBodySchema);
```
bickelj, line 67: I haven't tested it but will this compilation take place at server startup or at first POST to /canonicalFields? If the latter, for more consistent performance we might want to go to the common area. If it only takes a few milliseconds regardless then this is a non-issue.

src/types/QueryContextWithError.ts
```diff
@@ -0,0 +1,82 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+// This type is defined to fill a type void in the tinypg library
+// See: https://github.com/dynajoe/tinypg/issues/40
+export interface QueryContextWithError {
```
bickelj, line 6: Nice.

src/__tests__/canonicalFields.int.test.ts
slifty: Both great points -- I'll make an issue to clean that (and other things you've flagged) up

bickelj (commented):
LGTM

PR #77, opened 2022-09-06
Add GET /applicants endpoint

slifty:
This PR adds the `GET /applicants` endpoint.

You can test it using:

```
curl --request GET \
  --url http://localhost:3000/applicants \
  --header 'content-type: application/json'
```

Resolves #36

PR #78, opened 2022-09-07
Add POST `/applicants` endpoint

### LINKED ISSUE #37: Implement `POST /applicants` endpoint
Allows the creation of an applicant.

slifty:
This PR adds the ability to create an applicant via the API.

To test:

```
curl --request POST \
  --url http://localhost:3000/applicants \
  --header 'content-type: application/json' \
  --data '{
  "externalId": "12345"
}'
```

Resolves #37 
Relies on #77 (don't review until 77 is merged and this is rebased)

### DIFF
--- a/src/test/utils.ts
+++ b/src/test/utils.ts
@@ -1,1 +1,18 @@
+import { db } from '../database';
+
 export const isoTimestampPattern = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/;
+
+interface TableMetrics {
+  count: number;
+  maxId: number;
+  now: Date;
+}
+export const getTableMetrics = async (tableWithId: string): Promise<TableMetrics> => {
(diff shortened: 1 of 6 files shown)


src/test/utils.ts
```diff
@@ -1 +1,18 @@
...
+import type { Result } from 'tinypg';
+
 export const isoTimestampPattern = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/;
+
+interface TableMetrics {
+  count: number;
+  maxId: number;
+  now: Date;
+}
+export const getTableMetrics = async (tableWithId: string): Promise<Result<TableMetrics>> => (
```
slifty, line 11: Might as well return the metrics rather than the result since I don't think the result is ever going to be interesting (famous last words maybe)

bickelj (approved):
Looks good.

PR #80, opened 2022-09-09
Fix JsonObject type guard

slifty:
This PR fixes an edge case where `isJsonObject` would return true for `null` values.  It also adds unit tests for good measure.

### DIFF
--- a/src/types/JsonObject.ts
+++ b/src/types/JsonObject.ts
@@ -1,5 +1,10 @@
+import { ajv } from '../ajv';
 import type { JsonObject } from 'swagger-ui-express';
+import type { JSONSchemaType } from 'ajv';
 
-export const isJsonObject = (obj: unknown): obj is JsonObject => (
-  typeof obj === 'object'
-);
+const jsonObjectSchema: JSONSchemaType<JsonObject> = {
+  type: 'object',
+  properties: {},
+  required: [],
+};
+export const isJsonObject = ajv.compile(jsonObjectSchema);
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -4,7 +4,10 @@ module.exports = {
       tsconfig: 'tsconfig.dev.json',
     },
   },
-  collectCoverageFrom: ["src/**/*.ts"],
+  collectCoverageFrom: [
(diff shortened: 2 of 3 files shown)


src/types/JsonObject.ts
```diff
@@ -1,5 +1,14 @@
...
+// in the swagger-ui-express library.  Currently that definition is:
+//
+// export interface JsonObject {
+//     [key: string]: any;
+// }
+//
+// ^ this is essentially the same as the definition of a non-null Object
 export const isJsonObject = (obj: unknown): obj is JsonObject => (
   typeof obj === 'object'
+  && obj !== null
```
bickelj, line 13: This is still disconnected from the actual definition of `JsonObject` right? If it works, that's fine, but we're basically defining our own `JsonObject` rather than using an existing `JsonObject`, right?

bickelj: Suppose the definition within `swagger-ui-express` changes. Now we have a separate definition of our own. I suppose an error at runtime would occur. But then in that case what was the point of the type checking?

bickelj: I now see your comments in the #71 area discussing the same. My summary: library user type guards are a necessary hack for libraries that do not supply them. If the type guard were not necessary there would be no disconnect. So whenever we introduce a type guard for a library this same disconnect happens. Close?

slifty: Want to be sure you saw my explanation over in that other thread!

bickelj: The question is less important for this simple instance and more important for my understanding or a more complex/impactful instance. I am satisfied with your explanation even if I keep treating TypeScript like a monkey with a hammer.

bickelj (approved):

src/types/JsonObject.ts
slifty: I definitely want to make sure you walk away understanding -- this may be an "ah HA!" moment once we're through it!

That said, your question did get me to do the more correct thing and lean into ajv for this (that way if the definition changes AJV will let us know right away in the form of a static analysis error).

PR #81, opened 2022-09-12
Add GET /applicationForms endpoint

### LINKED ISSUE #65: Implement `GET /applicationForms` endpoint
Retrieves a list of application forms (each having a list of fields) for a given opportunity.

slifty:
This PR adds the Application Forms type as well as the endpoint for retrieving a list of all application forms.

Resolves #65

### DIFF
--- a/src/__tests__/applicationForm.int.test.ts
+++ b/src/__tests__/applicationForm.int.test.ts
@@ -1,0 +1,78 @@
+import request from 'supertest';
+import { app } from '../app';
+import { db } from '../database';
+import type { Result } from 'tinypg';
+
+const agent = request.agent(app);
+
+describe('/applicationForms', () => {
+  describe('GET /', () => {
+    it('returns an empty array when no data is present', async () => {
+      await agent
+        .get('/applicationForms')
+        .expect(200, []);
+    });
+
+    it('returns all application forms present in the database', async () => {
+      await db.query(`
+        INSERT INTO opportunities (
+          title,
+          created_at
+        )
+        VALUES
+          ( 'Tremendous opportunity 👌', '2525-01-01T00:00:05Z' ),
+          ( 'Good opportunity', '2525-02-01T00:00:05Z' );
+      `);
+      await db.query(`
+        INSERT INTO application_forms (
+          opportunity_id,
+          version,
+          created_at
+        )
+        VALUES
+          ( 1, 1, '2022-07-20 12:00:00+0000' ),
+          ( 1, 2, '2022-08-20 12:00:00+0000' ),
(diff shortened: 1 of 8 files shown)


src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,44 @@
...
+  Request,
+  Response,
+} from 'express';
+import type { Result } from 'tinypg';
+import type { ApplicationForm } from '../types';
+
+const logger = getLogger(__filename);
+
+const getApplicationForms = (req: Request, res: Response): void => {
+  db.sql('applicationForms.fetchAll')
```
slifty, line 19: This should be `selectAll` now

src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,44 @@
...
+    .then((fetchAllQueryResult: Result<ApplicationForm>) => {
```
slifty, line 20: The name of this var should be `applicationFormsQueryResult`

src/__tests__/applicationForm.int.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    it('returns all application forms present in the database', async () => {
+      await db.query(`
+        INSERT INTO opportunities (
+          title,
+          created_at
+        )
+        VALUES
+          ( 'Tremendous opportunity 👌', '2525-01-01T00:00:05Z' );
+      `);
+      await db.query(`
```
slifty, line 25: This test was half baked.  It passes but only because it isn't actually testing the right endpoint.

src/database/migrations/0004-create-application_forms.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE application_forms (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  opportunity_id INTEGER NOT NULL,
+  version INTEGER NOT NULL DEFAULT 1,
```
slifty, line 4: I'm not sure we should actually define a default here

bickelj: When combined with the unique constraint below it means that the version is optional on the first (majority case) post but required on the second and following versions of that same opportunity. It seems like a slightly annoying burden to require a version for the first version posted. But it is probably simpler and clearer to require it every time. Requiring it every time tells the caller to be aware that multiple versions are supported. Then again, just because you have fancy options available (in this case versions of opportunities) does not mean we need to burden the simple case (a single opportunity) with the complexity of the complex case (multiple opportunities). In short, there is good reason to do it either way, I'm up for either.

PR #82, opened 2022-09-12
Add POST / applicationForms endpoint

### LINKED ISSUE #64: Implement `POST /applicationForms` endpoint
Allows creation of an application form for a given opportunity. `/applications` depends on a previous post to `/applicationForm` (this) which in turn depends on a previous post to `/opportunities`.

slifty:
This PR adds the ability to create an application form in the PDC

It builds on #81 so should not be reviewed until that has been merged and this rebased.

Resolves #64

### DIFF
--- a/src/database/db.ts
+++ b/src/database/db.ts
@@ -3,11 +3,17 @@ import { TinyPg } from 'tinypg';
 import { isQueryContextWithError } from '../types';
 import type { TinyPgError } from 'tinypg';
 
+// See https://www.postgresql.org/docs/8.2/errcodes-appendix.html
+export enum PostgresErrorCode {
+  FOREIGN_KEY_VIOLATION = '23503',
+  UNIQUE_VIOLATION = '23505',
+}
+
 export const db = new TinyPg({
   root_dir: [path.resolve(__dirname, 'queries')],
 });
 
 export const detectUniqueConstraintViolation = (error: TinyPgError): boolean => (
   isQueryContextWithError(error.queryContext)
-  && error.queryContext.error.routine === '_bt_check_unique'
+  && error.queryContext.error.code === PostgresErrorCode.UNIQUE_VIOLATION
 );
--- a/src/database/queries/applicationForms/insertOne.sql
+++ b/src/database/queries/applicationForms/insertOne.sql
@@ -1,0 +1,19 @@
+INSERT INTO application_forms (
+  opportunity_id,
+  version
+) VALUES (
+  :opportunityId,
+  COALESCE(
+    (
+      SELECT MAX(af.version) + 1
+      FROM application_forms as af
+      WHERE af.opportunity_id = :opportunityId
+    ),
+    1
+  )
+)
+RETURNING
+  id as "id",
+  opportunity_id as "opportunityId",
+  version as "version",
+  created_at as "createdAt"
--- a/src/handlers/applicationFormsHandlers.ts
+++ b/src/handlers/applicationFormsHandlers.ts
@@ -1,9 +1,14 @@
+import { TinyPgError } from 'tinypg';
+import { ajv } from '../ajv';
 import { getLogger } from '../logger';
 import {
   db,
+  PostgresErrorCode,
 } from '../database';
 import {
+  isApplicationForm,
   isApplicationFormArray,
+  isQueryContextWithError,
 } from '../types';
 import { ValidationError } from '../errors';
 import type {
@@ -11,6 +16,7 @@ import type {
   Response,
 } from 'express';
 import type { Result } from 'tinypg';
+import type { JSONSchemaType } from 'ajv';
 import type { ApplicationForm } from '../types';
 
 const logger = getLogger(__filename);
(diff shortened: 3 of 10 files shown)


src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,125 @@
...
+          'The database responded with an unexpected format.',
+          isApplicationForm.errors ?? [],
+        );
+      }
+    })
+    .catch((error: unknown) => {
+      if (error instanceof TinyPgError
+      && isQueryContextWithError(error.queryContext)) {
+        switch (error.queryContext.error.code) {
+          case PostgressErrorCode.FOREIGN_KEY_VIOLATION:
```
slifty, line 96: This might be a nicer approach than the DB error detection than using the function since it allows us to use a switch statement, but of course having the complexity in a function has some perks too!  Offering this in this PR for consideration.

src/database/db.ts
```diff
@@ -3,11 +3,17 @@ import { TinyPg } from 'tinypg';
 import { isQueryContextWithError } from '../types';
 import type { TinyPgError } from 'tinypg';
 
+// See https://www.postgresql.org/docs/8.2/errcodes-appendix.html
+export enum PostgressErrorCode {
```
bickelj, line 7: Small typo, it should be postgres.

src/database/db.ts
```diff
@@ -3,11 +3,17 @@ import { TinyPg } from 'tinypg';
...
+}
+
 export const db = new TinyPg({
   root_dir: [path.resolve(__dirname, 'queries')],
 });
 
 export const detectUniqueConstraintViolation = (error: TinyPgError): boolean => (
   isQueryContextWithError(error.queryContext)
-  && error.queryContext.error.routine === '_bt_check_unique'
+  && error.queryContext.error.code === PostgressErrorCode.UNIQUE_VIOLATION
```
bickelj, line 18: Hooray, this is a clearer approach than the previous approach!

src/database/queries/applicationForms/insertOne.sql
```diff
@@ -0,0 +1,19 @@
+INSERT INTO application_forms (
+  opportunity_id,
+  version
+) VALUES (
+  :opportunityId,
+  COALESCE(
```
bickelj, line 6: I wondered if coalesce was needed here due to the default value, but I guess the situation is "this might be the first form for the given opportunity" and therefore the subquery returns null, so it makes sense. I guess this makes the versions autoincrementing within a given opportunity and that's convenient.

src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,125 @@
...
+    })
+    .catch((error: unknown) => {
+      logger.error(error);
+      res.status(500)
+        .contentType('application/json')
+        .send(error);
+    });
+};
+
+const postApplicantionFormsBodySchema: JSONSchemaType<Omit<ApplicationForm, 'createdAt' | 'id' | 'version'>> = {
```
bickelj, line 48: Small typo as well, should be postApplicationFormsBodySchema.

src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,125 @@
...
+  properties: {
+    opportunityId: {
+      type: 'number',
+    },
+  },
+  required: [
+    'opportunityId',
+  ],
+};
+const isPostApplicantionFormsBody = ajv.compile(postApplicantionFormsBodySchema);
```
bickelj, line 59: Small typo again, isPostApplicationFormsBody.

src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,125 @@
...
+    })
+    .catch((error: unknown) => {
+      if (error instanceof TinyPgError
+      && isQueryContextWithError(error.queryContext)) {
+        switch (error.queryContext.error.code) {
+          case PostgressErrorCode.FOREIGN_KEY_VIOLATION:
+            res.status(409)
+              .contentType('application/json')
+              .send({
+                message: 'Unique key constraint violation.',
```
bickelj, line 100: The message should mention either foreign key violation and/or what can be done about it (e.g. "please create the opportunity first because there is no opportunity with ID=x").

bickelj (changes requested):
Looks good, just a couple typos.

src/handlers/applicationFormsHandlers.ts
slifty: That detail is built into the `errors` object (the error has more specifics) -- I think that might be better, since we don't necessarily know what field triggered the error without parsing the error (or without coupling this message with the current schema in ways we might forget to update later).

Does that info resolve the concern or do you want it in the message too?

slifty: (Oh -- this actually incorrectly says Unique key constraint whoops!  I misread what you were saying!)

src/handlers/applicationFormsHandlers.ts
```diff
@@ -0,0 +1,125 @@
...
+        switch (error.queryContext.error.code) {
+          case PostgresErrorCode.FOREIGN_KEY_VIOLATION:
+            res.status(409)
+              .contentType('application/json')
+              .send({
+                message: 'Unique key constraint violation.',
+                errors: [error.queryContext.error],
+              });
+            return;
+          case PostgresErrorCode.UNIQUE_VIOLATION:
```
slifty, line 104: Actually, since the unique constraint is based on `version` and we don't let the user specify a version / that is maintained on our end, I don't believe a unique validation *can* occur (and if it does it is a 5xx error not a 4xx error).  given that, I don't think it's possible to write a test that generates this case and so I'm gonna just take it out and if somehow that comes through it would go down to the 500 error logic below.

slifty: I've moved this to a simple if statement to cover that single known likely; in a future PR I'm going to propose an alternative way to handle these errors anyway so that it's always robust regardless of what we understand as possible in a single case..

src/handlers/applicationFormsHandlers.ts
slifty: I think I've resolved this by updating "Unique key" to "Foreign key".

Related note, as mentioned in another comment thread I have a proposed refactor to this logic block (since it ends up being needed in a bunch of places / endpoints) so I think that will be a good time to revisit exploring more detail in the message.

slifty:
I gave a "last call" ping for the re-review but also I do think that @bickelj gave the green light once those typos were fixed (I'll do one more pass before hitting merge but ultimately will hit merge)

bickelj:
Yes, you got it. My only other thought "what can be done about it" is a more general comment regarding error messages a la Don Norman, though I cannot find the quote now, something about not giving error messages but rather what-you-can-do-about-it messages. This is a high bar though and sometimes it is most straightforward to have just the error.

PR #84, opened 2022-09-12
Create fields on form creation

### LINKED ISSUE #83: Add the ability to create an application form with fields
Instead of creating endpoints for `POST applicationFields` I think we should allow the user to create the application form in one swoop, passing an array of application fields.

This means expanding the applicationFields POST endpoint to support a list of fields.

Since we're working with "versions" I think it makes sense to have that list of fields be required on POST.  In other words, the fields cannot be added (or edited) ad-hoc after the creation of a form.

slifty:
This PR updates the POST `/applicationForm` endpoint to require a list of fields to store as part of the form's definition.

It also updates the shape of the response so that it contains those fields.

GET `/applicationForm` does NOT populate the results with fields, but eventually `GET /applicationForm/{applicationFormId}` should do so.

This PR builds on #82 and should not be reviewed until that is rebased

Resolves #83

### DIFF
--- a/src/database/migrations/0005-create-application_form_fields.sql
+++ b/src/database/migrations/0005-create-application_form_fields.sql
@@ -1,0 +1,21 @@
+CREATE TABLE application_form_fields (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  application_form_id INTEGER NOT NULL,
+  canonical_field_id INTEGER NOT NULL,
+  position INTEGER NOT NULL,
+  label VARCHAR,
+	created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  CONSTRAINT fk_application_form
+    FOREIGN KEY(application_form_id)
+      REFERENCES application_forms(id),
+  CONSTRAINT fk_canonical_field
+    FOREIGN KEY(canonical_field_id)
+      REFERENCES canonical_fields(id),
+  UNIQUE(application_form_id, position)
+);
+
+COMMENT ON TABLE application_form_fields is
+  'Application form fields represent a specific piece of data being collected as part of a given application form.';
+COMMENT ON COLUMN application_form_fields.canonical_field_id is 'The id of the canonical field associated with the data being collected by this form field';
+COMMENT ON COLUMN application_form_fields.position is 'The order that this field should appear on the application form';
+COMMENT ON COLUMN application_form_fields.label is 'The language presented to an applicant for this field';
--- a/src/types/ApplicationFormField.ts
+++ b/src/types/ApplicationFormField.ts
@@ -1,0 +1,76 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+export interface ApplicationFormField {
+  readonly id: number;
+  applicationFormId: number;
+  canonicalFieldId: number;
+  position: number;
+  label: string;
+  readonly createdAt: Date;
+}
+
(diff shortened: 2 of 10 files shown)


src/types/ApplicationFormField.ts
```diff
@@ -0,0 +1,76 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+export interface ApplicationFormField {
+  readonly id: number;
+  applicationFormId: number;
+  canonicalFieldId: number;
```
slifty, line 7: This should be the canonical field *CODE* -- not ID -- since the code is what has meaning to external users.

slifty:
As part of the work on this commit I shaved a yak, which I think is large enough to go into its own separate PR so it can be discussed in isolation.  It basically creates the concept of a standard response body for errors and creates the first step toward middleware to generate the right status code / body.  (I'll explain more in the PR but since I committed it to this branch in the short term I wanted to at least comment here for now)

src/database/migrations/0005-create-application_form_fields.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE application_form_fields (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  application_form_id INTEGER NOT NULL,
+  canonical_field_short_code VARCHAR NOT NULL,
```
bickelj, line 4: I think either of two changes would make sense here:
1. Drop the generated integer id column entirely in canonical_fields and make the short code the primary key, or
2. Replace this column and foreign key with a reference to the id field of canonical_fields.

The reason I would use a generated integer id is when I don't have anything better and/or the performance of joins on integer columns. If the user is going to supply the short code (not the integer id) and the join will happen on the varchar column, there is no purpose to having the id column. If the purpose is consistency across the API, then we should require the integer id here as well.

It would be nice to use all natural keys wherever possible rather than the synthetic instance-specific keys but I think we already went in the integer id direction across the board.

If we wish do provide the convenience of the short code, another resolution might be to allow the caller to supply the short code at the API level but then do the lookup on the backend at the database level.

src/types/ApplicationFormField.ts
```diff
@@ -0,0 +1,76 @@
...
+export const applicationFormFieldSchema: JSONSchemaType<ApplicationFormField> = {
+  type: 'object',
+  properties: {
+    id: {
+      type: 'integer',
+    },
+    applicationFormId: {
+      type: 'integer',
+    },
+    canonicalFieldShortCode: {
```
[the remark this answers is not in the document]
slifty: This makes sense -- two things to confirm that I mean it:

1) Your point around the importance of having internal consistency around key relationships is well taken; regardless of API this should store that relationship in terms of ID and not shortCode

2) Your point around the API consistency is also well taken; I think at this stage providing a shortCode convenience method would be a premature optimization AND result in inconsistent API definition.  That could always be added at a later date as an optional alternative to canonicalFieldId if we determined the need.

---

So the next step would be for me to revert the shortCode change!  If that sounds good to you I'll knock that out as part of the rebase to main.

src/database/migrations/0005-create-application_form_fields.sql
slifty: I've updated this to only reference ID and never shortCode!

src/types/ApplicationFormField.ts
slifty: (For future archeologists: we decided not to do this)

slifty:
Rebased against main -- no code changes since last week's approval!

PR #87, opened 2022-09-27
Minor doc fixes after I stood up a local instance

kfogel:
These are actually from 2022-09-19, when I was standing up an instance locally to generate some screenshots for John Mohr.  These fixes just touch the commands I happened to run into.  There might be other similar tweaks needed elsewhere -- since "npm foo" and "npm run foo" are sometimes both valid (but not equivalent) commands, I wasn't sure; someone more knowledgeable that I am should check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ To run migrations:
 
 ```bash
 npm build
-npm migrate
+npm run migrate
 ```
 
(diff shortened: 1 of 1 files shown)


bickelj (approved):
Thank you!

PR #89, opened 2022-09-27
Fix field name

kfogel:
Fix field name

This is against branch `noissue-september-demo-lock` (commit eac0b8835267), not against `main`.


slifty:
Thank you!!

PR #90, opened 2022-09-28
Refactor error handling

### LINKED ISSUE #61: Use middleware for Express error handling
We [had a thread about express error handling](https://github.com/PhilanthropyDataCommons/service/pull/55/files/e4c36fd8a3d6d88b2e0f70fa118fba0c93cf5230#r931263680) which ultimately resolved in the idea that it is probably best to have custom Middleware that will consistently handle errors raised in express route handlers.

Express has [default middleware](https://expressjs.com/en/guide/error-handling.html) but unfortunately it returns HTML rather than a `json` response.  [It looks like](https://stackoverflow.com/questions/53931490/how-to-default-to-returning-errors-as-json-instead-of-html-with-express) the best way to have the error handler return json is custom middleware.

If we go this route we should go through and modify any handlers that return `500` errors directly so that they lean on the middleware.

slifty:
This issue consolidates our error handling logic to a new piece of [express middleware](https://expressjs.com/en/guide/error-handling.html).

Resolves #61


slifty:
(sorry for the early review request -- I didn't realize I had to rebase first to bring up to date with main!)

slifty:
Apparently this refactor decimated our test coverage BUT I think it actually did so by making the control flow clearer and, as a result, exposing an untested case.  We didn't have coverage to test what would happen in the event that tinypg actually throws an error!  I'll add some tests for that, plus some of the missing tests for making sure `errorHandler` can handle unknown errors!

(I think this is still worth reviewing though; those will be added test cases rather than changes to the code itself)

slifty:
Update: complete tests have now been added.

slifty:
I think maybe the unit test coverage isn't being considered by codecov -- It's too late for me to think about this right now but I'll try to resolve tomorrow.

bickelj (approved):
This is great.

Something that crossed my mind while seeing the `next()` call in multiple endpoints is that some of these errors that have a common message could allow the message to be optional. Then again we probably should make specific-to-that-endpoint or specific-to-this-exact-error helpful messages. Either way, this is something that could be done in future (i.e. should not stop merge of this code).

Another thought is the awkwardness of the `return` statements here and there after calling `next()`. This appears to be a drawback in the express middleware design and not something we have any control over. As opposed to saying `throw X` now I say `next(X); // pause and carefully review code to see if I have to call return or not.... ok no not here.` But again, not something in our design but express' design.

bickelj:
As for code coverage, perhaps a more complex passage rule would make sense. Instead of "decreasing coverage fails the coverage stage" maybe something like "decreasing the coverage by more than 2% or decreasing the coverage below 80% fails the coverage stage" would help.

bickelj:
While looking over the code, I also ran the tests locally, they passed, built it locally and made a couple of requests that exercised the error handling. It all looked great.

slifty:
I *think* this has 100% coverage; I don't quite understand what codecov is complaining about.  I'm... gonna merge since it's approved and we can figure out the rules another time / in another issue later!

> Another thought is the awkwardness of the return statements here and there after calling next()

I agree completely X(

Some of the problem here is actually because of the way *promises* work in node / javascript / typescript.  I had originally hoped I could avoid using `next`, which we actually can possibly do by leveraging a promise-based router, but we can make that change later and I think this is a less invasive next.

Would be worth a voice-based discussion some day I think!

Issue #23, opened 2022-07-05
Avoid use of PAT in CI / CD


bickelj:
Current CI / CD makes use of a Personal Access Token which is less than ideal, because they should expire but then when they do expire it breaks CI / CD (also the CI / CD is a project thing not a personal thing). See https://github.com/PhilanthropyDataCommons/service/pull/22#issuecomment-1175283021 and following.

bickelj:
While looking for something else, I found this https://docs.github.com/en/developers/overview/managing-deploy-keys#server-to-server-tokens

bickelj:
The solution to https://github.com/PhilanthropyDataCommons/deploy/issues/36 may be of interest.

bickelj:
If it will expire in 30 days and that was 29 days ago, expiration is imminent.

bickelj:
I updated the `REPO_ACCESS_TOKEN` at https://github.com/PhilanthropyDataCommons/service/settings/secrets/actions with one that will expire in one year, buying some more time.

bickelj:
https://medium.com/prompt/trigger-another-github-workflow-without-using-a-personal-access-token-f594c21373ef was updated to mention that the rules have changed. https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/ and https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

bickelj:
I haven't looked at this since #91 but I doubt the situation has changed. A PAT is still needed. It will need renewal in July 2023.

bickelj:
Perhaps there is a way to have the `deploy` project watch for a new image (i.e. a pull-to-`deploy`-from-`service`) rather than have the `service` project notify `deploy` (i.e. a push-to-`deploy`-from-`service`).

bickelj:
GitHub now offers [fine-grained Personal Access Tokens](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#fine-grained-personal-access-tokens) associated with organizations. See [solution now live in #343](https://github.com/PhilanthropyDataCommons/service/issues/343#issuecomment-1538894718).

PR #91, opened 2022-10-03
Use secrets.GITHUB_TOKEN for workflow, not a PAT

bickelj:
Issue #23 Avoid use of PAT in CI / CD

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - 'main'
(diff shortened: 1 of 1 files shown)


bickelj:
I am not confident that this will work since the token is associated with a different repo, but it is worth the short experiment.

bickelj:
Indeed: https://github.com/PhilanthropyDataCommons/service/actions/runs/3176001529/jobs/5174768236#step:11:19
> Error: Resource not accessible by integration

bickelj:
So nevermind.

Issue #79, opened 2022-09-09
Update our swagger spec


slifty:
I mistakenly wasn't updating the swagger spec as I defined new endpoints!

The canonicalFields endpoints should be added in, along with populated schema information.

bickelj:
While poking at this, I found the same thing @slifty did here: https://github.com/ajv-validator/ajv/issues/1814#issuecomment-1239574448

slifty:
@bickelj I did a bad thing and didn't link to this issue in a bunch of documentation I created :( -- sorry for the redundant effort; check out this branch to see if there's anything useful in there: https://github.com/PhilanthropyDataCommons/service/blob/noissue-september-demo-lock/src/openapi.json

bickelj:
@slifty Nice, I pushed that version instead.

slifty:
I think this is finished now (we still need to keep it up to date but I believe the current state reflects current implementation)

PR #92, opened 2022-10-03
Add docs for some recently added API endpoints

bickelj:
Issue #79 Update our swagger spec

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -7,18 +7,588 @@
   },
   "components": {
     "schemas": {
-      "Opportunity": {
-        "type": "object"
+      "Applicant": {
+        "type": "object",
+        "properties": {
+          "id": {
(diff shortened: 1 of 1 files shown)


bickelj:
This is a manual update to the swagger spec until we can figure out a better integration with Ajv. I tried to simply import the Ajv `JSONSchema` objects and write them to disk but that failed with the error mentioned in #79.

slifty:
Sorry!  There is a refactored spec in this branch: https://github.com/PhilanthropyDataCommons/service/blob/noissue-september-demo-lock/src/openapi.json

Note that it uses the `readOnly` fields to avoid the need for Post and Get versions of each schema!

bickelj:
@slifty I updated it with the stuff from the other branch.

PR #93, opened 2022-10-05
Authentication: Register and login user with keycloak 

gmurtaza00:
This PR adds the api routes for user authentication. 
- User login ( base_url/users/login ) route requires the following payload
`{
    "username": "",
    "password": "",
}`

- User register ( base_url/users/register ) route requires the following payload
`{
    "username": "",
    "email": "",
    "password": "",
    "confirm_password": ""
}`


slifty:
Thank you for doing all the research and including an implementation!

I'll give this a closer look tomorrow with an actual review but in the mean time one higher level question I have is around whether we should be in the business of doing registration and login ourselves.  The Keycloak documentation says this about the SSO functionality:

```
Users authenticate with Keycloak rather than individual applications. This means that your applications don't have to deal with login forms, authenticating users, and storing users. Once logged-in to Keycloak, users don't have to login again to access a different application. 
```

What do you think of a model where Keycloak handles all that, and all *we* do is accept a `bearer-auth` header token that our middleware checks and converts into a set of roles according to that token?  One benefit of that is we leverage more of what Keycloak has to offer (and therefore we don't have to maintain / test those pieces of the user flow).

In that world I think this PR would be much smaller; it would include:

1. Documentation on how to set up Keycloak (ideally pointing to Keycloak's documentation rather than replicating it in our repo wherever possible)
2. Definition for express middleware that processes bearer tokens and decorates the request with the user's roles

Let me know if talking through by voice would be helpful!

bickelj:
It seems we agree keycloak is the right free/libre open source software auth solution. We are still working on the best way to integrate keycloak. However, the keycloak team recently marked the `keycloak-connect` software deprecated, so this PR is unlikely to be the one.

PR #97, opened 2022-10-17
Add a high-level overview to README.md

kfogel:


kfogel:
@jim-mcgowan (and Sam Caplan, if you're here -- sorry, I don't know your GitHub nick if you have one):

The easiest way to review this is probably to just look directly at the proposed [new version of the PDC Build README.md](https://github.com/PhilanthropyDataCommons/service/blob/noissue-overview-documentation/README.md).  Everything from *"This is the data-handling service layer..."* to *"See also the technical architecture diagram."* is the material I just added.  It is my attempt at a high-level description of what the PDC Build is all about.  Suggestions/edits welcome, and of course there might be some supplemental material we could add that doesn't necessarily belong in the README but would still be useful for conference presentations and such -- I'm happy to help draft such material, if you'd like.  Let me know.

slifty:
Side note: there's clearly something up with our code coverage script -- I'll try to fix that separately but we can just ignore it for this PR since it's a documentation change

README.md
```diff
@@ -5,7 +5,25 @@
 
 # Philanthropy Data Commons Service
 
-This project is related to [Philanthropy Data Commons](https://philanthropydatacommons.org) (PDC). Please visit https://philanthropydatacommons.org for an overview of PDC.
+This is the data-handling service layer for the [Philanthropy Data Commons](https://philanthropydatacommons.org) (PDC).
+
+The PDC is an access-controlled environment in which changemakers and funders can share funding proposals, both for improved efficiency (e.g., offering a "common grant application" in specific domains) and for opening up new possibilities in partnering and alliance-building.  The PDC is designed to enable cross-organizational data sharing while allowing organizations to maintain their own systems, practices, and data standards.
```
slifty, line 10: I know the target audience of this is the existing PDC community -- wondering if we should define "changemakers" somewhere for those who might not be in that loop? (e.g. we could create a glossary of terms in `/docs/GLOSSARY.md` and link to it?)

README.md
```diff
@@ -5,7 +5,25 @@
...
+
+To do this, the PDC maintains a **mapping** between various organizations' data fields and the PDC's internal data representation.  For example, if one organization uses `Proposal Name` and another uses `Title of Proposal`, both of those might map to the PDC field `ProposalTitle`.  The PDC remembers this mapping, translating back and forth as needed so that data flow in and out seamlessly, a unified search interface can be offered, etc.
+
+With the above overview in mind, we can now summarize what this service layer does:
+
+* Store **opportunities** (analogous to RFPs / CFPs) and the **fields** (which hold applicants' responses) associated with those opportunities.
+* Store **proposals** -- the actual responses to opportunities, with long-term consistency provided via the above-described mapping.
```
slifty, line 17: *gulp* we actually call these applications in PDC :( (proposal would be a more clear name IMHO, but we should probably be consistent)

slifty (approved):
Looks great -- one potential change is that you reference `proposals` but in PDC we call those `applications`.  Marked as approved since I'm not asking for the change so much as raising the possibility of the change.

I also wonder if we might want to make a GLOSSARY.md to define various terms of art.

README.md
kfogel: @slifty I like the idea of a glossary -- I don't think this PR needs to block on it, but it's a good idea.  The immediate audience is PDC folks, you are right, and they'll know the terms I'm using.  (The choice of "proposals" instead of "applications" was also deliberate, so that there would be no confusion between an application form and the submitted response to that form.  I realize that within the PDC code base "application" is the official term.)

README.md
kfogel: @slifty We can call them anything we want within the PDC code and documentation, but this high-level explanation has to be clear to outsiders.  We could add a parenthetical saying that within the code they're actually called "applications", but that that means "proposal".  Or we could just switch the PDC internally to using "proposal" (except I presume that there's some reason we didn't use that word in the first place, and whatever that reason is it still holds...?).

kfogel: @slifty "Application" by itself is ambiguous because it could mean either the blank form (the opportunity) or the submitted response that fills in the fields in the blank form (the proposal).  Even in the nearby presence of the word "opportunities", I'm worried that people -- especially those skimming -- might misunderstand, so I used "proposals" here.

kfogel: (Btw, I'll wait until this conversation seems to have resolved itself before merging... I realize you're not blocking the merge, but I want to make sure I'm not doing something dumb before I merge.)

slifty: Re application vs proposal: I don't remember the reasoning; it might have been someone said "proposal is not a good word to use for political reasons" or it might have been just the choice that was made and nobody applicationed -- err, proposed -- an alternative!

This is fine either way.  I think the bolding of it in the README is what threw me; it makes it sounds like "proposal" is our term for a system level / schema level / API endpoint level concept (opportunity, for instance is; as is field), and so it struck me as potentially confusing.

slifty: Looked at the history and indeed: it was originally `Proposal` and I believe Jim requested we make a change on behalf of stakeholders: https://github.com/PhilanthropyDataCommons/service/issues/2#issuecomment-1159186802

slifty: UPDATE: I'm implementing the application endpoint now -- I'm going to implement it with `proposals` since I like that better (and we can change it to application in review if needed)

kfogel: So should we merge this now, @slifty?

kfogel: Rebased just now; I'll merge.

PR #99, opened 2022-10-17
Set up CI test configuration

### LINKED ISSUE #98: Fix code coverage
We have two sets of tests -- unit tests and integration tests -- something about the way code coverage is being generated for those tests seems to be causing complications.

I say seems because maybe it isn't!  Either way this is something that needs looking into to fix (or, if nothing is broken, not fix!)

slifty:
This PR creates a new test configuration for coverage generation / CI.  This allows us to have a consistent comprehensive coverage report regardless of whether the coverage is generated from a dev environment or a CI environment.

Resolves #98

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,19 +36,13 @@ jobs:
       - name: Run build
         run: npm run build
 
-      - name: Run unit tests
-        run: npm run test:unit -- --coverage
-
-      - name: Upload unit test coverage
-        uses: codecov/codecov-action@v2
-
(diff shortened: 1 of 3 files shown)


slifty:
There's a bit more context / thinking in the commit message as well so please be sure to check that out.

slifty:
Any coverage report diffs here are a reflection of the correction / simplification -- not actually a change in what was covered.

The patch coverage seems to be incorrectly configured, so I'll check that out (since if not now... when?)

bickelj (approved):
No objection!

Issue #35, opened 2022-07-12
Implement `POST /applications` endpoint


slifty:
Allows the creation of a new application in the system.

bickelj:
Depends on #64 and #65

slifty:
In working on this I needed to think about how versioning works.  A few pieces on that:

1. An applicant might submit more than one distinct proposal!
2. An applicant might update an existing proposal (create a new version).

To support this I believe we will need to store an "external proposal ID" with a given proposal, so that when the GMS posts the proposal fields our system can manage versioning directly.

slifty:
A parallel question is how will GMSs indicate which field a given value is associated with.

In other conversations we decided that we will not allow clients to talk to us in terms of shortCode; this means they're going to have to look up the application forms, map the short codes themselves, and get our ID.  I think we should stay consistent there (again we may some day decide to provide a quality of life ability to pass a shortCode optionally).  It's also possible that some applications will allow more than one copy of the same canonical field, so this is particularly reasonable to require up front while we're still in prototype mode.

With all that said, for the initial implementation the proposalFieldValues will need the client to pass the `applicationFormFieldId`.

bickelj:
Perhaps the read/GET on one specified entity can return the whole tree of objects for the one requested versus the read/GET for all instances only returns the immediate objects. If that's too inconsistent across the API then maybe an option on the read/GET for one specified entity can return the whole tree of objects. I think for POST we will always want the new instance to have exactly what is needed for that new instance. Those are some of my initial thoughts in response to "A parallel question is how will GMSs indicate which field a given value is associated with."

slifty:
I think that's a great plan (and can be consistent with some of the initial thinking over at https://github.com/PhilanthropyDataCommons/service/issues/101).  And yeah I think that's a fine design rule that we'll want POST to have the correct IDs rather than short codes.

slifty:
I wanted to codify some of the thinking around the POST endpoint here (I'll include it in commits too).

I think our philosophy so far has been to design the first version of these endpoints with a priority towards RESTful design (within reason; note we don't follow this approach for proposalFieldValues for instance).

With that bias in mind, if a GMS wants to create a new version of a proposal they would need to POST to the `proposalVersions` endpoint for a given proposal ID, rather than posting to the `proposals` endpoint.

We may eventually decide to create some quality of life functionality such as:

(A) Allowing a POST to /proposals/{id} to create a new version, or
[...]

PR #100, opened 2022-10-18
Implement GET /proposals endpoint

### LINKED ISSUE #34: Implement `GET /applications` endpoint
Returns a list of all applications int he system.

This should support an optional `opportunityId` querystring parameter which filters the returned applications by opportunityId.

slifty:
This PR creates a new `GET /proposals` endpoint and updates the relevant documentation to reflect the proposed change in name from `Applications` to `Proposals`.

It also sneaks in two little fixes that I noticed when implementing (one is documentation related, the other is just updating a type guard name).

Resolves #34

### DIFF
--- a/docs/ENTITY_RELATIONSHIP_DIAGRAM.md
+++ b/docs/ENTITY_RELATIONSHIP_DIAGRAM.md
@@ -11,7 +11,7 @@ erDiagram
   }
   Applicant {
     int id
-    string externalId 
+    string externalId
     bool hasGivenPermission
   }
   ExternalSource {
@@ -35,15 +35,15 @@ erDiagram
     string title
     datetime dateCreated
   }
-  ApplicationSchema {
+  ApplicationForm {
     int id
     int opportunityId
     int version
     datetime dateCreated
   }
-  ApplicationSchemaField {
+  ApplicationFormField {
     int id
-    int applicationSchemaId
+    int applicationFormId
     int canonicalFieldId
     int order
     string label
@@ -66,7 +66,7 @@ erDiagram
   ApplicationFieldValue {
     int id
     int applicationVersionId
-    int applicationSchemaFieldId
+    int applicationFormFieldId
     int index
     string value
     datetime dateCreated
@@ -77,16 +77,16 @@ erDiagram
     string log
     datetime dateCreated
   }
-  
+
   Applicant ||--o{ Application : submits
   Application }|--|| Opportunity : "responds to"
-  Opportunity ||--|{ ApplicationSchema : establishes
+  Opportunity ||--|{ ApplicationForm : establishes
   Application ||--o{ Outcome : "has"
-  ApplicationSchema ||--|{ ApplicationSchemaField : has
-  ApplicationSchemaField }o--|| CanonicalField : represents
+  ApplicationForm ||--|{ ApplicationFormField : has
+  ApplicationFormField }o--|| CanonicalField : represents
   Application ||--|{ ApplicationVersion : has
   ApplicationVersion ||--|{ ApplicationFieldValue : contains
-  ApplicationFieldValue }o--|| ApplicationSchemaField : populates
+  ApplicationFieldValue }o--|| ApplicationFormField : populates
   Applicant ||--o{ ExternalFieldValue : "is described by"
   ExternalFieldValue }o--|| CanonicalField : "contains potential defaults for"
   ExternalSource ||--o{ ExternalFieldValue : "populates"
@@ -96,15 +96,15 @@ erDiagram
 
 1. An `Applicant` submits an `Application`
 2. An `Application` is a response to an `Opportunity`.  An `Opportunity` represents a given challenge, RFP, etc.
-3. An `Opportunity` establishes an `Application Schema`. An application schema is the set of fields that make up an application.  An `Opportunity` might update its `Application Schema` over time, which is why an `Opportunity` can have many `Application Schemas`.
-4. An `Application Schema` will define many `Application Schema Fields`.
-5. An `Application Schema Field` represents a `Canonical Field`.
+3. An `Opportunity` establishes an `Application Form`. An application form is the set of fields that make up an application.  An `Opportunity` might update its `Application Form` over time, which is why an `Opportunity` can have many `Application Forms`.
+4. An `Application Form` will define many `Application Form Fields`.
+5. An `Application Form Field` represents a `Canonical Field`.
 
 Meanwhile...
 
 6. An `Application` can have more than one `Application Version`.  This occurs as an application is updated or revised.
 7. An `Application Version` contains a set of `Application Field Values`.  These are the responses that were provided by the `Applicant`.
-8. An `Application Field Value` contains a response to a given `Application Schema Field`.  Some fields might allow multiple responses, which is why we provide an `index`.
+8. An `Application Field Value` contains a response to a given `Application Form Field`.  Some fields might allow multiple responses, which is why we provide an `index`.
 
 The thinking is that when a new application is being written, a Grant Management System could ask the PDC "is there any pre-populated data we should use for this organization?"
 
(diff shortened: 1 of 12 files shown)


slifty:
Please note that this PR includes the proposed (ha!) change from `Application` to `Proposal`.

This is based on a discussion in #97; if there's a reason not to make this change then please feel free to push back (it's super easy to make the right decision now, a lot harder to change it later).

src/database/migrations/0006-create-proposals.sql
```diff
@@ -0,0 +1,19 @@
...
+  CONSTRAINT fk_opportunity
+    FOREIGN KEY(opportunity_id)
+      REFERENCES opportunities(id)
+);
+
+COMMENT ON TABLE proposals IS
+  'Sets of values submitted by applicants in responses to an opportunity stored in the PDC.';
+COMMENT ON COLUMN proposals.applicant_id IS
+  'The applicant that submitted this proposal.';
+COMMENT ON COLUMN applicants.opted_in IS
```
slifty, line 18: This snuck in from copy + paste

slifty:
As discussed in https://github.com/PhilanthropyDataCommons/service/issues/35#issuecomment-1284309012 I think we need to have an `externalProposalId` field so that GMS's don't need to worry about tracking our proposal IDs in the event that an applicant updates a proposal.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -11,15 +11,16 @@ erDiagram
...
     bool hasGivenPermission
   }
   ExternalSource {
     int id
     string name
   }
-  Application {
+  Proposal {
     int id
+    string externalId
```
bickelj, line 23: What is the reasoning behind the redundant externalId? One externalId is already present via `int applicationId` below.

bickelj: Ah, I see, it is not the same as the other externalId at all, it is a proposal external ID, not an applicant external ID.

slifty: This is the external ID for the proposal (e.g. not `applicantExternalId`, not the applicant (unless I'm misunderstanding the question).  I didn't add `proposal` to the field name since it's already in the table.  Does that make sense?

bickelj: Yes it makes sense. In the same way that we do not have `applicants.applicantId` or `proposals.proposalId` but rather `applicants.id` and `proposals.id` we have `applicants.externalId` and `proposals.externalId`.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -57,16 +58,16 @@ erDiagram
     int sequence
     datetime dateCreated
   }
-  ApplicationVersion {
+  ProposalVersion {
     int id
     int applicationId
```
slifty, line 63: Whoops, this should say `proposalId`

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -77,61 +78,61 @@ erDiagram
...
+3. An `Opportunity` establishes an `Application Form`. An application form is the set of fields that make up an application.  An `Opportunity` might update its `Application Form` over time, which is why an `Opportunity` can have many `Application Forms`.
+4. An `Application Form` will define many `Application Form Fields`.
+5. An `Application Form Field` represents a `Canonical Field`.
 
 Meanwhile...
 
-6. An `Application` can have more than one `Application Version`.  This occurs as an application is updated or revised.
-7. An `Application Version` contains a set of `Application Field Values`.  These are the responses that were provided by the `Applicant`.
-8. An `Application Field Value` contains a response to a given `Application Schema Field`.  Some fields might allow multiple responses, which is why we provide an `index`.
+6. A `Proposal` can have more than one `Proposal Version`.  This occurs as an application is updated or revised.
```
bickelj, line 106: `... as a proposal is updated...`

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -171,30 +172,30 @@ sequenceDiagram
   API ->>- GMS :  Here they are
 ```
 
-When an applicant begins to fills out an application, the Grant Management System would request all field values known for that `Applicant`.  Which values are returned could be based on business logic; it could be the complete set; it could be restricted to just the fields associated with a given application schema -- these would be implementation details but the schema would support any of them.
+When an applicant begins to fills out an application, the Grant Management System would request all field values known for that `Applicant`.  Which values are returned could be based on business logic; it could be the complete set; it could be restricted to just the fields associated with a given application form -- these would be implementation details but the form would support any of them.
```
bickelj, line 175: `...begins to fill out a proposal...`

src/database/migrations/0006-create-proposals.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE proposals (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  external_id VARCHAR NOT NULL,
```
bickelj, line 3: It might be better to have the `varchar` fields follow the `integer` fields for packing efficiency. Or if the logical order as found in the unique index below is applicant_id then opportunity_id then external_id to make it consistent with that.

src/database/migrations/0006-create-proposals.sql
```diff
@@ -0,0 +1,21 @@
...
+  applicant_id INTEGER NOT NULL,
+  opportunity_id INTEGER NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  UNIQUE(applicant_id, opportunity_id, external_id),
```
bickelj, line 7: I'm trying to guess as the best order here and it depends on what the caller will do. But in general I think the "largest granularity to finest granularity" order would also make sense, which I think is represented here with `applicant_id` followed by `opportunity_id` followed by `external_id`.

src/database/migrations/0006-create-proposals.sql
```diff
@@ -0,0 +1,21 @@
...
+  CONSTRAINT fk_applicant
+    FOREIGN KEY(applicant_id)
+      REFERENCES applicants(id),
+  CONSTRAINT fk_opportunity
+    FOREIGN KEY(opportunity_id)
+      REFERENCES opportunities(id)
+);
+
+COMMENT ON TABLE proposals IS
+  'Sets of values submitted by applicants in responses to an opportunity stored in the PDC.';
```
bickelj, line 17: `...in response to...`

src/database/migrations/0006-create-proposals.sql
slifty: Yeah I think that's what I had in my mind (granularity gradient).  Happy to shift it around if there's a better approach.

src/openapi.json
```diff
@@ -484,27 +488,27 @@
         }
       },
       "post": {
-        "summary": "Add a new application. (In Development)",
+        "summary": "Add a new proposal. (In Development)",
```
bickelj, line 491: I think the "version = 0.1.0" and "pilot" etc. all communicate the idea that this whole API is in development, so I would leave out this comment. At the same time, when exploring the API using the OpenAPI/swagger interface, it is nice to know which parts are expected to work and not. Then again, if it's not going to work, we probably shouldn't have it in the `openapi.json` yet, right?

bickelj: In any case this comment does not block merging in any way. Just a thought.

bickelj (approved):
I skimmed the code changes here, pulled the code, built it, ran migrations, ran tests, explored a little bit with the OpenAPI interface, but I did not try building up an applicant, application form, and proposal soup-to-nuts. From what I can see there is nothing stopping merge here.

src/openapi.json
slifty: Ah -- this was in there for the demo since I wanted to help clarify which API endpoints were working and which were just documentation placeholders.  I'd like to keep it insofar as folks might be looking at the swagger spec and wanting to be able to predict which endpoints they can trust actually exist.  It's about to be taken out though since it's almost done!

src/database/migrations/0006-create-proposals.sql
slifty: Oh; I don't know if this matters at all but I just noticed that we had been putting UNIQUEs after the CONSTRAINTs.

I suspect it doesn't but I'll teak for consistency since why not.

Issue #102, opened 2022-10-19
Make base fields and application forms from spreadsheets


bickelj:
There are several aspects:
* A semi-automated workflow to take spreadsheets and make JSON canonical fields from the data.
* A draft of canonical fields from the spreadsheets given so far.
* Notes of potential issues or conflicts along the way.

bickelj:
I updated test and prod with the slightly better versions of auto-generated shortCode and applicationForm.

bickelj:
This script gets it part of the way there (no special handling for quote characters, inefficient, fails lint, etc.):
```TypeScript
import fs from 'fs';
import CsvReadableStream from 'csv-reader';

const psvInput = fs.createReadStream('file.psv', 'utf8');

console.log('insert into canonical_fields (id, label, short_code, data_type, created_at) overriding system value values')
let firstRowArrived = false;
psvInput.pipe(
  new CsvReadableStream({
    parseNumbers: true,
    parseBooleans: true,
    trim: true,
    delimiter: '|',
    asObject: true,
  }),
).on('data', (row: any) => {
  if (firstRowArrived) {
    console.log(`,(${row.id}, '${row.label}', '${row.shortCode}', '${row.dataType}', '${row.createdAt}' )`);
  }
  else {
    console.log(`(${row.id}, '${row.label}', '${row.shortCode}', '${row.dataType}', '${row.createdAt}' )`);
  }
  console.error('A row arrived: ', row);
  firstRowArrived = true;
}).on('end', () => {
  console.error('No more rows!');
  console.log(';');
});
```

bickelj:
I updated the script to be a little more complete.
<details>
<summary>Updated script</summary>
```TypeScript
import fs, { WriteStream } from 'fs';
import os from 'os';
import CsvReadableStream from 'csv-reader';
import { parse } from 'ts-command-line-args';

interface Args {
  inputFile: string;
  outputFile: string;
}

const args = parse<Args>({
  inputFile: String,
  outputFile: String,
});

const psvInput = fs.createReadStream(args.inputFile, 'utf8');
const sqlOutput = fs.createWriteStream(args.outputFile, 'utf8');

sqlOutput.write(`INSERT INTO canonical_fields (id, label, short_code, data_type, created_at) OVERRIDING SYSTEM VALUE VALUES${os.EOL}`);

let firstRowArrived = false;
psvInput.pipe(
  new CsvReadableStream({
    parseNumbers: true,
    parseBooleans: true,
[...]

Issue #95, opened 2022-10-06
Opportunity data fields for the Miami Foundation/Submittable


jim-mcgowan:
Attached are the opportunity data fields for the Miami Foundation/Submittable.<br>
[Miami Foundation-Submittable Data Fields.xlsx](https://github.com/PhilanthropyDataCommons/service/files/9728072/Miami.Foundation-Submittable.Data.Fields.xlsx)

bickelj:
Thank you!

Issue #94, opened 2022-10-06
Opportunity data fields for the MacArthur Foundation/Fluxx


jim-mcgowan:
Attached are the opportunity data fields for the MacArthur Foundation/Fluxx.<br>
[MacArthur Foundation-Fluxx Data Fields.xlsx](https://github.com/PhilanthropyDataCommons/service/files/9728067/MacArthur.Foundation-Fluxx.Data.Fields.xlsx)

bickelj:
Thank you!

bickelj:
Without objection, this can be considered part of #102

PR #103, opened 2022-10-19
Add a script that helps seed canonical fields

bickelj:
The script accepts one pipe-separated values file as arg1.

Example usage of the script and more is in its comments.

An example output from one of those usages is included in the file named generatedCanonicalFields.json.

Issue #102 Make canonical fields from spreadsheets Issue #94 Opportunity data fields for the MacArthur Foundation/Fluxx Issue #95 Opportunity data fields for the Miami Foundation/Submittable


docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "shortCode": "Proposal Type",
+    "dataType": "string"
+  },
+  {
+    "label": "Proposal Fiscal Sponsorship Org Name",
+    "shortCode": "Proposal Fiscal Sponsorship Org Name",
+    "dataType": "string"
+  },
+  {
+    "label": "Proposal Location of Work (free-form string)",
```
bickelj, line 158: This one looks out of place. The parenthetical note in here looks it is in the upstream spreadsheets.

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+  },
+  {
+    "label": "Org Administrative Lead",
+    "shortCode": "Org Administrative Lead",
+    "dataType": "string"
+  },
+  {
+    "label": "Org Annual Operating Budget",
+    "shortCode": "Org Annual Operating Budget",
+    "dataType": "number"
```
bickelj, line 70: These types are JSON types. If the type does not appear to be an integer, number, boolean, then it becomes a string.

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "dataType": "string"
+  },
+  {
+    "label": "Sub-grantees",
+    "shortCode": "Sub-grantees",
+    "dataType": "string"
+  },
+  {
+    "label": "Accounting Period End Month",
+    "shortCode": "fiscal_month",
```
bickelj, line 189: This is the first Funder-supplied field, so you see the script currently takes the Funder Technical Field Name rather than the (empty) PDC Field Name as the shortCode. For this funder it is more human-friendly, another funder uses universally-unique identifiers.

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "dataType": "number"
+  },
+  {
+    "label": "Purpose Statement",
+    "shortCode": "purpose_statement",
+    "dataType": "string"
+  },
+  {
+    "label": "Org Name",
+    "shortCode": "Org Name",
```
bickelj, line 529: Here we see a duplicate field because two spreadsheets were processed having the same core/PDC fields. When using this JSON to post to the PDC service to seed canonical fields, this is no big deal because the first one to be posted wins. These should be identical in all cases except when there are different types determined via the Funder Field Format.

docs/examples/scripts/extractCanonicalFields.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+# Example that creates JSON from the PSVs and posts each JSON map to the locally
+# running PDC service straightaway without jq and without an intermediate file:
+#
+#     for f in *.psv; do ./extractCanonicalFields.sh $f; done | sed 's/\(.*\),$/\1/g' | while read data; do curl -H "Content-Type: application/json" --data-binary "${data}" localhost:3000/canonicalFields; done
+
+# Strip header, get core fields and perhaps their data types, JSON-ize.
+tail -n +2 $1 \
+    | grep '^\(Organization\|Proposal\)|[a-zA-Z0-9].*|' \
+    | cut -d'|' -f 2,5 \
```
bickelj, line 40: At the moment this uses column 5 (Funder Field Format) to figure out the data type of core fields. But it might be better to have a separate column for the JSON field type so there is no ambiguity.

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
+[
+  {
+    "label": "Org Name",
+    "shortCode": "Org Name",
```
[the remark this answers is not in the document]
slifty: +1

FWIW I was imagining it being `camelCase` (and generally inspried by the label / matching the label).  That has been what we've demoed a few times and it seemed intuitive enough / nobody raised an issue with it.

docs/examples/generatedCanonicalFields.json
slifty: I would suggest we NOT determine the short code based on the funder's technical field names (they'll be mapping those to our short codes either way, and better for us to be consistent in our formatting)

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
+[
+  {
+    "label": "Org Name",
```
slifty, line 3: We will probably want to figure out a naming convention / style guide for canonical fields.  I'd recommend we have a rule that we don't use shortened versions of words in our canonical field names (e.g. this should be "Organization Name" rather than "Org Name")

We might also say that labels should expand acronyms AND define them.  E.g. we might say that "SSN" should be "Social Security Number (SSN)".

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "shortCode": "Org Website",
+    "dataType": "string"
+  },
+  {
+    "label": "Org Entity Type",
+    "shortCode": "Org Entity Type",
+    "dataType": "string"
+  },
+  {
+    "label": "Org Registration Number: EIN / TIN",
```
slifty, line 23: Similar style guide comment -- should we use `:` or should we use `()` for this type of thing.

Maybe this should be `Organization Tax ID Number (TIN / EIN)

docs/examples/generatedCanonicalFields.json
slifty: Specifically these, right? https://json-schema.org/understanding-json-schema/reference/type.html

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "shortCode": "Org Grant Agreement Signatory",
+    "dataType": "string"
+  },
+  {
+    "label": "Org Fiscal Year End Date",
+    "shortCode": "Org Fiscal Year End Date",
+    "dataType": "string"
+  },
+  {
+    "label": "Proposal Name(s)",
```
slifty, line 103: I don't think we need to have implied plurals in the label (*any* of these fields could be duplicated in a given proposal, we have the mechanism for allowing that).  Given that I think we'll want the labels / canonical fields to be singular as a rule.

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "shortCode": "Proposal Name(s)",
+    "dataType": "string"
+  },
+  {
+    "label": "Proposal Primary Contact Name",
```
slifty, line 108: I think the `Proposal` scope could (should?) be considered the default for all canonical fields (e.g. we can remove the `Proposal` prefix).

docs/examples/generatedCanonicalFields.json
slifty: Ah ha! Our first address!  The big kahuna!  This is the moment we must ask ourselves: are field transformations part of the pilot.

If the answer is "no" then we face a choice: do we store atomic address components as separate fields and put the burden of parsing / collating on the GMS's that collect addresses as free form strings, or do we store full address strings as single fields and put the burden of parsing / collating on the GMS that collect addresses as atomic strings.

I think this might warrant a separate issue / discussion.

One approach could be saying "we store EVERYTHING in non-atomic pieces but will eventually provide API utilities for parsing out atoms".  The logic here being that it's trivial for GMS's to construct a string from atoms; it's harder for GMS's to parse atoms from a string.

This also applies to Full Name vs First / Last name.

docs/examples/generatedCanonicalFields.json
```diff
@@ -0,0 +1,822 @@
...
+    "dataType": "string"
+  },
+  {
+    "label": "Proposal Involves Lobbying or Electioneering?",
+    "shortCode": "Proposal Involves Lobbying or Electioneering?",
+    "dataType": "string"
+  },
+  {
+    "label": "IRB review needed?",
+    "shortCode": "IRB review needed?",
```
slifty, line 169: More potential style guide ideas:

1. Name boolean fields the way we would name a variable (e.g. `isBlah` or `hasBlah` or `needsBlah`)
2. Don't include acronyms in short codes

So this shortCode would be `needsInternalReviewBoard`.

docs/examples/generatedCanonicalFields.json
bickelj: I had been going for pure JSON types rather than JSON schema types but JSON schema types would be better. I think the key difference is whether integer is recognized as integer or number, right?

docs/examples/generatedCanonicalFields.json
bickelj: I am on board. Maybe the automated pass can limit the word or character count (so we do not end up with "whatStepsDoYouTakeOrProcessesDoYouHaveInPlaceToEnsureThatTheIndividualsOrOrganizationsThatYouMayPayWithMacArthurGrantFundsOtherThanYourOwnStaffAreNotOnTerroristWatchListsArePaidInComplianceWithApplicableAntiMoneyLaunderingStatutesAndTheFundsWillNotBeUsedForCriminalOrIllegalActivity" as a short code) before a manual pass.

bickelj:
2338adb has the updated script (with new and "improved" regular expressions) that gets closer to the general goals above. I suppose we could have a script that applies replacements for individual shortCodes as a sort of stage two in the processing once this first script pass seems good enough.

bickelj:
dea4892 removes "Proposal".

docs/examples/generatedCanonicalFields.json
slifty: I believe so - yeah why don't we plan for that; heck in some future iteration we could even have our "type" field here contain a full blown json schema definition (which could even help solve our validation question / problem!)

bickelj:
176be1a has changes to the format and usage and is more of a progress commit. The goal is to use the pieces present script out these steps:
0. transform PSVs to (now larger) JSON,
1. post needed canonical fields, 
2. get all existing canonical fields,
3. then post an application form.

bickelj:
2d6bc0c includes a script that takes an xlsx and creates an application form in the PDC end to end.

docs/examples/generatedCanonicalFields.json
bickelj: Suppose a new column were created in the original spreadsheet(s) that includes a PDC shortCode? Then the translation from PDC canonicalField to shortCode would be clearer end-to-end. What do you think?

docs/examples/generatedCanonicalFields.json
bickelj: What do you think of naming conventions being worked out in the original spreadsheets rather than in downstream processing?

docs/examples/generatedCanonicalFields.json
bickelj: I agree this should be a separate issue or discussion.

docs/examples/generatedCanonicalFields.json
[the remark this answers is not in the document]
bickelj: 2338adb generated camelCase shortCodes using up to the first 9 words in the label.

docs/examples/generatedCanonicalFields.json
bickelj: We now keep `integer` as `integer`.

docs/examples/generatedCanonicalFields.json
bickelj: 2338adb has the fix.

docs/examples/generatedCanonicalFields.json
bickelj: This is still kind of an issue where the original spreadsheets are missing the JSON schema type everyone prefers for PDC.

docs/examples/generatedCanonicalFields.json
bickelj: See https://github.com/PhilanthropyDataCommons/service/pull/103#issuecomment-1289402724

docs/examples/generatedCanonicalFields.json
bickelj: See https://github.com/PhilanthropyDataCommons/service/pull/103#issuecomment-1289394036

bickelj:
The downside of yesterday's commit 2d6bc0c is the difficulty of resuming the process from one data transformation stage or running only through a desired stage. For example, when one runs the `buildApplicationForm.sh` script twice, two application forms will be created in PDC. It might be better to run all stages but the last couple (the "post opportunity" and "post applicationForm" stages) and save the JSON, then be able to review and/or manually edit the JSON document, then run the last stage with the cleaned up document.

bickelj:
Some of the "shortCode" values get cut short by non-alphanumeric characters such as ":", and for some of these fields that's a feature and for others it's a bug, it might be better to let them be longer to avoid collisions for super-specific application fields.

bickelj:
Also, now that there is an authentication piece in place, this needs to be updated to provide an API key.

bickelj:
3d284e4 did not update the corresponding examples, fcba217 includes updates.

slifty:
One item that came up on a call today: we should be sure to update the README with documentation on how to run the canonical field import script (and maybe even add an `npm` command as well)

slifty:
So I want to throw out this concept for this PR and hopefully not get boo'd off the stage for only figuring out how to articulate this now!

This is excellent work and the script is super useful as a work product.  I think the ultimate goal here is *NOT* a reproducible script, but rather the set of canonical fields (which in many ways were always intended to be manually generated / more akin to data-driven documentation with functionality impact than anything else).  For the pilot we knew we were going to have to manually map application form fields to canonical fields.

With that in mind, I'd like to suggest a very slight tweak to our final form here...

We don't need the script to generate exactly the canonical fields that are going to be inserted; we DO want the script to generate a processed version of the canonical fields which can then be manually edited to reflect the exact canonical fields we wanted.

I think we may also want a script that creates the associated application form (and application form fields) -- and THAT script would ideally involve no manual intervention (aside from defining field maps)

---

So, all this is to say, I'm going to propose a few changes to the resulting canonical fields but the intent is NOT that this would result in any changes to the script.  In fact, those canonical fields could even become official `seed` data that gets put into the project's source under something like `database/seed/pilot/0001-insert-canonical-fields.sql`

The premise here is that unlike ANY of the other tables, canonical fields are *defined by the PDC* rather than being defined by third party data.  They are then MAPPED to the third party data.

bickelj:
> One item that came up on a call today: we should be sure to update the README with documentation on how to run the canonical field import script (and maybe even add an npm command as well)

I am not sure how I missed this. Yes, this would be helpful. Let me see if I can find the best starting place if it's already there or add one if not.

bickelj:
I think the most straightforward place to start is `docs/examples/scripts/buildApplicationForm.sh` combined with the files available in #94 and #95. If we want it to be run from `npm` I suppose it should be converted to TypeScript or JavaScript. I can at least add to the README now.

bickelj:
> So I want to throw out this concept for this PR and hopefully not get boo'd off the stage for only figuring out how to articulate this now!

Boo! Just kidding!

> This is excellent work and the script is super useful as both a solid example of data processing related to PDC AND a work product. I think the ultimate goal here is _NOT_ a reproducible script, but rather the set of canonical fields (which in many ways were always intended to be manually generated / more akin to data-driven documentation with functionality impact than anything else). For the pilot we knew we were going to have to manually map application form fields to canonical fields.

Yes, I understood this before and it is clearer now. A data product rather than a software product, if you will, is the end goal. As for when that manual generation needs to happen I think I was mistaken on it. I hoped to get more of a sample than N=2 before trying to perform a manual merge but perhaps now is the time!

> With that in mind, I'd like to suggest a very slight tweak to our final form here...
> 
> We don't need the script to generate exactly the canonical fields that are going to be inserted; we DO want the script to generate a processed version of the canonical fields which can then be manually edited to reflect the exact canonical fields we wanted.

Yes and that I think can be extracted from these or adapted from these. The trick is that latter bit and you're right that it must be manual if the canonical field cannot be generated. I think my unspoken until now approach has been to defer the manual stuff until we have more data. In other words, let some fields be duplicated and allow some weird-looking fields for the moment and see what folks with more domain knowledge say about them. I think you have much more domain knowledge than me so I am less sure of myself in making manual data edits.

> I think we may also want a script that creates the associated application form (and application form fields) -- and THAT script would ideally involve no manual intervention (aside from defining field maps)

Right and this might be harder to extract but wouldn't be too hard to create, I suppose.

> So, all this is to say, I'm going to propose a few changes to the resulting canonical fields but the intent is NOT that this would result in any changes to the script. In fact, those canonical fields could even become official `seed` data that gets put into the project's source under something like `database/seed/pilot/0001-insert-canonical-fields.sql`
> 
> The premise here is that unlike ANY of the other tables, canonical fields are _defined by the PDC_ rather than being defined by third party data. They are then MAPPED to the third party data.

Yes, this makes sense, and as you pointed out also forces a manual step into the process because the machine has no way to know which canonical field it should belong (without that manually-created map).

> I should note: I'm also totally comfortable merging this PR as is (with the requested documentation) and exploring the above discussion as a new / separate issue. Whatever you'd prefer @bickelj

It's a work in progress and completely up for review or rework, keeping or throwing, etc. I think there is value in having scripts that demonstrate use of the API so I would probably lean toward adding a couple replacements prior to throwing it all out.

bickelj:
I apparently forgot to update the example script execution in `buildApplicationForms.sh` to have an API key, fixed in a21ebdb.

There probably is a bunch of missing context as well, let me know. For example, should the docs mention the commands to install the required deps, e.g. `apt install jq xlsx2csv`?

slifty:
@bickelj Hmmm.  so I don't run an OS with apt and right now the project does not require that; I'm not sure we want to make apt a dependency.  That said, we *should* list dependencies (and ideally those dependencies would work across multiple platforms).

slifty:
Oh for sure -- (also we can hash this out in Zulip and summarize here) -- yeah the approach you picked makes sense; and the answer will probably just be highlighting that these are example scripts and defining their context.

It might also be the case that we *should* make a separate repository of pilot-data-processing that is in typescript and freely able to add dependencies / etc without impacting the service.

bickelj:
> It might also be the case that we _should_ make a separate repository of pilot-data-processing that is in typescript and freely able to add dependencies / etc without impacting the service.

I lean in this direction now too.

bickelj:
There is a new approach in #164 that supercedes this work. A spreadsheet designed to be used to generate canonical fields and application forms is the new approach.

bickelj:
There might still be some utility in the code in this branch, however, so leaving it until it is fully replaced by other code.

PR #106, opened 2022-10-27
Create POST `/proposals` and POST `/proposalVersions`

slifty:
This PR adds two related endpoints: POST `/proposals` and POST `/proposalVersions`

These endpoints reflect the bread and butter of the PDC API for this phase of PDC integration -- it is the way a third party would register an applicant's proposal in the system in response to a given opportunity / application form.

There were three smaller changes to the overall system required for this, but mostly to the way we ran tests:

1. There was a bug with the way the test schema was set in which the full test pool was not using the same schema.
2. The API schema definition for `Error` was to narrow; the `errors` field has been renamed to `details` to be more generic to different types of errors (there is room for improvement on making our error response format more consistent)
3. We were letting TypeScript act a little unsafe around arrays; that's no longer the case.

Also take note that I found a much better way to mock database responses resulting in less brittle tests!  We have a few tests using the old way which I'll update in a separate PR.

This looks like a huge PR but most of those lines are test lines (I am sticking with our philosophy of "tests should be explicit" which means a lot of redunadnt looking boilerplate for DB setup; some day we might want to consider a better pattern somehow).


slifty:
(Added some missing test coverage that codecov uncovered)

src/database/migrations/0007-create-proposal_versions.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE proposal_versions (
```
bickelj, line 1: What is the advantage of this table versus adding the columns `application_form_id` and `version` to the `proposals` table?

bickelj: I think I see it (going back to the original ER diagram): the proposal outcome can relate to the proposal, not to a particular version of the proposal.

slifty: That's one, example yes.  Also an applicant can submit more than one distinct proposal to the same opportunity which would in turn have multiple versions!

src/__tests__/opportunities.int.test.ts
```diff
@@ -118,15 +118,18 @@ describe('/opportunities', () => {
         RETURNING id, title;
       `);
       logger.debug('opportunityIds: %o', opportunityIds);
-      const sparkleOpportunityId = opportunityIds.rows
-        .filter((it) => it.title === '✨')[0].id;
-      logger.debug('sparkleOpportunityId: %d', sparkleOpportunityId);
+      const sparkleOpportunity = opportunityIds.rows
+        .filter((it) => it.title === '✨')[0];
+      if (sparkleOpportunity === undefined) {
```
bickelj, line 123: This would help in the situation where no rows came back at all, right? Then it would fail with the below message rather than attempting to get a property on undefined? If so, yes, this is much better.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -0,0 +1,1272 @@
...
+describe('/proposalVersions', () => {
+  describe('POST /', () => {
+    it('creates exactly one proposal version', async () => {
+      await db.query(`
+        INSERT INTO opportunities (
+          title,
+          created_at
+        )
+        VALUES
+          ( '🔥', '2525-01-02T00:00:01Z' )
```
bickelj, line 25: I never mentioned it before, and it is definitely not a blocker, but I think it helps to have completely distinct values in test data such that if we are looking at a failed test with db state having a value there is exactly one place in the test code that could have produced that value. Maybe on a rainy day after we have much more important things accomplished I might make updates to this effect.

src/database/migrations/0008-create-proposal_field_values.sql
```diff
@@ -0,0 +1,17 @@
...
+  value VARCHAR NOT NULL,
+  sequence INTEGER NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  CONSTRAINT fk_proposal_version
+    FOREIGN KEY(proposal_version_id)
+      REFERENCES proposal_versions(id),
+  CONSTRAINT fk_application_form_field
+    FOREIGN KEY(application_form_field_id)
+      REFERENCES application_form_fields(id)
+);
```
bickelj, line 14: We eventually will want an index of some kind in here. Maybe a unique constraint/index across (proposal_version_id, application_form_field, sequence)? Not that this is particularly restrictive or meaningful but it would be some index instead of no index. You could still have the same form field id repeated using sequence.

src/database/migrations/0008-create-proposal_field_values.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE proposal_field_values (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  proposal_version_id INTEGER NOT NULL,
+  application_form_field_id INTEGER NOT NULL,
+  value VARCHAR NOT NULL,
+  sequence INTEGER NOT NULL,
```
bickelj, line 6: Four 32-bit integers in a row followed by varchar is probably better packed than three 32-bit integers followed by varchar (64-bit?) followed by 32-bit followed by 64-bit timestamp. In short, sequence and value could be swapped.

src/database/queries/proposalVersions/insertOne.sql
```diff
@@ -0,0 +1,22 @@
+INSERT INTO proposal_versions (
+  proposal_id,
+  application_form_id,
+  version
+) VALUES (
+  :proposalId,
+  :applicationFormId,
+  COALESCE(
+    (
+      SELECT MAX(pv.version) + 1
```
bickelj, line 10: There is probably a race condition here, but I am not sure what bad thing (if any) would happen even when it is encountered. It should just be a unique constraint violation on one of the transactions or something. And I doubt we will see such a thing any time soon.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -0,0 +1,224 @@
...
+  Proposal,
+  ProposalVersion,
+  ProposalVersionWrite,
+  ProposalFieldValue,
+  ProposalFieldValueWrite,
+} from '../types';
+
+const logger = getLogger(__filename);
+
+const assertApplicationFormExistsForProposal = async (
```
bickelj, line 35: On the one hand, this is a good validation and it looks like it keeps things correctly synchronized. On the other hand, it seems like the relations are wonky if the validation is needed. I probably had the same question about this in the past that I cannot find in GH comments, sorry, so I'll ask again: what would go missing if a proposal was in response to an application form rather than in response to an opportunity? The opportunity can be deduced from the application form and the possibility of mixing things up vanishes. A similar thing might be possible for application form fields, though (posting a field value and field that does not belong to that form), and I cannot come up with a similar suggestion there.

I have a vague memory of you suggesting that we would relate proposalVersion to applicationForm and that is done here and is probably fine.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -0,0 +1,224 @@
...
+        entityType: 'ApplicationForm',
+        entityId: applicationFormId,
+        contextEntityType: 'Proposal',
+        contextEntityId: proposalId,
+      },
+    );
+  }
+};
+
+const assertProposalFieldValuesMapToApplicationForm = async (
```
bickelj, line 94: 👍

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -0,0 +1,224 @@
...
+    assertApplicationFormExistsForProposal(
+      req.body.applicationFormId,
+      req.body.proposalId,
+    ),
+    assertProposalFieldValuesMapToApplicationForm(
+      req.body.applicationFormId,
+      req.body.fieldValues,
+    ),
+  ]).then(() => {
+    db.transaction(async (transactionDb) => {
```
bickelj, line 159: Fancy!

bickelj (approved):
This is excellent!

src/__tests__/opportunities.int.test.ts
slifty: Correct!

src/__tests__/proposalVersions.int.test.ts
slifty: This is a good suggestion (appreciate the note to not block on it for now, as I agree it would be worth fixing en masse)

src/database/migrations/0008-create-proposal_field_values.sql
slifty: This is a good call; I'll add the unique.

src/database/queries/applicationForms/selectLatestByOpportunity.sql
```diff
@@ -0,0 +1,9 @@
+SELECT af.id AS "id",
+  af.opportunity_id as "opportunityId",
+  af.version AS "version",
+  af.created_at AS "createdAt"
+FROM application_forms af
+WHERE af.opportunity_id = :opportunityId
+LIMIT 1
+ORDER BY af.created_at DESC
```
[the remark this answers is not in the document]
slifty: Thank you -- this was cruft from a local iteration, I'm taking it out!

src/handlers/proposalVersionsHandlers.ts
slifty: So, ERD aside, at a conceptual level a proposal (the concept of "hey I want to apply!") is always in response to an opportunity and that a proposal version (the concept of "here is some information!") will always be in responce  to a specific request for information defined within that opportunity (e.g. specific application form).

By linking a *PROPOSAL* to an application form we couple the desire to apply itself to a specific request for information as opposed to the more general opportunity.  This doesn't necessarily result in lost data on surface, but if we think about potential life cycles of the data I think it becomes a bit more clear: if someone wanted to delete an obsolete application form that would break relationships between a proposal and an opportunity if we link through that entity. (We won't be doing deletions at this point, but from a ERD perspective I think that helps clarify?)

A non-deletion example: if an opportunity updates its application forms and wants to collect some additional information / adds a question to an application form, this would appropriately make a proposal *version* obsolete, but the *proposal* is still in fine order (it's just that it doesn't yet have a version that is responsive to the most recent request for information / application form).

With the current model the proposal is linked to an opportunity (which may have any number of application forms / application form versions) and the proposal version is the only thing locked into a specific application form version.

On a related / side note; in one of our feedback sessions the concept of having more than one *ACTIVE* application form associated with an opportunity was raised (e.g. people should fill out form X, Y, and Z which individually might have evolving version!)  This would mean creating another entity under "ApplicationForm" and allowing a single proposal to be responsive to more than one application form (thus making it I think more clear why the opportunity is important, though we don't need to think about this here / this is not the leading reason for the proposed design).

... I am worried that none of the above makes any sense 😭 

---

I don't know if psql allows for foreign key relationships *THROUGH* other tables.  I'll look into that, as it would allow us to avoid this check.

bickelj: That all makes sense, yes, and explains the utility of modeling and relating a proposal to an opportunity separately from application form.

src/database/queries/proposalVersions/insertOne.sql
slifty: Oh good point -- does psql not prevent multiple writes at a time / is that why sequences are implemented the way they are 😬.

There is another instance of this type of logic (over in application forms) -- so I think it's already a problem that needs to be solved.  I'll create an issue to capture it.

src/handlers/proposalVersionsHandlers.ts
slifty: I spent a bit of time trying to think through this (and pinging some friends about the broader problem space).  The verdict seems to be there is no great solution.

It would be possible to create a trigger that validates this relationship on insert.  there is also the concept of a `CHECK` but psql documentation is pretty clear that CHECK should not be used to validate anything outside of the columns being directly inserted:

```
PostgreSQL does not support CHECK constraints that reference table data other than the new or updated row being checked. While a CHECK constraint that violates this rule may appear to work in simple tests, it cannot guarantee that the database will not reach a state in which the constraint condition is false (due to subsequent changes of the other row(s) involved). This would cause a database dump and restore to fail. The restore could fail even when the complete database state is consistent with the constraint, due to rows not being loaded in an order that will satisfy the constraint. If possible, use UNIQUE, EXCLUDE, or FOREIGN KEY constraints to express cross-row and cross-table restrictions.

If what you desire is a one-time check against other rows at row insertion, rather than a continuously-maintained consistency guarantee, a custom trigger can be used to implement that. (This approach avoids the dump/restore problem because pg_dump does not reinstall triggers until after restoring data, so that the check will not be 
```

I think that maybe for now the problem might not have a good solution and this runtime check is our best bet (it does mean that, for instance, a proposal version COULD be associated with an invalid application form in the event that the application form's opportunity relationship is changed...)

src/__tests__/proposalVersions.int.test.ts
slifty: Captured in https://github.com/PhilanthropyDataCommons/service/issues/108

src/database/queries/proposalVersions/insertOne.sql
slifty: Captures in: https://github.com/PhilanthropyDataCommons/service/issues/109

PR #107, opened 2022-10-28
Keycloak: middleware to verify api-key against keycloak

gmurtaza00:
- This middleware function receives the user api key from request headers through _x-api-key_ header and verifies it against keycloak by hitting a custom keycloak endpoint.

- It returns 200 response upon correct api-key or error message otherwise.


src/errors/ApiKeyError.ts
```diff
@@ -0,0 +1,3 @@
+import { ValidationError } from './ValidationError';
+
+export class ApiKeyError extends ValidationError {}
```
slifty, line 3: I don't think we need this to extend ValidationError (those are errors related to data structure validation)

I would probably just call this an `AuthenticationError` (and have it extend the built in `Error` object).

src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,23 @@
+import http from 'http';
+import { ApiKeyError } from '../errors';
+import type {
+  NextFunction,
+  Request,
+  Response,
+} from 'express';
+
+export const checkApiKey = (req: Request, res: Response, next: NextFunction): void => {
```
slifty, line 9: Could you add some integration tests for this endpoint in order to get 100% diff coverage?  You can check out the other integration tests for some good patterns,

src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,23 @@
...
+  const apiKey = req.headers['x-api-key'] ?? '';
+  const authServer = process.env.AUTH_SERVER_URL ?? '';
+  const realmName = process.env.REALM_NAME ?? '';
+  http.get(`http://${authServer}/realms/${realmName}/check?apiKey=${apiKey.toString()}`, (authResponse: http.IncomingMessage) => {
```
slifty, line 13: There is a lot going on in this line, and if the dev doesn't set those env vars it'll result in some confusing behavior -- I'd suggest:

1. Creating a utility function that returns the authServer base URL *AND* throws an error if either of the environment variables are set, (with a message explaining that a key environment variable isn't set).

2. Use [URLSearchParams](https://stackoverflow.com/a/49701878/159522) to define the querystring in that way rather than constructing the string manually (this becomes a bit more legible and also less error prone int he long run).

slifty (changes requested):
This is looking close and the approach is excellent!

I have some suggestions around the error taxonomy as well as a suggested changes to the way the auth URL is defined in the name of covering our error bases.

Also please add tests so we have 100% coverage on the diff!

src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (apiKey === '') {
+    next(new AuthenticationError(
+      'API key not provided in the header',
+    ));
+  } else if (authServer === '' || realmName === '') {
+    next(new EnvVariableError(
+      'Environment variables not set.',
+    ));
+  } else {
+    const url = new URL(`http://${authServer}/realms/${realmName}/check`);
```
bickelj, line 25: This probably should leave the protocol (http or https) to the environment variable such that either can be specified.

src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,41 @@
...
+  NextFunction,
+  Request,
+  Response,
+} from 'express';
+
+export const checkApiKey = async (req: Request, res: Response, next: NextFunction):
+Promise<void> => {
+  const apiKey = req.headers['x-api-key'] ?? '';
+  const authServer = process.env.AUTH_SERVER_URL ?? '';
+  const realmName = process.env.REALM_NAME ?? '';
```
bickelj, line 15: `REALM_NAME` could be `AUTH_REALM_NAME`, what do you think?

slifty:
Wanted to mention this here -- I would suggest using [nock](https://www.npmjs.com/package/nock) to mock the KeyCloak API responses so that the tests can run without a working keycloak server.  This is appropriate since these are `unit` tests not `int` tests so integration isn't expected.  (We may some day consider integration tests as well)

bickelj:
It seems we agree keycloak is the right free/libre open source software auth solution. We are still working on the best way to integrate keycloak. However, I think the team (maybe I'm just speaking for myself) wasn't 100% comfortable with this approach, so I doubt this PR will be the one that integrates keycloak. If this is incorrect, please re-open this PR.

PR #113, opened 2022-11-08
GMS apikey authentication through local file

gmurtaza00:
As discussed in the meeting, for a temporary solution keys will be stored and verified from a local file.
I have added some keys (50 characters alphanumeric) for the sake of test cases, please change them in production.


src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,29 @@
...
+  NextFunction,
+  Request,
+  Response,
+} from 'express';
+
+export const checkApiKey = (req: Request, res: Response, next: NextFunction): void => {
+  const apiKey = req.headers['x-api-key'] ?? '';
+  if (apiKey === '') {
+    next(new AuthenticationError(
+      'API key not provided in the header',
```
bickelj, line 13: Add the exact header expected in the message, e.g. 'API key not provided in header "x-api-key"'.

bickelj (commented):
@slifty should this be merged separately from the changes to make use of it in the whole app or should it be one PR? I can see it either way.

src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const checkApiKey = (req: Request, res: Response, next: NextFunction): void => {
+  const apiKey = req.headers['x-api-key'] ?? '';
+  if (apiKey === '') {
+    next(new AuthenticationError(
+      'API key not provided in the header "x-api-key"',
+    ));
+  } else {
+    try {
+      const data = fs.readFileSync('test_keys.txt', 'utf8').split('\n');
```
bickelj, line 17: OK, now I see why it was not originally `test_keys.txt` :smile:

Hmm, I'm still thinking about how to resolve this. What went wrong with the tests when using an env var?

slifty: FWIW we are able to use an environment variables in tests using the type of mocking technique described in https://www.webtips.dev/how-to-mock-processenv-in-jest

src/middleware/__tests__/apiKeyChecker.unit.test.ts
```diff
@@ -0,0 +1,47 @@
+import fs from 'fs';
+import { AuthenticationError } from '../../errors';
+import { checkApiKey } from '../apiKeyChecker';
+import type { NextFunction, Request, Response } from 'express';
+
+describe('Authorization middleware', () => {
```
bickelj, line 6: Only because I heard someone else say it earlier: authentication here rather than authorization :nerd_face:

src/middleware/apiKeyChecker.ts
bickelj: I'm trying it out locally to see what you mean.

src/middleware/apiKeyChecker.ts
```diff
@@ -0,0 +1,29 @@
...
+      if (data.includes(apiKey.toString())) {
+        next();
+      } else {
+        next(new AuthenticationError(
+          'Invalid api key provided',
+        ));
+      }
+    } catch (err) {
+      next('Internal server error');
```
slifty, line 26: Looks like we have a path that needs a test.  Let's make sure this case processes as expected as well.

src/middleware/__tests__/apiKeyChecker.unit.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  let mockRequest: Partial<Request> = {};
```
slifty, line 7: Is there value in defining this outside the scope it is used?

I'm a little worried that this creates the possibility for odd outcomes when we break out of bands / start running things in parallel though I don't know if that worry is justified.

src/middleware/__tests__/apiKeyChecker.unit.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  const mockResponse: Partial<Response> = {};
+  const nextFunction: NextFunction = jest.fn();
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('without x-api-key header', async () => {
+    const expectedResponse = new AuthenticationError('API key not provided in the header "x-api-key"');
```
slifty, line 16: This is a minor point and please feel free to ignore it -- defining the expected response up here removes it a little from when the value is actually used / checked.  It isn't reused, so doing it like this doesn't save us lines of code, for instance.  Maybe worth just having the expected response defined inline on line 23 (this is how our other tests are written for what it's worth).

At the very least I think that we should follow the pattern of:

1. Arrange
2. Act
3. Assert

Described [here](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/)

Specifically we would have `Arrange` be about defining the values needed to act (not to assert).

TL;DR let's either move this definition down to be part of the "assert" block of the tests, OR (ideally IMO) just define them inline as part of the assert.

bickelj:
I pushed a commit that has a sketch of what I was going for with the environment variable but obviously I didn't run lint locally, my mistake. Feel free to overwrite that commit (force push the same branch without that change), I just wanted to get it out there before calling it a night.

src/middleware/apiKeyChecker.ts
slifty: Good call -- and yes agreed they should supply the key

src/middleware/__tests__/apiKeyChecker.unit.test.ts
slifty: (I will get this wrong a billion times plz correct me every time!)

bickelj:
It's a little tricky with the environment variable merge that happens in tests. If you do not specify a variable in the `.env.test` but do specify one in `.env` then the tests will pick up that one, which in this case causes things to go awry. You're better off leaving both out or specifying both. I added something to the README.md in my next push that should help.

bickelj:
I am still a little worried about the inclusion of a static `test_keys.txt` despite the name and despite the documentation. It is still too easy to use those accidentally. Maybe if the keys could be generated automatically for tests it would mitigate the risk of accidental use. Do you have other ideas, @slifty or @gmurtaza00?

gmurtaza00:
> I am still a little worried about the inclusion of a static `test_keys.txt` despite the name and despite the documentation. It is still too easy to use those accidentally. Maybe if the keys could be generated automatically for tests it would mitigate the risk of accidental use. Do you have other ideas, @slifty or @gmurtaza00?

Can you please elaborate 'accidentally' ?

bickelj:
> Can you please elaborate 'accidentally' ?

Sure, it would be like setting `API_KEYS_FILE=test_keys.txt` in `.env.test` then pasting the same thing in `.env` instead of taking the step of generating a separate file and making sure they're different. Less accidentally, since the `test_keys.txt` file is already there it is a lower barrier than making a new file.

bickelj:
Thanks to @slifty suggestion earlier I think I was able to figure out coverage for the remaining missing lines, see 8396205.

src/middleware/apiKeyChecker.ts
bickelj: This should be covered in the most recent commit.

bickelj:
I think we want the bulk of the routes to be protected, right? At least every POST and almost every GET. I can add another commit to do this.

bickelj:
Getting the routes protected is straightforward, getting OpenAPI docs to make it clear and easy is less straightforward but I am figuring it out.

bickelj:
I also added redaction of the x-api-key in the logger setup, tried out the OpenAPI setup on all routes, it looks OK now. You can see a little lock next to each protected route and you can authenticate by clicking it. Also there is an "authorize" [sic] button available at the top to authenticate once and for all on that page.

Another missing thing is more specific HTTP response codes, e.g. 401 instead of 500 when the x-api-key header is not in the request (as well as documentation of the same).

bickelj:
I think pushing what I have so far is still worthwhile to get review from @gmurtaza00 and @slifty.

bickelj:
That last push was a rebase on main.

bickelj:
Rebased again.

bickelj (approved):
Good enough. But some of the changes are my own, so it would be good to get others as well.

Issue #114, opened 2022-11-09
Missing sequence in ApplicationFormField in openapi.json


mortimal1:
ApplicationFormField  - looks like position was changed to sequence but not changed in openapi.json
- also in ProposalFieldValue

bickelj:
Thank you for finding this as well. I'm taking a look.

bickelj:
In the main branch as of 6ba6f6817c66c5d686ad4b36318ff71768cd73d7 it looks like `openapi.json` and the database tables from `src/database/migrations` are in sync for ApplicationFormField and I think I see the same table definition in 14d21ba49bcf5053feb78bfe48ead58b138d63a0 (where it was added). But I see the `docs/ENTITY_RELATIONSHIP_DIAGRAM.md` is not in agreement.

In ProposalFieldValue I see "sequence" in the ER diagram, "sequence" in the db, but then "position" in the `openapi.json`.

bickelj:
I added a fix in 39ac792 for the specific issues above, but obviously we need to review the rest of the fields more carefully too, which I will do next. For some reason the "create pull request" button failed to work as well (I'll try again soon).

bickelj:
261061c overwrites e9ac792 and makes the ER diagram, OpenAPI documentation, database schema, and service code more consistent. See PR #115.

Thank you for taking the time to report these issues and sorry for the inconsistency!

bickelj:
Now that #115 is merged, it should be resolved in the main branch. Thanks again.

PR #115, opened 2022-11-09
Make fields in docs, code, and schema consistent

bickelj:
Make fields in docs, code, and schema consistent

* Instead of "order" or "sequence", use the word "position."
* Update some of the fields in the ER diagram.
* Make the order of Proposal fields in code more consistent with schema.

Issue #114 Missing sequence in ApplicationFormField in openapi.json


bickelj:
There is more to come on this one.

Issue #112, opened 2022-11-07
Proposal Version `version` attribute should be read only


slifty:
Version number is not provided by the client but is instead assigned by the PDC on creation of the new version; we just need to update the OpenAPI spec to reflect this.

PR #117, opened 2022-11-11
Mark proposalVersion.version read-only in OpenAPI

bickelj:
Issue #112

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -103,7 +103,8 @@
           },
           "version": {
             "type": "integer",
-            "example": 1
+            "example": 1,
+            "readOnly": true
           },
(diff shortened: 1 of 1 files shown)


src/openapi.json
```diff
@@ -103,7 +103,7 @@
           },
           "version": {
             "type": "integer",
-            "example": 1
```
slifty, line 106: There is still value in providing an example (since the schema is used for GET as well and the example populates sample responses)

bickelj: Got it. I put it back in.

bickelj:
Since I updated based on the comment, I infer that this is OK : )

Issue #96, opened 2022-10-11
Authentication


slifty:
We need:

1) The ability for users of our API to authenticate (with a role).
2) ~~The ability for our API to reject access to specific endpoints depending on the role of the authenticated user.~~

We're thinking [KeyCloak](https://www.keycloak.org/) is the way to go here.

For this iteration I think our goal is ~~*API key based authentication*~~ OAuth 2 authentication.

The intent would be that tell API clients to pass a [`Authorization: Bearer ######`](https://swagger.io/docs/specification/authentication/bearer-authentication/) header token to *all calls* to our system.  That token is then used to authenticate the user and identify their roles.

~~This article could be useful for inspiration (but may not be the exact pattern to use): http://www.zakariaamine.com/2019-06-14/extending-keycloak/~~

bickelj:
Much of the related work can be found in PR #93, PR #97, https://github.com/PhilanthropyDataCommons/keycloak-extensions/pull/1, https://github.com/PhilanthropyDataCommons/deploy/pull/50.

To summarize what I have learned when integrating with keycloak:
* It was straightforward to get keycloak up and running using the guides at https://www.keycloak.org
* If we terminate TLS at the load balancer only (but have keycloak serve plain HTTP within the docker overlay network) then we cannot seem to activate keycloak "production mode" because it wants the backend to also serve via TLS.
* If I use spaces in a realm name in keycloak, I get "Invalid parameter: redirect_uri", so instead of "Philanthropy Data Commons" it is better to use "pdc" as the realm.
* When I included the jar from https://github.com/PhilanthropyDataCommons/keycloak-extensions/pull/1 in the keycloak "providers" directory with keycloak 20.0.0, I got the following warnings: `2022-11-03 22:32:33,608 WARN  [org.keycloak.services] (build-36) KC-SERVICES0047: check (org.philanthropydatacommons.providers.ApiKeyResourceProviderFactory) is implementing the internal SPI realm-restapi-extension. This SPI is internal and may change without notice`, `2022-11-03 22:32:34,722 WARN  [org.keycloak.services] (build-36) KC-SERVICES0047: api-key-registration-generation (org.philanthropydatacommons.providers.RegisterEventListenerProviderFactory) is implementing the internal SPI eventsListener. This SPI is internal and may change without notice`
* It seems like we should not have to extend keycloak to get the functionality we need.
* It also seems like we should not have to use private or internal SPIs to get the functionality we need.
[...]

PR #124, opened 2022-11-23
Validate Bearer JWT for authentication

bickelj:
<details><summary>(Outdated original message)</summary>

~~As of this commit, only for authentication, but eventually should be authorization as well.~~

~~This is a progress commit demonstrating a change using the /applicants endpoint. It should work in conjuction with a Bearer token issued by keycloak using the OAuth 2.0 Device Authorization Grant.~~
</details>

Expect a JSON Web Token (JWT) in the `Authorization` header for
protected resources. Validate the JWT using the configured realm key.
This allows the standard OpenID Connect protocol to be used for
authentication to an auth server, namely Keycloak, and the resulting
token to be valid for calls to this service. For example, a user of a
browser-based UI can log in using the standard Authorization Code flow
and that UI can make calls to this back-end service using the JWT.

For audit trail purposes, it is still important to log some portion of
the JWT. A JWT, while expected to be short-lived, is still useful as
credentials in its complete form. Part of the validation of a JWT is
signature validation. To keep the audit trail while preventing replay
from a server log, redact the signature of the JWT in logs.

The OpenAPI documentation is a normal browser-based Client. This
change expects the documentation Client to have a particular name,
'pdc-openapi-docs', to be configured in the 'pdc' realm of the
authentication service.

Issue #96 Authentication


bickelj:
There are some awkward things in here that will need cleanup, such as the TypeScript types coming into production/non-dev dependencies, the weird need to wrap the expressjwt function or else fail linting, and probably more. Also it does not check if a user is in a group but it validates that the token was indeed issued by the issuer configured at server launch time. I pushed in order to show progress as well as get feedback on this approach.

bickelj:
8e0bd2e has a rebase and also an additional commit demonstrating authorization (on top of authentication) from the application perspective.

bickelj:
It seems we agree keycloak is the right free/libre open source software auth solution. We are still working on the best way to integrate keycloak. This PR might be the way to go but there is also an alternative (or complement) https://authjs.dev/ to consider.

bickelj:
The push of 83285ab includes a rebase on main, updates to dependencies, and removal of the `ErrorWithStatus` (to be considered separately if needed).

package-lock.json
```diff
@@ -13,6 +13,10 @@
         "ajv-keywords": "^5.1.0",
         "dotenv": "^16.0.3",
         "express": "^4.18.2",
+        "dotenv": "^16.0.1",
+        "express": "^4.18.1",
```
bickelj, line 17: Repeated dependency declarations here, whoops. I am surprised npm did not error on this.

I am still trying to figure out why several `@types` dependencies are ending up outside the dev dependencies area when doing `npm install`.

package-lock.json
```diff
@@ -3328,6 +3366,24 @@
         "node": ">= 0.10.0"
       }
     },
+    "node_modules/express-jwt": {
+      "version": "8.4.0",
+      "resolved": "https://registry.npmjs.org/express-jwt/-/express-jwt-8.4.0.tgz",
+      "integrity": "sha512-AvIkHI6+wuwHQTgfnl4kEPWeMKo5yw4FnXJJK+jf/PRWAflmuJKTs06ENRNJ6sCQceIUVqAi/fy8Nav8alnv0w==",
+      "dependencies": {
+        "@types/jsonwebtoken": "^9",
```
bickelj, line 3374: This could be the culprit, new non-dev dependency `express-jwt` declares dependency on `@types/jsonwebtoken` without saying that one should be dev dependency.

package-lock.json
```diff
@@ -5139,6 +5210,56 @@
...
+        "safe-buffer": "^5.0.1"
+      }
+    },
+    "node_modules/jwks-rsa": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/jwks-rsa/-/jwks-rsa-3.0.1.tgz",
+      "integrity": "sha512-UUOZ0CVReK1QVU3rbi9bC7N5/le8ziUj0A2ef1Q0M7OPD2KvjEYizptqIxGIo6fSLYDkqBrazILS18tYuRc8gw==",
+      "dependencies": {
+        "@types/express": "^4.17.14",
+        "@types/jsonwebtoken": "^9.0.0",
```
bickelj, line 5244: Likewise here: `jwks-rsa`, a dependency, does not say these are dev dependencies.

bickelj: Hmm, found this comment: https://github.com/auth0/node-jwks-rsa/pull/80#pullrequestreview-244880314
Which references this: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies
I thought most libraries let downstream libraries use the typescript types optionally, though.

bickelj:
a9b72d1 rebases on main, updates dependencies, and manually puts the `@types/*` dependencies in dev dependencies by adding the dev flag in `package-lock.json` because it sounds like auth0 is going to keep those in production dependencies.

bickelj:
The branch at 45db64a has been rebased on main again.

src/auth/jwtOptions.ts
```diff
@@ -0,0 +1,34 @@
...
+    jwksRequestsPerMinute,
+    jwksUri,
+    handleSigningKeyError: (err, cb) => {
+      if (err instanceof SigningKeyNotFoundError) {
+        cb(new Error('Could not fetch certificates from JWKS URI'));
+        return;
+      }
+
+      cb(err);
+    },
```
jasonaowen, line 27: Can you say more about this? Why is this better than the default?

jasonaowen (commented):
Could you please rebase this on `main` and resolve the merge conflicts, @bickelj?

bickelj:
@jasonaowen Thank you for the review. This last push was not yet in response to your review but may have incidentally resolved a couple of your comments. I will address your comments after taking care of an update to #164.

src/__tests__/applicants.int.test.ts
```diff
@@ -1,30 +1,58 @@
...
+      realm_access: {
+        roles: [
+          'users-role',
+          'default-roles-pdc',
+        ],
+      },
+    });
+    dummyApiKey = { Authorization: `Bearer ${accessToken}` };
+    logger.debug(`${process.env.AUTH_SERVER_REALM_URL ?? 'NA'}`);
+    logger.debug(`${accessToken}`);
```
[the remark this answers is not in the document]
bickelj: I don't see them in my most recent push, so probably not.

package.json
```diff
@@ -53,9 +54,11 @@
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-sort-exports": "^0.8.0",
     "jest": "^28.1.1",
+    "jsonwebtoken": "^9.0.0",
```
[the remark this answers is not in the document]
bickelj: Before [the suggested change above to the type of jwtOptions](https://github.com/PhilanthropyDataCommons/service/pull/124#discussion_r1128674483) there was only a `type` import from `jsonwebtoken`, right? So it would come from `@types/jsonwebtoken` and all the `@types` are only `devDependencies`. Granted, `jsonwebtoken` is a transitive dependency of one of our `dependencies` (`espress-jwt`), so it does end up in the `dependencies` in `package-lock.json`.

Now with [the suggested change above to the type of jwtOptions](https://github.com/PhilanthropyDataCommons/service/pull/124#discussion_r1128674483) there is no direct dependency on `@types/jsonwebtoken` nor `jsonwebtoken` in `src/auth/jwtOptions.ts`, but there is one import in `src/middleware/jwtTokenChecker.ts` of the `JwtPayload` type which is needed at compile time. Therefore `@types/jsonwebtoken` can stay at compile time (`devDependencies` in package.json) but we don't need a reference to plain `jsonwebtoken` at all in the `package.json` It might still arrive in `package-lock.json` in non-dev `dependencies` due to being a transitive dependency, though.

src/auth/jwtOptions.ts
bickelj: I am trying to find where I got this (or why I put it there), but looking again, I do not see why it should be included. The only possible advantage we could get would be if we used our own Error type but we do not. I will see if anything bad happens when removing it.

src/middleware/jwtTokenChecker.ts
```diff
@@ -0,0 +1,41 @@
...
+  realm_access: {
+    roles: string[];
+  };
+  preferred_username: string;
+}
+
+const verifyRoleMembership = (
+  req: JwtRequest<JwtPayload & KeycloakAuthPayload>,
+  requiredRole: string,
+): void => {
```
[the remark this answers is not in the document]
jasonaowen: I do still prefer, please! You're probably right that we do need authz as part of the pilot, but either way I think it'll be easier if they're separated.

package.json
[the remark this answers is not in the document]
jasonaowen: I'll look at this more deeply later, but in general the policy is: if you `import` (not `import type`) from a dependency in non-test code, it goes in `dependencies`, even if it's already a transitive dependency; if you `import` from a dependency in test code, or `import type` from a dependency in non-test code, it goes in `devDependencies`; if you never import it, then leave it to your direct dependencies to bring it in as a transitive dependency (except in some special cases).

I think you already know that, but I mention it just in case; more likely, what's happening is that there's been a bit of thrash on this branch and we're talking about different states of the world!

jasonaowen:
We'll also need to update the Swagger docs as part of this, since it currently knows about the API key header.

bickelj:
160cf6f should pass lint but not tests. It is still a work in progress.

package.json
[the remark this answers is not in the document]
bickelj: It sounds like we're on the same page. I think this is resolved. If not, we can revisit.

README.md
```diff
@@ -126,15 +126,44 @@ LOG_LEVEL=trace npm run test
...
+### Authentication and Authorization
+
+A valid Bearer JSON Web Token (Bearer JWT) is required in requests to the PDC service. The exception is the API documentation page. These environment variables configure the expected realm and JWKS issuer keys:
+- `AUTH_SERVER_ISSUER`: the URL for the issuer (this is the "realm" URL in Keycloak terminology).
+- `AUTH_SERVER_JWKS_URI`: the URI for this service to get the JSON Web Key Sets (JWKS) to validate the signature of the Bearer JWT against the public key of the issuer. Set this to `${AUTH_SERVER_ISSUER}/protocol/openid-connect/certs` if Keycloak is your auth service.
```
jasonaowen, line 133: I've found that it is clearer and more sustainable to put environment variable documentation in the `.env.example` instead of the README. It's closer to what it's documenting, it's more likely to be updated correctly as environment variables change, and it simplifies the instructions in the README. I did that recently in the [data-viewer `.env.example`](https://github.com/PhilanthropyDataCommons/data-viewer/blob/main/.env.example), for reference.

README.md
```diff
@@ -126,15 +126,44 @@ LOG_LEVEL=trace npm run test
...
+
+The PDC service assumes there will be three roles in the JWT with ascending levels of privilege, named `pdc-viewer`, `pdc-user`, and `pdc-admin`, respectively. These are expected inside an array with key `roles` in an object under key `realm_access` in the JWT object. In other words, for a JWT named `jwt`, there is an array at `jwt.realm_access.roles`.
+* A user having role `pdc-viewer` can read data with `GET` requests.
+* A user having role `pdc-user` inherits the `pdc-viewer` role and can also write to most endpoints with `POST` requests.
+* A user having role `pdc-admin` inherits the `pdc-user` and `pdc-viewer` roles and can also `POST` to all endpoints.
```
jasonaowen, line 138: This is no longer needed as we're not including authorization at this time, right?

```suggestion
```

README.md
```diff
@@ -126,15 +126,44 @@ LOG_LEVEL=trace npm run test
...
+
+#### An example with a Keycloak authentication and authorization service
```
jasonaowen, line 140: I'm a bit uncomfortable with this level of detail - instructions on how to configure Keycloak feels out of scope for this repo, and a lot of the steps are for authorization, which we're not doing yet. I very much appreciate the urge to document, and am happy to hear pushback, but I would suggest cutting this entire example section.

src/routers/opportunitiesRouter.ts
```diff
@@ -1,11 +1,11 @@
...
+import { verifyJwt } from '../middleware/verifyJwt';
 
 const opportunitiesRouter = express.Router();
 
-opportunitiesRouter.post('/', checkApiKey, opportunitiesHandlers.postOpportunity);
-opportunitiesRouter.get('/:id', checkApiKey, opportunitiesHandlers.getOpportunity);
-opportunitiesRouter.get('/', checkApiKey, opportunitiesHandlers.getOpportunities);
+opportunitiesRouter.post('/', verifyJwt, opportunitiesHandlers.postOpportunity);
+opportunitiesRouter.get('/:id', verifyJwt, opportunitiesHandlers.getOpportunity);
+opportunitiesRouter.get('/', verifyJwt, opportunitiesHandlers.getOpportunities);
```
jasonaowen, line 9: Some of the routers have a `thingRouter.use(verifyJwt)` and some have `thingRouter.method(path, verifyJwt, ...)`. Why the split? Can we just use `use` everywhere?

src/middleware/verifyJwt.ts
```diff
@@ -0,0 +1,16 @@
...
+} from 'express';
+
+const expressJwtMiddleware = expressjwt(jwtOptions);
+
+export const verifyJwt = (req: Request, res: Response, next: NextFunction): void => {
+  expressJwtMiddleware(req, res, next)
+    .catch((error: unknown) => {
+      throw error;
+    });
+};
```
jasonaowen, line 16: This function looks very redundant. Why `catch` an error only to immediately `throw` it? In fact, why wrap the `expressJwtMiddleware` at all? Perhaps this is an artifact of the development process that led here.

Can we simplify this to:
```suggestion
export const verifyJwt = expressjwt(jwtOptions);
```

jasonaowen (commented):
I haven't found the test problem that you've been working on yet, I'm afraid!

bickelj:
> I haven't found the test problem that you've been working on yet, I'm afraid!

That is odd but turned out to help get this resolved.

Jason and I spoke. We found that the commit above seems to work when running the service (e.g. `npm run start`) but had trouble when running the mock JWKS service at test time for many tests (e.g. `npm run test:integration`). The failures happened reliably but the errors themselves were misleading. When running with a single mock JWKS service at test time for a single test file, the failures went away. The above commit helped us diagnose the issue.

The next step is to make authentication-only branch with two atomic commits. The first commit renames awkwardly named variables. The second commit removes ad-hoc API key authentcation and replaces it with JWT authentication.

src/test/mockJwt.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const getMockJwks = (): JWKSMock => createJWKSMock(
+  issuer,
+  '/protocol/openid-connect/certs',
+);
+
+export const getMockJwt = (jwksMock: JWKSMock): string => {
+  const aMomentAgo = Math.round(new Date().getTime() / 1000);
+  return jwksMock.token({
+    exp: aMomentAgo + 100000,
```
bickelj, line 13: More re-use of the mock JWKS server and mock JWT resolves the issue at test time. Given the re-use, it makes sense to increase the expiration duration.

package-lock.json
jasonaowen, line 1: I believe the package lock file is out of date; running `npm install` on this branch leads to changes. Can you please include those changes in this commit?

src/openapi.json
```diff
@@ -7,11 +7,21 @@
   },
   "components": {
     "securitySchemes": {
-      "preSharedApiKey": {
-        "type": "apiKey",
-        "description": "A pre-shared key required to create or access protected resources via the API.",
-        "name": "x-api-key",
-        "in": "header"
+      "auth": {
+        "type": "oauth2",
```
jasonaowen, line 11: I tried for a while to get this to work with `"type": "openIdConnect",` as the [docs say it should](https://swagger.io/docs/specification/authentication/openid-connect-discovery/), but I wasn't able to before I reached the end of the timebox I set for myself. :disappointed:

jasonaowen (changes requested):
Great work, @bickelj! I was able to run this, authenticate into the OpenAPI docs, and make authenticated requests.

We need to fix the `package-lock.json` problem, and I would appreciate enabling PKCE, but please consider everything else optional!

package-lock.json
bickelj: It will pollute our production (dist) dependencies, but I will do it. Sigh.

jasonaowen: Ah, I hadn't realize that was an intentional decision. I see now how and why you arrived at it, but unfortunately staying in that out-of-sync state is not sustainable; either developers will have to be constantly careful about not committing `package-lock.json` changes, or they'll sneak in to an unrelated commit. That sucks, and I appreciate you opening the upstream issue, but for now I think we'll just have to live with it - thank you for doing so!

src/auth/jwtOptions.ts
```diff
@@ -0,0 +1,32 @@
...
+export const issuer = AUTH_SERVER_ISSUER;
+
+export const jwtOptions: Params = {
+  // Because of different types from version-to-version in express-jwt (see jwks-rsa notes), cast
+  // to the later express-jwt API's GetVerificationKey.The jwks-rsa notes can be found in
+  // `node_modules/jwks-rsa/index.d.ts` or here:
+  // https://github.com/auth0/node-jwks-rsa/blob/83d0202a9f462451c0e1e2adb1d91280ee1ef30c/index.d.ts#L60
+  secret: expressJwtSecret({
+    cache: false,
+    rateLimit: false,
```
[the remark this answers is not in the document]
bickelj: Instead, I will remove these lines.

src/openapi.json
bickelj: I was able to use `"type": "openIdConnect"` locally. The user experience was not good because an exhaustive litany of capabilities or options showed up. With the `oauth2` type, I was able to present exactly one known-good option.

src/auth/jwtOptions.ts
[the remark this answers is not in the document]
jasonaowen: Thanks!

Looks like the defaults are to enable the cache and disable the rate limit ([source](https://github.com/auth0/node-jwks-rsa/blob/59373c5cd5b357fbb56b621c943f002e9f3076ca/src/JwksClient.js#L7-L14)). I think we want the rate limit on, too, to prevent a DDoS attack from spreading from the service to the IdP - right?

src/openapi.json
jasonaowen: Fair enough! Thanks for following up!

src/routers/documentationRouter.ts
```diff
@@ -1,13 +1,49 @@
...
+  const authCodeWithNewIssuer = {
+    ...documentation.components.securitySchemes.auth.flows.authorizationCode,
+    authorizationUrl: `${issuer}/protocol/openid-connect/auth`,
+    tokenUrl: `${issuer}/protocol/openid-connect/token`,
+    refreshUrl: `${issuer}/protocol/openid-connect/token`,
+  };
+  documentation.components.securitySchemes.auth.flows.authorizationCode = authCodeWithNewIssuer;
+  documentationRouter.get('/', swaggerUi.setup(documentation, options));
+} else {
+  throw new InternalValidationError(
```
jasonaowen, line 43: Now that I look more closely, I think I do actually agree with Codecov here.

Adding an error handler for improperly formatted JSON is a great idea, and I'm +1 to stopping the service over it, but it's not integral to setting up auth. Let's split it out to its own commit, put the `throw` logic into the `isJsonObject` function, rename it to something like `validateOpenApiConfiguration`, and add to the existing tests!

Then, later, let's add - and test - a function that will take that already-validated configuration and substitute in the issuer. Something like:

```ts
const getSwaggerDocsWithIssuer = (documentation: JsonObject, issuer: string): JsonObject => {
  const authCodeWithNewIssuer...
  return documentationWithNewIssuer;
};
documentationRouter.use('/', swaggerUi.serve);
documentationRouter.get('/', swaggerUi.setup(getSwaggerDocsWithIssuer(documentation, issuer), options));
```

That should be a bit easier to read, much easier to test, and will address the Codecov complaint this change introduced.

jasonaowen (commented):
One of the failing code coverage checks are, I believe, not worth fixing. There are two: some new code to handle the case that the OpenAPI JSON file is well-formed, and the non-promise-returning wrapper around `expressjwt`. The former is fixable - see below - but the latter seems like it would take too much effort for too little payoff.

I propose that, once this is otherwise ready, we merge anyway. @slifty, if you disagree with that plan, please speak up!

package-lock.json
bickelj: I think this was really a misunderstanding on my part, though, because at the moment it doesn't appear to make any difference in `dist` or `node_modules` or the resulting docker image. I think we make no distinction and package everything from `node_modules` and `dist` at the moment. So it is more trouble than it's worth to try to pick that nit. Furthermore, I see the issue it creates. We ought to be able to `npm i` anytime without worry!

jasonaowen: > I think we make no distinction and package everything from node_modules and dist at the moment.

That sounds like a deployment problem, but yes I agree it's out of scope for this PR.

src/auth/jwtOptions.ts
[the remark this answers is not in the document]
bickelj: I recall hitting rate limits when I had `mock-jwks` set up and torn down multiple times during tests (if I recall correctly, even with only one `mock-jwks` per test suite/file). Now that we use one `mock-jwks` and one mock bearer JWT, I doubt that issue would occur with rate limits set. I will set the rate limit low during tests to see if it is exceeded (I do not expect it to, even at `1`) to double-check it is working the way I expect, and then set it to high if that works fine.

src/routers/documentationRouter.ts
bickelj: Is it OK to have that commit in a separate PR? For now I can simply remove it.

src/auth/jwtOptions.ts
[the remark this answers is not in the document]
bickelj: Ten per minute seemed low to me. So I set it to ten times that.

src/routers/documentationRouter.ts
bickelj: I removed the `else` in the most recent push.

jasonaowen (approved):
Great work, @bickelj! Thanks for getting this to done; I know how much effort was involved.

I was able to run this locally, authenticate into the docs (using PKCE!), and make an authenticated request. I was also able to get the expected errors with an invalid token and with an expired token.

The linter is happy and CI passed. Not sure what's up with Codecov, but I'm fine to merge either without it or over its objections, and previously discussed. One small suggestion below; feel free to either fix & merge or ignore & merge.

Yay auth! :closed_lock_with_key: :balloon:

src/auth/jwtOptions.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const jwtOptions: Params = {
+  // Because of different types from version-to-version in express-jwt (see jwks-rsa notes), cast
+  // to the later express-jwt API's GetVerificationKey.The jwks-rsa notes can be found in
+  // `node_modules/jwks-rsa/index.d.ts` or here:
+  // https://github.com/auth0/node-jwks-rsa/blob/83d0202a9f462451c0e1e2adb1d91280ee1ef30c/index.d.ts#L60
+  secret: expressJwtSecret({
+    jwksUri: `${AUTH_SERVER_ISSUER}/protocol/openid-connect/certs`,
+    rateLimit: true,
+    jwksRequestsPerMinute: 100,
```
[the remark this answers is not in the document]
bickelj: Right. This is a compromise between "do not enable rate limiting" and "enable rate limiting with default value of 10 requests per minute." I think the auth server should have no trouble handling 1-2 requests per second from the service on top of the rest of the requests from data-viewer, etc. I see your point though, realistically 10 is fine too.

Issue #129, opened 2022-11-29
Generating API keys fails on certain dev environments


reefdog:
OS: macOS 13.0.1 (Ventura)

Attempting to follow the setup instructions, I ran into a problem [generating API keys](https://github.com/PhilanthropyDataCommons/service#api-keys):

```
service git:(main) for i in {1..5}; do tr -dc a-zA-Z0-9 </dev/random | head -c 80; echo ''; done > secret_api_keys.txt
tr: Illegal byte sequence
tr: Illegal byte sequence
tr: Illegal byte sequence
tr: Illegal byte sequence
tr: Illegal byte sequence
```

The `secret_api_keys.txt` file was then populated with this content:
```

87




```

bickelj:
Oh right, the example assumed (without mentioning it) using GNU coreutils. With BSD there is probably different syntax.

bickelj:
I don't have a mac or BSD machine handy, but does this resolve it?
`for i in {1..5}; do LC_ALL=C tr -dc a-zA-Z0-9 </dev/random | LC_ALL=C head -c 80; echo ''; done > secret_api_keys.txt`

bickelj:
@reefdog Great! I guess it was a locale issue then rather than syntax, but yes, it also worked fine with GNU coreutils. I will update the readme.

PR #130, opened 2022-11-29
Use C locale when generating API keys

bickelj:
Including LC_ALL=C helps make this command work on more systems.

Issue #129 Generating API keys fails on certain dev environments


bickelj:
86f637a is a rebase on main.

reefdog (approved):
🙏 Thanks for doing this. Confirmed it works on macOS!

PR #133, opened 2022-12-01
Update node to 16.18.1

bickelj:


reefdog (changes requested):
Could you update `.node-version` as well?

bickelj:
`.node-version` updated in 9642542, I knew I missed something!

reefdog (approved):
Switched my local Node to 16.18.1. Tests ran cleanly, app booted and API calls succeeded. As far as I can tell, it all works!

I don't know what else is needed to ensure this survives the workflow, but it seems fine from my limited perspective.

bickelj:
Thanks for the review!

slifty:
I know this is merged and it's not WRONG by any means but I would note that generally I'd recommend we not do patch level updates on the node version (or really any updates on the node version) unless there is a project-driven purpose for it.  For instance: a security patch, a feature, or ensuring that we are using a supported / stable version.

Updates to the node engine come with overhead for anybody who is using the system (including the dev team)!

For what it's worth, I believe most node projects do not specify patch level versions but rather major (or sometimes minor) versions.  We might consider adopting that practice!

bickelj:
There must be some kind of disconnect in expectations somewhere. This update was from node major version 16 to node major version 16 and applies to the primary instance of the PDC. Anyone running a secondary or private instance controls exactly which node version they would like to run. So while I politely disagree on the merits of updates and using exact versions (we can discuss that more separately), a more fundamental question I have is what pain is being created by this update? I am hoping to understand more here.

Issue #134, opened 2022-12-02
Node.js engine versioning


slifty:
We just updated the node engine version ~~for the first time~~ (#133) and that led to a realization that it's probably time to document a policy on when / how the Node.JS version of this project will be updated!  I'm making this issue so we can discuss that policy and ultimately we can just document somewhere in `/docs`

First the easy part: Node releases a new major stable version once a year (the even numbers).  I would recommend that we commit to at least keeping up to date with that.

A more interesting question comes down to whether we should pin to minor / patch versions both in our developer oriented `.node-version` as well as our tests / deploy workflows (we do both right now, so a decision to continue that would just be a matter of deciding when / how those are updated).

The benefits to this are of course around ensuring predictable builds and deployments since things are so closely pinned.  The downside is the additional overhead related to updating our versions manually (it looks like patch releases happen once every two weeks or so).  

Externally the most important decision is related to the value of `.node-version` since we have third parties who are running this software and by specifying a version we're signaling that those third parties should adhere to the version at whatever granularity we indicate there.

We can also, for instance, pin `.node-version` to a major and pin to the reproducable for tests / deployment.

bickelj:
I think this is a fair summary, although I recall updating the node before (after some digging found it in https://github.com/PhilanthropyDataCommons/service/commit/4f55064b1cbdfa12b23d9d7dd568c4adc2ece562).

Updating the node version in most cases should adversely affect no one, including the developers, as the common case will be working code for both the old and new versions. The same should be true for libraries (though less commonly). If an update does cause an adverse effect, the tests or the developers (and in the worst case the users) should flag it.

The benefit to keeping exact versions under source control (top to bottom) is that we know exactly which version combination we tested, can deploy with exactly those versions, can reproduce issues more easily with exactly that version of the deployed artifact, and can have some confidence of how to go in the other direction from binary to source. The repository is more likely to have almost exactly the same software after running the same commands at that same revision.

As for making updates to those exact versions we have not done a great job of keeping up with updates. The updates ideally would be tested as soon as any new version comes out. Any change to a dependency regardless of the declared version numbers is a potentially breaking change. Ideally they would be atomic, one at a time, but that is very cumbersome, so it is easier to do it in batches where at least the only change is to dependent libraries. If something breaks, we can narrow it down. Ideally, also, we could automate some of the dependency updates. I suppose that's what dependabot can do.

[...]

PR #136, opened 2022-12-02
Try dependabot

bickelj:
Updating to recent versions of software is a chore. The reason to
update is to get ahead of CVEs or take advantage of improvements. It
is best to do these changes atomically and as frequently as new
versions are available in order to tell which update might cause an
issue. Our libraries and base images can be scanned automatically for
updates using dependabot and pull requests automatically created. If
all goes well with the proposed change, we should see the tests pass
in that pull request and can simply merge. If an issue is revealed by
the updated software library, it is visible in the pull request.

Somewhat related to issue #134 (Node.js engine versioning)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+open-pull-requests-limit: 10
+


bickelj:
If this yml works, dependabot literally only took 10 minutes to get set up including this comment.

slifty:
This might be relevant to include in the commit message but I actually didn't realize dependabot needed a yml to run!  It would be cool to link in the comment to just the page of documentation that explains how this file is populated for future versions of us who might wanna modify / improve it (or who might not know what dependabot is).

Does this expand the range of what it looks at? Or did something change recently about dependabot?  Either way: neat!  I had no idea dependabot configuration was a thing!

slifty:
Update: just a link to this would be enough https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

AND regarding that opening question, TIL: https://stackoverflow.com/a/69211343/159522

I'm betting I've just always seen dependabot running on projects with it being enabled via github interface.  I definitely prefer the `.yml` approach, how fun!

slifty (approved):
Welcome to our new robotic overlords!

Outside of review I noted that it might be useful to add a bit more context about what dependabot is / what this yml is doing and enabling in the commit message, but it's definitely not a blocking thought if you don't feel the commit message muse calling.

bickelj:
I've never used it, only seen evidence of it on other projects, but just followed https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates

I suppose the commit message could be slightly more descriptive, fixing that.

slifty (approved):

bickelj:
Oh dear, syntax error!

bickelj:
Dependabot works better than I imagined it would. It is rebasing, including lots of notes, etc. I am sure there will be a few kinks to work out such as getting the github actions included as well such that `node` updates are more-or-less synchronized across actions, docker images, and `.node-version`.

Thirty minutes well spent!

PR #137, opened 2022-12-03
Correct the syntax for dependabot

bickelj:
The open-pull-requests-limit is per package ecosystem, not overall.

Somewhat related to issue #134


bickelj:
The main question with the previous PR was whether or not to try dependabot. Therefore I assume it is OK to merge this.

Issue #135, opened 2022-12-02
Logging and creating an audit trail


slifty:
There is a requirement that the PDC supports the use case of knowing (in rough terms):

1. Who has accessed a given piece of data
2. Who has inserted a given piece of data
3. Changes to data over time
4. The provenance of a given piece of data

The last two items are already roughly covered by our existing system.  The first will involve a logging system with these trails in mind.  @bickelj has set up a great logger in the existing system.  We also discussed the potential of using something like Loki (or Prometheus) which provides log exploration interfaces (but adds a level of architecture complexity that is hard to justify at this point).

On our call today we talked about the value of making that data accessible via API so it could be rendered in the data viewer.  This would involve:

1. The design and creation of a new table (which is currently mocked into the ERD but that should be considered a placeholder).
2. The creation of new endpoints for reading those logs.
3. Code to write to the table on various endpoint calls.

reefdog:
Our [use cases](https://pad.opentechstrategies.com/p/2022-12-02-pdc-logging-conversation#L5) include:

> * Funder A wants to see which other funders are looking at Funder A's proposals.
> * ChangeMaker A wants to see which funders have been looking at their stuff.

To support these, communicating "who" has accessed a piece of data to the browser will require knowing some metadata about the user, including their organizational attachments (i.e., Jane Doe _of the XYZ Foundation_ looked at…). So we either need to snapshot a bunch of user metadata alongside the log (which makes it stale, but accurate for auditing purposes) or support looking up user metadata using whatever our identifier is (either directly from the client application, or perhaps as part of the API response generation by the service).

bickelj:
On this issue there is a need to unredact some portion of the API keys in the logs until we have a full auth system.

slifty:
I'm going to close this as stale (I don't believe these are stated needs in phase 2) -- also @reefdog hey look we *DID* have use cases!

PR #158, opened 2022-12-07
For audit capabilities, log subset of API keys

bickelj:
When using x-api-key for auth, there is only a set of authentication keys. To have a clearer picture of which entity made which API call, log some subset of that key, assuming that it is long enough to not compromise it by this logging.

Issue #135 Logging and creating an audit trail

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -11,9 +11,21 @@ import type {
 // here in the logger code.
 dotenv.config();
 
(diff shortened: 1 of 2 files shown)


gmurtaza00 (approved):

bickelj:
Many thanks, @gmurtaza00!

PR #159, opened 2022-12-07
Expand API key checks to remainder of endpoints

bickelj:
We want all the endpoints (except the documentation) to be protected. This commit does that.

Issue #96 Authentication and authorization


bickelj:
If I understand correctly, @slifty is mostly away this week. Merging.

Issue #104, opened 2022-10-27
Should PDC store atomic fields or composite fields for names, addresses, etc.?


bickelj:
From https://github.com/PhilanthropyDataCommons/service/pull/103#discussion_r1001854443

System A might use two fields for family name and given name. System B might use one field for full name.
System A might use five fields for house number, street, city, state, and zip code. System B might use one field for address.

If PDC uses atomic fields (e.g. System A style) then composite field callers of PDC (e.g. System B) are burdened with splitting apart their fields. If PDC uses composite fields (e.g. System B style) then atomic field callers of PDC (e.g. System A) are burdened with joining their fields.

slifty:
Just a note that in either choice *someone* will need to parse atoms (e.g. if we store as composite fields, a system that uses atomic will need to parse our responses.)

I'd propose that we store with atomic BUT that we provide a utility API that is capable of breaking certain types of composite fields into atoms.

bickelj:
@slifty Right you are. My comment above was when a system calls to post something, but when a system calls to get something, the converse is true too.

bickelj:
This issue raises a more fundamental issue which is the process by which we establish the initial set of canonical fields. Regardless of which direction we go it has to somehow get integrated into that process. Do we keep the bash scripts alive until we have several more spreadsheets and then ditch the sheets and go with some manually curated JSON after that? Do we ditch the bash scripts now and replace them with more sophisticated software? Do we ditch the bash scripts now and go with fully manually generated JSON? These are some options.

bickelj:
From yesterday's team meeting, it sounded like everyone was OK with the approach in #164. That approach is to have the initial set of canonical fields seeded by a SQL migration script. The migration script gets generated from a manually-curated spreadsheet. At the moment, the conflict raised in this issue is not addressed, but if we are OK with #164 then we have a path forward for setting the expectations in the manually-curated spreadsheet. As for multiplexing the fields to and from proposals, that's going to have to be part of the software that does the ingest.

bickelj:
[...]

PR #164, opened 2022-12-08
Use a db migration script to seed canonical fields

bickelj:
With this approach, many canonical fields will already be present in the database following database migration. This makes it easier to get started working on the PDC service software by having portable canonical field ids, i.e. ids are consistent across the official instance of PDC and also local development instances. It also lowers the barrier to creation of an application form.

If we do not have canonical fields present, it is more challenging to generate an application form, e.g. from a spreadsheet. This approach assumes that an original spreadsheet can be viewed and/or edited by folks with domain knowledge and then used to generate migration scripts as well as application form creation JSON. In fact this is how the initial SQL script here was created.

This approach replaces the previous attempt at fully automatic generation of canonical fields from unedited spreadsheets of fields. As such it is a kind of breaking change at the database data level. In order for migration to succeed with this change, an empty canonical_fields table is expected. So before running migrate, run a 'delete from canonical_fields;' query.

As time goes on and fields are modified, two approaches could make sense:
1. Refine the initial insert script and have it delete all the data in the database and re-insert it. This is the easier approach, assuming all form fields and other data can be re-inserted via automation.
2. Use additional scripts to carefully update and preserve data. This approach will need to be taken at some point in the future once more data is present.

This commit takes approach (1) but could potentially take (2), e.g. start with what is present in the pilot PDC instance and carefully insert and update where appropriate (with additional work).

See the script used to generate the canonical fields SQL file here:
https://github.com/PhilanthropyDataCommons/data-scripts/blob/ba13a1d4b92319cc72f0d96e1c5fa888bb56088d/src/generateCanonicalFieldsInserts.ts#L1

Issue #102 Make canonical fields and application forms from spreadsheets


src/database/migrations/0010-insert-canonical_fields.sql
```diff
@@ -0,0 +1,283 @@
+INSERT INTO canonical_fields (id, label, short_code, data_type, created_at) OVERRIDING SYSTEM VALUE VALUES
+(1, 'Organization Name', 'organizationName', '{ type: "string" }', '2022-12-05 11:03:49'),
```
reefdog, line 2: Do we actually need to be setting timestamps manually?

```suggestion
(1, 'Organization Name', 'organizationName', '{ type: "string" }', NOW()),
```

bickelj: Technically not. The reason for doing it here is so that everybody who runs this migration will have exactly the same first N rows in the `canonical_fields` table. The timestamp gets generated automatically in the source spreadsheet when adding a label. In the migration sense, it's manual. In the spreadsheet sense, it's somewhat automatic.

bickelj: In other words, this timestamp would represent when the field was added by who edited the sheet. But I am not totally tied to this approach either and it would be fine to keep it consistent with the other tables.

reefdog: (Sorry, just now noticing you responded!)

👍 If we're hoping to maintain data consistency between migrations, that's a good enough reason! If, in practice, we won't make use of that and/or other records will be added in ways that throw a wrench in this -- i.e. if we can't guarantee timestamp consistency always -- then may as well do away with that implicit commitment here by doing `NOW()`. I'll leave that up to you thought, just wanted to flag it!

jasonaowen:
What's going on with this PR? I'm not sure I have enough context to usefully comment on the approach! @bickelj, is this still useful / needed? If so, could you please rebase and resolve the merge conflicts?

src/database/migrations/0010-insert-canonical_fields.sql
bickelj: The newer script at https://github.com/PhilanthropyDataCommons/data-scripts/blob/ba13a1d4b92319cc72f0d96e1c5fa888bb56088d/src/generateCanonicalFieldsInserts.ts#L41 based on the newer spreadsheet format lets the created date be generated on `insert` instead.

bickelj:
@jasonaowen I still intended for this approach to be merged, yes, but we were working out the details of the spreadsheet that helped generate the base canonical fields as well as the code to generate them. If this approach is undesirable, of course we can try another approach.

jasonaowen:
> As time goes on and fields are modified, two approaches could make sense:
> 
> 1. Refine the initial insert script and have it delete all the data in the database and re-insert it. This is the easier approach, assuming all form fields and other data can be re-inserted via automation.
> 
> 2. Use additional scripts to carefully update and preserve data. This approach will need to be taken at some point in the future once more data is present.

Having spent some more time studying this repo, I want to caution against the first approach.

We're using [postgres-migrations](https://www.npmjs.com/package/postgres-migrations), which has a few strong opinions; one of them is that database migrations are immutable: if you introduce a migration, it should _never_ change, and the library builds in some safeguards to ensure that. From the section "Hash checks for previous migrations":

> Previously run migration scripts shouldn't be modified, since we want the process to be repeated in the same way for every new environment.
>
> This is enforced by hashing the file contents of a migration script and storing this in migrations table. Before running a migration, the previously run scripts are hashed and checked against the database to ensure they haven't changed.

In general, I find adding data in migrations to be a thorny issue. Having consistent data between different environments is convenient, but not so convenient to justify breaking environments.

I would rather we keep database migrations limited to schema changes. As an alternative, could we commit some seed scripts that use the API to load some sample or initial data? That should be easier to keep up-to-date, both in terms of the data and the schema, and should be a useful bit of dog-fooding.

bickelj:
@jasonaowen Regarding database migrations: I understand and understood on the way into this PR. The first approach can only works o long as we have not told people "go" on the official instance and requires manual steps to start from scratch again. It is akin to getting one of our initial database schema migration scripts correct. Once we cross a particular threshold of exposure or use, we have to make sure the migrations are immutable, backward compatible, etc.

Let me review some of the history and counterpoints.

My first approach in #103 was to treat the canonical fields like any other entity in the database. That is to say the approach in #103 assumed everyone with access could get and post canonical fields just like they could get and post proposals. Under this assumption in order to add an application form you need to make sure the canonical fields you need exist. All well and good: simply post them. All well and good the first time you create an application form, but what about the second time? Easy enough: post them again, and if some of those posts fail due to the field already being present (as a result of the unique key on `shortCode`), that's not a big deal except that the id increments which is an aesthetic annoyance but does not break anything. Just make sure to look up the canonical fields prior to posting the application form. Perhaps one could look up the canonical fields first before even posting canonical fields.

So what's the problem? The purpose of the project (as I understand it) is to curate the list of fields used across all application forms, to compare and contrast the use of fields, and to aim at field re-use such that foundations in the long term arrive at a list of common fields used by all foundations. This means that at least on the official instance of PDC, we expect updates to the canonical fields, which implies an update to the application forms as well. For example, we might have one canonical field called "Surname" and one canonical field called "Family Name" and during exploration of the proposal data we see these fields are used for the same things, so we need to point the existing application forms (either by insert or update) to the preferred canonical field and deprecate the old canonical field (either by delete or another convention). My sense is that the script that performs such a change should not be a script off to the side but an official migration script, tested via automated testing, etc., therefore a migration script. Technically I suppose it could be done either way: a script off to the side, but if it is a migration script it would be less awkward to have the canonical fields already present.

Another problem arises when someone tries to create application forms in a local development or test instance prior to posting data to the official instance. I admit this is an ease-of-use or user experience issue rather than a strict technical barrier. It is both simpler and easier to clone, build, and run the software and then post an application form with existing known canonical field IDs. The alternative is several more steps. Furthermore, suppose you start fresh and add canonical fields, and then afterward attempt to add the "official" canonical fields from script and get a collision on shortCode. There is no guarantee that your new canonical field matches that canonical field and now you have to go and start from scratch locally again (or do some manual deletes on the database). Suppose there is a canonical fields seed script and you add your own canonical fields locally, now when your script fails, your canonical fields are still consistent with the official instance. Adding canonical fields in this way on the official instance, though, creates problems with the approach here in #164 which is why it assumes zero or limited access to the post route on canonical fields (and why I had gone ahead with authorization in #96, #124 as well).

In short, the canonical fields are fundamental to the PDC and the changes to canonical fields need to be done carefully, so canonical fields are special. I agree that migration scripts should generally be reserved for schema changes and I am open to going back to that approach if my concerns about canonical fields are overblown. Hopefully I have explained some of the reasoning behind this PR and why canonical fields would end up in a migration script.

jasonaowen:
> Let me review some of the history and counterpoints.
> 
> [...] Hopefully I have explained some of the reasoning behind this PR and why canonical fields would end up in a migration script.

@bickelj, this is fantastic background information that I did not have; thank you so much. I didn't realize this was a conversation you and @slifty had already had!

Let me think about this for a little bit and get back to you.

In terms of your work, @bickelj, is #124 blocking on this?

src/database/migrations/0010-insert-canonical_fields.sql
```diff
@@ -0,0 +1,285 @@
+INSERT INTO canonical_fields (id, label, short_code, data_type) OVERRIDING SYSTEM VALUE VALUES
+(1, 'Organization Name', 'organization_name', '{ type: "string" }' ),
```
jasonaowen, line 2: What is this `'{ type: "string" }'` value? It looks vaguely JSON-ish, but like JavaScript notation for JSON - it is not valid JSON:

```sh
$ echo '{ type: "string" }' | jq .
parse error: Invalid literal at line 1, column 7
```

The `canonical_fiels.data_type` database field is declared as `character varying`, rather than `jsonb`, and I don't see any parsing code in the application - it looks like the API just passes the value in the database back to the HTTP client.

src/database/migrations/0010-insert-canonical_fields.sql
```diff
@@ -0,0 +1,285 @@
+INSERT INTO canonical_fields (id, label, short_code, data_type) OVERRIDING SYSTEM VALUE VALUES
```
jasonaowen, line 1: @bickelj [wrote](https://github.com/PhilanthropyDataCommons/service/pull/164#issuecomment-1466445071) (and I edited for brevity):

> Another problem arises when someone tries to create application forms in a test instance prior to posting data to the official instance. It is both simpler and easier to clone, build, and run the software and then post an application form with existing known canonical field IDs.

I don't think I fully understand the ID concern! Is that not the purpose of the `shortCode` field? https://github.com/PhilanthropyDataCommons/service/blob/f585cf8712bbc54e3c65247e788abf4a5c98cbe4/src/database/migrations/0001-create-canonical_fields.sql#L12

> ... suppose you start fresh and add canonical fields, and then afterward attempt to add the "official" canonical fields from script and get a collision on `shortCode`. There is no guarantee that your new canonical field matches that canonical field and now you have to go and start from scratch locally again (or do some manual deletes on the database).

This sounds like a question of assigning short codes, and in particular is an artifact of a [current technical limitation](https://github.com/PhilanthropyDataCommons/service/issues/104#issuecomment-1419687999):

> At the fundamental level currently all fields need to have a canonical field underneath them.

Taking that constraint at face value, it sounds like we'd need to either be mixing static and dynamic values, or writing, reviewing, merging, and deploying a database migration each time a partner wanted to import data with a new field. In each case, managing canonical fields by way of database migrations sounds untenable! Am I overstating the problem? Or, what am I missing?

src/database/migrations/0010-insert-canonical_fields.sql
bickelj: It is intended to be a JSON schema type, but yeah, it looks like we need to shore it up one way or another.

src/database/migrations/0010-insert-canonical_fields.sql
bickelj: > I don't think I fully understand the ID concern! Is that not the purpose of the `shortCode` field?
> 
> https://github.com/PhilanthropyDataCommons/service/blob/f585cf8712bbc54e3c65247e788abf4a5c98cbe4/src/database/migrations/0001-create-canonical_fields.sql#L12

Sure, we can use `shortCode` instead, and go a roundabout way of getting the necessary auto-generated IDs in order to make the post to `/applicationForms`. In other words, we can require "first, get canonical fields, then second post an application form" and that is technically OK but requires more steps. Originally, I think Dan had the `canonicalFields` as a special table in that rather than referencing auto-generated IDs we would join on `shortCode`. In all other places we use the auto-generated IDs. In general, auto-generated IDs are necessarily exposed so that callers can use them in endpoints such as `GET /applicationForm/{id}`. For consistency, I argued that we should use the canonical field auto-generated IDs like any other. Or perhaps we could go another way and use natural keys throughout, i.e. something akin to `shortCode` in every table, but that was an early decision we decided against. All that said, the whole system from top to bottom is up for revision at any time. It is better to get it closer to what is needed sooner than later. I am not stuck on this approach if there is a better one. Perhaps the complexity (from a caller attempting to create an application form perspective) of the approach you described is still better overall. Some of that complexity will be borne by our team (e.g. in the https://github.com/PhilanthropyDataCommons/service/pull/164 repo rather than the actual caller).

> This sounds like a question of assigning short codes, and in particular is an artifact of a [current technical limitation](https://github.com/PhilanthropyDataCommons/service/issues/104#issuecomment-1419687999):
> 
> > At the fundamental level currently all fields need to have a canonical field underneath them.

It is a modeling or design decision that creates a constraint, for sure. Constraints are often a good thing.

> Taking that constraint at face value, it sounds like we'd need to either be mixing static and dynamic values, or writing, reviewing, merging, and deploying a database migration each time a partner wanted to import data with a new field. In each case, managing canonical fields by way of database migrations sounds untenable! Am I overstating the problem? Or, what am I missing?

This is the clearest and best objection to the approach in this PR, yes.

The goal of the PDC as I understand it is to provide a means to curate a list of common application form fields. If the data model is getting in the way, we need a fresh data model. If this approach gets in the way, we need a fresh approach.

In OTS chat you mentioned that the problem is similar to that of tags for large data archives. Tags could be totally free-form which results in an explosion of very specific or duplicated tags. Tags could be totally managed and controlled by a central authority which results in missing or bad tags. Or there could be some middle ground where a group of moderators links or updates free-form tags into the most useful form. I agree this problem is similar here.

jasonaowen:
@slifty, @bickelj, and I talked after the team meeting today. Here's where we landed:

Short term, this will be a SQL file (and instructions; something to the effect of `psql -f seed.sql pdc`) that a developer can run to get a local environment that is somewhat similar to prod. It will *not* assign IDs. This is the least work necessary to move us forward without locking us into any design or policy decisions long term.

Medium term, we would like to move this to a script (probably in the [data-scripts repo](https://github.com/PhilanthropyDataCommons/data-scripts)) that seeds the initial canonical field list via the API, which would give us a more stable interface for seeding data (that is, less sensitive to schema changes), give us a chance to dogfood the API, and build an example for other partners on how to write clients.

Long term, once the instance at `api.pdco` is in use and people with domain expertise have carefully reviewed and curated the canonical fields, we'll want a script that can import canonical fields from one instance to another. That will allow a new instance to have prod-like data without requiring developers in the loop for maintaining that data.

I propose we switch from the SQL seed to an API seed (that is, switch from the short- to medium-term plan) if we make any schema changes that necessitate any major alterations to the SQL seed, and if at that time `api.pdco` doesn't have good enough data yet.

src/database/migrations/0010-insert-canonical_fields.sql
bickelj: I updated the origin spreadsheet but that does not address the validation at post or at insert time.

src/database/migrations/0010-insert-canonical_fields.sql
bickelj: See below. The team agreed we can use a SQL script off to the side, to be superseded by a script that interacts with the API.

bickelj:
See above comment. The work can be further tracked in #102 and/or https://github.com/PhilanthropyDataCommons/data-scripts

Issue #170, opened 2022-12-13
OpenAPI docs use integer 1 for almost all IDs


bickelj:
It might communicate the relationships more clearly to pick a different id for the various entities and use that id consistently throughout the docs. The danger is that if they are unique values one might get the mistaken impression that they share a single set of IDs but each entity has its' own (low risk here, I think).

The context of where this issue came from: discussion regarding the process that a data-viewer would use to get data from the service. It is one of those small changes I almost pushed in order to get the code coverage to run again but decided to send it in its own issue.

PR #171, opened 2022-12-13
Use a distinct ID for each entity in OpenAPI docs

bickelj:
When one entity uses another, use the same ID from the latter in the example of the former so that the relationship is clearer.

Issue #170 OpenAPI docs use integer 1 for almost all IDs


bickelj:
Hearing no objection, the "next day at the same time" window has passed, and it is a pretty minor change. Merging.

PR #178, opened 2022-12-15
Update node to 16.19.0, make @types/node use v16.x

bickelj:
Keep the node versions for build, lint, and test in sync with the version used in production (Dockerfile). These are exact versions.

Downgrade the @types/node back to a 16 version until we upgrade to 18.

The workflows still use a different docker image than we use in prod (and an ambiguous "latest" one at that), but I think this is required for GH workflows. Otherwise we would use the same exact bitnami image in tests that we use in production.

Set the .node-version to only the major version since this is a signal to developers about which node version(s) should work.

See issue #134 "Node.js engine versioning" for summary discussion


bickelj:
Hearing no objection...

PR #186, opened 2022-12-20
Update to node 18

bickelj:
Issue #134 Node.js engine versioning


bickelj:
I'm applying the "OK to merge without review on the next business day at same time" rule.

Issue #190, opened 2022-12-29
`npm ci` reports vulnerabilities that dependabot did not catch


bickelj:
```
$ npm ci

added 614 packages, and audited 615 packages in 11s

96 packages are looking for funding
  run `npm fund` for details

5 high severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
$ npm audit
# npm audit report

json5  <2.2.2
Severity: high
Prototype Pollution in JSON5 via Parse Method - https://github.com/advisories/GHSA-9c47-m6qq-7p4h
fix available via `npm audit fix --force`
Will install eslint-plugin-import@2.24.1, which is a breaking change
node_modules/json5
node_modules/tsconfig-paths/node_modules/json5
  tsconfig-paths  3.5.0 - 3.9.0 || 3.11.0 - 3.14.1
  Depends on vulnerable versions of json5
  node_modules/tsconfig-paths
    eslint-plugin-import  >=2.24.2
    Depends on vulnerable versions of tsconfig-paths
    node_modules/eslint-plugin-import
      eslint-config-airbnb-base  >=15.0.0
      Depends on vulnerable versions of eslint-plugin-import
      node_modules/eslint-config-airbnb-base
        eslint-config-airbnb-typescript  >=16.0.0
        Depends on vulnerable versions of eslint-config-airbnb-base
        Depends on vulnerable versions of eslint-plugin-import
        node_modules/eslint-config-airbnb-typescript

5 high severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force
```

bickelj:
It looks like `npm audit` now shows `0 [known/published/CVE-listed] vulnerabilities`

PR #191, opened 2022-12-29
Updates from `npm update`

bickelj:
This update attempt was spurred by an alleged json5 vulnerability. It is a dev dependency and therefore should not be included in production code and therefore should not affect deployed instances of the software.

This commit includes an update to json5 v1 which should be compatible with eslint plugin and removes the vulnerability.

See https://github.com/import-js/eslint-plugin-import/pull/2447#issuecomment-1367078871
See https://github.com/json5/json5/pull/298

Issue #190 `npm ci` reports vulnerabilities...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,30 +70,30 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.18.8",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.18.8.tgz",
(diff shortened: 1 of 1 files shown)


bickelj:
Can the vulnerability be exploited somehow during build time here with the GH actions? Not sure. Regardless, it sounds like it should be fixed upstream soon enough.

bickelj:
By the way, `npm audit fix --force` did not seem to fix the issue. So another `npm update` later will be needed (and pushed to this branch) or a dependabot update may push a separate PR and then this can be closed.

bickelj:
aaccf51 has updates to both json5 v1 and v2. The v1 backport happened a few minutes ago. It still appears `npm audit` thinks there is a problem with the v1 version, perhaps this is just a CVE reporting issue, though (e.g. it says "all versions below v2.2.2 and needs to reflect the backport).

Issue #101, opened 2022-10-18
Implement GET `/proposals/{proposalId}` endpoint


slifty:
As mentioned in #34 we need the ability to load a specific proposal.  We have explored the pattern of populating deep fields when loading a single ID and I think it would be appropriate to do so here as well, which would mean the returned proposal would include the various proposal versions and proposal field values.

bickelj:
Had we originally decided to return deep objects on all the single-id endpoints? Or should we default to shallow objects on all endpoints and then offer deep objects on some, as suggested around https://github.com/PhilanthropyDataCommons/service/pull/157#issuecomment-1349388808? PR #192 assumes shallow objects to be the default. The next step would be to allow deep objects when specified.

bickelj:
I am still working on the deep objects code and tests. But here is an example response for when asking to include fields and values with `curl ... http://host/proposals/1?includeFieldsAndValues=true | jq`:
```json
{
  "id": 1,
  "applicantId": 1,
  "opportunityId": 1,
  "externalId": "AnIdGeneratedByAGms",
  "createdAt": "2022-11-10T20:07:04.695Z",
  "versions": [
    {
      "id": 2,
      "proposalId": 1,
      "applicationFormId": 4,
      "version": 1,
      "createdAt": "2023-01-04T16:28:47.276Z",
      "fieldValues": [
        {
          "id": 1,
          "proposalVersionId": 2,
          "applicationFormFieldId": 3,
          "position": 1,
          "value": "Jones",
          "createdAt": "2023-01-04T16:29:57.877Z",
          "applicationFormField": {
            "id": 3,
            "applicationFormId": 4,
            "canonicalFieldId": 280,
            "position": 1,
            "label": "Family Name",
            "createdAt": "2023-01-04T16:27:40.175Z"
          }
        },
        {
          "id": 2,
          "proposalVersionId": 2,
          "applicationFormFieldId": 4,
          "position": 2,
          "value": "Jim",
          "createdAt": "2023-01-04T16:30:08.974Z",
          "applicationFormField": {
            "id": 4,
            "applicationFormId": 4,
            "canonicalFieldId": 279,
            "position": 2,
            "label": "Given Name",
            "createdAt": "2023-01-04T16:27:40.175Z"
          }
        }
      ]
    }
  ]
}
```

bickelj:
[...]

PR #192, opened 2022-12-30
Add GET /proposals/{proposalId} endpoint

bickelj:
A call to /proposals/{proposalId} returns a single proposal with its immediate attributes.


src/handlers/proposalsHandlers.ts
```diff
@@ -56,6 +56,44 @@ const getProposals = (
...
+      }
+      const proposal = proposalsQueryResult.rows[0];
+      if (isProposal(proposal)) {
+        res.status(200)
+          .contentType('application/json')
+          .send(proposal);
+      } else {
+        next(new InternalValidationError(
+          'The database responded with an unexpected format.',
+          isProposalArray.errors ?? [],
```
bickelj, line 81: Whoops, typo here.

bickelj:
It needs OpenAPI docs as well. Doing that.

src/handlers/proposalsHandlers.ts
```diff
@@ -56,6 +56,44 @@ const getProposals = (
...
+      }
+      const proposal = proposalsQueryResult.rows[0];
+      if (isProposal(proposal)) {
+        res.status(200)
+          .contentType('application/json')
+          .send(proposal);
+      } else {
+        next(new InternalValidationError(
+          'The database responded with an unexpected format.',
+          isProposal.errors ?? [],
```
reefdog, line 81: Is `isProposal.errors` correct?

I've glanced through the code and see this pattern repeatedly, but I'm a little mystified how it works. `isProposal` is a schema validation method, I take it, so how does calling `.errors` on it return errors specific to the `isProposal(proposal)` check that was done on [L74](https://github.com/PhilanthropyDataCommons/service/pull/192/files#diff-77b7da863212124379e28ebf124a964abd60d0e05f56a0d7fdb0a077a1061e22R74)?

src/__tests__/proposals.int.test.ts
```diff
@@ -146,6 +146,115 @@ describe('/proposals', () => {
...
+        .mockImplementationOnce(async () => ({
+          rows: [{ foo: 'not a valid result' }],
+        }) as Result<object>);
+      const result = await agent
+        .get('/proposals/2')
+        .set(dummyApiKey)
+        .expect(500);
+      expect(result.body).toMatchObject({
+        name: 'InternalValidationError',
+        details: expect.any(Array) as unknown[],
```
reefdog, line 213: Dropping a pin here because if [my other comment](https://github.com/PhilanthropyDataCommons/service/pull/192/files#r1065075986) ends up having legs, then I would expect this test is only passing because we're always falling back to the empty array, and need to actually check that whatever `isProposal.errors` is expected to return is actually being returned.

If that comment _doesn't_ have legs, this one can be dismissed too.

src/handlers/proposalsHandlers.ts
bickelj: It is the AJV way, I suppose. https://ajv.js.org/api.html#validation-errors says this:
> In case of validation failure, Ajv assigns the array of errors to errors property of validation function

I agree it's kind of odd. It might be a source of trouble for multiple calls, is that what you are seeing?

src/__tests__/proposals.int.test.ts
bickelj: I'm not sure I fully understand the suggestion but I know this test can pass when you call the wrong `isWhatever.errors` because we're not actually looking at the exact expected error message, so I agree there is some improvement to be had here.

As for it being an array, I think it's always going to be an array due to the `ValidationError` type defined in `src/errors/ValidationError.ts` and used by `src/middleware/errorHandler.ts` to structure the response.

reefdog: What I meant was, if it turned out that `isProposal.errors` _was_ a broken invocation (which turns out not to be the case), and was always failing / falling back to `[]`, then we wouldn't be catching that, and probably should be. But that's not the case so this comment can be resolved.

But yeah, to your point, we should probably check to make sure an expected error is returned.

src/handlers/proposalsHandlers.ts
reefdog: Holy moly, that's wild! Clearly you're correct and this is the AJV way so I'll mark this resolved, but what a bananas design IMO. Presumably there are some performance benefits they're gaining from it.

bickelj: Looking further down it says:
> In case of asynchronous validation, the returned promise is rejected with exception Ajv.ValidationError that has errors property.

So perhaps they really expect us to use async validation because that seems better. Speculating.

src/__tests__/proposals.int.test.ts
bickelj: Oh, I see, back in the handler code that fallback to empty array. I think I got it.

reefdog: Right! Like as-is, this test wouldn't catch if we had accidentally typed `isProposal.erors || []` back in the handler (note the `erors` typo). Instead we'd just… never be returning the errors. Dunno if that's a useful addition of test coverage or not tho.

slifty (approved):
This looks great!

bickelj:
8b47346 re-re-bases on main.

bickelj:
I see these workflow runs completed via the `Actions` tab:
* https://github.com/PhilanthropyDataCommons/service/actions/runs/3894034260
* https://github.com/PhilanthropyDataCommons/service/actions/runs/3894034262

It is odd that the PR does not see the status yet.

bickelj:
bddae03 re-re-re-bases on main. Hopefully the lint/test reports will arrive here this time.

PR #198, opened 2023-01-09
Add param to deep GET /proposals/{proposalId}

bickelj:
Add optional param to GET /proposals/{proposalId}

When optional query parameter includeFieldsAndValues is set to true on GET /proposals/{proposalId}, include all proposal versions, associated values, and associated application form fields in the response.

By returning this almost-fully-deep object tree, it is more convenient for the caller and more efficient in terms of request, query, and response counts. The assumption is that the purpose of the PDC is to see and compare which fields are used for what purpose for proposals.

### DIFF
--- a/src/handlers/proposalsHandlers.ts
+++ b/src/handlers/proposalsHandlers.ts
@@ -5,6 +5,7 @@ import {
   isProposalWrite,
   isProposal,
   isTinyPgErrorWithQueryContext,
+  isProposalRowWithFieldsAndValuesArray,
 } from '../types';
 import {
   DatabaseError,
@@ -19,7 +20,10 @@ import type {
 import type { Result } from 'tinypg';
 import type {
   Proposal,
+  ProposalFieldValue,
+  ProposalRowWithFieldsAndValues,
   ProposalWrite,
+  ProposalVersion,
 } from '../types';
 
 const logger = getLogger(__filename);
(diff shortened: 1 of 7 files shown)


src/middleware/errorHandler.ts
```diff
@@ -102,6 +105,8 @@ export const errorHandler = (
   res: Response,
   next: NextFunction, // eslint-disable-line @typescript-eslint/no-unused-vars
 ): void => {
+  logger.debug(err);
+  logger.trace(req.body);
```
bickelj, line 109: The changes to errorHandler are not really related to the change, but I find myself frequently adding this to the error handler regardless. If desired, I can put this in a separate change.

src/types/ProposalFieldValue.ts
```diff
@@ -38,6 +41,14 @@ export const proposalFieldValueSchema: JSONSchemaType<ProposalFieldValue> = {
       required: [],
       instanceof: 'Date',
     },
+    applicationFormField: {
+      type: 'object',
+      /* eslint-disable @typescript-eslint/no-unsafe-assignment */
+      properties: applicationFormFieldSchema.properties,
+      /* eslint-enable @typescript-eslint/no-unsafe-assignment */
```
bickelj, line 48: I would really like to figure out a way to include the application form field schema without having to disable an `eslint` rule. This seemed to me the least worst option that I could come up with. I am eager to hear about better ways to achieve the goal, though.

bickelj (commented):
This pull request has two commits. The first is identical to the one in PR #192 and the second commit builds on the first. The assumption is that #192 is relatively uncontroversial because it adds basic functionality. PR #192 still needs review, however, but both changes could be applied in this single PR instead, in which case #192 would be obsoleted. If the additional commit here turns out to require extensive revision, #192 could be merged separately.

reefdog:
@bickelj Is it too late to change the base branch for this PR from `main` to `101-add-proposal-by-id-endpoint`? When laddering PRs like this, that ensures this one only shows the commits added that aren't in that other commit. (GitHub's also smart enough to change the base on this PR to `main` once that other PR is merged in.)

If you _do_ change something about the earlier PR, then you'll have to do some cherry-picking and force-pushing and whatnot to keep this one up to date. But that's usually rare and worthwhile.

No insistence here, just alerting you to a process that I've found useful in this scenario! For now, I can just drill into the delta commit (5582675ef96270eb798ea6f5f21a2f99fe2d67ff) to view the diff.

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +60,160 @@ const getProposals = (
...
+  if (row.proposalVersionId === undefined
+    || row.proposalVersionId === null
+    || row.proposalVersionApplicationFormId === undefined
+    || row.proposalVersionApplicationFormId === null
+    || row.proposalVersionVersion === undefined
+    || row.proposalVersionVersion === null
+    || row.proposalVersionCreatedAt === undefined
+    || row.proposalVersionCreatedAt === null
+    || row.proposalFieldValueId === undefined
+    || row.proposalFieldValueId === null
```
reefdog, line 74: I think about right here is where I would have made a utility method to check all these, lol.

bickelj:
> @bickelj Is it too late to change the base branch for this PR from `main` to `101-add-proposal-by-id-endpoint`? When laddering PRs like this, that ensures this one only shows the commits added that aren't in that other commit. (GitHub's also smart enough to change the base on this PR to `main` once that other PR is merged in.)

Ah, I probably should have rebased #192 before starting here, I had rebased this one on main in an effort to make the PR #192 easier to review (e.g. less spammy, less volatile with all the force pushing). I should probably merge a bunch of updates to dependencies first and then get it in order. I'll do that.

src/handlers/proposalsHandlers.ts
bickelj: Fair enough haha.

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +60,160 @@ const getProposals = (
...
+      applicationFormId: row.proposalVersionApplicationFormId,
+      canonicalFieldId: row.applicationFormFieldCanonicalFieldId,
+      position: row.applicationFormFieldPosition,
+      label: row.applicationFormFieldLabel,
+      createdAt: row.applicationFormFieldCreatedAt,
+    },
+  };
+};
+
+const getProposalWithFieldsAndValues = (
```
reefdog, line 111: I have a not terribly helpful comment about this method, which is: I spent quite a lot of time reading through it and still couldn't figure out what it was so complicated. It seems like it should be more concise to simply collect all the versions into an array, and yet…

I trust that there is a reason for this verbosity and some of the gymnastics (like the dummy `currentProposalVersion` that gets replaced as needed). I think I would also you to do these things:

1. Consider if this method of collecting proposal versions can be simplified / made more legible.
2. If not, and/or if any improvements still aren't self-documenting, add more comments about why we're doing things.
3. Consider if anything can be abstracted out to named utility methods — even if not to reduce redundancy, just to have a clearer and more readable flow.

I leave it up to you to decide if that's worthwhile, but the long-term concern here is that this method may be difficult to maintain / mask errors.

bickelj: It is a helpful comment, thank you! One thing that comes to mind is the optional nature of the deeper objects and the mandatory (non-null) requirements of some of our types. I think the combination of those might explain some of it. But I am also unsatisfied with the way this one turned out and wish for a better way. One utility method is present, namely `getValueWithFullFieldFromRow` but there is still plenty to be desired here.

src/types/ProposalFieldValue.ts
slifty: You can actually just say:

```
applicationFormField: applicationFormFieldSchema
```

Which should solve your problem AND avoid some repeated code!

bickelj: That is much more elegant! Yes!

Unfortunately it doesn't "just work" without some more finagling. I'll figure out what else needs to be updated to get it to work.

bickelj:
FWIW, I purposely aimed for one database call per one HTTP request here.

bickelj:
Now that the related #192 has been merged,  01a0e71 rebases on main and this is one commit rather than two.

slifty:
Heya -- copying this here from a post I made in Zulip:

O(1) database calls is the metric I'd like us to target (rather than literally 1); one call per type of data returned.

I agree we need to avoid any situation where the number of SQL calls is arbitrary, but I think the cost associated with trying to do just a SINGLE call is incredibly high (both in terms of added code / complexity as well as technical debt, but also in that we're now re-implementing a little extra of what a database is already optimized to do in terms of data packaging and parsing).

So this would look like:

one new SQL query per data type (getProposalVersionsByProposalId, getProposalFieldValuesByProposalId).

Then we can use the existing schema validators to parse the results of those queries, do a quick loop on their results to put them in the correct buckets, and be happy!

bickelj:
The good news is it sounds like the API design is agreeable, namely the idea of a deep GET on `/proposals/{proposalId}` and the structure of that response looks OK.

The bad news is the implementation leaves some things to be desired, namely the idea of one SQL query per web request with several joins followed by rolling up the data in TypeScript.

@slifty said:
> I agree we need to avoid any situation where the number of SQL calls is arbitrary, but I think the cost associated with trying to do just a SINGLE call is incredibly high (both in terms of added code / complexity as well as technical debt, but also in that we're now re-implementing a little extra of what a database is already optimized to do in terms of data packaging and parsing).

Can you elaborate a bit on the high costs as compared to the alternative?

* Added code
* Added complexity
* Added technical debt
* Re-implementing what a db does

@slifty also said:
> one new SQL query per data type (getProposalVersionsByProposalId, getProposalFieldValuesByProposalId).
>
> Then we can use the existing schema validators to parse the results of those queries, do a quick loop on their results to put them in the correct buckets, and be happy!

I hear what you're saying but this wouldn't technically be O(1) because the last query is actually by proposal version id, so it would be O(N) where N is the count of proposal versions in a given proposal. I also see this as implementing a join in code rather than letting the DB do the join. It would be more efficient in terms of data per row, total data sent from db to service, for sure, but less efficient in terms of round trips (query count).

slifty:
This may answer the question, and if not maybe my best bet would be to show via implementation.

The approach I'm proposing would involve:

1. No new types
2. No new validation scripts
3. Three new .sql files
4. Two utility methods for "decorating" some of the parent objects with children.

It *would* truly be O(1), specifically it would involve three separate queries instead of one single query, which I don't think reflects a particularly high burden from a computational POV.

I'll take a stab at explaining better (probably will require code examples to really show what I'm tlaking about) tomorrow!

slifty (commented):
These were comments that apparently didn't get published so hitting send now!

bickelj:
I hereby officially withdraw this PR in favor of #210 which supercedes it and is much simpler.

Issue #174, opened 2022-12-14
Add ability to retrieve deep associations


reefdog:
The API currently returns only one level of requested objects. E.g., `/proposals` returns a list of proposals, but without any associated objects nested within. (The Swagger documentation about nesting `versions` is not accurate.)

So if, for example, you want all the field data associated with the most recent version of a proposal, you have to make many sequential calls to the API. That's fine as the default behavior, but in real-world use, API consumers are greatly advantaged by being able to opt in to including some of the nested associations as part of a single call. In the above example, the proposal would be returned with the application form fields already included.

We need to determine the preferred ergonomics of this. For instance, are our needs for each endpoint obvious enough that a simple boolean (`deep=true`) only has one reasonable outcome? Or do we need a query param language for the API consumer to request specific related objects (`include=proposalVersion,applicationForm,applicationFormField`)? Etc.

bickelj:
> We need to determine the preferred ergonomics of this. For instance, are our needs for each endpoint obvious enough that a simple boolean (`deep=true`) only has one reasonable outcome? Or do we need a query param language for the API consumer to request specific related objects (`include=proposalVersion,applicationForm,applicationFormField`)? Etc.

Having a simple `deep=true` might be confusing because it raises the question "how deep?" If I were to set `deep=true`, I might expect it to be fully deep. For example, I might expect `GET /applicationForms?deep=true` to return all the way through to and including canonical fields.

A simple `shallow=true` would be unambiguous: it always returns only the direct attributes of the entity.

The principle I used in https://github.com/PhilanthropyDataCommons/service/pull/157 up through b71998d was to return to a `GET` call the same depth as was expected in a `POST` call (plus the wrapping list). We probably agreed earlier to keep it shallow in some discussion several weeks/months ago and then yesterday @slifty made the same suggestion on `GET`.

A compromise might be that the default is the shallow, direct attributes of the entity, but there is a boolean parameter to suggest "same depth as POST." I don't have a great name for that other than `returnSameDepthAsPost=true|false`.

reefdog:
> Having a simple `deep=true` might be confusing because it raises the question "how deep?"

Right. I would only suggest we consider this if we determine it's actually unambiguous once we are implementing. (I assume that's _not_ the case, but keep reading…)

> A simple `shallow=true` would be unambiguous: it always returns only the direct attributes of the entity. […] A compromise might be that the default is the shallow, direct attributes of the entity…

[...]

PR #210, opened 2023-01-20
Add option to deep GET /proposals/{proposalId}

bickelj:
When optional query parameter includeFieldsAndValues is set to true on GET /proposals/{proposalId}, include all proposal versions, associated values, and associated application form fields in the response.

By returning this almost-fully-deep object tree, it is more convenient for the caller and more efficient in terms of request, query, and response counts. The assumption is that the purpose of the PDC is to see and compare which fields are used for what purpose for proposals.

(Alternative approach vs #198)

### DIFF
--- a/src/handlers/proposalsHandlers.ts
+++ b/src/handlers/proposalsHandlers.ts
@@ -4,12 +4,16 @@ import {
   isProposalArray,
   isProposalWrite,
   isProposal,
+  isProposalVersionArray,
   isTinyPgErrorWithQueryContext,
+  isProposalFieldValueArray,
+  isApplicationFormFieldArray,
 } from '../types';
 import {
   DatabaseError,
   InternalValidationError,
   InputValidationError,
+  NotFoundError,
 } from '../errors';
 import type {
   Request,
@@ -18,8 +22,11 @@ import type {
 } from 'express';
 import type { Result } from 'tinypg';
 import type {
+  ApplicationFormField,
   Proposal,
+  ProposalFieldValue,
   ProposalWrite,
+  ProposalVersion,
 } from '../types';
 
 const logger = getLogger(__filename);
@@ -56,7 +63,129 @@ const getProposals = (
     });
 };
 
-const getProposal = (
+export const joinProposalFieldValuesToProposalVersion = (
+  proposalVersion: ProposalVersion,
+  values: ProposalFieldValue[],
+): ProposalVersion => {
+  const newVersion = structuredClone(proposalVersion);
+  values.forEach((proposalFieldValue) => {
+    if (newVersion.fieldValues === undefined) {
+      newVersion.fieldValues = [];
+    }
+    if (proposalFieldValue.proposalVersionId === newVersion.id) {
+      newVersion.fieldValues.push(proposalFieldValue);
+    }
+  });
+  return newVersion;
+};
+
+export const joinApplicationFormFieldsToProposalFieldValues = (
+  values: ProposalFieldValue[],
+  fields: ApplicationFormField[],
+): ProposalFieldValue[] => {
+  const newValues = structuredClone(values);
+  if (newValues.length !== fields.length) {
+    // It is invalid to try to zip values and fields of unequal length.
+    throw new Error(`Given arrays must be of equal length. ${values.length} !== ${fields.length}`);
+  }
+
+  return newValues.map((value, index) => {
+    // Look in the fields array at the same index as the values array.
+    const field = fields[index];
+    if (field === undefined) {
+      throw new Error(
+        `Expected only defined fields and values, copiedFields[${index}] was undefined`,
+      );
+    }
+    if (value.applicationFormFieldId !== field.id) {
+      throw new Error(
+        `Given values and fields must be sorted such that they correspond. index=${index}: ${value.applicationFormFieldId} !== ${field.id}.`,
+      );
+    }
+    // When all is well, add the field to the value.
+    return { ...value, applicationFormField: field };
+  });
+};
+
+const getProposalWithFieldsAndValues = (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): void => {
+  // All queries could technically be sent concurrently but start with a proposal query to confirm
+  // there is a proposal first before sending the others concurrently.
+  db.sql('proposals.selectById', { id: req.params.id })
+    .then((proposalsQueryResult: Result<Proposal>) => {
+      if (proposalsQueryResult.row_count === 0) {
+        throw new NotFoundError(
+          'Not found. Find existing proposals by calling with no parameters.',
+        );
+      }
+      const baseProposal = proposalsQueryResult.rows[0];
+      if (!isProposal(baseProposal)) {
+        throw new InternalValidationError(
+          'The database responded with an unexpected format.',
+          isProposal.errors ?? [],
+        );
+      }
+      // Run the remaining queries concurrently.
+      Promise.all([
+        db.sql('proposalFieldValues.selectByProposalId', { proposalId: req.params.id }),
+        db.sql('applicationFormFields.selectByProposalId', { proposalId: req.params.id }),
+        db.sql('proposalVersions.selectByProposalId', { proposalId: req.params.id }),
+      ]).then(([
+        proposalFieldValuesQueryResult,
+        applicationFormFieldsQueryResult,
+        proposalVersionsQueryResult,
+      ]) => {
+        if (!isProposalFieldValueArray(proposalFieldValuesQueryResult.rows)) {
+          throw new InternalValidationError(
+            'The database responded with an unexpected format.',
+            isProposalFieldValueArray.errors ?? [],
+          );
+        }
+        if (!isApplicationFormFieldArray(applicationFormFieldsQueryResult.rows)) {
+          throw new InternalValidationError(
+            'The database responded with an unexpected format.',
+            isApplicationFormFieldArray.errors ?? [],
+          );
+        }
+        if (!isProposalVersionArray(proposalVersionsQueryResult.rows)) {
+          throw new InternalValidationError(
+            'The database responded with an unexpected format.',
+            isProposalVersionArray.errors ?? [],
+          );
+        }
+        const valuesWithFields = joinApplicationFormFieldsToProposalFieldValues(
+          proposalFieldValuesQueryResult.rows,
+          applicationFormFieldsQueryResult.rows,
+        );
+        const versions = proposalVersionsQueryResult.rows.map(
+          (proposal) => joinProposalFieldValuesToProposalVersion(
+            proposal,
+            valuesWithFields,
+          ),
+        );
+        const proposal = { ...baseProposal, versions };
+        res.status(200)
+          .contentType('application/json')
+          .send(proposal);
+      }).catch((error: unknown) => {
+        throw error;
+      });
+    }).catch((error: unknown) => {
+      if (isTinyPgErrorWithQueryContext(error)) {
+        next(new DatabaseError(
+          'Error retrieving proposal.',
+          error,
+        ));
+        return;
+      }
+      next(error);
+    });
+};
+
+const getProposalShallow = (
   req: Request,
   res: Response,
   next: NextFunction,
@@ -65,10 +194,7 @@ const getProposal = (
     .then((proposalsQueryResult: Result<Proposal>) => {
       logger.debug(proposalsQueryResult);
       if (proposalsQueryResult.row_count === 0) {
-        res.status(404)
-          .contentType('application/json')
-          .send({ message: 'Not found. Find existing proposals by calling with no parameters.' });
-        return;
+        throw new NotFoundError('Not found. Find existing proposals by calling with no parameters.');
       }
       const proposal = proposalsQueryResult.rows[0];
       if (isProposal(proposal)) {
@@ -92,6 +218,18 @@ const getProposal = (
       }
       next(error);
     });
+};
+
+const getProposal = (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): void => {
+  if (req.query.includeFieldsAndValues !== undefined && req.query.includeFieldsAndValues === 'true') {
+    getProposalWithFieldsAndValues(req, res, next);
+  } else {
+    getProposalShallow(req, res, next);
+  }
 };
 
 const postProposal = (
@@ -135,7 +273,7 @@ const postProposal = (
 };
 
 export const proposalsHandlers = {
-  getProposals,
   getProposal,
+  getProposals,
   postProposal,
 };
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -151,7 +151,15 @@ describe('/proposals', () => {
       await agent
         .get('/proposals/9001')
         .set(dummyApiKey)
-        .expect(404, { message: 'Not found. Find existing proposals by calling with no parameters.' });
+        .expect(404, {
+          name: 'NotFoundError',
+          message: 'Not found. Find existing proposals by calling with no parameters.',
+          details: [
+            {
+              name: 'NotFoundError',
+            },
+          ],
+        });
     });
 
     it('returns the one proposal asked for', async () => {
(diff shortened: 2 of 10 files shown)


bickelj:
fcd5ea6 rebased on main, moved a comment, clarified a comment, made SQL more consistent (caps `AS`).

bickelj (commented):
I will try to improve the ugliness in `mergeApplicationFormFields`.

bickelj:
abc0291 uses `map` instead of `for`, eliminating the gymnastics around convincing TypeScript that elements are truly defined.

bickelj:
6ce9d01 improves test coverage by removing a couple of redundant lines and adding some unit tests.

bickelj:
c9e413b improves test coverage and improves a couple of comments.

bickelj:
42a1e98 improves test coverage and rebases on main.

src/errors/ErrorWithStatus.ts
```diff
@@ -0,0 +1,12 @@
+export class ErrorWithStatus extends Error {
```
slifty, line 1: Love this new addition -- minor organization point but it would be helpful to have it (and the related changes) in a separate commit (same PR is fine) since it represents an atomic improvement / not directly coupled with the deep query endpoint.

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +62,175 @@ const getProposals = (
     });
 };
 
-const getProposal = (
+/**
```
slifty, line 65: We haven't tended to use docblocks in this repo and part of that is that we try to make sure the names (and resulting intellisense) are self explanatory -- I think that applies here.

For instance if you rename the parameters to `shallowProposal` and `fieldValues` and if you name the function `decorateProposalWithFieldValues` I think that becomes auto-documenting and you can ditch the doc block.  (one benefit of this approach is we don't have to update two places if we make changes!)

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +62,175 @@ const getProposals = (
...
+ * It is OK if there are no versions or zero field values.
+ * @param proposal The proposal with versions but no values.
+ * @param values The values to merge into the given proposal.
+ * @returns A new proposal with the values merged into their respective versions.
+ */
+export const mergeProposalFieldValues = (
+  proposal: Proposal,
+  values: ProposalFieldValue[],
+): Proposal => {
+  const newProposal = structuredClone(proposal);
```
slifty, line 76: oh snap `structuredClone` that's some cutting edge stuff!  Love it :D

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +62,175 @@ const getProposals = (
...
+  const { versions } = newProposal;
+  if (versions === undefined) {
```
slifty, line 78: I actually think that avoiding the destructuring above might make this check more legible since the context of what is being checked would be right there in the line (though see below for another idea!)

```
if (proposal.versions === undefined)
```

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +62,175 @@ const getProposals = (
...
+    // In this case there are no versions associated, return a simple copy.
```
slifty, line 79: This comment is re-stating what the code already says and so we can remove it (or if it seems like the code isn't clear, create an appropriately named helper function that makes it explicit what we're checking for, e.g. `proposalHasVersions` if that's truly what's being checked.

That said if that's the check then I think it isn't enough to only check `!==undefined` -- we also need there to be a version in the array.  If it's something else being checked then we'd want to name it appropriately to convey the true intent.)

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +62,175 @@ const getProposals = (
...
+    return newProposal;
+  }
+  // Because both queries are sorted by proposal version, we don't need to do N*M loops.
```
slifty, line 82: This function can't make assumptions about its input without validating those assumptions -- it has to be able to handle un-sorted input (or error if the inputs aren't sorted).  This could mean running an O(n) sort on each before hitting this code block though!

On a related note though we should avoid needing this kind of comment; I'm not quite sure what it means and it would be great to craft an algorithm that is self explanatory

src/handlers/proposalsHandlers.ts
```diff
@@ -56,7 +62,175 @@ const getProposals = (
...
 
-const getProposal = (
+/**
+ * Merges the given field values into a new copy of the given proposal.
+ * It is OK if there are no versions or zero field values.
+ * @param proposal The proposal with versions but no values.
+ * @param values The values to merge into the given proposal.
+ * @returns A new proposal with the values merged into their respective versions.
+ */
+export const mergeProposalFieldValues = (
```
slifty, line 72: Instead of merging field values onto proposals we could have a simpler function that merges field values onto proposal VERSIONS.  This would have the added benefit of being more universally relevant (and would / could go into a `/utils` and be used on other handlers that involve deeply populated proposal versions).

I think this would result in a dramatically less complex operation; you would then be able to do something like:

```
proposalVersions.map((proposalVersion) => decorateProposalVersionWithFieldValues(
  proposalVersion,
  proposalFieldValues,
));
```

^ and you could have the decorator just skip any field values whose proposalVersionId doesn't align with the proposal version.

Your deep lookup function could then be something as simple as this:

```typescript
Promise.all(
  proposalQuery,
  proposalVersionQuery,
  proposalFieldValuesQuery,
).then(([proposals, proposalVersions, proposalFieldValues]) => {
  const decoratedProposalVersions = proposalVersions.map(
    (proposalVersion) => decorateProposalVersionWithFieldValues(
      proposalVersion,
      proposalFieldValues,
    )
  );
  const decoratedProposals = proposals.map((proposal) => decorateProposalWithVersions(
    proposal,
    decoratedProposalVersions,
  ))
  res.send(decoratedProposals) // or whatever
})
```

slifty (changes requested):
Some initial feedback -- I have some more but wanted to start with this to see what you think first.

I believe that things could be a bit simpler -- we don't need the decorator / merge functions to be in terms of the *PROPOSAL* but rather in terms of the object actually being decorated (e.g. the version, the field value, etc).  This then saves a whole lot of internal checks and assumptions.

I think this would result in much more concise code with less need for checking / thrown errors while ALSO making those decorator functions universal / relevant to future deep object queries (and the deep query endpoint implementations will just be a matter of getting the data and calling the decorators in the right order!)

src/handlers/proposalsHandlers.ts
bickelj: Fair enough, I see a place where it unsorted versions could slip through below, after looking more carefully. The comment should be clearer too. For example it can describe why it does what it does rather than talking about what it might have done.

src/handlers/proposalsHandlers.ts
bickelj: So then you are iterating over the proposal field values (having count N) for each proposal version (having count M), so N*M instead of N. Such N*M iteration is what the comment below was trying to communicate that we avoid. I think your suggestion or something like it isl worth doing because it truly is simpler, more targeted, etc., but there will need to be a separate step of getting the proposal versions split up or accepting M queries per request (yuck), or I suppose letting the iterations happen because the expensive part is over (e.g. expensive to get from a disk, we're merely iterating over heap memory objects now more cheaply).

src/handlers/proposalsHandlers.ts
bickelj: A noble goal. Now we merely have to agree on a name that communicates clearly. It sounds like `merge` is not the word you would use and I don't think `decorate` is quite right either. Perhaps `join`?

src/handlers/proposalsHandlers.ts
bickelj: I had a feeling that `tsc` is more able to grasp a top-level variable undefinedness check rather than a property on a variable or index into a variable, but I haven't experimented enough to say that with any confidence. I will try it and see.

src/handlers/proposalsHandlers.ts
bickelj: It doesn't need to verify that there is a version in the array because `forEach` below should skip it in that case. Maybe this should be the comment instead.

src/handlers/proposalsHandlers.ts
slifty: Yeah I think we still stick with JUST the three queries (getting ALL data at once).

We could then have something like `filterFieldValuesByProposalVersionId` that just filters out the ones we want -- which I know is slightly more wasteful in runtime since it would filter the array N * M times, but I also don't think we need to optimize to that level at this point in the pilot (and to your point the thing to really avoid is hitting disk / running a query for each)

If we wanted to optimize (but this would get a little more code complex) then we could also pre-process the query result with a generic utility like "generateGroupsByAttribute" which would return an object of `Record<attributesType, T[]>`

```
generateGroupsByAttribute<T>(items: T[], attribute: string) => items.reduce(
  (groups, item) => {
    if (!item.hasOwnProperty(attribute)) {
      return groups;
    }
    return {
     ...groups,
     [item[attribute]]: [
       ...(groups[item[attribute]] ?? []),
       item
  }),
  {},
);
```

^ this (untested, written in github) function would split an array into groups with keys of whatever attribute.  So you could just do something like this:

```
  const groupedProposalFieldValues = generateGroupsByAttribute(proposalFieldValues, 'proposalVersionId');
  const decoratedProposalVersions = proposalVersions.map(
    (proposalVersion) => decorateProposalVersionWithFieldValues(
      proposalVersion,
      groupedProposalFieldValues[proposalVersion.id] ?? [],
    );
  );
```

^^ Again that's a bit more complex (though O(N)) than something like: 

```
  const decoratedProposalVersions = proposalVersions.map(
    (proposalVersion) => decorateProposalVersionWithFieldValues(
      proposalVersion,
      groupedProposalFieldValues.filter((proposalFieldValue) => proposalFieldValue.proposalVersionId === proposalVersion.id)
    );
  );
```

src/handlers/proposalsHandlers.ts
slifty: merge is fine too, but I do like `join` since it alludes to the parallel DB operation -- FWIW I decorate is an allusion to the Decorator pattern -- https://en.wikipedia.org/wiki/Decorator_pattern -- which takes an object and dynamically adds (decorates) functionality to it.

src/errors/ErrorWithStatus.ts
bickelj: I am rethinking this one because I probably gave feedback in the past that said "don't re-include an HTTP status in the body" and here I go including HTTP status in the body. The construct is nice but I hate seeing an HTTP status in a body, so even if I have to strip it out of the returned object I will. I agree it should be in a separate commit or PR. I am not entirely sold on squashing at all so I am still feeling my way through what should or should not be a separate commit (with regard to improvements) but your hint about atomicity makes good sense.

src/handlers/proposalsHandlers.ts
bickelj: Yes, the old GoF. I tend to think of code as different from data, process as different from product, etc. In functional language communities, there need not be such a distinction. Then again, Design Patterns talked about OOP and mentions adding behavior, so that is why I think "decorator" is a little bit off: these are data.

src/handlers/proposalsHandlers.ts
bickelj: I tend not to genericize in advance but rather after a thing is used more than once. The whole discussion or review surrounds code that we expect to be copied/pasted so I think your anticipation of re-use is well-founded.

bickelj:
Going way back to the higher level questions: what do you think of the parameter name `includeFieldsAndValues`? This differs, for example, from the parameter `includeFields` proposed for a different endpoint. The discussion is really over in https://github.com/PhilanthropyDataCommons/service/issues/174.

src/handlers/proposalsHandlers.ts
slifty: 100% agree with the general "don't-over-generalize" philosophy!  +1 to the idea that we can anticipate it to some degree here so there's a balance worth striking!

src/handlers/proposalsHandlers.ts
bickelj: I tried without destructuring and `tsc` was unhappy with it later on (these line numbers are after removing the docstring so do not correspond exactly to what was pushed):
```
src/handlers/proposalsHandlers.ts:79:27 - error TS18048: 'newProposal.versions' is possibly 'undefined'.

79     let proposalVersion = newProposal.versions[versionIndex];
                             ~~~~~~~~~~~~~~~~~~~~

src/handlers/proposalsHandlers.ts:86:27 - error TS18048: 'newProposal.versions' is possibly 'undefined'.

86       if (versionIndex >= newProposal.versions.length) {
                             ~~~~~~~~~~~~~~~~~~~~

src/handlers/proposalsHandlers.ts:89:25 - error TS18048: 'newProposal.versions' is possibly 'undefined'.

89       proposalVersion = newProposal.versions[versionIndex];
                           ~~~~~~~~~~~~~~~~~~~~
```
If I keep the destructuring present but then check `if (newProposal.versions === undefined)` then I get this later on:
```
src/handlers/proposalsHandlers.ts:80:27 - error TS18048: 'versions' is possibly 'undefined'.

80     let proposalVersion = versions[versionIndex];
                             ~~~~~~~~

src/handlers/proposalsHandlers.ts:87:27 - error TS18048: 'versions' is possibly 'undefined'.

87       if (versionIndex >= versions.length) {
                             ~~~~~~~~

src/handlers/proposalsHandlers.ts:90:25 - error TS18048: 'versions' is possibly 'undefined'.

90       proposalVersion = versions[versionIndex];
                           ~~~~~~~~
```
I'd rather not check it four times, nor check it once before destructuring and once after destructuring, so I think what we have there is still least worst.

src/handlers/proposalsHandlers.ts
bickelj: The reason for the more fine-grained `await` calls was also perhaps premature optimization. The idea is like this: it is safe to guess that selecting one proposal row will be quicker than selecting M proposal versions rows which will be quicker than selecting N proposal field values rows (and N application form fields rows). After the first two queries complete, there is no reason we can't do the `join` or mash those together while waiting for the remaining two to complete. `Promise.all` will block until the slowest query completes. That said, your suggestion is much easier to reason about when reading the code so I will try it. It will be slightly more complex with the Ajv validations.  I am not sure about whether the types will be preserved either, I will try it and see.

bickelj: The merge that happens first uses the two biggest query results after all. `Promise.all` it is!

bickelj:
c602b6a has most of the suggested changes. I spent way too long trying to figure out how to convince jest to only throw an error on the second call to `db.sql` rather than on the first (the naive "chain another `.mockImplementationOnce` call did not work for me) but just pushed without those two more lines being covered. This commit should be very close to the test coverage goal regardless.

src/handlers/proposalsHandlers.ts
bickelj: I think this one becomes moot by the refactoring in c602b6a.

src/handlers/proposalsHandlers.ts
bickelj: I think this became moot with the refactor in c602b6a as well.

src/handlers/proposalsHandlers.ts
bickelj: It is still unclear to me which situations `tsc` can deduce successfully potential `undefined` vars and which it cannot. I am still surprised at some of the situations it cannot. And for the purposes of this PR, I think this also is moot with the refactor in c602b6a.

src/handlers/proposalsHandlers.ts
bickelj: I used `join` but I am OK with going back to `merge`. For the reason above (a connotation of adding behavior, not data) I dislike `decorate`.

src/errors/ErrorWithStatus.ts
bickelj: I ditched the Error hierarchy changes as part of this PR because I would like to give it more thought when making it and also apply it more consistently/thoroughly in the codebase. In short, it might not happen haha.

bickelj:
ca13026 cleans up `src/ProposalFieldValue.ts` a bit, also based on previous feedback. I think it's time to ring the "good enough" bell. What do you think @slifty ?

bickelj:
0d56542 rebased on main.

bickelj:
9b3bdb1 rebased on main.

slifty (approved):
This looks really great thank you so much for all the iteration!

slifty (changes requested):
Ack so sorry -- so the code looks excellent, but I'm now seeing test coverage isn't 100% (unless I'm being lied to)

That said, it occurs to me that we might consider helper functions that invoke the query and parse the result, which are then tested in isolation and reused across our endpoints rather than having to test every single endpoint...  If you prefer we could just open an issue to take on that debt separately.

bickelj:
Since I could not get that extra one percent in 4 hours I judged it not worth it because we have 80% coverage already, this also has >80% coverage on new code, and does not reduce the coverage by 2%.

The utility/helper function idea might be a more straightforward way to get the coverage rather than mocking only the second call (which I could not get working after many hours of trying).

slifty:
Yeah absolutely -- I'll update back to approve and create an issue!  Thanks for considering it (and for the very legit pushback)

slifty (approved):

bickelj:
551be26 squashed the two commits into one.

PR #215, opened 2023-01-26
Update dependencies that dependabot missed

bickelj:
Dependabot doesn't seem to care about transitive deps. This commit is from a simple `npm update`.


bickelj:
The passage of four tests counts as the review.

PR #232, opened 2023-02-14
Fix and document GET /proposals/{id} endpoint

bickelj:
Add more complete details to the OpenAPI documentation, specifically about the `includeFieldsAndValues` query parameter. Also fix a potential bug where a promise could be left hanging.

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -567,7 +567,7 @@
           {
             "name": "includeFields",
             "in": "query",
-            "description": "Include the fields associated with the application form when 'true', otherwise not.",
+            "description": "Include the fields associated with the application form when `true`, otherwise do not include them.",
             "required": false,
             "schema": {
               "type": "boolean"
@@ -587,7 +587,7 @@
                 },
                 "examples": {
                   "default": {
-                    "summary": "The default, shallow response with no parameters or where includeFields=false",
+                    "summary": "The default, shallow response with no parameters or where `includeFields=false`.",
                     "value": {
                       "id": 3529,
                       "opportunityId": 3203,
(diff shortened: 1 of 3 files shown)


bickelj:
I forgot to mention in the commit message that this also cleans up the tests by moving them into the appropriate scopes.

src/openapi.json
```diff
@@ -732,6 +732,15 @@
             "schema": {
               "type": "integer"
             }
+          },
+          {
+            "name": "includeFieldsAndValues",
+            "in": "query",
+            "description": "Include the application form fields and the associated proposal values when 'true'. This necessarily includes all proposal versions because form fields are related to a particular proposal version.",
```
reefdog, line 739:
```suggestion
            "description": "Include the application form fields and the associated proposal values when `true`. This necessarily includes all proposal versions because form fields are related to a particular proposal version.",
```



This risks appearing like finding fault just to find fault, lol, but I really like using literal backticks in place of the single-quotes around `true` in the description. I doubt Swagger will actually format that as `<code></code>` — aka I don't think it's doing Markdown transformations — but it's a nice way to inline code, boolean values, etc.

reefdog (approved):
This looks fine to me, tests run, the API docs indeed show the explanation and allow using that parameter. 👍

I left a tiny note about a personal preference (with some value) about how to quote literal code or values, but it's a real take-or-leave suggestion.

src/openapi.json
bickelj: This is a great suggestion, let me try it out and see what works best with OpenAPI!

bickelj:
Updates in f76cde9 include backticks in several places when referring to literal values or variable names and also fills out two missing summaries.

src/openapi.json
reefdog: Ooh, it sure does, neat!

image

PR #242, opened 2023-02-24
Upgrade GitHub Actions checkout and setup-node

jasonaowen:
Use version 3 of [actions/checkout](https://github.com/actions/checkout) and [actions/setup-node](https://github.com/actions/setup-node). Some of the workflows were already using version 3, so this makes us consistent.


slifty (approved):
Thank you for this improvement!

PR #253, opened 2023-03-07
Use typescript-eslint recommended rules

jasonaowen:
The typescript-eslint project has [three sets of recommended rules](https://typescript-eslint.io/linting/configs#recommended-configurations). The [`all` configuration](https://typescript-eslint.io/linting/configs#all) is actually called out by the project as a bad idea!

> We do not recommend a TypeScript projects extend from `plugin:@typescript-eslint/all`. Many rules conflict with each other and/or are intended to be configured per-project.

I think some of those potential conflicts are smoothed out by our later `extends`, but let's start from a better base!

Happily, we are already following all of these rules, so we do not need to make any code changes to adopt these rule sets.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,7 +9,9 @@
     "sort-exports"
   ],
   "extends": [
-    "plugin:@typescript-eslint/all",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:@typescript-eslint/recommended-requiring-type-checking",
+    "plugin:@typescript-eslint/strict",
     "airbnb-base",
     "airbnb-typescript/base"
   ],


slifty (approved):
Oh neat -- this is good to know!  I suspect you're right about the smoothing (since so far I don't think we've been bit by anything), but regardless let's not do the explicitly-not-recommended thing.

The only thing I'm left wondering is if there are any rules we *should* configure for our project that aren't being configured by these other items...  But I'll admit I'm not wondering so hard that I feel like it's worth spending time on at this stage of the project; maybe for a future one.

jasonaowen:
@slifty we could get more intense about [arrow functions](https://www.npmjs.com/package/eslint-plugin-prefer-arrow), but I'm a little worried that the package hasn't been updated in a few years. Otherwise, I think this is a pretty good list!

PR #256, opened 2023-03-10
Configure CORS

jasonaowen:
We expect clients, including the [data-viewer](https://github.com/PhilanthropyDataCommons/data-viewer), to make requests from a different domain than the API itself. For security, browsers will not allow [cross-origin requests](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) that include credentials - as are required by every endpoint of this API - without permission from the target server, in order to protect against malicious third-party websites from exploiting a logged-in user's session.

Install and use the [CORS middleware](https://expressjs.com/en/resources/middleware/cors.html) with the default options, which allows requests from any origin. This is not the ideal security policy from a defense-in-depth standpoint, but is useful during development. The current lack of destructive endpoints is one mitigation, and the lack of cookie authentication is another: by default, browser requests do not include the authentication headers we [are planning to] use, so an attacker would need to directly compromise the credentials from the user's browser, not just fire off a request.

I advocate for this initial laxity for the sake of speed; it of course does not preclude a later move to limit CORS requests to a known good list of domains. I suspect that configuration will be somewhat complex, though - either a hardcoded list in the code (which I don't like), some complicated environment variable parsing (which can make deployment more difficult, with string quoting and related issues), or runtime configuration via the database (which is ideal, but effortful).

Issue #96 Authentication
Issue PhilanthropyDataCommons/data-viewer#6 Proof of concept authentication via Keycloak

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "ajv": "^8.12.0",
         "ajv-keywords": "^5.1.0",
+        "cors": "^2.8.5",
         "dotenv": "^16.0.3",
         "express": "^4.18.2",
         "pino": "^8.11.0",
@@ -21,6 +22,7 @@
       },
       "devDependencies": {
         "@tsconfig/node16": "^1.0.3",
+        "@types/cors": "^2.8.13",
         "@types/express": "^4.17.17",
         "@types/jest": "^28.1.2",
         "@types/node": "^18.11.17",
@@ -1371,6 +1373,15 @@
       "integrity": "sha512-t73xJJrvdTjXrn4jLS9VSGRbz0nUY3cl2DMGDU48lKl+HR9dbbjW2A9r3g40VA++mQpy6uuHg33gy7du2BKpog==",
       "dev": true
     },
+    "node_modules/@types/cors": {
+      "version": "2.8.13",
+      "resolved": "https://registry.npmjs.org/@types/cors/-/cors-2.8.13.tgz",
+      "integrity": "sha512-RG8AStHlUiV5ysZQKq97copd2UmVYw3/pRMLefISZ3S1hK104Cwm7iLQ3fTKx+lsUH2CE8FlLaYeEA2LSeqYUA==",
+      "dev": true,
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/express": {
       "version": "4.17.17",
       "resolved": "https://registry.npmjs.org/@types/express/-/express-4.17.17.tgz",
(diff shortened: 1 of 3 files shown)


bickelj (approved):
I see nothing that gives me pause. At the same I didn't find a quick and easy way to test it on a development machine. I suppose it will be straightforward to compare the results on our test machine versus prod machine once this gets automatically deployed to the test environment.

Aside: I also am jealous that the `express-jwt` library hints that it should be this simple but in fact returns a `Promise<void>` rather than `void` and therefore needs a wrapper.

jasonaowen:
Oh right, testing - sorry, I should have included that in the original description!

The simplest demonstration is to make an `OPTIONS` request, and look for the `Access-Control-Allow-*` headers in the response. On `main`:

```sh
$ curl -v -X OPTIONS http://localhost:3001/
*   Trying 127.0.0.1:3001...
* Connected to localhost (127.0.0.1) port 3001 (#0)
> OPTIONS / HTTP/1.1
> Host: localhost:3001
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< X-Powered-By: Express
< Allow: GET,HEAD
< Content-Type: text/html; charset=utf-8
< Content-Length: 8
< ETag: W/"8-ZRAf8oNBS3Bjb/SU2GYZCmbtmXg"
< Date: Tue, 14 Mar 2023 20:43:33 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< 
* Connection #0 to host localhost left intact
GET,HEAD
```

On this branch:

```sh
$ curl -v -X OPTIONS http://localhost:3001/
*   Trying 127.0.0.1:3001...
* Connected to localhost (127.0.0.1) port 3001 (#0)
> OPTIONS / HTTP/1.1
> Host: localhost:3001
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 204 No Content
< X-Powered-By: Express
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: GET,HEAD,PUT,PATCH,POST,DELETE
< Vary: Access-Control-Request-Headers
< Content-Length: 0
< Date: Tue, 14 Mar 2023 20:43:17 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< 
* Connection #0 to host localhost left intact
```

jasonaowen:
Follow up issue for doing CORS right: https://github.com/PhilanthropyDataCommons/service/issues/261

Issue #227, opened 2023-02-07
Add an `externalId` for ApplicationsForms and ApplicationFormFields


slifty:
We want to make it easy for a third party to map their application forms to the versions stored in the PDC.

This will require us to add `externalId` to both the Application Form *and* the Application Form Field.

bickelj:
An alternative that does not require adding columns: use the canonical spreadsheet (or a subset) in CSV form to map these outside the software, e.g. in `data-scripts` repository.

bickelj:
Adding `externalId` will be useful, though.

bickelj:
The reason for addressing this now is it came up with regard to data ingest, e.g. from spreadsheets in #102.

bickelj:
See [comments in the PR](https://github.com/PhilanthropyDataCommons/service/pull/264), either moot or not needed. @slifty if you disagree and think this is still needed, please re-open and clarify.

PR #264, opened 2023-03-20
Add `externalId` to application forms and fields

bickelj:
For data platform providers to see which fields in PDC map to the technical field names in their systems, store an `externalId` on both `applicationForms` and `applicationFormFields`. Supplying the field is mandatory (for consistency) but the value may be `null`.

Issue #227 Add an externalId for ApplicationForms and ApplicationFormFields


src/types/ApplicationForm.ts
```diff
@@ -13,6 +14,7 @@ export interface ApplicationForm {
   readonly id: number;
   opportunityId: number;
   version: number;
+  externalId: string | null;
```
jasonaowen, line 17: The commit message says 

> Supplying the field is mandatory (for consistency) but the value may be `null`.

Can you please expand on that a bit? What is this consistent with, and why is that consistency desirable? `fields`, below, is both optional and nullable.

src/database/migrations/0010-alter-application_forms-external_id.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE application_forms ADD COLUMN external_id VARCHAR;
+
+COMMENT ON COLUMN application_forms.external_id is 'An ID for this application form used by an external system.';
```
jasonaowen, line 3: This comment does not add any new information: we know it's for an application form because of the table, and we know it's an external ID because of the column. "System" is a new word, but that's essentially a stopword in this context.

I'm not against adding useful comments, but I think we'd need to add more detail to qualify. If we don't have any, let's just skip the comment!

```suggestion
```

src/database/migrations/0011-alter-application_form_fields-external_id.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE application_form_fields ADD COLUMN external_id VARCHAR;
+
+COMMENT ON COLUMN application_form_fields.external_id is 'An ID for this field used by an external system.';
```
jasonaowen, line 3: Likewise:

```suggestion
```

src/database/migrations/0010-alter-application_forms-external_id.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE application_forms ADD COLUMN external_id VARCHAR;
```
jasonaowen, line 1: I notice that there is an `external_id` column on both `applicants` and `proposals`. The columns on those tables are both `NOT NULL` and `UNIQUE` (when combined with other columns, in the case of `proposals`). Why are these new `external_id`s nullable and not unique? Or, when would an `application_form` or an `application_form_field` not have an external ID?

jasonaowen (changes requested):
Can you please rebase this onto `main`, @bickelj?

src/types/ApplicationForm.ts
bickelj: Will do in the comments in the code.

src/database/migrations/0011-alter-application_form_fields-external_id.sql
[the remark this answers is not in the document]
jasonaowen: Yes, in `postgres-migrations`, one file is one migration - and each migration is run in its own transaction.

I believe that guidance stems from the constraint that Liquibase has to work with other database management systems, where schema changes cannot be made inside a transaction. PostgreSQL does support data definition statements in transactions - including reverting those changes with `ROLLBACK` - and as the name suggests, this library does not have to worry about any other DBMS.

> I think I will go for the separate changes because an application field is a different thing from an application form even though conceptually we have the same purpose behind both changes.

That sounds good to me! Thanks, @bickelj.

src/types/ApplicationForm.ts
jasonaowen: That will be good, but I did mean it as a genuine question, not only - or even primarily - as a nudge!

bickelj: 1. The `application_forms` table has the `external_id` column (unlike `fields` which is another relation via join).
2. We don't want many types for `ApplicationForm` if we can avoid it.
3. I am not certain that the `externalId` will always be available.
4. The `externalId` is for convenience more than a requirement for PDC to work.

Then again, we do not always have the original order, but we mandate the original order. I am now leaning toward requiring the value even if we make one up on the spot. Having too many optional fields can be a pain.

src/database/migrations/0010-alter-application_forms-external_id.sql
bickelj: Currently they do not. So the application forms and form fields in the database today do not have it. Therefore we need some kind of default value, even if null.

bickelj: There probably exists an ID somewhere (in the external system) but that does not mean we have it. So we either need to delete and start over in order to constrain it to `NOT NULL` or allow `NULL`.

bickelj: This raises the question about whether every data platform provider will definitely have such an ID. I do not know the answer.

bickelj:
This change is probably not urgent and requires a bit more thought, e.g. whether to mandate those fields, and if so what the default values should be for existing forms.

bickelj:
This is pretty stale and nothing bad has happened without the feature yet.

Issue #179, opened 2022-12-15
Paginate API results


reefdog:
_Note: This is essentially a performance optimization and can wait until we need it. I'm opening it anyway to capture a conversation that started in another ticket._

@reefdog in https://github.com/PhilanthropyDataCommons/service/issues/175#issue-1497031419:
> As the dataset grows and we have to support things like pagination, there are going to be request metadata useful to an API consumer, e.g. number of total objects, number of objects in the current response, number of batches available, position of current batch, timestamp of most recent object change, etc.

@bickelj in https://github.com/PhilanthropyDataCommons/service/issues/175#issuecomment-1351986273:
> For pagination, for example, can it not be handled with a pair of query parameters `?lastId=42&count=50?`

@reefdog, after [initially liking the idea](https://github.com/PhilanthropyDataCommons/service/issues/175#issuecomment-1353336191), in https://github.com/PhilanthropyDataCommons/service/issues/175#issuecomment-1353402688:
> I'm realizing `?lastId=42&count=50` limits us to nothing but "give me the next `count` results", which isn't sufficient for a true pagination interface, where we'll also need to go backwards and target a batch of results that aren't contiguous with the last requested batch. (I.e., from page 1 I might want to skip to page 10.)

Our pagination syntax should allow indicating the **starting position** and the **batch count**¹. The API consumer should already know the **total object count**, and ideally this would be returned on every call anyway (either as an HTTP header or a `metadata` sidecar object, TBD in #175), since it may change between calls.

The order of the results should always be consistent between calls; that is, using a stable default order when none is specified so that unless underlying data change, repeated calls to `?start=100&count=50` always return the same 50 objects.

(I don't think we need to worry over-much about inconsistencies caused by new data having arrived in the system since the last request. If the API consumer requests `?start=100&count=50`, and then the database receives 10 new objects inserted at position `0`, and then the API consumer requests `?start=150&count=50`, then the first 10 results will be the last 10 from the previous call. I think that's acceptable.)

----

¹This is really more of a batch _maximum_ since, at the end of the result set, the API may return less than an entire batch.

bickelj:
@reefdog Blast. You're right: true pagination requires a bit more than those two.

bickelj:
How should we paginate when using UUIDs (#72)? Order by created date? Order by UUID?

reefdog:
[...]

PR #265, opened 2023-03-21
Add Pagination to GET /proposals

slifty:
This PR adds pagination to the `GET /proposals` endpoint.

Related to #179

### DIFF
--- a/src/__tests__/openapi.unit.test.ts
+++ b/src/__tests__/openapi.unit.test.ts
@@ -1,0 +1,25 @@
+import apiSpecification from '../openapi.json';
+
+describe('openapi.json', () => {
+  describe('page parameter', () => {
+    it('should have a non-negative default value', () => {
+      expect(apiSpecification.parameters.pageParam.default).toBeGreaterThanOrEqual(1);
+    });
+
+    it('should have an integer default value', () => {
+      const integerValue = Math.floor(apiSpecification.parameters.pageParam.default);
+      expect(integerValue).toEqual(apiSpecification.parameters.pageParam.default);
+    });
+  });
+
+  describe('count parameter', () => {
+    it('should have a non-negative default value', () => {
+      expect(apiSpecification.parameters.countParam.default).toBeGreaterThanOrEqual(1);
+    });
+
+    it('should have an integer default value', () => {
+      const integerValue = Math.floor(apiSpecification.parameters.countParam.default);
+      expect(integerValue).toEqual(apiSpecification.parameters.countParam.default);
+    });
+  });
+});
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -24,7 +24,7 @@ describe('/proposals', () => {
         .expect(200, []);
     });
 
-    it('returns all proposals present in the database', async () => {
+    it('returns proposals present in the database', async () => {
       await db.query(`
         INSERT INTO opportunities (
           title,
@@ -61,20 +61,87 @@ describe('/proposals', () => {
           200,
           [
             {
-              id: 1,
-              externalId: 'proposal-1',
-              applicantId: 1,
-              opportunityId: 1,
-              createdAt: '2525-01-01T00:00:05.000Z',
-            },
-            {
               id: 2,
               externalId: 'proposal-2',
               applicantId: 1,
               opportunityId: 1,
               createdAt: '2525-01-01T00:00:06.000Z',
             },
+            {
+              id: 1,
+              externalId: 'proposal-1',
+              applicantId: 1,
+              opportunityId: 1,
+              createdAt: '2525-01-01T00:00:05.000Z',
+            },
           ],
+        );
+    });
+
+    it('returns according to pagination parameters', async () => {
+      await db.sql('opportunities.insertOne', {
+        title: '🔥',
+      });
+      await db.sql('applicants.insertOne', {
+        externalId: '12345',
+        optedIn: 'true',
+      });
+      await Array.from(Array(20)).reduce(async (p, _, i) => {
+        await p;
+        await db.sql('proposals.insertOne', {
+          applicantId: 1,
+          externalId: `proposal-${i + 1}`,
+          opportunityId: 1,
+        });
+      }, Promise.resolve());
+      await agent
+        .get('/proposals')
+        .query({
+          _page: 2,
+          _count: 5,
+        })
+        .set(authHeader)
+        .expect(200)
+        .expect(
+          (res) => expect(res.body).toEqual(
+            [
+              {
+                id: 15,
+                externalId: 'proposal-15',
+                applicantId: 1,
+                opportunityId: 1,
+                createdAt: expect.stringMatching(isoTimestampPattern) as string,
+              },
+              {
+                id: 14,
+                externalId: 'proposal-14',
+                applicantId: 1,
+                opportunityId: 1,
+                createdAt: expect.stringMatching(isoTimestampPattern) as string,
+              },
+              {
+                id: 13,
+                externalId: 'proposal-13',
+                applicantId: 1,
+                opportunityId: 1,
+                createdAt: expect.stringMatching(isoTimestampPattern) as string,
+              },
+              {
+                id: 12,
+                externalId: 'proposal-12',
+                applicantId: 1,
+                opportunityId: 1,
+                createdAt: expect.stringMatching(isoTimestampPattern) as string,
+              },
+              {
+                id: 11,
+                externalId: 'proposal-11',
+                applicantId: 1,
+                opportunityId: 1,
+                createdAt: expect.stringMatching(isoTimestampPattern) as string,
+              },
+            ],
+          ),
         );
     });
 
--- a/src/database/queries/proposals/selectWithPagination.sql
+++ b/src/database/queries/proposals/selectWithPagination.sql
@@ -3,4 +3,6 @@ SELECT p.id AS "id",
   p.external_id AS "externalId",
   p.opportunity_id AS "opportunityId",
   p.created_at AS "createdAt"
-FROM proposals p;
+FROM proposals p
+ORDER BY p.id DESC
+OFFSET :offset FETCH NEXT :limit ROWS ONLY
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -1,2 +1,3 @@
 export * from './db';
 export * from './migrate';
+export * from './parameters';
--- a/src/database/parameters/__tests__/getLimitValues.unit.test.ts
+++ b/src/database/parameters/__tests__/getLimitValues.unit.test.ts
@@ -1,0 +1,47 @@
+import {
+  getLimitValues,
+} from '..';
+import {
+  InternalValidationError,
+} from '../../../errors';
+
+describe('getLimitValues', () => {
+  it('should return calculated values when valid parameters are provided', () => {
+    const limitValues = getLimitValues({
+      count: 5,
+      page: 2,
+    });
+    expect(limitValues).toEqual({
+      limit: 5,
+      offset: 5,
+    });
+  });
+
+  it('should error if count is a float value', () => {
+    expect(() => getLimitValues({
+      count: 100.5,
+      page: 42,
+    })).toThrow(InternalValidationError);
+  });
+
+  it('should error if count is less than 1', () => {
+    expect(() => getLimitValues({
+      count: 0,
+      page: 42,
+    })).toThrow(InternalValidationError);
+  });
+
+  it('should error if page is less than 1', () => {
+    expect(() => getLimitValues({
+      count: 100,
+      page: 0,
+    })).toThrow(InternalValidationError);
+  });
+
+  it('should error if page is a float vlaue', () => {
+    expect(() => getLimitValues({
+      count: 100,
+      page: 42.5,
+    })).toThrow(InternalValidationError);
+  });
+});
--- a/src/database/parameters/getLimitValues.ts
+++ b/src/database/parameters/getLimitValues.ts
@@ -1,0 +1,28 @@
+import { isPaginationParameters } from '../../types';
+import { InternalValidationError } from '../../errors';
+import type {
+  PaginationParameters,
+} from '../../types';
+
+interface LimitValues {
+  limit: number;
+  offset: number;
+}
+
+export const getLimitValues = (
+  paginationParameters: PaginationParameters,
+): LimitValues => {
+  if (!isPaginationParameters(paginationParameters)) {
+    throw new InternalValidationError(
+      'Invalid pagination parameters',
+      isPaginationParameters.errors ?? [],
+    );
+  }
+  const zeroIndexedPage = paginationParameters.page - 1;
+  const limit = paginationParameters.count;
+  const offset = paginationParameters.count * zeroIndexedPage;
+  return {
+    limit,
+    offset,
+  };
+};
--- a/src/database/parameters/index.ts
+++ b/src/database/parameters/index.ts
@@ -1,0 +1,1 @@
+export * from './getLimitValues';
--- a/src/handlers/proposalsHandlers.ts
+++ b/src/handlers/proposalsHandlers.ts
@@ -1,5 +1,8 @@
 import { getLogger } from '../logger';
-import { db } from '../database';
+import {
+  db,
+  getLimitValues,
+} from '../database';
 import {
   isProposalArray,
   isProposalWrite,
@@ -15,6 +18,9 @@ import {
   InputValidationError,
   NotFoundError,
 } from '../errors';
+import {
+  extractPaginationParameters,
+} from '../queryParameters';
 import type {
   Request,
   Response,
@@ -36,7 +42,13 @@ const getProposals = (
   res: Response,
   next: NextFunction,
 ): void => {
-  db.sql('proposals.selectAll')
+  const paginationParameters = extractPaginationParameters(req);
+  db.sql(
+    'proposals.selectWithPagination',
+    {
+      ...getLimitValues(paginationParameters),
+    },
+  )
     .then((proposalsQueryResult: Result<Proposal>) => {
       logger.debug(proposalsQueryResult);
       const { rows: proposals } = proposalsQueryResult;
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -5,6 +5,26 @@
     "description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
     "version": "0.1.0"
   },
+  "parameters": {
+    "pageParam": {
+      "in": "query",
+      "name": "_page",
+      "required": false,
+      "type": "integer",
+      "minimum": 1,
+      "default": 1,
+      "description": "The page number being retrieved."
+    },
+    "countParam": {
+      "in": "query",
+      "name": "_count",
+      "required": false,
+      "type": "integer",
+      "minimum": 1,
+      "default": 100,
+      "description": "The maximum number of items that should appear per page or request."
+    }
+  },
   "components": {
     "securitySchemes": {
       "auth": {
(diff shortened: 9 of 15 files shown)


src/utils/extractPaginationParameters.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const parseIntegerQueryParameter = (
+  req: Pick<Request, 'query'>,
+  parameterName: string,
+  defaultValue: number,
+): number => {
+  const rawValue = req.query[parameterName];
+  if (typeof rawValue === 'string') {
+    const parsedValue = parseInt(rawValue, 10);
+    return Number.isNaN(parsedValue)
```
slifty, line 16: Should we throw an error if an invalid string is provided or should we treat it as though it wasn't passed in the first place?

Error would be more explicit, so I'm leaning towards changing this to throw a validation error.

slifty:
I'm gonna move the validation to AJV, since we're already using AJV

slifty:
(Those rebase issues are now resolved!)

src/utils/__tests__/generateLimitValues.unit.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      count: 0,
+      page: 42,
+    })).toThrow(InternalValidationError);
+  });
+
+  it('should error if page is less than 1', () => {
+    expect(() => generateLimitValues({
+      count: 100.5,
+      page: 0,
+    })).toThrow(InternalValidationError);
```
jasonaowen, line 38: This would fail validation for multiple reasons, wouldn't it?

```suggestion
  it('should error if page is less than 1', () => {
    expect(() => generateLimitValues({
      count: 100,
      page: 0,
    })).toThrow(InternalValidationError);
```

src/__tests__/proposals.int.test.ts
```diff
@@ -78,6 +78,103 @@ describe('/proposals', () => {
...
 
+    it('returns according to pagination parameters', async () => {
+      await db.query(`
+        INSERT INTO opportunities (
+          title,
+          created_at
+        )
+        VALUES
+          ( '🔥', '2525-01-02T00:00:01Z' )
+      `);
```
jasonaowen, line 89: Do we care about this timestamp? Can we use the existing insert script instead?

```suggestion
      await db.sql('opportunities.insertOne', { title: '🔥' });
```

src/__tests__/proposals.int.test.ts
```diff
@@ -78,6 +78,103 @@ describe('/proposals', () => {
...
+      await db.query(`
+        INSERT INTO applicants (
+          external_id,
+          opted_in,
+          created_at
+        )
+        VALUES
+          ( '12345', 'true', '2022-07-20 12:00:00+0000' ),
+          ( '67890', 'false', '2022-07-20 12:00:00+0000' );
+      `);
```
jasonaowen, line 99: Likewise:

```suggestion
      await db.sql('applicants.insertOne', { externalId: '12345', optedIn: true });
      await db.sql('applicants.insertOne', { externalId: '67890', optedIn: false });
```

Oh and do we need both? I only see references to the first in the test below.

jasonaowen (changes requested):
Cool! I was able to try it in both the UI and via curl.

src/database/queries/proposals/selectWithPagination.sql
bickelj, line 1: What do you think about having a single query to replace both `selectWithPagination` and `selectAll` with default values set for `:limit` and `:offset` when no pagination is requested? For example, default to some very high limit (like max bigint or whatever JavaScript's max integer is) and offset 1. The advantage would be fewer queries to manage/debug later, the same indices that are used for `order by` in the one query would probably stay cached, and perhaps it's fewer query plans so that same plan could get re-used (I am not certain on this last one though). The drawback is it's one more complex query with default values rather than two targeted queries. And now I'm vaguely recalling that we had this conversation in the past. :thinking:

src/utils/index.ts
bickelj, line 1: I prefer not to have `utils` directories because they tend to accumulate unrelated things. Perhaps `src/pagination/`?

bickelj (commented):
Thank you for paginating! I did not dive as deeply as @jasonaowen but I was able to compile, lint, test, and run locally. I didn't have multiple proposals locally, yet, so I lean on your tests there. The only thing that I think needs a change is what @jasonaowen mentioned already, the `ORDER BY` and all my other comments are food for thought.

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -3,4 +3,6 @@ SELECT p.id AS "id",
   p.external_id AS "externalId",
   p.opportunity_id AS "opportunityId",
   p.created_at AS "createdAt"
-FROM proposals p;
+FROM proposals p
+LIMIT :limit
+OFFSET :offset
```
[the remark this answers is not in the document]
slifty: Oh right good call!  Yes I think it would make most sense to add a sort by primary key as the default and we can add more control later as desired.

src/utils/__tests__/generateLimitValues.unit.test.ts
slifty: Good catch -- my initial implementation was zero indexed and I didn't update this test.

src/utils/index.ts
slifty: My concern with `src/pagination/` is it seems very narrow in scope compared to, say, `database` and `middleware` -- this becomes more clear why it might matter when thinking about how similar functions for things like `extractFilterParameters` and `extractSortParameters` may come out down the line.

Would something like `src/utils/pagination` meet the need, or maybe there's a name out there that's more specific than `utils` but still more analogous in scope to `middleware`.

I'll think on it and update!

src/database/queries/proposals/selectWithPagination.sql
slifty: I don't think there's a use case for `selectAll` any more so I just renamed it and added pagination.  Happy to keep it `selectAll` and add default parameters if folks feel like it's better to keep that possibility; alternatively we could implement that at a later date if we find the need for `selectAll` somewhere.

src/database/queries/proposals/selectWithPagination.sql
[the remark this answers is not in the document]
slifty: Neat!  Will update!

src/__tests__/proposals.int.test.ts
slifty: Oh hey I like that way better.

src/utils/__tests__/extractPaginationParameters.unit.test.ts
```diff
@@ -0,0 +1,49 @@
...
+} from '../../errors';
+
+describe('extractPaginationParameters', () => {
+  it('should return default values according to documented defaults', () => {
+    const paginationParameters = extractPaginationParameters({
+      query: {},
+    });
+    expect(paginationParameters).toEqual({
+      page: 1,
+      count: 100,
```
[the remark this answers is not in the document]
slifty: Possibly... we haven't been treating openapi.json as a form of documentation less than a driver of functionality (part of this is because typescript is not great at importing types from json files).

Given that, this would be the first (and only) formal coupling.  It might be surprising behavior for future developers to have *just this* aspect of the openapi spec be a direct driver of code, when nothing else does (I know there's a risk of stale documentation, but again that is a constant risk for all other aspects of the openAPI spec right now).

So the question is: is the DRY / staleness protection worth the "this is different than all the other API functionality's relationship with the spec"?

src/database/queries/proposals/selectWithPagination.sql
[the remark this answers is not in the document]
bickelj: If the sort is by `id DESC`, then there might be some proposals that show up on two pages when a proposal is posted between another user's get requests, right? If it's by `id ASC`, there would be no repetition. If we're eyes-open OK with having occasional overlap on pages due to new proposals, no problem.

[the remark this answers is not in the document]
jasonaowen: Sorting by `id ASC` would mean that new stuff is always at the end, and I think that would probably be a surprising default sort to folks; I feel like most web UIs put most recent stuff first these days. I'm fine either way, though!

src/utils/__tests__/extractPaginationParameters.unit.test.ts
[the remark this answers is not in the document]
slifty: I'll take a look at the oath config process and see if I can leverage that approach.

The bigger question of how we can connect the two in a more formal way is something that I think will warrant a separate issue.  Right now we're defining *every schema* twice -- once in the spec, once in our type.   I spent a few cycles early on trying to find a better way but they all felt pretty brittle at the time.  Worth revisiting.

src/database/queries/proposals/selectWithPagination.sql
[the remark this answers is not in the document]
slifty: We talked about the "might be on two pages if there's a post in between" issue briefly and I believe the verdict was: that's OK!  This is intended as an endpoint to drive *user facing* pagination (if you load page 3 after a post was made, it should reflect what page 3 actually contains.)

If / when we shift to a machine-driven pagination process we will likely want to provide something other than "page" and instead specify something like "_fromId"

src/utils/__tests__/extractPaginationParameters.unit.test.ts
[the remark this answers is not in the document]
jasonaowen: Here's the relevant code: https://github.com/PhilanthropyDataCommons/service/blob/bb25a75dcb218aca886435f5449f39ff6a76d6ee/src/routers/documentationRouter.ts#L36-L45

I thought about suggesting some kind of deep merge of a nested object, but there isn't great first-class support for that in JavaScript and it didn't seem worthwhile to bring in a new dependency specifically for that. If we'll be doing more in that direction, though, maybe it is worthwhile!

> Right now we're defining _every schema_ twice -- once in the spec, once in our type.

I think it is worth differentiating values from types here; I saw you doing some of that work, and I've done similar work before, and yeah unifying types is very very difficult. Default values, though - and whatever other hardcoded constants we might have - seem much more achievable!

jasonaowen:
Part of what we were discussing in #179 was giving the front end the additional clues it will need to build a UI, such as how many pages there are total, and perhaps referenced entities. Is this the right place to continue that discussion, @slifty?

slifty:
@jasonaowen Good question -- I'd lean towards that being a separate PR (I think that the general shape of this PR is relatively self contained even though that is the next step).

If you feel it should be part of this PR I'm happy to do so though.

src/database/queries/proposals/selectWithPagination.sql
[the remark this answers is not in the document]
slifty: I'm starting with `DESC` but will make an issue about adding sort parameters to these endpoints overall so it's easier for us to adjust.

src/database/queries/proposals/selectWithPagination.sql
slifty: I'm leaning towards the rename-for-specificity, with the intention being that if there's a use case for a more vanilla `selectAll` we can look into modifying it to be more complex / support optional parameters..

src/utils/__tests__/extractPaginationParameters.unit.test.ts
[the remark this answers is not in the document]
slifty: Ahh.... digging into this now.

I see why this approach makes lots of sense for a value that *will* be necessarily overriden by a local copy, but I'm not sure that's what's going on here.  I don't think we want these values to be configurable by the server, right?  e.g. should it be that the PDC API might have different defaults API values in different implementations?  It feels to me that default values are actually driven by specification, not by the server.

So while I could have this driven by a constant / env var, that constant should never change between instances.

Meanwhile, we want our spec to be a useful artefact without having to run the server, so I'm going to want to provide the value in the `openapi.json` as well (We want someone to be able to just download the swagger spec from our repo and throw it into their favorite tool / use it as an artefact in their client / etc.)

This means we'd still have redundancy.

All this said I'll see if I can just load it from the swagger spec as you described; to your point this is not the same as a type so I think we should be able to more safely define and access it.  Will circle back with an update in a moment!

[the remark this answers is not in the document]
slifty: ^^
Using these values for the tests are going to get a little weird.

The tests will be *entirely* dynamic, and start to become a bit obscured in logic.  I feel we might be taking steps back here in long term legibility in the name of avoiding redundancy.  When these values change it's a change to the API spec -- just like if we were to rename a parameter name, or a path name, etc.

This might be a strawman, but we don't have parameter names driven by the spec -- it's a dev / review task to make sure the implementation is updated to reflect the spec!  Maybe this is the same concept?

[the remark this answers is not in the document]
slifty: UPDATE: it's super trivial to pull out the value from the spec (the thing that was not trivial was pulling schema definitions / types.)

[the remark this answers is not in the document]
slifty: One note: I'm a bit worried that using the same `apiSpecification.parameters.pageParam.default` value for both the implementation *and* the test is brittle.  This means the test would fail if the api spec was somehow incorrect.

Ultimately by loading the spec in our implementation we are making the spec part of the implementation -- that means if we load the spec into the test we're writing a test that is defined in terms of the implementation, making it possible to do something like break the spec in a way that breaks the implementation and have the tests still pass.

Would it be OK for me to have the test hard coded? There's no real risk of the test becoming out of date since it will be loud if the spec ever changes without a corresponding test update.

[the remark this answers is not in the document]
jasonaowen: > UPDATE: it's super trivial to pull out the value from the spec

Yay! I'm glad!

> Ultimately by loading the spec in our implementation we are making the spec part of the implementation -- that means if we load the spec into the test we're writing a test that is defined in terms of the implementation, making it possible to do something like break the spec in a way that breaks the implementation and have the tests still pass.

It sounds like you're describing a need for tests for the spec! I think there's value in the tests validating that the default value comes from the spec, and perhaps there's also value in having tests that validate that the values in the spec are reasonable - like that they are non-negative numbers, or whatever.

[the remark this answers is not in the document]
slifty: Hmmm fair -- and I suppose the main case I'm worried about (param renaming) would cause a test failure because `apiSpecification.parameters.pageParam.default` would now be trying to get the `default` attribute of a null value, so the test would simply error.

FWIW: I do agree with you Jesse that it'd be better to just make this spec a build artifact in the long run.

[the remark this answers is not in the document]
slifty: Tests updated (and spec tests added!)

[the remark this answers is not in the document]
jasonaowen: > On the issue of whether to generate docs from code or code from docs, I prefer the former. [...] the creation of an `openapi.json` would be a build step [...] with the resulting documentation included in the build artifact.

This sounds plausible to me; can we talk about the conditions that would lead to that? Static assets are simpler (both computationally and conceptually) than built assets, so I think it's worthwhile to avoid that if we can. Does the auth stuff merit this step, or are we still okay with the local overrides at run time? If not auth stuff, what kind of dynamism are you imagining would make us want to construct the OpenAPI docs at build time?

src/__tests__/openapi.unit.test.ts
```diff
@@ -0,0 +1,25 @@
+import apiSpecification from '../openapi.json';
+
+describe('openapi.json', () => {
```
jasonaowen, line 3: Nice!

src/__tests__/proposals.int.test.ts
```diff
@@ -61,23 +61,90 @@ describe('/proposals', () => {
...
+        optedIn: 'true',
+      });
+      await Array.from(Array(20)).reduce(async (p, _, i) => {
+        await p;
+        await db.sql('proposals.insertOne', {
+          applicantId: 1,
+          externalId: `proposal-${i + 1}`,
+          opportunityId: 1,
+        });
+      }, Promise.resolve());
```
jasonaowen, line 96: I haven't seen this pattern before! It's a little hard to parse, although I think I got there.

If you don't mind a little code golf, how does this look to you?

```suggestion
      await Promise.all(Array.from(Array(20), (_, i) => db.sql('proposals.insertOne', {
          applicantId: 1,
          externalId: `proposal-${i + 1}`,
          opportunityId: 1,
        })));
```

A few lines of code shorter, avoids `reduce` which is a difficult function to reason about, and runs the `INSERT`s all in parallel instead of sequentially to hopefully save a bit of test time.

That said, what you have is fine, and I'm 100% okay with you silently resolving this discussion without any changes!

src/__tests__/proposals.int.test.ts
```diff
@@ -61,23 +61,90 @@ describe('/proposals', () => {
...
+        .expect(200)
+        .expect(
+          (res) => expect(res.body).toEqual(
+            [
+              {
+                id: 15,
+                externalId: 'proposal-15',
+                applicantId: 1,
+                opportunityId: 1,
+                createdAt: expect.stringMatching(isoTimestampPattern) as string,
```
jasonaowen, line 113: Nice!

jasonaowen (approved):
Awesome! Nice work, @slifty - thanks again!

src/utils/index.ts
slifty: this is hopefully a step in the right direction but kicked out `/utils` and made:

1. A new `/database/parameters` for the limit value generation script <-- Totally open to a better name, but the intent is that it provides tools for generating certain common parameters for various queries.  I also swapped `generateLimitValues` to `getLimitValues` since I think generate means more things, especially in a db context.

2. A new `/queryParameters` folder for queryparam parsers.  I don't know if this is too specific (e.g. we may find ourselves with common POST or HEADER parameters that need extracting, but we can always change the bucket later and this is better than `utils`)

src/__tests__/proposals.int.test.ts
slifty: The reason I did the obnoxious sequential version is because I (think?) I need the test to have deterministic IDs -- otherwise I won't be able to know for sure which rows are supposed to be returned.

That said, maybe the IDs would be deterministic in parallel!  I was just assuming it would introduce a race condition into the test.

jasonaowen: Except a matcher then wouldn't test the core assertion: that we're sorting by ID. Ugh. Okay, thanks for bearing with me, I'm all caught up now!

slifty: FWIW what I really wanted to do was a for loop (I mean, want is a strong word), but our linter hate hate hated that.

I could just guiltlessly disable the linter since it's a test, and the thing becomes instantly legible.  Thoughts?

jasonaowen: Maybe step back and reconsider what we actually are trying to test, and if there are alternative approaches?

I think the intention of this test is to make sure of is that we're returning the correct subset of results. Except, we're relying on PostgreSQL to do that for us... is this a worthwhile test? Under what circumstances would it meaningfully fail?

Maybe it'd make more sense to test that PostgreSQL will accept the queries we're generating, that we are able to get as many pages as we expect, and that we're not introducing any off-by-one errors in the query construction. None of those require in-order proposal creation, or even a clean database!

slifty: Well, as a for instance... I do want to be sure that our pagination implementation will constructs the right query to return the X most recently created proposals (not simply X proposals).

We're validating our query construction and then using the output of psql's engine (which we trust) to test that query.

At any rate, I know this is resolved; we can always rip out the test later as desired.

src/utils/__tests__/extractPaginationParameters.unit.test.ts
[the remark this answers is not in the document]
bickelj: It's not dynamism as much as specifying everything twice: once in the code and once in the docs. It's really a separate issue from this, though, and the ship may have already sailed.

PR #296, opened 2023-04-11
Support the Bundle wrapper

### LINKED ISSUE #266: Time for a changelog + versions?
It is probably a few moments too soon for a changelog but soon we might have people using the API so I wanted to raise the issue for discussion.

slifty:
This PR introduces a `Bundle` wrapper around the `proposal` collection response.

Once this pattern is approved it will be applied to other collections (along with pagination support)

Related to #179 
Resolves #266


frankduncan:
Schultz's virtuosic technique does not, in the end, make up for his unfamiliarity with the composer's style or his apparent lack of anything approaching an architectural vision for the structure of this notoriously complex piece. Nonetheless, there were fleeting moments of beauty: he drew a gorgeous hushed tone from his instrument in the meditative central section of the adagio, and achieved a dashing, pearly lightness of touch in the scherzo that many a lesser player would envy. Would that his obvious physical gifts were placed in the service of a correspondingly thorough musical conception. But perhaps we ask for too much.

jmergy:
Um

frankduncan:
@jmergy Ah!  Sorry, there was a bit of banter on the OTS zulip that led to that comment.  I removed it.

src/database/accessors/loadBundle.ts
```diff
@@ -0,0 +1,23 @@
...
+import type {
+  Bundle,
+} from '../../types';
+
+export const loadBundle = async <T extends object>(
+  tinyPgQueryName: string,
+  tinyPgQueryParameters: TinyPgParams,
+  tableName: string,
+): Promise<Bundle<T>> => {
+  const { rows: entries } = await db.sql<T>(
```
slifty, line 13: After reflection, I'd like to move this logic to its own `loadObjects` function, which I think is going to be more flexible for future for non-bundle developer use cases.  I'd also want that function to verify the type of the response, which means it (and this) would take a type guard for the objects expected in the return.

jasonaowen: +1 to moving type verification here, but I'm not sure I follow your intentions for `loadObjects`; can you say more about where you'd want to use that?

jasonaowen (commented):
This looks promising in my first pass! A few questions below, and I'll test it this afternoon.

src/openapi.json
```diff
@@ -348,6 +348,42 @@
...
           "createdAt"
         ]
+      },
+      "Bundle": {
+        "type": "object",
+        "properties": {
+          "entries": {
+            "type": "array"
+          },
+          "total": {
```
bickelj, line 358: 3

See (1), we need a detailed definition here regardless of the term chosen.

src/handlers/proposalsHandlers.ts
```diff
@@ -43,26 +43,18 @@ const getProposals = (
...
-          .send(proposals);
-      } else {
-        next(new InternalValidationError(
-          'The database responded with an unexpected format.',
-          isProposalArray.errors ?? [],
-        ));
-      }
-    })
+  (async () => {
+    const proposalBundle = await loadBundle<Proposal>(
```
bickelj, line 47: 4

I like how the use of generics turned out. We can grab a `Bundle<Proposal>`. That is nice and satisfying.

As for the term "load", I think I have seen "load" mean "write" and I have seen "load" mean "read." So I would rather it be "get" or "select" or "read" for reads and "post" or "insert" or "write" for writes. This comment about "load" applies to several more spots, not only here.

src/database/accessors/loadTableMetrics.ts
```diff
@@ -0,0 +1,27 @@
+import { db } from '../db';
+import { isTableMetrics } from '../../types';
+import { InternalValidationError } from '../../errors';
+import type { TableMetrics } from '../../types';
+
+export const loadTableMetrics = async (tableName: string): Promise<TableMetrics> => {
+  const metricsQueryResult = await db.query<TableMetrics>(`
+    SELECT COUNT(id) AS "count",
```
bickelj, line 8: 5

See (2). It might be premature optimization to change this query to explicitly request an estimate, but maybe not. It's good as-is but maybe a comment in the code to note the potential future optimization if this query gets heavy.

bickelj (changes requested):
0

My comments start with an integer indicating the order of review and intended order of reading.

I skimmed https://build.fhir.org/bundle.html because that is the inspiration for this bundle object, correct? Or I suppose that standard and our inspiration share a common ancestor of metadata-in-the-response-object. I am still on the fence about including metadata in a response body because when I go to `/applicants` I expect some form of `Applicant` response as statelessly/deterministically as practical without ever-changing or echoed/repeated/extraneous state included. Also the FHIR looks overly complex for our purposes. But we don't have to include everything, we're cherry-picking that which helps and leaving that which hurts, right?

I didn't review the tests in detail because my skimming indicated no significant changes, only those naturally needed for the `Bundle<T>` change.

I compiled, ran tests, and ran the server locally. I saw the top-level response (I don't have proposals locally at the moment) and trust that the proposals under `entries` would populate as usual.

Overall, I am pretty sure this is what we agreed to and what we want. Having a `Bundle<T>` wrapper further disconnects "what you POST" from "what you GET" but I see the utility and that discussion is still out of scope for this PR which successfully achieves the goal. If we use the `Bundle<T>` consistently across all entities it is easy enough to understand. And if it helps the front-end, I am all for it!

src/types/Bundle.ts
```diff
@@ -0,0 +1,4 @@
+export interface Bundle<T> {
+  total: number;
```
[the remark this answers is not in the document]
jasonaowen: > The `total` can be misinterpreted as the count of entries included in this response, i.e. identical to `entries.length`, but `total` actually includes the overall count of `T` on the server, apart from the arguments given in this request, right?

Correct! Can you clarify if this is a misinterpretation you experienced, or if you are worried about the possibility for misinterpretation? I ask because I would place much more weight on the former!

In general, I would hope we are building an API that does not needlessly repeat information, and hope that developers looking at this API for the first time would give us the grace of assuming we have that goal. Which is to say: we should document it ([as you mention below](https://github.com/PhilanthropyDataCommons/service/pull/296#discussion_r1164549895)), but I actually find `total` to be the clearest name for this concept, in part because it matches what FHIR is doing.

src/types/TableMetrics.ts
```diff
@@ -0,0 +1,32 @@
...
+export const tableMetricsSchema: JSONSchemaType<TableMetrics> = {
+  type: 'object',
+  properties: {
+    count: {
+      type: 'integer',
+    },
+    maxId: {
+      type: 'integer',
+    },
+    now: {
```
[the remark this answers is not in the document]
jasonaowen: If we were to include it in the response, we should use [`timestamp`](https://www.hl7.org/fhir/bundle-definitions.html#Bundle.timestamp) rather than `now`. But we're not currently:

https://github.com/PhilanthropyDataCommons/service/blob/84bc24afedf0383bf9cfff76faa1a463033f2013/src/database/accessors/loadBundle.ts#L19-L22

That said, the argument in favor of including the timestamp in the response rather than just the headers is to allow clients to more easily store responses offline and process them later - including humans who just want to `curl > proposals.json` and then later open that in their editor. This is an important use case in FHIR, but I don't know how important it is to us!

src/handlers/proposalsHandlers.ts
jasonaowen: `load` meaning "write" is very surprising! Where did you see that, @bickelj?

src/database/accessors/loadTableMetrics.ts
jasonaowen: Can you say more here? I'm not sure I understand what you're referencing in [the comment you labeled as 2](https://github.com/PhilanthropyDataCommons/service/pull/296#discussion_r1164520000).

I think the complexity of asking for a [count estimate](https://wiki.postgresql.org/wiki/Count_estimate) is premature. If this becomes a bottleneck, we can discover that through the kind of performance analysis we'll need to be doing at that point anyway - a point I would expect to be far in the future.

src/types/TableMetrics.ts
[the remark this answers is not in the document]
bickelj: If we were ever to include it in the response I would be fine with the `timestamp` name, but would rather it mean something more like "the datetime when the total count was estimated or found" or the "last modified" of the contents of the bundle rather than the FHIR "bundle assembled" definition. If callers can use an API, surely they can get the current datetime locally and/or read the `Date` header in the server response. For example, I similarly would object to including the HTTP status in the body. A [response object should have it all](https://axios-http.com/docs/res_schema), usually, right?

Anyway, not an issue at the moment, as @jasonaowen pointed out, because we aren't including it, thankfully!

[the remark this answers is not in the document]
slifty: Yeah, to be clear -- the response body currently is just using part of the table metrics type (but table metrics is used in testing and the original implementation had it so I maintained it in the refactor)

I would advocate for us to *not* include timestamp in any part of the response right now just in the sense that it isn't necessary to accomplish the intended scope of this PR and there is not an immediate use for it.

src/handlers/proposalsHandlers.ts
slifty: FWIW I'm comfortable using `get` instead of `load`

I think of get as an http verb and load as a database verb, but that distinction is probably invented and might not matter.

src/database/accessors/loadTableMetrics.ts
slifty: I would prefer to not leave a comment in the code but rather open an issue in our issue tracker.  Happy to open that issue, or also comfortable waiting until it becomes a bottleneck of some kind.

src/database/accessors/loadBundle.ts
slifty: > can you say more about where you'd want to use that?

Yes!  I have it implemented in a WIP (which I was really hoping to get pretty and ready for an update to this PR before hitting the road but it is going to have to wait)

Basically, there are instances where we want to get collections of objects but NOT a bundle; in particular when loading collections of objects internally for the purposes of decorating other objects.  For this case I'm writing a `loadObjects` (well, it will probably be `getObjects` based on our future conversation about load vs get) which can be called in those places without the unnecessary "count" overhead / bundle wrapper.

If this doesn't make sense just stay tuned for the commit that adds it -- I might merge the commits later but can at least start with it being separate for the sake of the review process.

src/handlers/proposalsHandlers.ts
```diff
@@ -43,26 +43,18 @@ const getProposals = (
...
-  db.sql(
-    'proposals.selectWithPagination',
-    {
-      ...getLimitValues(paginationParameters),
-    },
-  )
-    .then((proposalsQueryResult: Result<Proposal>) => {
-      logger.debug(proposalsQueryResult);
-      const { rows: proposals } = proposalsQueryResult;
-      if (isProposalArray(proposals)) {
```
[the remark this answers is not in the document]
slifty: The check will continue (when we re-add the verification to the loadBundle accessor)

src/openapi.json
```diff
@@ -348,6 +348,42 @@
           "title",
           "createdAt"
         ]
+      },
+      "Bundle": {
```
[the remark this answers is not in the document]
slifty: I am a bit concerned it could be confusing to link to someone else's documentation from our documentation (since it might cause folks to be unsure of which parts of the FHIR spec we are / are not implementing, and while we are inspired by it we are not committing to follow it now or necessarily in future).

What do you think of instead maintaining that association in the developer logs / the commit message (I should include that in the commit message either way!).

src/handlers/proposalsHandlers.ts
bickelj: Yeah, `get` does have an http ring, and it would feel weird to `post` to the database. However, `load` and `store` work fine. And `read` and `write` would work, though technically all these include both reading from one place and writing to another. :fist_oncoming: :skull_and_crossbones: :horse:

src/types/Bundle.ts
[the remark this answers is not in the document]
slifty: Let me know if the description clears this up.

src/openapi.json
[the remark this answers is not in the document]
slifty: For now I've at least added the link to the comment.  Marking this as resolved BUT please un-resolve if you're feeling it's worth having more discussion about including it in the documentation directly.

src/handlers/proposalsHandlers.ts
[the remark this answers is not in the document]
slifty: This check has been added back -- marking this as resolved.

src/database/accessors/loadBundle.ts
slifty: I ended up just merging it into one commit, since I think it is part of a single complete change.

src/__tests__/applicants.int.test.ts
```diff
@@ -134,16 +134,16 @@ describe('/applicants', () => {
           externalId: '🆔',
         })
         .expect(201);
-      const after = await getTableMetrics('applicants');
+      const after = await loadTableMetrics('applicants');
       logger.debug('after: %o', after);
-      expect(before.count).toEqual('0');
+      expect(before.count).toEqual(0);
```
slifty, line 139: Just wanted to call this out since it is a minor but pleasing improvement: the ajv validation will validate AND coerce the results for us (so a valid number from the database will become a number rather than a string.)

src/handlers/proposalsHandlers.ts
bickelj: Further beating the dead horse: what does the "L" in ETL mean?

src/__tests__/proposals.int.test.ts
```diff
@@ -104,49 +100,52 @@ describe('/proposals', () => {
...
-                id: 11,
-                externalId: 'proposal-11',
-                applicantId: 1,
-                opportunityId: 1,
-                createdAt: expect.stringMatching(isoTimestampPattern) as string,
-              },
-            ],
+            {
+              total: 20,
+              entries: [
```
bickelj, line 105: I like this example of the 5 entries with the 20 total.

src/openapi.json
```diff
@@ -348,6 +348,44 @@
...
+        "type": "object",
+        "properties": {
+          "entries": {
+            "type": "array",
+            "description": "The collection of requested resources."
+          },
+          "total": {
+            "type": "integer",
+            "description": "The total number of requested resources that exist in the database, across all pages."
```
bickelj, line 361: Great, thank you!

src/openapi.json
bickelj, line 1: Now that folks use the API, we should update the documented version. Because `Bundle` is a breaking change, I would call this API version `1.0` or `1.0.0`.

src/types/Bundle.ts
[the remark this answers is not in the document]
bickelj: Yes, the OpenAPI doc resolves it.

src/handlers/proposalsHandlers.ts
slifty: image
image

src/openapi.json
slifty: SemVer does not have the concept of breaking changes before version 1 -- I agree we should update the version, but we should not release v1

Basically: we're on a 0 version which is an indication that every change can be a breaking change; let's keep it that way!

slifty: Updated to 0.2.0 -- I spent a moment seeing if there was a way to explicitly link the api version with our package version but it doesn't look like that's a thing without dynamically generating either `package.json` (lol) or `openapi.json`

src/handlers/proposalsHandlers.ts
```diff
@@ -43,26 +43,19 @@ const getProposals = (
...
-    })
+  (async () => {
+    const proposalBundle = await loadBundle(
+      'proposals.selectWithPagination',
+      {
+        ...getLimitValues(paginationParameters),
+      },
+      'proposals',
+      isProposal,
+    );
```
jasonaowen, line 54: I wonder if this is worth promoting into a named, exported function... we're only doing it twice - and only once in production code - but that's getting to be a lot of stuff that needs to be kept in sync, and it kind of feels like not the handler's responsibility for that.

src/openapi.json
```diff
@@ -348,6 +348,44 @@
...
+          }
```
jasonaowen, line 362: Right now, we are not providing any placeholder data for this field, which apparently inspires OpenAPI to just dream up `total: 0` in its "Example Value" section. It's super cool that it can do that! ...but perhaps somewhat unhelpful in this specific case. Would it be hard to add a non-zero example?

jasonaowen (approved):
This looks good to me! Thank you, @slifty.

I was able to successfully test this locally.

I've made a companion PR https://github.com/PhilanthropyDataCommons/data-viewer/pull/80 . That should be deployed around the same time as this, as deploying one without the other will break the data-viewer.

src/openapi.json
slifty: Super easy -- I'll add that now.

src/openapi.json
slifty: You know what... I'm going to go one step further here and implement a changelog, which will also address https://github.com/PhilanthropyDataCommons/service/issues/266

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog for @pdc/service
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
bickelj, line 6: Given that this is an incompatible change, we would expect a major version update, right? That or we should not mention we follow semver.

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+## [Unreleased]
```
bickelj, line 8: What does this mean? Is it a placeholder for future changes? If empty, perhaps leave it out.

bickelj (commented):
Just a quibble about semver.

src/openapi.json
bickelj: Missed this, thanks, yes, you are right that semver allows for some things I would call silliness.

bickelj (approved):

CHANGELOG.md
slifty: This is something of a permanent placeholder, but I'll take it out for now since we need to figure out how and when we tack on a full blown version.

I would normally only update versions *ON RELEASE* not simply on merge.  This would mean that a PR like this would *not* update a version, but would just document changes in `[Unreleased]`.

When we were ready to deploy to production we would follow a flow like this:

1. Open a PR to update the version based on the collective impact of the unreleased functionality.  This would also involve creating a new version header in the changelog.
2. Tag the commit that updates the version, using the appropriate version tag
3. perform the deployment

This is a little different from the continuous deployment philosophy we talked about last week, but it does allow us to batch the changelog.  Maybe the way to accomplish both is to have a "version number" feature flag that gets baked into our code base.

Anyway, this is all out of scope for this PR I think; I'll just strike this line and then we can talk more about versions / changelogs later!

PR #302, opened 2023-04-19
Configure dependabot to update github actions

slifty:
This PR gets dependabot to keep our actions up to date (in theory!)


bickelj (approved):
Let's try it!

I assume we don't need to limit open PRs from dependabot because we have only a handful of workflows.

PR #305, opened 2023-04-19
Add data seed helper script 2

bickelj:
Developers or users of the API can use this script to add base fields to their local databases. It is intended to run after the first script. In other words, even if you are just starting out, run the script starting with `0001` followed by `0002` and so forth.

This script ran on the pilot database around 2023-04-19T15:08-0600. No errors occurred.

Issue #102 Make canonical fields and application forms from spreadsheets


src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+('Organization Conflict of Interest Policy Document', 'organization_conflict_of_interest_policy_document', '{ "type": "uri" }' ),
+('Organization Human Subjects Research Policy Document', 'organization_human_subjects_research_policy_document', '{ "type": "uri" }' ),
+('Organization IRS Tax Letter Document', 'organization_irs_tax_letter_document', '{ "type": "uri" }' ),
+('Organization Latest Unaudited Financial Statement Document', 'organization_latest_unaudited_financial_statement_document', '{ "type": "uri" }' ),
+('Organization Charter Document', 'organization_charter_document', '{ "type": "uri" }' ),
+('Organization Policy Regarding Data Gathering from Children Document', 'organization_policy_regarding_data_gathering_from_children_document', '{ "type": "uri" }' ),
+('Organization Whistleblower Policy Document', 'organization_whistleblower_policy_document', '{ "type": "uri" }' ),
+('Organization Latitude EPSG:4326', 'organization_latitude_epsg_4326', '{ "type": "number" }' ),
+('Organization Longitude EPSG:4326', 'organization_longitude_epsg_4326', '{ "type": "number" }' ),
```
jasonaowen, line 184: This feels like another example of a split data type: having one without the other is useless, and mixing values from one version to another is actively incorrect. This is actually a useful place for a URI! Specifically, a [`geo` URI](https://en.wikipedia.org/wiki/Geo_URI_scheme).

src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+('Organization Neighborhood', 'organization_neighborhood', '{ "type": "string" }' ),
+('Proposal Abstract', 'proposal_abstract', '{ "type": "string" }' ),
+('Proposal Purpose Statement', 'proposal_purpose_statement', '{ "type": "string" }' ),
+('Proposal Agreement with Fiscal Agent Docume', 'proposal_agreement_with_fiscal_agent_document', '{ "type": "uri" }' );
```
jasonaowen, line 188: Looks like the label got clipped here:

```suggestion
('Proposal Agreement with Fiscal Agent Document', 'proposal_agreement_with_fiscal_agent_document', '{ "type": "uri" }' );
```

src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+-- Add new fields. Generated from 2023-04-19 base fields sheet
+INSERT INTO canonical_fields (label, short_code, data_type) VALUES
+('Organization Location', 'organization_location', '{ "type": "string" }' ),
+('Proposal Additional Contact Name', 'proposal_additional_contact_name', '{ "type": "string" }' ),
+('EIN Foreign Entities', 'ein_foreign_entities', '{ "type": "string" }' ),
+('Proposal Fiscal Sponsor Location', 'proposal_fiscal_sponsor_location', '{ "type": "string" }' ),
+('Proposal Fiscal Sponsor Signatory', 'proposal_fiscal_sponsor_signatory', '{ "type": "string" }' ),
+('Authorized Banking Contacts', 'authorized_banking_contacts', '{ "type": "string" }' ),
+('Organization Information Change', 'organization_information_change', '{ "type": "string" }' ),
+('Updated Contact Information', 'updated_contact_information', '{ "type": "string" }' ),
```
jasonaowen, line 146: What is the intent of this field? Is this handled by versioning in the PDC?

src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+DELETE FROM canonical_fields WHERE short_code = 'when_was_this_overall_rating_created';
+DELETE FROM canonical_fields WHERE short_code = 'who_they_serve';
+DELETE FROM canonical_fields WHERE short_code = 'working_capital_ratio';
+DELETE FROM canonical_fields WHERE short_code = 'working_capital_score';
+DELETE FROM canonical_fields WHERE short_code = 'yes_no_indicator_this_rating_is_a_donor_advisory';
+
+-- Add new fields. Generated from 2023-04-19 base fields sheet
+INSERT INTO canonical_fields (label, short_code, data_type) VALUES
+('Organization Location', 'organization_location', '{ "type": "string" }' ),
+('Proposal Additional Contact Name', 'proposal_additional_contact_name', '{ "type": "string" }' ),
```
jasonaowen, line 140: We can encode multiple values for a single field; does this need to be separate from `proposal_primary_contact_name`?

Alternatively, do we also need the `proposal_additional_contact_{title,phone,email}` fields?

jasonaowen (changes requested):
> This script ran on the pilot database around 2023-04-19T15:08-0600. No errors occurred.

I see these changes in https://api.philanthropydatacommons.org/canonicalFields. I hope it's not too late to review them!

src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+DELETE FROM canonical_fields WHERE short_code = 'what_is_the_percent_of_their_independant_board';
+DELETE FROM canonical_fields WHERE short_code = 'when_was_this_overall_rating_created';
+DELETE FROM canonical_fields WHERE short_code = 'who_they_serve';
+DELETE FROM canonical_fields WHERE short_code = 'working_capital_ratio';
+DELETE FROM canonical_fields WHERE short_code = 'working_capital_score';
+DELETE FROM canonical_fields WHERE short_code = 'yes_no_indicator_this_rating_is_a_donor_advisory';
+
+-- Add new fields. Generated from 2023-04-19 base fields sheet
+INSERT INTO canonical_fields (label, short_code, data_type) VALUES
+('Organization Location', 'organization_location', '{ "type": "string" }' ),
```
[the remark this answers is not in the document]
bickelj: Here's the information I have on this one:
> Concatenated field that displays city/state for US orgs and city for non-US based orgs | This is a field that displays the city/state for US orgs and country for non-US orgs

src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+-- Run this in sequence, in other words, run 0001-...sql followed by 0002-...sql, etc.
+
+-- Usage example 1: `psql -d pdc -U pdc -p 5432 -f 0002-update-fields_and_forms.sql`
+-- Usage example 2: `cat 0002-update-fields_and_forms.sql` then copy/paste into a `psql` session.
+
+-- Fix a typo in an existing field.
+UPDATE canonical_fields
+  SET label = 'Proposal Fiscal Sponsor Contact Last Name',
+    short_code = 'proposal_fiscal_sponsor_contact_last_name'
+  WHERE short_code = 'proposal_fiscal_sponsor_cantact_last_name';
```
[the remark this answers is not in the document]
bickelj: I agree this still needs to be done. It will need to be in another change.

src/database/seeds/0002-update-canonical_fields.sql
```diff
@@ -0,0 +1,188 @@
...
+('Proposal Collects New or Personal Data Explanation', 'proposal_collects_new_or_personal_data_explanation', '{ "type": "string" }' ),
+('Proposal Research Will be Published', 'proposal_research_will_be_published', '{ "type": "boolean" }' ),
+('Organization has Human Subjects Research Policy', 'organization_has_human_subjects_research_policy', '{ "type": "boolean" }' ),
+('Proposal Research Involves Minors', 'proposal_research_involves_minors', '{ "type": "boolean" }' ),
+('Organization has Child Protection Policy', 'organization_has_child_protection_policy', '{ "type": "boolean" }' ),
+('Proposal Funds Fiscal Agent Duties Research or Conference', 'proposal_funds_fiscal_agent_duties_research_or_conference', '{ "type": "boolean" }' ),
+('Proposal Funds Fiscal Agent Duties Research or Conference Explanation', 'proposal_funds_fiscal_agent_duties_research_or_conference_explanation', '{ "type": "string" }' ),
+('Additional Information From Applicant', 'additional_information_from_applicant', '{ "type": "string" }' ),
+('Organization Articles of Incorporation Document', 'organization_articles_of_incorporation_document', '{ "type": "uri" }' ),
```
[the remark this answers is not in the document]
bickelj: Hmm, I don't see that JSON Schema string type available at https://json-schema.org/draft/2020-12/json-schema-validation.html#name-resource-identifiers, I only see `uri`, `uri-reference`, `iri`, `iri-reference`, and `uuid` in that section. I don't see `url` anywhere on the page. But I suppose the JSON Schema here is incorrect since `uri` is supposed to apply to a `string`? Corrected it would be more like this, right?
```json
{ "type": "string", "format": "uri" }
```

src/database/seeds/0002-update-canonical_fields.sql
bickelj: Good call, yes, a geo URI would be better here.

src/database/seeds/0002-update-canonical_fields.sql
bickelj: Nice catch. Yes, this needs to be fixed in another set of changes.

src/database/seeds/0002-update-canonical_fields.sql
bickelj: I was kind of puzzled by this one too. The best description I have is this:
> Description of the changes the grant applicant would like made to their own contact information in the Foundation's database

src/database/seeds/0002-update-canonical_fields.sql
bickelj: Another good call. It is past time to figure out a way to encode these many-to-one and one-to-many relationships in our base fields spreadsheet.

bickelj:
Jason said:
> I see these changes in https://api.philanthropydatacommons.org/canonicalFields. I hope it's not too late to review them!

It's never too late to review. But maybe in the sense of too late to edit this exact script that already ran in production, perhaps it's too late. The changes will need to be in another script.

These changes were intended to unblock a customer from using the API to post proposal data. We expect to continue to improve on all of the values here and also the validation, collapsing, etc. I see you already created issues for most of these.

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
bickelj: I found a good-enough way to deal with this with JSON Schema. I started with this:
```json
{ "type": "string", "allOf": [  { "format": "uri" }, { "pattern": "^https?://.*" } ] }
```
And eventually simplified down to this:
```json
{ "type": "string", "format": "uri", "pattern": "^https?://" }
```

jasonaowen:
> It's never too late to review. But maybe in the sense of too late to edit this exact script that already ran in production, perhaps it's too late. The changes will need to be in another script.

How do you propose we move forward with this PR? And, more generally, how do you propose we handle similar such updates in the future?

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
jasonaowen: Aha, I missed that this was being fed directly into JSON Schema! Yes, that simplification looks good - thanks, @bickelj!

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
jasonaowen: That sounds duplicative of the existing `organization_{city,state_province,country}` fields; can we give the feedback that those should be used instead?

bickelj:
> > It's never too late to review. But maybe in the sense of too late to edit this exact script that already ran in production, perhaps it's too late. The changes will need to be in another script.
> 
> How do you propose we move forward with this PR? And, more generally, how do you propose we handle similar such updates in the future?

In this case, it would make sense to add a third seed helper script to this PR and review that before merge. As for the future, I think most of the time we should review before running in prod, but sometimes the same tack here may be needed. As for the deletes they will fail if they are attempted on a row that is in use by an application form. As we add validation to both the base fields as well as the proposal field values the risks of running something like this drops.

As more folks use the API, more application forms appear, and more folks are looking at proposals, more coordination at the social level will be needed before running these.

bickelj:
c0c3899 rebases on main and includes an updated progress commit, but the idea of data validation should be resolved in #309 or #102 before moving forward.

bickelj:
6fa255f re-rebases on main.

bickelj:
54d0254 rebases on main and adds the 3rd and 4th scripts that ran in production on 2023-05-17 and 2023-05-30 respectively.

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
bickelj: We could, I suppose, but we are not (or were not) in a position to tell other orgs how to organize their data internally, and at the time of the pilot our goal was to get data into PDC.

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
bickelj: Again, the goal is/was to get data into PDC, so I agree with my past self that some other change would be required to get these combined, but that new change will need to be in accord with new approach(es) to fields.

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
bickelj: While I think we have decided to ditch JSON Schema and make our own Domain-Specific Language for the schema, this change is what happened back then and still should be merged.

src/database/seeds/0002-update-canonical_fields.sql
bickelj: Again, I agree that in the long run we want to collapse some data fields and shore them up, that will need to be done with our new approach(es) to fields. This change happened in the pilot production and is still more useful merged than unmerged.

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
bickelj: I created https://github.com/PhilanthropyDataCommons/service/issues/603

src/database/seeds/0002-update-canonical_fields.sql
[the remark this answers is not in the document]
bickelj: Created https://github.com/PhilanthropyDataCommons/service/issues/604

src/database/seeds/0002-update-canonical_fields.sql
bickelj: It looks like this still needs improvement because I see it in the prod base fields. I created https://github.com/PhilanthropyDataCommons/service/issues/605 to track this one.

src/database/seeds/0002-update-canonical_fields.sql
bickelj: #603 might be the right place to consider this too.

src/database/seeds/0002-update-canonical_fields.sql
bickelj: We've made some strides toward figuring out these relationships, I think the issues that capture them are likely https://github.com/PhilanthropyDataCommons/service/issues/396 and https://github.com/PhilanthropyDataCommons/service/issues/333.

bickelj: Created https://github.com/PhilanthropyDataCommons/service/issues/607 for this.

slifty (approved):
I'm fine with merging this for the reasons articulated -- this has happened on prod!  We're going to be looking closely at base fields in general in the coming weeks, and this does reflect the current reality.

src/database/seeds/0002-update-base_fields.sql
jasonaowen, line 1: Was this file both renamed and changed? The diff for this commit is showing that `src/database/seeds/0002-update-canonical_fields.sql` was deleted and `src/database/seeds/0002-update-base_fields.sql` was created, but isn't tracking it as a move operation.

Since it hasn't been merged yet, that rename and doc changes (I assume just `s/canonical/base/`, right? I can't actually tell from the web view) should be squashed into the previous commit.

src/database/seeds/0003-insert-base_fields.sql
jasonaowen, line 1: It looks like both this and the next script have already been run in production. Why split them into two files? If this is intended to document that they were run in production at different times, and to support different sets of partner data, can we document that via the commit message and/or via comments in the script(s)?

jasonaowen (approved):
Thanks for following up on this, @bickelj. I agree that it makes sense to fix this facts-on-the-ground discrepancy. I'd appreciate some additional cleanup before merging, please!

jasonaowen (changes requested):
This needs to be rebased on #639, and descriptions need to be added. (Or, well, probably just empty strings, but...)

bickelj:
We're going for a new direction with #756: users will be adding/editing data in the PDC application using the PDC API and/or application, we will treat the central PDC instance as canonical, and to seed a local database it will be a process of synchronizing base fields from an instance URL (usually the canonical instance). There is no need to throw good stones after bad here, then, and this can be closed. For historical curiosity purposes I think this abandoned PR is sufficient.

Issue #306, opened 2023-04-20
Update the build's workflow dispatch task


bickelj:
A recent build failed: https://github.com/PhilanthropyDataCommons/service/actions/runs/4755734776/workflow

It appears that the workflow dispatch plugin used has not been updated in some time and is using deprecated versions of software.

https://github.com/aurelien-baudet/workflow-dispatch/pull/7 would probably fix it.

bickelj:
It worked. See https://github.com/PhilanthropyDataCommons/service/actions/runs/4756702287/jobs/8452605978#step:11:15, https://github.com/PhilanthropyDataCommons/deploy/actions/runs/4756710333, and  https://chat.opentechstrategies.com/#narrow/stream/75-PDC/topic/ci.2Fcd/near/162614

PR #307, opened 2023-04-20
Use a different workflow dispatch during build

bickelj:
This attempts to resolve the issue of deprecated workflow code. The workflow https://github.com/marketplace/actions/workflow-dispatch-plus appears mostly compatible with the previously used workflow.

Issue #306 Update the build's workflow dispatch task


bickelj:
Because this code runs only on merge to main, because the automated build depends on it, and because the build is currently broken, I am merging without review.

Issue #308, opened 2023-04-21
Validate proposal field values using base field type


bickelj:
The current data model affords the application a `dataType` on base (formerly canonical) fields. This means the application could validate incoming data against the valid `dataType`. ~~The `dataType` is intended to be valid [JSON Schema](https://json-schema.org/draft/2020-12/json-schema-validation.html) except that instead of validating a JSON document it will validate a single value in the proposal.~~ not anymore, see below.

As of this issue, there is no validation of proposal field values posted  to `proposalVersions`. This issue will be resolved by the `proposalVersions` endpoint performing that validation. Making this change will help us resolve (or make progress on) other existing validation issues such as:
https://github.com/PhilanthropyDataCommons/service/issues/282
https://github.com/PhilanthropyDataCommons/service/issues/285
https://github.com/PhilanthropyDataCommons/service/issues/297
https://github.com/PhilanthropyDataCommons/service/issues/298
https://github.com/PhilanthropyDataCommons/service/issues/299

bickelj:
Copying @slifty's comment from https://github.com/PhilanthropyDataCommons/service/pull/309#discussion_r1176843605 here for convenience:
> the type field stores a string (not json) and ideally even an DB level enum. The valid values of the field, for instance, could be:
>
>    string
>    integer
>    phoneNumber
>    url
>    email
>    taxId
>
> ^ those literal strings are what would be in the database.
>
> Then we'd have a switch statement somewhere against those enum values, which then route to the appropriate validator.

bickelj:
Superseded by #760.

PR #309, opened 2023-04-24
Validate base field `dataType` values as JSON

bickelj:
The goal is to use the values of `dataType` for a given base field to validate values of `value` for given proposal field values. The change in this commit is a stepping stone along the way to that goal by first validating the value of `dataType`. We would like `dataType` to be a valid JSON Schema validation value. An example validation value is `{ "type": "string", "format: "date" }`. See other valid values at https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

In order to validate the `dataType` values, we can first exploit the `jsonb` postgresql type. The database itself will therefore validate that the value for `dataType` is JSON. All JSON Schema values are JSON but not all JSON values are JSON Schemas. At the application level, we use Ajv to validate that the `dataType` is not only valid JSON but also a valid JSON Schema.

Issue #308 Validate proposal field values using base field JSON Schema type

### DIFF
--- a/src/__tests__/proposalVersions.int.test.ts
+++ b/src/__tests__/proposalVersions.int.test.ts
@@ -121,8 +121,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
       `);
       await db.query(`
         INSERT INTO application_form_fields (
@@ -520,8 +520,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
       `);
       await db.query(`
         INSERT INTO application_form_fields (
@@ -800,8 +800,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
       `);
       await db.query(`
         INSERT INTO application_form_fields (
@@ -894,8 +894,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
       `);
       await db.query(`
         INSERT INTO application_form_fields (
@@ -1059,8 +1059,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
       `);
       await db.query(`
         INSERT INTO application_form_fields (
@@ -1229,8 +1229,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
       `);
       await db.query(`
         INSERT INTO application_form_fields (
--- a/src/database/migrations/0010-alter-canonical_fields-data_type.sql
+++ b/src/database/migrations/0010-alter-canonical_fields-data_type.sql
@@ -1,0 +1,3 @@
+ALTER TABLE canonical_fields
+  ALTER COLUMN data_type TYPE jsonb
+  USING to_jsonb(data_type);
--- a/src/handlers/canonicalFieldsHandlers.ts
+++ b/src/handlers/canonicalFieldsHandlers.ts
@@ -11,6 +11,7 @@ import {
   InputValidationError,
   InternalValidationError,
 } from '../errors';
+import { jsonSchemaObject } from '../types/JsonSchemaObject';
 import type { JSONSchemaType } from 'ajv';
 import type {
   Request,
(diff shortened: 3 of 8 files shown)


src/types/JsonSchemaObject.ts
```diff
@@ -0,0 +1,27 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+// The goal is to test an object is a valid JSON Schema (not against a schema, IS a schema).
+// This is a temporary interface. There is likely a better JSON Schema interface from ajv.
```
bickelj, line 5: In the interest of time and getting this change out for review, I didn't dig very deeply into Ajv or other libraries to find an existing JSON Schema, er, schema.

jasonaowen: I think instead of defining our own type here, we can use AJV's `JSONSchemaType`.

src/types/JsonSchemaObject.ts
```diff
@@ -0,0 +1,27 @@
...
+      type: 'string',
+      nullable: true,
+    },
+  },
+  required: [
+    'type',
+  ],
+};
+
+export const isJsonSchemaObject = ajv.compile(jsonSchemaObject);
```
jasonaowen, line 27: This does not appear to be called anywhere. If it were, I believe we should instead call [`validateSchema`](https://ajv.js.org/api.html#ajv-validateschema-schema-object-boolean).

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -1059,8 +1059,8 @@ describe('/proposalVersions', () => {
           created_at
         )
         VALUES
-          ( 'First Name', 'firstName', 'string', '2022-07-20 12:00:00+0000' ),
-          ( 'Last Name', 'lastName', 'string', '2022-07-20 12:00:00+0000' );
+          ( 'First Name', 'firstName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' ),
+          ( 'Last Name', 'lastName', '{ "type": "string" }', '2022-07-20 12:00:00+0000' );
```
jasonaowen, line 1063: Can we use the `insertOne` SQL script for all of these? I believe `pg` knows how to convert JavaScript objects into JSONB data, so ideally we should be able to

```ts
db.sql('canonicalFields.insertOne', {
  label: 'First Name',
  shortCode: 'firstName',
  dataType: { type: 'string' },
});
```

I think it's worth splitting out a commit just for switching all these direct SQL calls into uses of existing scripts!

src/database/migrations/0010-alter-canonical_fields-data_type.sql
jasonaowen, line 1: This would actually be a good place for a comment indicating that the contents of the column are to be JSON Schema values!

src/handlers/canonicalFieldsHandlers.ts
```diff
@@ -63,9 +64,7 @@ const postCanonicalFieldBodySchema: JSONSchemaType<Omit<CanonicalField, 'created
     shortCode: {
       type: 'string',
     },
-    dataType: {
-      type: 'string',
-    },
+    dataType: jsonSchemaObject,
```
jasonaowen, line 67: Oh - I don't know if there's an existing, exported schema schema that AJV provides to allow this kind of reference.

src/types/JsonSchemaObject.ts
jasonaowen, line 1: There's a decision being made in this PR that I think is worth discussing: should validations be driven by dynamic JSON Schema definitions that live in the database, or should we provide a static, hardcoded set of validations to choose from?

This PR proposes the dynamic approach. There's merit to that: in some respects it is very simple, it's very flexible, and to technical clients, it clearly communicates (with some additional documentation...) whether a given value will fit a particular base field.

However, there are drawbacks to the dynamic approach, as well. It requires a higher level of technical aptitude in creating new base fields. I don't know if there are any risks to validating against user-supplied schemas; that is somewhat mitigated by the trust we have in the partners we give access to, but is still a concern. It also requires repetition of commonly used schemas. Finally, it has some technical difficulties in validating such schema, as other discussions in this file show.

An alternative is to have a handful of schemas statically defined in code, requiring that base fields refer to a known schema, and validating data using those schemas. It is less flexible, but potentially clearer: it is easier - for everyone, I think - to read `string` or `phone-number` or `url` or whatever than it is to read `{ "type": "string" }` or whatever the JSON schema values for the other two examples would be. It avoids any potential risks to user-supplied schemas, and reduces repetition in the data. Finally, as we're exploring in #305, reviewing and modifying base field data is difficult, and this approach reduces that.

The values in the `data_type` column are currently exposed via the API, so I think it would be a non-trivial change to switch later. The flexibility in the dynamic approach is compelling, but I think on balance I'm still leaning towards the static approach. @slifty, @bickelj, what do you think?

Thanks again @bickelj for prototyping this!

jasonaowen (commented):
Great job getting this ball rolling, @bickelj!

src/types/JsonSchemaObject.ts
bickelj: It looks like `validateSchema` would remove the need for this file entirely. Good call.

src/types/JsonSchemaObject.ts
slifty: +1 to the prototyping thanks!

I had the same question related to validation approach.  My original assumption was that we would create a list of documented and hard coded validations and `type` would basically reference a documented schema definition.  I agree with Jason's items, but want to add some other points in support of the static approach:

1. It ensures consistency within each "type" (we don't risk a situation where a phone number is validated one way for one field, and another way for another field).
2. Validations become part of the documented API specification, and API validation for types of fields is more predictable across environments.
3. It is possible to write unit tests for the validations.
4. Validation behavior (and changes over time) become review-able (in this repository) rather than being a data seed / data entry process.

I would also at this point prefer the hard coded / enum-of-approved-types approach.  It feels more predictable / stable for now, and to Jason's point if we found it limiting we could eventually move to dynamic in future.

src/handlers/canonicalFieldsHandlers.ts
bickelj: There must still be a better Ajv-ish way but I am glad it's not just me that couldn't easily find it.

bickelj: I wish https://ajv.js.org/api.html#ajv-validateschema-schema-object-boolean took the missing `JsonSchema` type instead of an `object`.

src/types/JsonSchemaObject.ts
bickelj: At first glance I still lean toward using the `dataType` as a JSON Schema type as proposed here. The alternative raises more questions and problems than answers or solutions.

While base fields are technically capable of being dynamic, in practice they are created exclusively by a programmer and pushed to a git repo for review by other programmers. I would like to think that a non-programmer could create base fields but I do not foresee this happening any time soon, perhaps ever, under the current assumptions of our schema, etc. A non-programmer should definitely be able to say "something ain't right about this field" and initiate the change. And it should be relatively easy for a programmer to make the change happen. But to make it possible or even easy for a non-programmer to change the base fields is a non-trivial problem that requires much more thought, design, undoing, and redoing. In other words, the ease of updating base fields point is out of scope for this change.

To create a map of short names to type definitions means we are creating our own domain-specific language (DSL) for grant. Such indirection makes it harder to make a base field. First, I must look over the existing custom names to see if there is one that looks OK. Second, I must refer to the definitions of those custom names to see if it might work, and if not, third, I must create a new name and definition in code before I can add the base field. JSON Schema already exists with the purpose of validating data, has wide tool support, and the simple data validations are relatively simple. Even a non-technical person could look at an existing field and copy-paste the JSON Schema for a new field. For people unfamiliar with JSON Schema, instead of learning a PDC-specific language, they would learn a more general and portable tool.

If we did map our DSL names to technical definitions in code, then we'd lose out on the first level of database validation (that the JSON Schema is at least JSON). If we add our DSL names to the database, now we'd have to create a database enum type or lookup table (more joins) and maintain that code too.

The security concern of user-supplied JSON Schema documents is valid. I think it is mitigated (as suggested above) by strong authentication and trust in our customers.

The concern about having technical difficulties is not compelling. I think we'll find something in Ajv or elsewhere to make the validation of a JSON Schema straightforward.

I don't think we can't write unit tests for validations. We can add tests for whatever calls to [ajv.validate](https://ajv.js.org/api.html#ajv-validateschema-schema-object-boolean) we want. Or to whatever wrapper we make around that.

Overall, I wonder if the objections to this approach are hinting at some deeper problem with base fields and their maintenance. Some deeper issue feels at hand rather than "let's validate this way vs that way."

slifty: I suppose I'm not thinking that the database should store a json schema at all, so there's no database validation needed, and at least on my end I'm not talking about a language!

I'm saying we should say a field is type:

1. string
2. integer
4. phoneNumber
5. url
6. email
7. taxId

^ those literal strings are what would be in the database.

Then we'd have a switch statement somewhere on those strings, which routes to the appropriate validator.

Those validators could definitely be unit tested to make sure they reject invalid data and accept valid data.

bickelj:
After discussion today, it is OK to go with the application-level validation and a short list of home-grown types. The reasons above plus the semantic meaning associated with the words in the enum aka DSL (as opposed to mere validity) can be helpful to the API. "I have a 'phoneNumber'" is more meaningful than "I have a '{ "type": "string", "pattern", "\+[0-9]+"}'".

jasonaowen:
Thank you for that summary, @bickelj! And thank you again for this prototype and for the conversation - I appreciate your hard work!

PR #314, opened 2023-04-26
Single Query Approach: Return deep proposals from GET `/proposals`

### LINKED ISSUE #274: Return deep objects on `/proposals`
The `/proposals` endpoint returns shallow objects.

We want the ability to access deep objects from this endpoint, similar to the behavior of `/proposals/{proposalId}`.

We also might want to make this "deep object" return be the default / singular behavior.

There was some discussion about exploring the use of "related resources" in the base "bundle" object that was discussed in #179, but our initial conversation led us to feel this would be an inconvenient approach since clients would need to reconstruct deep relationships in order to get any value.

There was also some discussion about having the deep object return proposal version values as a dictionary / object instead of an array of key value pairs.  This feature would apply to other endpoints as well so it is captured in #276.

slifty:
This PR changes the way we load proposal data to load all related data alongside, and then manually join the results into a single deep object.

Resolve #274

### DIFF
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -35,6 +35,11 @@ describe('/proposals', () => {
         externalId: '12345',
         optedIn: true,
       });
+      await db.sql('canonicalFields.insertOne', {
+        label: 'Summary',
+        shortCode: 'summary',
+        dataType: 'string',
+      });
       await db.sql('proposals.insertOne', {
         applicantId: 1,
         externalId: 'proposal-1',
@@ -44,6 +49,25 @@ describe('/proposals', () => {
         applicantId: 1,
         externalId: 'proposal-2',
         opportunityId: 1,
+      });
+      await db.sql('applicationForms.insertOne', {
+        opportunityId: 1,
+      });
+      await db.sql('proposalVersions.insertOne', {
+        proposalId: 1,
+        applicationFormId: 1,
+      });
+      await db.sql('applicationFormFields.insertOne', {
+        applicationFormId: 1,
+        canonicalFieldId: 1,
+        position: 1,
+        label: 'Short summary',
+      });
+      await db.sql('proposalFieldValues.insertOne', {
+        proposalVersionId: 1,
+        applicationFormFieldId: 1,
+        position: 1,
+        value: 'This is a summary',
       });
       await agent
         .get('/proposals')
@@ -60,6 +84,7 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
                 {
                   id: 1,
@@ -67,6 +92,29 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [{
+                    id: 1,
+                    proposalId: 1,
+                    version: 1,
+                    applicationFormId: 1,
+                    createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                    fieldValues: [{
+                      id: 1,
+                      applicationFormFieldId: 1,
+                      proposalVersionId: 1,
+                      position: 1,
+                      value: 'This is a summary',
+                      createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                      applicationFormField: {
+                        id: 1,
+                        applicationFormId: 1,
+                        canonicalFieldId: 1,
+                        label: 'Short summary',
+                        position: 1,
+                        createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                      },
+                    }],
+                  }],
                 },
               ],
             },
@@ -109,6 +157,7 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
                 {
                   id: 14,
@@ -116,6 +165,7 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
                 {
                   id: 13,
@@ -123,6 +173,7 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
                 {
                   id: 12,
@@ -130,6 +181,7 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
                 {
                   id: 11,
@@ -137,6 +189,7 @@ describe('/proposals', () => {
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
               ],
             },
--- a/src/database/accessors/loadBundle.ts
+++ b/src/database/accessors/loadBundle.ts
@@ -1,7 +1,6 @@
 import { loadTableMetrics } from './loadTableMetrics';
 import { loadObjects } from './loadObjects';
 import type { TinyPgParams } from 'tinypg';
-import type { ValidateFunction } from 'ajv';
 import type {
   Bundle,
 } from '../../types';
@@ -10,12 +9,10 @@ export const loadBundle = async <T extends object>(
   tinyPgQueryName: string,
   tinyPgQueryParameters: TinyPgParams,
   tableName: string,
-  entryValidator: ValidateFunction<T>,
 ): Promise<Bundle<T>> => {
-  const entries = await loadObjects(
+  const entries = await loadObjects<T>(
     tinyPgQueryName,
     tinyPgQueryParameters,
-    entryValidator,
   );
   const metrics = await loadTableMetrics(tableName);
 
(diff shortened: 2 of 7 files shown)


src/database/accessors/loadProposalBundle.ts
```diff
@@ -1,16 +1,222 @@
...
+);
+
+const extractApplicationFormFields = (
+  applicationFormFieldPieces: ApplicationFormFieldPiece[],
+): ApplicationFormField[] => (
+  applicationFormFieldPieces.reduce<ApplicationFormField[]>(
+    (applicationFormFields, applicationFormFieldPiece) => [
+      ...applicationFormFields,
+      {
+        id: applicationFormFieldPiece.applicationFormField_id,
```
slifty, line 61: These extraction functions are unfortunately repeating the shape of the objects they are extracting to -- I don't know of any better way to do this, but it does mean we now have to update the type AND this extraction function.

src/database/accessors/loadObjects.ts
```diff
@@ -1,26 +1,14 @@
...
   tinyPgQueryName: string,
   tinyPgQueryParameters: TinyPgParams,
-  objectValidator: ValidateFunction<T>,
 ): Promise<T[]> => {
   const { rows } = await db.sql<T>(
     tinyPgQueryName,
     tinyPgQueryParameters,
   );
 
-  rows.forEach((row) => {
```
slifty, line 16: We don't want this check anyway (though we should make sure we're OK with the lack of schema-based coersion, as the validators were performing coersion / ensuring that the resulting properties types aligned with the expected type.)

src/database/accessors/loadBundle.ts
```diff
@@ -10,12 +9,10 @@ export const loadBundle = async <T extends object>(
   tinyPgQueryName: string,
   tinyPgQueryParameters: TinyPgParams,
   tableName: string,
-  entryValidator: ValidateFunction<T>,
 ): Promise<Bundle<T>> => {
-  const entries = await loadObjects(
+  const entries = await loadObjects<T>(
```
slifty, line 13: I think only reason for `loadObjects` outside of `loadBundle` is if we were using an enrichment pattern.  I'm not sure we have need to load objects outside of a bundle context if we are going with the "one complex query per type of object selection" approach.

src/database/accessors/loadProposalBundle.ts
```diff
@@ -1,16 +1,222 @@
...
+        position: applicationFormFieldPiece.applicationFormField_position,
+        label: applicationFormFieldPiece.applicationFormField_label,
+        createdAt: applicationFormFieldPiece.applicationFormField_createdAt,
+      },
+    ],
+    [],
+  ).filter(filterByDistinctId)
+);
+
+const extractProposalFieldValues = (
```
slifty, line 73: These extractors + their "Piece" correlated types stand alone outside of proposals.  I'm not certain how it should be organized yet, but would do so before merging this PR if we go this route.

src/database/accessors/loadProposalBundle.ts
```diff
@@ -1,16 +1,222 @@
...
+        opportunityId: proposalPiece.proposal_opportunityId,
+        externalId: proposalPiece.proposal_externalId,
+        createdAt: proposalPiece.proposal_createdAt,
+      },
+    ],
+    [],
+  ).filter(filterByDistinctId)
+);
+
+const joinProposalFieldValuesWithApplicationFormFields = (
```
slifty, line 128: Similar to the piece extraction: these joiners (which maybe we could / should still call `enrichers`) belong in their own directory as well.

src/__tests__/proposals.int.test.ts
```diff
@@ -60,13 +84,37 @@ describe('/proposals', () => {
...
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [],
                 },
                 {
                   id: 1,
                   externalId: 'proposal-1',
                   applicantId: 1,
                   opportunityId: 1,
                   createdAt: expect.stringMatching(isoTimestampPattern) as string,
+                  versions: [{
```
slifty, line 95: Note that this structure / this test is identical to the test in #313 -- which is to say, both of these implementations result in the same API result.

src/database/accessors/loadProposalBundle.ts
```diff
@@ -1,16 +1,222 @@
...
+    queryParameters,
+    'proposals',
+  );
+  const deepProposalPieces = unprocessedBundle.entries;
+  const applicationFormFields = extractApplicationFormFields(
+    deepProposalPieces.filter(isApplicationFormFieldPiece),
+  );
+  const proposalFieldValues = joinProposalFieldValuesWithApplicationFormFields(
+    extractProposalFieldValues(
+      deepProposalPieces.filter(isProposalFieldValuePiece),
```
jasonaowen, line 202: I think I understand why the `filter`ing is done here - to get better type hints, right? - but it'd kinda be nice if that could happen in the `extract`or functions.

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -1,8 +1,33 @@
...
 FROM proposals p
+  LEFT JOIN proposal_versions pv on pv.proposal_id = p.id
+  LEFT JOIN proposal_field_values pfv on pfv.proposal_version_id = pv.id
+  LEFT JOIN application_form_fields aff on pfv.application_form_field_id = aff.id
+WHERE p.id = ANY(
+  SELECT inner_p.id
+  FROM proposals inner_p
+  ORDER BY inner_p.id DESC
+  OFFSET :offset FETCH NEXT :limit ROWS ONLY
+)
```
jasonaowen, line 32: As a small stylistic suggestion, I think this would be a tiny bit clearer as a [CTE](https://www.postgresql.org/docs/current/queries-with.html) than as a subquery:

```sql
WITH paginated_proposals AS (
  SELECT id
  FROM proposals
  ORDER BY id DESC
  OFFSET :offset FETCH NEXT :limit ROWS ONLY
)
SELECT ...
FROM paginated_proposals
  INNER JOIN proposals p ON paginated_proposals.id = p.id
  ...
```

It makes clearer that the subquery is not referring to any data in the enclosing query, and I find it a little easier to think through joins than subqueries.

Either way, I don't think the `inner_p` alias is necessary.

src/database/accessors/loadObjects.ts
jasonaowen: This is a great callout! I don't have all the relevant pieces in my head - does this mean that any `{ "type": "number" }` fields will start returning `string`s instead? Could/should we still apply these type conversions via validation on the narrowed objects?

jasonaowen (commented):
Having seen it in practice, I think I do still prefer this approach! It's got some rough edges, but hopefully the suggestions below will file some of those off.

Now that you've seen both, @slifty, which way are you leaning?

src/database/accessors/loadProposalBundle.ts
```diff
@@ -1,16 +1,222 @@
...
+        id: proposalPiece.proposal_id,
+        applicantId: proposalPiece.proposal_applicantId,
+        opportunityId: proposalPiece.proposal_opportunityId,
+        externalId: proposalPiece.proposal_externalId,
+        createdAt: proposalPiece.proposal_createdAt,
+      },
+    ],
+    [],
+  ).filter(filterByDistinctId)
+);
```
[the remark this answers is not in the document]
slifty: Oh yeah -- I had started with reduce thinking I would try to filter dupes at the same time but then decided that would be larger than O(n) and it'd be better to just do it in two steps.

src/database/accessors/loadProposalBundle.ts
slifty: It's required / the filter is a type guard.

The alternative would be to have the extract functions take in a `DeepProposalPiece` but then they would not be re-usable across other deep objects.

src/database/queries/proposals/selectWithPagination.sql
slifty: Oh that's much better (I admit my SQL skills these days are about 10 years out of date so thank you)

src/database/accessors/loadProposalBundle.ts
jasonaowen: Yeah, thank you for articulating that more clearly - that's what I was getting at. Kind of unfortunate.

src/database/accessors/loadProposalBundle.ts
slifty: It took me so long to find where on this interface Ic ould actually respond to this comment!

My concern with this is that, as I understand it, it would mean baking implicit magic /into attribute names.  I've kind of made it a blanket rule for me that if I'm ever building business logic that depends on a naming pattern, I'm doing something that should be more explicit.

Basically: this approach would trade explicitness for a level of dynamic mapping that makes me a bit uncomfortable that we're being a bit too clever...

I'm also not entirely sure that the above would pass type safety muster, which means we could never further decorate or preflight-validate these objects

For instance, taking this approach would mean if we ever got fancy and figured out how to map our openAPI specification to our handlers, the handlers would not be returning objects that could be statically analyzed to being the valid / expected type.

slifty: ^  All this is to say...

I think if we go this route we do need to be explicit and verbose :-/

I definitely don't love that part of course, for the same reasons you don't -- (which honestly is why I had lobbied for that enrichment pattern way back when) -- but I can't quite think of a good solution that maintains explicitness without having to be redundant...

Maybe instead of passing <T> going back to passing a type guard for T would be enough -- that way we're confirming that the mapping is real and correct, and that we do indeed have type safety from our magically named fields.

slifty: I just realized the json returns that are being explored [here](https://chat.opentechstrategies.com/#narrow/stream/75-PDC/topic/deep.20loading.20exploration/near/163035) might make this conversation moot to begin with!

Maybe we should hit pause on this for now until we resolve that?

slifty:
I *think* this is something to close now.  It's just there for example now (very stale) and when we explore the data model this might become moot anyway.

bickelj:
> I _think_ this is something to close now. It's just there for example now (very stale) and when we explore the data model this might become moot anyway.

I think so too. If we ever want to revisit it's still here and in the branch right?

slifty:
@bickelj it's still here, though I think I'll delete the branch since the branch represents clutter as well.

PR #317, opened 2023-04-28
Validate and fix our swagger specification

### LINKED ISSUE #316: openapi.json Issue
I'm seeing errors in the openapi.json file and client generation fails with the following error:
Errors: 
	-attribute parameters is unexpected
	-attribute components.schemas.Bundle.items is missing
	-attribute paths.'/proposals'(get).parameters.There are duplicate parameter values

	at org.openapitools.codegen.config.CodegenConfigurator.toContext(CodegenConfigurator.java:541)
	at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:568)
	at org.openapitools.codegen.cmd.Generate.execute(Generate.java:424)
	at org.openapitools.codegen.cmd.OpenApiGeneratorCommand.run(OpenApiGeneratorCommand.java:32)
	at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:66)

slifty:
This PR adds a validation tool to ensure our swagger specification is correctly formatted.  It also makes some corrections to the specification based on errors raised by the validator.

Resolves #316

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,6 +45,7 @@
         "pino-pretty": "^10.0.0",
         "rimraf": "^5.0.0",
         "supertest": "^6.3.3",
+        "swagger-spec-validator": "^5.0.0",
         "ts-jest": "^28.0.5",
         "ts-node": "^10.8.1",
         "typescript": "^5.0.4"
@@ -2456,6 +2457,15 @@
       },
       "engines": {
         "node": ">= 0.8"
+      }
+    },
+    "node_modules/commander": {
+      "version": "9.5.0",
+      "resolved": "https://registry.npmjs.org/commander/-/commander-9.5.0.tgz",
+      "integrity": "sha512-KRs7WVDKg86PWiuAqhDrAQnTXZKraVcCc6vFdL14qrZ/DcWwuRo7VoiYXalXO7S5GKpqYiVEwCbgFDfxNHKJBQ==",
+      "dev": true,
+      "engines": {
+        "node": "^12.20.0 || >=14"
       }
     },
     "node_modules/component-emitter": {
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -45,6 +45,7 @@
         "pino-pretty": "^10.0.0",
         "rimraf": "^5.0.0",
         "supertest": "^6.3.3",
+        "swagger-spec-validator": "^5.0.0",
```
bickelj, line 48: Thanks for the warning regarding the online server use. I thought I would recommend [redocly](https://www.npmjs.com/package/@redocly/cli) (from [openapi.tools](https://openapi.tools/#description-validators)) but when I ran it it gave like 25 more errors on this branch. I suppose if those are substantial (i.e. a generator won't work) then we should switch. In any case, this is certainly an improvement.

src/openapi.json
```diff
@@ -5,36 +5,40 @@
...
     "securitySchemes": {
       "auth": {
         "type": "oauth2",
         "description":  "The OAuth 2.0 (and OpenID Connect) Authorization Code flow.",
         "flows": {
           "authorizationCode": {
-            "authorizationUrl": "${AUTH_SERVER_ISSUER}/protocol/openid-connect/auth",
-            "tokenUrl": "${AUTH_SERVER_ISSUER}/protocol/openid-connect/token",
-            "refreshUrl": "${AUTH_SERVER_ISSUER}/protocol/openid-connect/token",
+            "authorizationUrl": "https://example.com/protocol/openid-connect/auth",
```
bickelj, line 39: I'm guessing this will fix lint but then point the auto-generated clients in the wrong direction. We probably will want to serve up the actual, final openAPI spec in production rather than pointing folks to the git repo.

bickelj (approved):
It's an improvement worth merging, but I expect we will need to keep improving it.

src/openapi.json
slifty: Agreed -- I'll open an issue for this!

package-lock.json
slifty: Absolutely -- I'll open an issue to capture this as well!

Issue #277, opened 2023-03-28
Search capability for proposal data


bickelj:
The ability to search within proposal field values has been suggested as an initial search capability.

We envision search to have (at least) two new components in the implementation:
1. Use/configuration of [postgresql search](https://www.postgresql.org/docs/14/textsearch.html).
2. Endpoint(s) in the API to perform the search queries.

jasonaowen:
For the first draft, we'll be okay with low [recall](https://en.wikipedia.org/wiki/Precision_and_recall) results - that is, it's more important to get some kind of search working that we can then iterate on, rather than spending time and effort preemptively maximizing the number of relevant search results found.

During today's call, @jim-mcgowan confirmed that - again, at least for a first iteration - it would be okay for multi-word search queries to *not* return proposals that had one search term in one field and another in another field. For example, if the search were "children climate", it would be acceptable for a proposal that had the fields "Organization Name: Children Prosperity Endowment" and "Proposal Title: Countering Climate Change" to not be returned in the search results, but a proposal that had the field "Proposal Title: Children and Climate Change" would be returned.

That suggests that we only need a single full-text-search index on `proposal_field_values.value`, plus the database queries to find it and the API endpoints/parameters to parse the request and return results. I'd suggest adding a `q` parameter to `/proposals`, so like `/proposals?q=children+climate` for the earlier example.

Finally, as @bickelj alluded to, [PostgreSQL Full Text Search is Good Enough](https://rachbelaid.com/postgres-full-text-search-is-good-enough/). Later we might decide that we need much deeper search functionality, and at that point we can reconsider our approach and evaluate Apache Solr or Meilisearch or whatever, but for now let's just use the very powerful tools we already have!

jasonaowen:
Another recent how-to post and discussion I saw recently, which I hope will be helpful: [PostgreSQL Full Text Search is better than...](https://admcpr.com/postgres-full-text-search-is-better-than-part1/) and [Reddit discussion](https://www.reddit.com/r/PostgreSQL/comments/12u30a7/postgres_full_text_search_is_better_than/).

slifty:
Resolved by #320

PR #320, opened 2023-04-28
Support search on GET `/proposals`

slifty:
This PR adds a new `_q` parameter to GET `/proposals`.

This does NOT attempt to optimize anything about full text search (that should happen next!).

Issue #277 Search capability for proposal data

### DIFF
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -123,6 +123,104 @@ describe('/proposals', () => {
         );
     });
 
+    it('returns a subset of proposals present in the database when search is provided', async () => {
+      await db.sql('opportunities.insertOne', {
+        title: '🔥',
+      });
+      await db.sql('applicants.insertOne', {
+        externalId: '12345',
+        optedIn: true,
+      });
+      await db.sql('canonicalFields.insertOne', {
+        label: 'Summary',
+        shortCode: 'summary',
+        dataType: 'string',
+      });
+      await db.sql('proposals.insertOne', {
+        applicantId: 1,
+        externalId: 'proposal-1',
+        opportunityId: 1,
+      });
+      await db.sql('proposals.insertOne', {
+        applicantId: 1,
+        externalId: 'proposal-2',
+        opportunityId: 1,
+      });
+      await db.sql('applicationForms.insertOne', {
+        opportunityId: 1,
+      });
+      await db.sql('proposalVersions.insertOne', {
+        proposalId: 1,
+        applicationFormId: 1,
+      });
+      await db.sql('proposalVersions.insertOne', {
+        proposalId: 2,
+        applicationFormId: 1,
+      });
+      await db.sql('applicationFormFields.insertOne', {
+        applicationFormId: 1,
+        canonicalFieldId: 1,
+        position: 1,
+        label: 'Short summary',
+      });
+      await db.sql('proposalFieldValues.insertOne', {
+        proposalVersionId: 1,
+        applicationFormFieldId: 1,
+        position: 1,
+        value: 'This is a summary',
+      });
+      await db.sql('proposalFieldValues.insertOne', {
+        proposalVersionId: 2,
+        applicationFormFieldId: 1,
+        position: 1,
+        value: 'This is a pair of pants',
+      });
+
+      await agent
+        .get('/proposals?_content=summary')
+        .set(authHeader)
+        .expect(200)
+        .expect(
+          (res) => expect(res.body).toEqual(
+            {
+              total: 2,
+              entries: [
+                {
+                  id: 1,
+                  externalId: 'proposal-1',
+                  applicantId: 1,
+                  opportunityId: 1,
+                  createdAt: expectTimestamp,
+                  versions: [{
+                    id: 1,
+                    proposalId: 1,
+                    version: 1,
+                    applicationFormId: 1,
+                    createdAt: expectTimestamp,
+                    fieldValues: [{
+                      id: 1,
+                      applicationFormFieldId: 1,
+                      proposalVersionId: 1,
+                      position: 1,
+                      value: 'This is a summary',
+                      createdAt: expectTimestamp,
+                      applicationFormField: {
+                        id: 1,
+                        applicationFormId: 1,
+                        canonicalFieldId: 1,
+                        label: 'Short summary',
+                        position: 1,
+                        createdAt: expectTimestamp,
+                      },
+                    }],
+                  }],
+                },
+              ],
+            },
+          ),
+        );
+    });
+
     it('returns according to pagination parameters', async () => {
       await db.sql('opportunities.insertOne', {
         title: '🔥',
--- a/src/database/queries/proposals/selectWithPagination.sql
+++ b/src/database/queries/proposals/selectWithPagination.sql
@@ -4,5 +4,15 @@ SELECT p.id AS "id",
   p.opportunity_id AS "opportunityId",
   p.created_at AS "createdAt"
 FROM proposals p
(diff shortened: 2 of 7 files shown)


src/openapi.json
```diff
@@ -28,6 +28,16 @@
...
         "required": false,
         "description": "The maximum number of items that should appear per page or request."
+      },
+      "searchParam": {
+        "schema": {
+          "type": "string",
+          "default": ""
+        },
+        "in": "query",
+        "name": "_q",
```
slifty, line 38: I just thought to actually look at FHIR (whoops for not doing this sooner) -- They reference `_search` [in their docs](https://www.hl7.org/fhir/search.html#searchinputs) though it's an indirect reference.  Anyway, I'd probably swap to that.

jasonaowen (commented):
This looks like a good first pass! I'm very happy with how simple the PostgreSQL side was; it's almost magical.

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -4,5 +4,15 @@ SELECT p.id AS "id",
   p.opportunity_id AS "opportunityId",
   p.created_at AS "createdAt"
 FROM proposals p
+  LEFT JOIN proposal_versions pv ON pv.proposal_id = p.id
+  LEFT JOIN proposal_field_values pfv on pfv.proposal_version_id = pv.id
+WHERE
+  CASE
+    WHEN :search::text != '' THEN
+      to_tsvector(pfv.value) @@ websearch_to_tsquery(:search::text)
```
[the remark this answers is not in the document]
slifty: I'll include that in this PR! I just wanted to make sure I was doing it right in the first place before adding it in.

src/queryParameters/extractSearchParameters.ts
```diff
@@ -0,0 +1,29 @@
...
+): SearchParameters => {
+  if (!isSearchParametersQuery(query)) {
+    throw new InputValidationError(
+      'Invalid search parameters.',
+      isSearchParametersQuery.errors ?? [],
+    );
+  }
+  /* eslint-disable no-underscore-dangle */
+  return {
+    search: query._q ?? apiSpecification.components.parameters.searchParam.schema.default,
```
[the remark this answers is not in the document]
slifty: It is!  But I also feel like we set this pattern and it would be surprising to have defaults driven by the spec in one place and not another...

src/types/SearchParameters.ts
```diff
@@ -0,0 +1,20 @@
...
+    search: {
+      type: 'string',
+    },
+  },
+  required: [
+    'search',
+  ],
+};
+
+export const isSearchParameters = ajv.compile(searchParametersSchema);
```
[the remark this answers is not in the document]
slifty: True that -- frankly I just made it a habit to always write type guards with types (this is rooted in a wish that it were standard practice for libraries to provide type guards!  but this isn't a library!).

src/types/SearchParameters.ts
```diff
@@ -0,0 +1,20 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+
+export interface SearchParameters {
```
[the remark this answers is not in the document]
slifty: Oh good call, I bet it would, I'll do that (unless the linter yells somehow but I doubt it).

src/openapi.json
slifty: `_content` it is!

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -4,5 +4,15 @@ SELECT p.id AS "id",
   p.opportunity_id AS "opportunityId",
   p.created_at AS "createdAt"
 FROM proposals p
+  LEFT JOIN proposal_versions pv ON pv.proposal_id = p.id
```
slifty, line 7: I don't know if this has a performance impact we need to worry about, but these joins are unnecessary if `:search` is undefined.  I wasn't quite sure how to have the joins themselves only occur in a case clause.

src/database/queries/proposals/selectWithPagination.sql
jasonaowen, line 1: Fun fact: `psql(1)` is a good way to play with PostgreSQL's stemming:

```psql
pdc=> select to_tsvector('propose proposal proposed');
  to_tsvector   
----------------
 'propos':1,2,3
(1 row)

pdc=> select to_tsvector('new newer newest');
         to_tsvector          
------------------------------
 'new':1 'newer':2 'newest':3
(1 row)
```

I'm finding that helpful in testing against my local data!

src/queryParameters/extractSearchParameters.ts
```diff
@@ -0,0 +1,12 @@
+import apiSpecification from '../openapi.json';
+import type {
+  Request,
+} from 'express';
+
+export const extractSearchParameters = (
+  { query }: Pick<Request, 'query'>,
```
jasonaowen, line 7: Why `Pick<Request...>` and not just `Request`?

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -4,5 +4,15 @@ SELECT p.id AS "id",
...
+  LEFT JOIN proposal_field_values pfv on pfv.proposal_version_id = pv.id
+WHERE
+  CASE
+    WHEN :search::text != '' THEN
+      pfv.value_search @@ websearch_to_tsquery(:search::text)
+    ELSE
+      true
+    END
+GROUP BY p.id
 ORDER BY p.id DESC
```
jasonaowen, line 17: Right now we're considering search results to be a filter, but probably we want to [rank the results](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-RANKING). This is another argument for splitting the search/nonsearch cases into separate queries. This is probably also future work, as I think we'd want to include the rank in the API response, and we don't need to blow up this PR or block on this suggested enhancement.

jasonaowen (approved):
I was able to run searches in the swagger docs, as well as locally modify data-viewer to send the search query to the server. Very cool! Thanks for leading the way on this, @slifty.

src/__tests__/proposals.int.test.ts
```diff
@@ -123,6 +123,104 @@ describe('/proposals', () => {
         );
     });
 
+    it('returns a subset of proposals present in the database when search is provided', async () => {
+      await db.sql('opportunities.insertOne', {
+        title: '🔥',
+      });
+      await db.sql('applicants.insertOne', {
```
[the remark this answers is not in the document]
slifty: Agreed but yes let's defer this.

src/database/queries/proposals/selectWithPagination.sql
slifty: I'll make an issue for this.

src/queryParameters/extractSearchParameters.ts
slifty: Ah yes, at this point Request is fine given that there is no validation occurring -- in a world with validation this would have signaled that the request is expecting the request to have a `query` value, I think was the logic!

src/__tests__/proposals.int.test.ts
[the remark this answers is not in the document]
slifty: Opened issue #331

src/database/queries/proposals/selectWithPagination.sql
slifty: Created issue #332

src/database/queries/proposals/selectWithPagination.sql
slifty: Understood -- I think this will be something to explore in more depth when we address #314

PR #321, opened 2023-05-01
Lint SQL with SQLFluff

jasonaowen:
One of the great advantages to using tinypg and postgres-migrations is the ability to separate our SQL queries and migrations into SQL files. We've already been benefiting from the editor support that approach gains us!

Another benefit is the ability to use static analysis to improve our SQL coding practices. I propose that we should start using [SQLFluff](https://docs.sqlfluff.com/en/stable/index.html), a SQL linter that works with PostgreSQL-flavored SQL. This is the best tool I've found to do so.

In the spirit of recent conversations, I wanted to open this as a draft PR to make sure the dev team is aligned on this. There's still work to be done to make this ready to merge, and I won't do that unless we agree that it's a good idea!

I don't think SQL is the strongest language of anyone on this team, so I think we can all learn from the rules this linter offers - I know in the past linters have been one of the main ways I've leveled up in other languages. In addition, I find it's always useful to have some automated enforcement of style guide stuff, such as capitalization, indenting, and so on; those are distracting during code reviews, and robots can find - and fix - such problems better than humans.

However, there are a few drawbacks to using SQLFluff. I think they're worthwhile tradeoffs, but I wanted to surface them up front:
- we'll need to add placeholder values for each `:parameter` in the SQLFluff configuration; there are a few demonstrated in the initial `.sqlfluff` configuration file, but we'd have to add the rest, and continue to add more as we add queries
- SQLFluff is written in Python; it's unfortunate to add a dependency from another language
- not all the syntax we're using is supported; for example, it is not yet able to parse SQL2011 `OFFSET / FETCH`, so we'd have to go back to the PostgreSQL-flavored `LIMIT / OFFSET`

You can play with this by following the [getting started instructions](https://docs.sqlfluff.com/en/stable/gettingstarted.html#gettingstartedref). I have the start of a configuration file included in this PR, so check out the branch and try running it: `sqlfluff lint src/database/queries/proposals/selectWithPagination.sql` or so.

If this feels like a worthwhile addition, here's what still needs to be done to be ready to merge:
- [x] write dev docs on how to install & run
- [x] run as part of CI
- [x] make sure we have it configured the way we want
  - [x] add placeholder values for all the template parameters
  - [ ] argue about [rules](https://docs.sqlfluff.com/en/stable/rules.html)
- [x] fix existing violations
- [x] ignore violations in existing migrations, since those can't be changed


bickelj:
@jasonaowen The following comments are prior to trying SQLFluff.

I am hesitant to introduce a Python dependency in a node project, but I am open to it for several reasons:
1. It appears to be a mature, active, widely-used tool.
2. The dependency is used at development and build time, not in a test or production environment.
3. I'm OK with a little Python wrangling and I expect others on the team are too, assuming there is no good substitute in TypeScript/JavaScript.

My other hesitation is the net value to the project. Is SQLFluff going to make a big difference? I know it will improve the SQL scripts aesthetically, probably will prevent a couple bugs, but considering the cost of bringing in this external dependency, my thought prior to actually running SQLFluff is "it is probably not worth the hassle."

In general, though, static analysis is great. So I am not really against it. My next step is to try SQLFluff locally and see what it does.

bickelj:
Already I have a bad User Experience when following the https://docs.sqlfluff.com/en/stable/gettingstarted.html#gettingstartedref verbatim (I have Python 3.9.2 and pip 20.3.4 installed).

First, this seemed to work OK:
```
$ pip install sqlfluff
Collecting sqlfluff
  Downloading sqlfluff-2.3.0-py3-none-any.whl (701 kB)
     |████████████████████████████████| 701 kB 1.4 MB/s 
Collecting tblib
  Downloading tblib-2.0.0-py3-none-any.whl (11 kB)
Requirement already satisfied: Jinja2 in /usr/lib/python3/dist-packages (from sqlfluff) (2.11.3)
Collecting regex
  Downloading regex-2023.8.8-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (771 kB)
     |████████████████████████████████| 771 kB 75 kB/s 
Requirement already satisfied: typing-extensions in /home/user/.local/lib/python3.9/site-packages (from sqlfluff) (4.7.1)
Collecting click
  Downloading click-8.1.7-py3-none-any.whl (97 kB)
     |████████████████████████████████| 97 kB 252 kB/s 
Requirement already satisfied: pyyaml>=5.1 in /usr/lib/python3/dist-packages (from sqlfluff) (5.3.1)
Requirement already satisfied: tqdm in /home/user/.local/lib/python3.9/site-packages (from sqlfluff) (4.65.0)
Requirement already satisfied: pytest in /usr/lib/python3/dist-packages (from sqlfluff) (6.0.2)
Collecting pathspec
  Downloading pathspec-0.11.2-py3-none-any.whl (29 kB)
Requirement already satisfied: toml in /usr/lib/python3/dist-packages (from sqlfluff) (0.10.1)
Requirement already satisfied: appdirs in /usr/lib/python3/dist-packages (from sqlfluff) (1.4.4)
Requirement already satisfied: chardet in /usr/lib/python3/dist-packages (from sqlfluff) (4.0.0)
Collecting colorama>=0.3
  Downloading colorama-0.4.6-py2.py3-none-any.whl (25 kB)
Collecting diff-cover>=2.5.0
  Downloading diff_cover-7.7.0-py3-none-any.whl (51 kB)
     |████████████████████████████████| 51 kB 1.8 MB/s 
Collecting Pygments<3.0.0,>=2.9.0
  Downloading Pygments-2.16.1-py3-none-any.whl (1.2 MB)
     |████████████████████████████████| 1.2 MB 283 kB/s 
Collecting pluggy<2,>=0.13.1
  Using cached pluggy-1.2.0-py3-none-any.whl (17 kB)
Installing collected packages: Pygments, pluggy, tblib, regex, pathspec, diff-cover, colorama, click, sqlfluff
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pyhanko 0.19.0 requires qrcode>=6.1, which is not installed.
Successfully installed Pygments-2.16.1 click-8.1.7 colorama-0.4.6 diff-cover-7.7.0 pathspec-0.11.2 pluggy-1.2.0 regex-2023.8.8 sqlfluff-2.3.0 tblib-2.0.0
```

I see the ERROR but `pip` also said it successfully installed `sqlfluff` so I tried it:
```
$ sqlfluff version
Traceback (most recent call last):
  File "/home/user/.local/bin/sqlfluff", line 5, in <module>
    from sqlfluff.cli.commands import cli
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/__init__.py", line 6, in <module>
    from sqlfluff.api import lint, fix, parse, list_rules, list_dialects
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/api/__init__.py", line 4, in <module>
    from sqlfluff.api.simple import lint, fix, parse, APIParsingError
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/api/simple.py", line 4, in <module>
    from sqlfluff.core import (
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/core/__init__.py", line 9, in <module>
    from sqlfluff.core.linter import Linter
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/core/linter/__init__.py", line 3, in <module>
    from sqlfluff.core.linter.common import (
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/core/linter/common.py", line 13, in <module>
    from sqlfluff.core.templaters import TemplatedFile
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/core/templaters/__init__.py", line 8, in <module>
    from sqlfluff.core.templaters.jinja import JinjaTemplater
  File "/home/user/.local/lib/python3.9/site-packages/sqlfluff/core/templaters/jinja.py", line 10, in <module>
    import jinja2.nodes
  File "/usr/lib/python3/dist-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/usr/lib/python3/dist-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/usr/lib/python3/dist-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/home/user/.local/lib/python3.9/site-packages/markupsafe/__init__.py)
```

I should be able resolve this in a few minutes, but the "out of the box" experience, especially for a non-Python developer, might leave something to be desired.

jasonaowen:
Thanks for reviewing, @bickelj!

> Is SQLFluff going to make a big difference?

I believe the answer is yes. Any time we can avoid arguing about lintable errors in pull requests is worthwhile; we've already spent some amount of time on that, which is what prompted this PR.

> I am also curious if any real bugs have been found with SQLFluff.

I believe the primary purpose of linters is not to find or prevent bugs directly, but to make bugs more easily found. Inconsistent formatting is a distraction, and standardizing the way we write code - any code - makes it easier for a human reviewer to notice problems. Anything we can do to help our pattern-matching brains better see both patterns and exceptions is a force multiplier.

We've since agreed to use a consensus process (which I still need to write up). Under that framework, I'll interpret your comments as having no blocking concerns, @bickelj.

I'll put getting this into shape on my agenda, although it might be a bit before it's ready, as it's not (yet) the highest priority. I do want to get it in before we start adding any major new SQL code!

bickelj:
@jasonaowen Yes, my interpretation is we agreed to "aim for consensus", and I am willing to use SQLFluff if someone wants to add it.

slifty:
Just to add my voice: I really dislike adding a python dependency to this project -- and the upkeep tasks / more complex dev setup that demands -- but not enough to block.

jasonaowen:
We discussed this further today, and @slifty suggested the possibility of using SQLFluff via docker rather than a locally managed Python installation.

README.md
```diff
@@ -182,6 +182,16 @@ For example: `0001-create-users` or `0001-modify-users`
 
 In `/src/databases/seeds` there is seed or starter data. The contents can be run manually to help developers get data in their databases. The scripts are not referenced by the software and are included for convenience. The migrations must run prior to using seed scripts.
 
+We use [SQLFluff](https://docs.sqlfluff.com/en/stable/index.html)
+to lint our SQL files. See
+[Getting Started with SQLFluff](https://docs.sqlfluff.com/en/stable/gettingstarted.html)
+for instructions on setting it up.
```
jasonaowen, line 188: I kind of punted on dev documentation; I know that some people like to use Docker, and some do not, and either way I'm not sure that we should proscribe a way for devs to do their work.

That said, in case it is convenient, the way I am using SQLFluff is with [`venv`](https://docs.python.org/3/library/venv.html) to avoid changing anything outside of this directory:

```sh
python3 -m venv venv
. venv/bin/activate
pip install sqlfluff
npm run lint:sqlfluff
```

jasonaowen (commented):
I think this is ready to review, @slifty @bickelj!

There are three pieces to this PR, and I'm very happy to split it along any lines that would make it easier to review:
- switch syntax to something SQLFluff can parse
- configure SQLFluff and add it to CI
- fix rule violations

If any of the rules or rule fix commits are at all controversial, we should split them out into a new PR after this - partial coverage is much better than no coverage.

.sqlfluff
```diff
@@ -0,0 +1,22 @@
+[sqlfluff]
+dialect = postgres
+templater = placeholder
+exclude_rules =
+  layout.indent,
```
jasonaowen, line 5: Indentation is more complicated, more controversial, and requires a bigger change than the others in this pull request, so I'm holding it for another PR.

README.md
slifty: Does this mean that every time I develop for this project I'd have to active venv (assuming I run lint as part of my pre-push process).  If so: I don't love that as opposed to being able to continue to just type `npm run lint` and have it just work.

I think it would be nice to be more opinionated in the name of dev convenience (specifically lowering the amount of recurring CLI setup that would need to be done). I think it's reasonable to have a `python3` requirement as part of the dev setup instructions since we have introduced it as a language requirement.

If `python3` is a documented dev requirement, could we have the npm script run `venv` as part of the `lint:sqlfluff` command, and have npm run `pip install` as part of an `npm install` prehook (and a pipfile that locks a version of sqlfluff)

Alternatively, if we think docker is a cleaner dev dependency than python3 we could update `lint:sqlfluff` to use docker.

package.json
```diff
@@ -13,6 +13,8 @@
 		"lint:swagger": "swagger-spec-validator src/openapi.json",
 		"lint:tsc": "tsc --noEmit -p tsconfig.dev.json",
 		"lint:prettier": "prettier . --check",
+		"lint:sqlfluff": "sqlfluff lint src/database/",
+		"lint:sqlfluff:fix": "sqlfluff fix src/database/",
```
slifty, line 17: I'd expect this to be pat of `format` rather than `lint`

So that would mean we have `format` split into `format:prettier` and `format:sqlfluff` and `format` itself would invoke both.

README.md
```diff
@@ -182,6 +182,16 @@ For example: `0001-create-users` or `0001-modify-users`
...
+
+Once you have SQLFluff installed,
```
slifty, line 190: We don't generally use this indentation style in our README -- let's keep with the current format of just having paragraphs as paragraphs in this project, if that's OK.

FWIW I do find myself reading readme files in raw form often enough that this format is notably harder for me to parse, so I also just don't really like overriding the end user's ability to just have their context determine the most appropriate wrap based on their accessibility needs (e.g. font size, screen size, etc).

.sqlfluffignore
```diff
@@ -0,0 +1,49 @@
+dist/
+node_modules/
+
+# Migrations are immutable, so already-committed migrations cannot be corrected.
+# https://github.com/zakpatterson/postgres-schema-migrations/blob/75f22a65d3a8ddd16356e74538316fec2ca0202e/README.md#hash-checks-for-previous-migrations
+src/database/migrations/0001-create-canonical_fields.sql
```
slifty, line 6: I was going to say "this approach adds a chore when we write a migration" but then realized "oh, no it doesn't! We don't *want* to ignore new migrations"

w00t!

.sqlfluffignore
```diff
@@ -0,0 +1,49 @@
...
+src/database/migrations/0034-alter-bulk_uploads-add-source_id.sql
+src/database/migrations/0035-alter-proposal_versions-add-created_by.sql
+src/database/migrations/0036-create-function-drop_function.sql
+src/database/migrations/0037-alter-users-change-authentication_id-to-keycloak_user_id.sql
+src/database/migrations/0038-alter-users-use-keycloak_user_id-as-primary-key.sql
+src/database/migrations/0039-rename-organizations-to-changemakers.sql
+src/database/migrations/0040-create-permission-tables.sql
+src/database/migrations/0041-rename-bulk_uploads-to-bulk_upload_tasks.sql
+
+# This file is very large, and is intended to be removed soon.
```
slifty, line 48: `soon` should be replaced with `Soon™`

(jk)

slifty (changes requested):
Wooo so much cleaner!

Asking that the fix be moved to `format` and that we strongly consider making `python3` a formal dependency in order to allow our npm scripts to more seamlessly support sql fluffing.

.sqlfluffignore
jasonaowen: Exactly!

The only exception is if we later change our rules, we'll have to ignore the migrations that passed the old rules. Hopefully that will be sufficiently rare to not need to worry about!

README.md
jasonaowen: I'm happy to look into that, but can that be done as part of a later PR? I'd like to get this merged sooner rather than later so that it'll start running in CI.

slifty: @jasonaowen that's absolutely fine by me, good point!

jasonaowen:
Rebased onto `main` and addressed the requested changes, except for the dev documentation, which I plan to follow up on in a subsequent PR.

slifty:
@jasonaowen Is there a way we could pin a version of sqlfluff (and make it dependabot tracked?) in this PR?

jasonaowen:
> @jasonaowen Is there a way we could pin a version of sqlfluff (and make it dependabot tracked?) in this PR?

Excellent point! I added a `requirements.txt` file, which I believe dependabot can update.

slifty (approved):
Yay!  I look forward to linting my open PR with this

bickelj (commented):
Weird. I see [the issue](https://github.com/sqlfluff/sqlfluff/issues/3725) but no activity on it. This is another strike against SQL fluff. I think that only makes two strikes, though, so we're OK : )

PR #334, opened 2023-05-03
Rename canonical fields to base fields

### LINKED ISSUE #275: Thinking through "Canonical Field" vs "Field"
We should think about whether `Canonical` means more blessing than we actually want to convey.

(this is an issue stub, to be filled out soon)

slifty:
This PR renames "canonical fields" to "base fields" throughout our API.

This is a breaking change, so I'm updating the version as well.

Resolves #275


src/__tests__/baseFields.int.test.ts
```diff
@@ -59,28 +58,13 @@ describe('/canonicalFields', () => {
         );
     });
 
-    it('should error if the database returns an unexpected data structure', async () => {
```
slifty, line 62: As we have talked about, I took out the implementation validation since we don't want to validate the shape of database returns.

That said, we are now no longer safely protecting against an issue where our SQL has not been properly formatted to select the correct shape of data.  This means we *could* introduce a bug that violates our documented API schema and our automated tests are no longer capturing that possibility.

bickelj (approved):
Bravo!

bickelj:
Locally I could not find any remaining inappropriate use of the word "canonical".

src/database/seeds/0001-insert-canonical_fields.sql
jasonaowen, line 1: Should we rename this file?

bickelj (commented):
I keep getting an error on a particular test but I need to look further:
```
FAIL  src/__tests__/proposals.int.test.ts (11.006 s)
  ● /proposals › GET / › returns a subset of proposals present in the database when search is provided

    expect(received).toEqual(expected) // deep equality

    - Expected  - 36
    + Received  +  1
```

jasonaowen (approved):
This looks good to me! Thanks for knocking this out, @slifty.

We'll need to coordinate a data-viewer release, as this is a breaking API change. We're also racing against a demo clock right now, and this might not be the most critical thing for me to work on - how would you feel if we held this for next week?

src/database/seeds/0001-insert-canonical_fields.sql
slifty: Oh this is a seed not a migration -- yes, I think we should!

slifty:
Very happy to not merge this until next week

jasonaowen (commented):
I found myself needing to revisit this PR to switch branches, so two more thoughts!

jasonaowen:
The front end PR (https://github.com/PhilanthropyDataCommons/data-viewer/pull/113) is ready to go; it just needs to be deployed at the same time as this.

Issue #336, opened 2023-05-05
Test failure on different pg text search settings


bickelj:
Locally, I got this test failure every time on main after the merge of #320:
```
 FAIL  src/__tests__/proposals.int.test.ts (12.825 s)
  ● /proposals › GET / › returns a subset of proposals present in the database when search is provided

    expect(received).toEqual(expected) // deep equality

    - Expected  - 36
    + Received  +  1

      Object {
    -   "entries": Array [
    -     Object {
    -       "applicantId": 1,
    -       "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/,
    -       "externalId": "proposal-1",
    -       "id": 1,
    -       "opportunityId": 1,
    -       "versions": Array [
    -         Object {
    -           "applicationFormId": 1,
    -           "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/,
    -           "fieldValues": Array [
    -             Object {
    -               "applicationFormField": Object {
    -                 "applicationFormId": 1,
    -                 "baseFieldId": 1,
    -                 "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/,
    -                 "id": 1,
    -                 "label": "Short summary",
    -                 "position": 1,
    -               },
    -               "applicationFormFieldId": 1,
    -               "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/,
    -               "id": 1,
    -               "position": 1,
    -               "proposalVersionId": 1,
    -               "value": "This is a summary",
    -             },
    -           ],
    -           "id": 1,
    -           "proposalId": 1,
    -           "version": 1,
    -         },
    -       ],
    -     },
    -   ],
    +   "entries": Array [],
        "total": 2,
      }

      181 |         .set(authHeader)
      182 |         .expect(200)
    > 183 |         .expect(
          |          ^
      184 |           (res) => expect(res.body).toEqual(
      185 |             {
      186 |               total: 2,

      at Object.<anonymous> (src/__tests__/proposals.int.test.ts:183:10)
      ----
      at src/__tests__/proposals.int.test.ts:184:37
      at node_modules/supertest/lib/test.js:308:13
[...]

PR #337, opened 2023-05-05
Specify the PG TS config `'english'` on search

bickelj:
While the postgres (PG) text search (TS) config 'english' is specified when building the text search column and index in the migration script `0010-alter-proposal_field_values-value_search.sql`, when the postgres `default_text_search_config` is set to `pg_catalog.simple` rather than to `pg_catalog.english`, the test named `returns a subset of proposals present in the database when search is provided` in file `proposals.int.test.ts` will fail. For consistency in search results across instances, this commit sets the query with search to use the same `english`. Without this commit, tests will fail with different PG settings for default text search config and inconsistent search results will appear between instances.

Issue #336 Test failure on different pg text search settings

### DIFF
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -175,7 +175,6 @@ describe('/proposals', () => {
         position: 1,
         value: 'This is a pair of pants',
       });
-
       await agent
         .get('/proposals?_content=summary')
         .set(authHeader)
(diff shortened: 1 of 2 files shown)


slifty (approved):
great catch!  (and hooray for tests!)

Out of curiosity... would it be possible to easily write an integration test that ensures this works when `pg_catalog.simple` is the configured setting?  It would be worth including a regression test if so.  If not, that's fine too.

bickelj:
It was pretty easy to add a single statement to the existing test, setting the default config to 'arabic', which then causes the test to fail if 'english' is not specified in the select query. Does this look OK?

bickelj:
Another thought I had was to use the exact bitnami image in our test that we use in prod. We don't have a non-manual way to synchronize that, however, and there is utility in testing with a recent postgres image. If I had to choose one or the other I'd test using the exact image we use in prod, but perhaps we should do both. What do you think, @slifty @jasonaowen ?

src/__tests__/proposals.int.test.ts
```diff
@@ -124,6 +124,8 @@ describe('/proposals', () => {
     });
 
     it('returns a subset of proposals present in the database when search is provided', async () => {
+      // This should pass even if the default text search config is not english.
+      await db.query('set default_text_search_config = \'arabic\';');
```
jasonaowen, line 128: > It was pretty easy to add a single statement to the existing test, setting the default config to 'arabic', which then causes the test to fail if 'english' is not specified in the select query. Does this look OK?

I'd prefer to have regression tests split out into separate tests - setting the default text search config isn't really a part of this basic search test, and tests shouldn't be doing multiple things. In this case, there should be a test that search works at all, and a test that search works with a config other than `english` set.

I'd also prefer that the regression test use the `simple` config rather than `arabic`, since that's where we stumbled. Pair that with a search term that is stemmed in the `english` config (as the search term used in this test, `'summary'`, happens to be) but not in `simple` (since `simple` doesn't do stemming) and test that there is one search result.

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -9,7 +9,7 @@ FROM proposals p
 WHERE
   CASE
     WHEN :search::text != '' THEN
-      pfv.value_search @@ websearch_to_tsquery(:search::text)
+      pfv.value_search @@ websearch_to_tsquery('english', :search::text)
```
jasonaowen, line 12: So glad this was such a simple fix!

jasonaowen (approved):
Good digging, @bickelj! Thanks for fixing this.

jasonaowen:
> Another thought I had was to use the exact bitnami image in our test that we use in prod. We don't have a non-manual way to synchronize that, however, and there is utility in testing with a recent postgres image. If I had to choose one or the other I'd test using the exact image we use in prod, but perhaps we should do both. What do you think, @slifty @jasonaowen ?

Well, I have a few thoughts, in no particular order:

- we should have alignment between dev and prod, but maybe we should move in the other direction - can we stop using the bitnami image altogether? It's already varied from our expectations in a way that impacted functionality; will it do so again?
- we could run the tests against multiple versions with a [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)
- rather than running a docker image, or running our own PostgreSQL at all, maybe we should be using Heroku or Digital Ocean; their tooling is better than anything we can build with the budget we have, and our engineering time is better spent on the application itself rather than on hosting

src/__tests__/proposals.int.test.ts
bickelj: I separated out the test that uses `'simple'`. My reasoning behind arabic was it would be more likely to trigger the failure in the future, but I think you're right about the regression test covering exactly what we experienced.

src/database/queries/proposals/selectWithPagination.sql
bickelj: It's not yet fixed, it has to get merged and deployed...

bickelj:
I rebased on main and implemented the suggestion of splitting out the regression test from the original test.

Jason said:
> * we should have alignment between dev and prod, but maybe we should move in the other direction - can we stop using the bitnami image altogether? It's already varied from our expectations in a way that impacted functionality; will it do so again?
> 
> * we could run the tests against multiple versions with a [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)
> 
> * rather than running a docker image, or running our own PostgreSQL at all, maybe we should be using Heroku or Digital Ocean; their tooling is better than anything we can build with the budget we have, and our engineering time is better spent on the application itself rather than on hosting

I like the matrix strategy. Multiple versions would be good, e.g. "what we use in prod today" as well as "a latest version", and/or a "stock pg image" and a "bitnami pg image".

I looked for the download button for Heroku and couldn't figure out how to download a Heroku :smiley:

Digital Ocean's default configuration also sets `simple`:
```quote
defaultdb=> select name, setting from pg_settings where name like '%default_text%';
            name            |      setting      
----------------------------+-------------------
 default_text_search_config | pg_catalog.simple
(1 row)
```

Therefore we likely would have run into the same issue with Digital Ocean as the bitnami image. I don't know Heroku's default setting.

Regardless, I think we agree that we should align the tests one way or another. Because we already use containers in test and prod and because GitHub already uses containers during workflows I still think the best path is to use that image during the workflow.

jasonaowen (approved):
Thanks for splitting that test out!

Issue #323, opened 2023-05-02
Implement GET /proposals by opportunity ID


bickelj:
This was requested by a user of the API.

bickelj:
Another user suggested this same thing again. That makes (at least) three of us who think it is useful.

slifty:
This will be very quick to implement, and is in line with other "related entity" filters

PR #349, opened 2023-05-11
Implement `GET /proposals` by opportunity ID

bickelj:
When posting new proposal data via the API, one gets existing proposal data before attempting to add a new one. Then one must filter through a lot of proposal data locally. This change allows filtering by opportunity ID on the server-side. This reduces the amount of client data filtering required and reduces overall data transfer.

Issue #323  Implement GET /proposals by opportunity ID

### DIFF
--- a/src/errors/SimpleInputValidationError.ts
+++ b/src/errors/SimpleInputValidationError.ts
@@ -1,0 +1,9 @@
+// In the case of non-ajv input validation, use this for input errors.
+export class SimpleInputValidationError extends Error {
+  public constructor(
+    message: string,
+  ) {
+    super(message);
+    this.name = this.constructor.name;
+  }
+}
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -38,6 +38,16 @@
         "name": "_content",
         "required": false,
         "description": "A web-search formatted string to select specific entities returned."
+      },
+      "opportunityIdParam": {
+        "schema": {
+          "type": "integer",
+          "default": 0
+        },
+        "in": "query",
+        "name": "opportunityId",
+        "required": false,
+        "description": "The PDC-generated ID of an opportunity by which to filter results."
       }
     },
     "securitySchemes": {
@@ -719,11 +729,12 @@
         "parameters": [
           { "$ref": "#/components/parameters/pageParam" },
           { "$ref": "#/components/parameters/countParam" },
-          { "$ref": "#/components/parameters/searchParam" }
+          { "$ref": "#/components/parameters/searchParam" },
+          { "$ref": "#/components/parameters/opportunityIdParam" }
         ],
         "responses": {
           "200": {
-            "description": "All proposals currently registered in the PDC.",
+            "description": "Proposals currently registered in the PDC matching the given criteria.",
             "content": {
               "application/json": {
                 "schema": {
--- a/src/queryParameters/extractOpportunityIdParameter.ts
+++ b/src/queryParameters/extractOpportunityIdParameter.ts
@@ -1,0 +1,19 @@
+import apiSpecification from '../openapi.json';
+import { SimpleInputValidationError } from '../errors';
+import type { Request } from 'express';
+
+export const extractOpportunityIdParameter = (
+  request: Request,
+) => {
+  if (request.query.opportunityId !== undefined
+    && request.query.opportunityId !== '') {
+    const id = Number(request.query.opportunityId);
+    if (!Number.isSafeInteger(id)) {
+      throw new SimpleInputValidationError(`opportunityId must be an integer, got '${id}'.`);
+    }
+    return { opportunityId: id };
+  }
+  return {
+    opportunityId: apiSpecification.components.parameters.opportunityIdParam.schema.default,
+  };
+};
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -395,6 +395,81 @@ describe('/proposals', () => {
         );
     });
 
+    it('returns a subset of proposals for a given opportunityId', async () => {
+      await db.sql('opportunities.insertOne', {
+        title: 'First opportunity',
+      });
+      await db.sql('opportunities.insertOne', {
+        title: 'Second opportunity',
+      });
+      await db.sql('opportunities.insertOne', {
+        title: 'Third opportunity',
+      });
+      await db.sql('applicants.insertOne', {
+        externalId: '5009',
+        optedIn: true,
+      });
+      await db.sql('proposals.insertOne', {
+        applicantId: 1,
+        externalId: 'proposal-5011',
+        opportunityId: 3,
+      });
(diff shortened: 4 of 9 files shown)


src/queryParameters/extractOpportunityIdParameter.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { Request } from 'express';
+
+export const extractOpportunityIdParameter = (
+  request: Request,
+) => {
+  if (request.query.opportunityId !== undefined
+    && request.query.opportunityId !== '') {
+    const id = Number(request.query.opportunityId);
+    if (!Number.isSafeInteger(id)) {
+      throw new SimpleInputValidationError(`opportunityId must be an integer, got '${id}'.`);
```
jasonaowen, line 12: In this domain, I think we should be narrower: we want a positive integer, as zero or a negative number will never be a valid opportunity ID. (Right?)

To that end, I think we might want to continue to use AJV here, rather than introduce a new `SimpleInputValidationError` type, with a schema like `{ type: 'integer', minimum: 1 }`.

src/openapi.json
```diff
@@ -38,6 +38,16 @@
         "name": "_content",
         "required": false,
         "description": "A web-search formatted string to select specific entities returned."
+      },
+      "opportunityIdParam": {
+        "schema": {
+          "type": "integer",
+          "default": 0
```
jasonaowen, line 45: Having a default of `0` feels very strange to me. I recognize that this is extending the pattern for the other query parameters on this endpoint, but it feels different.

One way I look at default values is by rewriting the call with all the values specified. That'd transform `/proposals` to `/proposals?_page=1&_count=100&_content=&opportunityId=0`.

The page and count parameters must have some kind of default, as they are required for processing a request, and the alternative is to make them non-optional.

The search parameter is optional, and its (non-empty) presence significantly affects the results; making it default to an empty string is a little bit weird, but it makes some sense: `?_content=` is clearly missing a value.

Perhaps the reason setting the default to `0` here is strange is because `0` is conceivably in the domain of opportunity IDs. Opportunity IDs are integers, 0 is an integer, therefore 0 could be an opportunity ID. This is incorrect, of course, but I feel like setting this default value - and communicating it in the documentation - encourages this confusion.

As an alternative, can we not set a default value, properly detect the parameter's presence or absence, and handle it appropriately?

jasonaowen (commented):
This seems like it could be a useful change! I haven't spent as much time thinking about the proposal creation process as you, so I'd appreciate some help understanding the context.

Can you talk more about the workflow where this is used, and the alternatives to this approach you've considered? It seems like you'd still need to filter the list client-side to find the specific proposal(s?) you want to add data to. How do you do that?

src/queryParameters/extractOpportunityIdParameter.ts
bickelj: Fair enough, yes.

src/openapi.json
bickelj: Yes, it is strange. I followed the established pattern but agree it feels different for an integer. The root of the default I think comes from wanting to have a single database query for `GET /proposals` with potentially many parameters. But as you mentioned, perhaps we can default somewhere between the client API call and the database code.

bickelj:
> This seems like it could be a useful change! I haven't spent as much time thinking about the proposal creation process as you, so I'd appreciate some help understanding the context.

One version of the process can be like [this](https://github.com/PhilanthropyDataCommons/data-scripts/blob/6abcd5cef16de25eb739d59ee86d0f16dc89845f/src/generateProposalVersionJson.ts#L166) where you have to call `GET /proposals` for every new proposal data row in a CSV. It's a lot of data if you're only interested in the proposals related to a given opportunity.

> Can you talk more about the workflow where this is used, and the alternatives to this approach you've considered? It seems like you'd still need to filter the list client-side to find the specific proposal(s?) you want to add data to. How do you do that?

Yes, there may still need to be filtering, but much less data transfer and less data filtering. Considering this is only the first of multiple parameters by which to filter proposals (the others being external ID, application form ID, applicant ID, and possibly time frame), it is useful on its own but will be more useful when combined with other parameters. Suppose you could ask for `GET /proposals?opportunityId=5&externalId=myProposal&applicantId=500` then you are going to filter it all the way down to exactly one if it exists (which, by the way, would still be a list of one element despite the unique constraint).

It was much less data overall (not fewer items to filter, though) when the default was to return the immediate proposals without all the accompanying proposal data.

src/queryParameters/extractOpportunityIdParameter.ts
slifty: We have a pattern for this kind of check which uses ajv in the pagination parameter tooling

src/errors/SimpleInputValidationError.ts
```diff
@@ -0,0 +1,9 @@
+// In the case of non-ajv input validation, use this for input errors.
+export class SimpleInputValidationError extends Error {
```
slifty, line 2: I mentioned this elsewhere but I'm not sure we should have non-ajv input validation.

We don't have to define a full blown type -- we can just use `ajv.validate({ type: 'integer'}, value), or whatever!

src/openapi.json
slifty: Could we just specify that the default is null?

bickelj: Having a default specified for search parameters also feels weird to me, yet [that is what happens](https://github.com/PhilanthropyDataCommons/service/blame/59e39aaae30d1dfd93261dfa11795ea155cc7d58/src/queryParameters/extractSearchParameters.ts#L10) and was approved in #320. Is that worth revisiting too?

jasonaowen: I'd be amenable to revisiting `_content`, but as I said it feels different. Are you offering to tackle that as part of this PR, @bickelj?

bickelj: Probably not but maybe it will be simple enough to include. I will revisit when refactoring the one under discussion here.

bickelj:
I don't know if this is still relevant, it depends on the uses of the `/proposals` endpoint. If there is no need for data scripts that use the API to send data to the PDC or if it's all handled in the `front-end` now or all handled in the bulk uploads flow, perhaps this is moot.

slifty:
@bickelj I do think that it might be ok to close this since opportunity doesn't have a huge number of use cases around it right now -- at the very least, we now have some new patterns around how to extract / validate / etc this kind of filter so it would probably be prudent to rewrite it with those patterns in mind.

bickelj:
@slifty I'll close this PR but keep the original issue #323 open.

bickelj:
@slifty I think three separate folks trying to post proposal data had use cases for it, looking at the original issue #323. If there is a different modeling of proposals or bulk upload resolves it in another way, I suppose we can close that original one.

slifty:
@bickelj makes sense!  I think it's fine to keep the issue open!

PR #355, opened 2023-05-15
Tell NPM about Node 18

slifty:
This PR defines the node engine in a way that npm, in theory, will honor when running updates.  My understanding is that this is what dependabot relies on when determining what PRs to open, and this in turn should resolve the issue where dependabot is proposing packages that aren't meant for Node 18.


slifty:
NOTE: I don't know for sure this will absolutely fix dependabot's attempt to open incompatible PRs, but my understanding is that it relies on npm to tell it what to do.

slifty:
This is context (I can add this to the comment) about why I added engine-strict: https://github.com/dependabot/dependabot-core/issues/5489

slifty:
Correct -- a byproduct of this is that we are requiring the use of node 18 / not allowing any other version (that is what `engineStrict` accomplishes).

Since this isn't a package, I think it's OK to require the expected version of node.

bickelj:
One benefit, I suppose, is that this signals to developers exactly which version to be on to avoid any potential conflict. It could also remind us to upgrade when a new node LTS version is out ("I checked out the current LTS version and it didn't work").

> Since this isn't a package, I think it's OK to require the expected version of node. Otherwise the type definitions are wrong (for the same reasons we don't want to use the node 20 types in a node 18 environment).

I would expect using `@types/node@18.6.9` to work with a node 20 runtime but not the other way around. But I have not tried it nor inspected the backward compatibility policies of node. For example, in Java-land, it is OK to target a Java 11 compilation level but you can still develop in Java 17 or run in Java 17. The other way around would not work, though.

In any case, I would be OK with the change because it makes clear what is expected and avoids any potential conflicts in addition to making dependabot work as expected.

slifty:
Makes sense!

As a point of information, I don't know if / when this would manifest in *type* incompatibilities but there are at least removals and deprecations between major versions which in theory could result in unexpected behavior if we were using things that changed (for example 20 doesn't have many, but you can see the section at least: https://nodejs.org/en/blog/release/v20.0.0#deprecations-and-removals)

bickelj (approved):
I'm still curious what @jasonaowen thinks but this seems good to me.

jasonaowen:
I think this makes sense. The only thing I'm sad to lose with this approach is something we're not doing, which is running the tests against the current version of Node.js in addition to the version running in production.

An alternative, for consideration, is to tell Dependabot not to open PRs for major version upgrades via the `version-update` key of the [ignore dependencies configuration](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore), either generally for all dependencies or specifically for the `@types/node` dependency. I still don't understand why it's opening major version PRs, to be honest; I find that surprising, and it makes me think there's something I'm missing.

slifty:
@jasonaowen re the alternative, making sure you saw my note in the PR:

```
An alternative approach I considered here was to directly specify a constraint on dependabot for the @types/node package. That felt like a worse version of this approach because (1) it would only prevent conflicts for that single package in a way that (2) would silently fail to upgrade versions in future if we forgot to update the rule.

By defining the engine in this way I believe when we ultimately would get errors if we only updated, for instance, .node-version since npm would complain that the version didn't match the one defined in package.json
```

I'd be most comfortable with a rule telling it to avoid major version PRs in general -- I suspect a types/node specific rule would be more likely to go forgotten (similar to how we currently have the tsconfig/16 instead of tsconfig/18)

jasonaowen:
> re the alternative, making sure you saw my note in the PR: [...]

@slifty ah sorry, it turns out I didn't originally understand what you meant by that! Thanks for clarifying.

> I suspect a types/node specific rule would be more likely to go forgotten (similar to how we currently have the tsconfig/16 instead of tsconfig/18)

I *think* we could configure Dependabot to have a special rule just for `@types/node` not to consider major version upgrades, without specifically saying "nothing newer than `^18`".

> I'd be most comfortable with a rule telling it to avoid major version PRs in general

After thinking about this more, and looking at the [previous dependabot PRs for this repo](https://github.com/PhilanthropyDataCommons/service/pulls?q=is%3Apr+author%3Aapp%2Fdependabot), I'm split. On the one hand, it is very convenient to have dependabot open PRs for major version upgrades, as it links to the changelog, runs CI, creates a branch for us to poke at, and reminds us to keep up to date on our dependencies. On the other hand, it encourages us to be cavalier about major version upgrades, while semver exists for a reason.

So, I guess we'll wait and see if this solves the immediate problem, and we'll cross the Node.js 20 or later bridge when we come to it.

bickelj:
Today an API user mentioned that pulling the code and building used to work but now doesn't, and I suspect it's because of this change and using node 20.

Given that this didn't end up satisfying dependabot, which was the impetus, can we go in either of these directions safely?
1. Undo the strict engines change, or
2. simply update to node 20 and change our declaration in https://github.com/PhilanthropyDataCommons/service#node-version to say "Active LTS or a soon-to-be-LTS (even numbered) release"?

bickelj:
Additional context: https://github.com/nodejs/release#release-schedule

It appears that there will be a few days without any "Active LTS" release. 18 will be the "maintenance" release as of 2023-10-18 but then 20 won't become "Active LTS" until 2023-10-24, about a week later.

slifty:
@bickelj would you be up for opening an issue (since this is a closed PR I'm not sure it's the most discoverable place to have the larger conversation, especially since it may culminate in an action item and broader decision).

I do think we should have a discussion about what versions of node we are utilizing as a part of the project, including the benefits and cons of each (asking API users to use our specified version of node seems like a reasonable thing to do, if that's the path we want to take -- e.g. an API user using version 16 of node might also face issues.)

PR #356, opened 2023-05-15
Use the TSConfig for Node.js 18

jasonaowen:
The TSConfig project hosts recommendations for configuring TypeScript. We were using the recommended configuration for Node.js 16, but we are targetting Node.js 18. Use the recommended configuration for our environment.

Found while reviewing #355.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "tinypg": "^7.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


slifty (approved):

jasonaowen:
Rebased onto `main` after #355 was merged.

bickelj:
Nice catch.

PR #358, opened 2023-05-15
Log 5XX errors appropriately

### LINKED ISSUE #350: Error logs do not contain useful information
When something is wrong with the service, error logs are the first place we look. Unfortunately, I've been finding that while they log the fact of the error - which is useful - they do not tend to include the cause of the error.

<details>
<summary>Example log message</summary>

```
[2023-05-11T19:54:52.375Z] INFO (1292163): request errored
    source: "src/app.ts"
    req: {
      "id": 7,
      "method": "GET",
      "url": "/baseFields",
      "query": {},
      "params": {},
      "headers": {
        "host": "localhost:3000",
        "user-agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/112.0",
        "accept": "application/json",
        "accept-language": "en-US,en;q=0.5",
        "accept-encoding": "gzip, deflate, br",
        "referer": "http://localhost:3001/",
        "authorization": "Bearer [omitted for brevity]",
        "origin": "http://localhost:3001",
        "dnt": "1",
        "connection": "keep-alive",
        "sec-fetch-dest": "empty",
        "sec-fetch-mode": "cors",
        "sec-fetch-site": "same-site"
      },
      "remoteAddress": "127.0.0.1",
      "remotePort": 33990
    }
    res: {
      "statusCode": 500,
      "headers": {
        "x-powered-by": "Express",
        "access-control-allow-origin": "*",
        "content-type": "application/json; charset=utf-8",
        "content-length": "224",
        "etag": "W/\"e0-kJB3uym6Y/jOI2j5KTf3FlXOZlk\""
      }
    }
    responseTime: 32
    err: {
      "type": "Error",
      "message": "failed with status code 500",
      "stack":
          Error: failed with status code 500
              at onResFinished (node_modules/pino-http/logger.js:110:39)
              at ServerResponse.onResponseComplete (node_modules/pino-http/logger.js:173:14)
              at ServerResponse.emit (node:events:525:35)
              at ServerResponse.emit (node:domain:489:12)
              at onFinish (node:_http_outgoing:967:10)
              at callback (node:internal/streams/writable:555:21)
              at afterWrite (node:internal/streams/writable:500:5)
              at afterWriteTick (node:internal/streams/writable:487:10)
              at processTicksAndRejections (node:internal/process/task_queues:81:21)
    }
```
</details>

We should be logging the underlying error that led to a 500 response. In this example, the cause is a failed database query as I'd switched to a branch that renamed tables but hadn't run the migrations, but that is not at all evident in the logs.

jasonaowen:
Our error handler is invoked in both routine and exceptional circumstances. The [HTTP status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) is a handy proxy for what should be logged: 1XX, 2XX, and 3XX codes should not invoke the error handler, 4XX codes are expected and should not be logged as an error, and 5XX codes are serious. Log such errors at the `error` level, which should be enabled in all environments.

Resolves #350 Error logs do not contain useful information

### DIFF
--- a/src/middleware/errorHandler.ts
+++ b/src/middleware/errorHandler.ts
@@ -113,9 +113,13 @@ export const errorHandler = (
   res: Response,
   next: NextFunction, // eslint-disable-line @typescript-eslint/no-unused-vars
 ): void => {
-  logger.debug(err);
   logger.trace(req.body);
   const statusCode = getHttpStatusCodeForError(err);
+  if (statusCode >= 500) {
+    logger.error({ err, statusCode });
+  } else {
+    logger.debug({ err, statusCode });
+  }
   res.status(statusCode)
     .contentType('application/json')
     .send({


slifty (approved):
I appreciate this change and the logic for it!

bickelj (approved):
This looks great and works great. When I turned off my DB locally but had a valid JWT, I see the error logged at `ERROR` level. When I logged out, I see the error logged at `DEBUG` level. Thank you for this!

Issue #228, opened 2023-02-07
We should consolidate DB result parsing


slifty:
Right now we in every endpoint we guard against the database response to ensure the database provided a valid type.

As we've implemented more endpoints (and more complex endpoints) we're increasingly duplicating those checks and, more meaningfully, duplicating the need for complicated mocks in our tests.

This came to a head in https://github.com/PhilanthropyDataCommons/service/pull/210

It would be good to explore creating a centralized helper function that will run a database query and properly parse / handle / throw in the event of an issue.  This could hopefully cut down on duplicated test coverage.

slifty:
On this note -- we can actually use TS generics to signal what type of object is going to be returned by a given `db.sql` call -- however this is just a type signal and if the query is malformed it could result in a lack of type safety.

I'm inclined to prefer to have type guards since they verify the returned value's type at runtime, but wanted to raise the option in case others felt that was overkill.

slifty:
The only place we use `db.sql` at this point is in the DB middleware we added.
There are tests that use `db.query` still, and once those are updated I think we can close this issue 🎉

PR #373, opened 2023-05-23
Add data platform provider endpoints

### LINKED ISSUE #368: Create temporary API for storing data platform provider data
We have the concept of "external data" in the ERD, but for the immediate future / very first iteration we have decided to simply store JSON data responses from Candid and Charity Navigator.  This is not intended to be something that lasts for the long term, but will allow the data viewer to access this data and get us some feedback from users and other stakeholders.

This will involve:

- A migration for storing the object
- An API for writing the object
- An API for accessing the object

slifty:
This PR adds the `GET /platformProviderResponses` and `POST /platformProviderResponses` endpoints.  These endpoints allow the PDC to store and serve raw data that has been pulled from data platform providers.

Resolves #368


src/handlers/platformProviderResponsesHandlers.ts
```diff
@@ -0,0 +1,124 @@
...
+    'platformProviderResponses.selectByExternalId',
+    { externalId: req.query.externalId },
+  ).then((platformProviderResponsesQueryResult) => {
+    const { rows: platformProviderResponses } = platformProviderResponsesQueryResult;
+    res.status(200)
+      .contentType('application/json')
+      .send(platformProviderResponses);
+  }).catch((error: unknown) => {
+    if (isTinyPgErrorWithQueryContext(error)) {
+      next(new DatabaseError(
```
slifty, line 57: *sigh* idk -- do we need to test these, should I give up on that dream.  It made sense for the first endpoint or two but now It's just so noisy and so much duplicated test code.  (See #228 which I think speaks to this issue indirectly).  Happy to add it, but also happy to just move on with our lives.

src/database/migrations/0012-create-platform_provider_responses.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE platform_provider_responses (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  external_id VARCHAR NOT NULL,
+  platform_provider VARCHAR NOT NULL,
+  data JSON NOT NULL,
```
jasonaowen, line 5: In general, prefer the `JSONB` data type over the `JSON` data type:

```suggestion
  data JSONB NOT NULL,
```

src/handlers/platformProviderResponsesHandlers.ts
```diff
@@ -0,0 +1,124 @@
...
+interface GetPlatformProviderResponseByExternalIdParams {
+  externalId: string;
+}
+const getPlatformProviderResponsesByExternalIdParamsSchema
+: JSONSchemaType<GetPlatformProviderResponseByExternalIdParams> = {
+  type: 'object',
+  properties: {
+    externalId: {
+      type: 'string',
+      pattern: '^\\d{9}$',
```
jasonaowen, line 25: This is the "we're storing EINs without dashes here" choice you called out verbally.

Just to talk through what I was suggesting: right now, in prod, every EIN we're storing in the `organization_tax_id` field has a dash (including the invalid ones!). Candid supports looking up by EINs that include a dash. My plan for the Candid data script in https://github.com/PhilanthropyDataCommons/data-scripts/pull/24 is to get the EIN from the proposal, look it up in Candid, and then upload the Candid response - without much parsing - to the PDC via the new write API.

Later, the data-viewer will make a request in the same fashion - get the EIN from the `organization_tax_id` field, and make a request to the new read API.

Requiring dashless EINs adds extra processing to every client, and I'm not clear on what the benefit is - can you talk more about the upside?

src/database/migrations/0012-create-platform_provider_responses.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE platform_provider_responses (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  external_id VARCHAR NOT NULL,
+  platform_provider VARCHAR NOT NULL,
```
jasonaowen, line 4: Are we wanting to keep a history? I don't think it'll be useful, and in the spirit of light-weightedness, may I instead propose:

```suggestion
  external_id VARCHAR,
  platform_provider VARCHAR,
  PRIMARY KEY (external_id, platform_provider),
```

In particular, that should make retrieval easier, as the client can assume one-or-none rather than having to pick through an array.

src/handlers/platformProviderResponsesHandlers.ts
```diff
@@ -0,0 +1,124 @@
...
+    },
+  },
+  required: [
+    'externalId',
+  ],
```
jasonaowen, line 30: Can the read API also require a `provider` parameter, so that the data viewer can request something like `/platformProviderResponses?externalId=12-3456789&provider=CharityNavigator`?

jasonaowen (commented):
Thanks for working on this, @slifty!

src/database/migrations/0012-create-platform_provider_responses.sql
slifty: I'm not sure we've spent enough time with this problem space to be able to confidently rule out history use cases as being a desired feature.  Most signs have pointed to "we always want an audit trail" so I would start off assuming we *DO* want history in the system.

On another note: I was under the impression that not having an surrogate primary key is an anti pattern, but that might be an out of date concept (I believe it's rooted in DB optimizations / faster joins).  This is to say, we could create a UNIQUE constraint on external_id and platform_provider as a natural key but I'd be wary to lose the `id` field.

Were you planning on doing one API call per platform provider (as opposed to just getting all the data in one call)?

src/handlers/platformProviderResponsesHandlers.ts
slifty: I think supporting a provider is a fine idea; requiring it is overly prescriptive to me.  What happens if you wanted to avoid two API calls?  What happens when there are 5 providers?

I realize this is a bespoke API endpoint, but I admit I assumed you'd want all of the provider responses in one swoop / you'd parse that single response, in the name of network efficiency.

src/handlers/platformProviderResponsesHandlers.ts
slifty: To be clear: it doesn't *quite*matter what candid's API supports since what we're defining here is an API that will only support that if we write the logic to convert the input provided.  To that end, it is much easier to just strip dashes, which means dash stripping is going to happen in order to normalize the storage / query (we don't want a situation where the candid data for an org is against a different external ID than the other data platform providers for that same org).

Which is to say: the processing has to happen regardless for every query / we are going to want to store without dashes.  It felt a little odd to me for someone to POST `{ externalId: '11-1111-111' }` and get back a modified value (`{ externalId: '111111111' }`) as opposed to requiring the client to adhere to a clearly documented API spec.

Does that make sense?  Happy to talk by voice about it if that's easier.

src/database/migrations/0012-create-platform_provider_responses.sql
slifty: ... actually, maybe the solution here is to not use external ID in the first place, and instead store this in relation to an applicant via an `applicant_id` foreign key.

src/handlers/platformProviderResponsesHandlers.ts
slifty: ... actually, maybe the solution here is to not use external ID in the first place, and instead store this in relation to an applicant via an applicant_id foreign key.

src/database/migrations/0012-create-platform_provider_responses.sql
jasonaowen: > On another note: I was under the impression that not having an surrogate primary key is an anti pattern, but that might be an out of date concept (I believe it's rooted in DB optimizations / faster joins). This is to say, we could create a UNIQUE constraint on external_id and platform_provider as a natural key but I'd be wary to lose the `id` field.

My understanding is the opposite: surrogate keys are only useful if the natural keys are very large, or likely to change, and should otherwise be avoided. [Josh Berkus's post Primary Keyvil](http://www.databasesoup.com/2015/03/primary-keyvil-reprised.html) is a great read; I strongly recommend it!

jasonaowen: @slifty and I just talked! I'll try to summarize our conversation.

My main concern was that as a consumer of the read API, I didn't want to have to pick among multiple versions of data platform provider responses. So long as the query only returns one-or-zero for each, that concern is addressed, and I don't have a strong opinion on keeping multiple versions.

The benefit of keeping multiple versions would be to have the option to someday show a history; that's something we baked in very deeply to proposals, and it's not unreasonable to think we might want this here.

However, the framing I think we arrived at is that this is a _cache_ of Charity Navigator / Candid's data, and nothing more. If we decide we want a history of that data in the PDC, then that will be one (of several) reasons to throw out this implementation. We had considered - and rejected - the option of being a proxy to those services, primarily because we didn't want our availability to be dependent on their availability, and secondarily because we didn't want that code to live in this repo; a cache - with the code living in the data-scripts repo - is a more conservative approach that resolves both of those concerns.

src/handlers/platformProviderResponsesHandlers.ts
jasonaowen: After discussion, I think we arrived at the decision that we do _not_ want to link this to applicants; this is a separate, standalone, tech-debt table that we are writing with the intention to delete it when we have the opportunity to build this feature out more fully and with less deadline pressure. Pilot, not MVP!

We also don't know what that fuller feature is going to look like: maybe there'll be a way to connect your data platform provider account? Maybe data platform providers will push data rather than have us pull data? Probably we'll map to our base fields?

So, for now, we're doing a solely application-side foreign key (that is, not in the database, just in client code), and we needn't normalize at all - whatever external ID the [data-scripts] client posts is what we save, and whatever external ID the [data-viewer] client looks up is what we query by, and if those differ than that's very sad.

src/handlers/platformProviderResponsesHandlers.ts
jasonaowen: > I assumed you'd want all of the provider responses in one swoop / you'd parse that single response, in the name of network efficiency (at the very least it seems like it'd be good for the API to support either approach).

Oh! Yes, I think you're right. I was initially worried about typing the response, because although the data-scripts and this repo don't need to care about the shape of the data platform providers' responses, the data-viewer does, but I don't think that's actually a significant difference in implementation.

src/handlers/platformProviderResponsesHandlers.ts
slifty: (Thank you for taking the time to summarize -- +1 to all you wrote!)

src/types/PlatformProviderResponse.ts
```diff
@@ -0,0 +1,6 @@
+export interface PlatformProviderResponse {
+  externalId: string;
+  platformProvider: string;
+  data: Record<string, unknown>;
+  createdAt: Date;
```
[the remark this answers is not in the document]
reefdog: FWIW — and this may just be specific to _my_ career/education and not universal — I _immediately_ associate `createdAt`/`updatedAt` specifically and only with local table insert/update timestamps. In fact if it ever referred to something else (like the remote created-at time of the fetched object), I'd object on those grounds.

src/database/queries/platformProviderResponses/selectByExternalId.sql
```diff
@@ -0,0 +1,9 @@
+SELECT
+  id AS "id",
+  external_id AS "externalId",
+  platform_provider AS "platformProvider",
+  data,
```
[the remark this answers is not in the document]
slifty: Ah yes -- any preference at all on which is better (I had intended to take out the AS "id" just because I thought it's what you and jason would have suggested; I probably would value consistency)?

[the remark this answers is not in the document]
bickelj: I'm OK with aiming at the concise form and only doing the "AS" as needed.

src/types/PlatformProviderResponse.ts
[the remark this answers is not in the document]
bickelj: I figured it out and given @reefdog feedback there is no need to waste any more time on this.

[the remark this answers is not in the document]
reefdog: @bickelj Whoops, you're right, I hadn't noticed that! I _still_ connect those terms to those concepts, but I was also considering them in a table context. Still, sounds like you're okay with it so no need for me to muse further.

src/database/queries/platformProviderResponses/selectByExternalId.sql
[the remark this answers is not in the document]
slifty: hahaha I'll take out the "AS" when it isn't needed!  ty for calling out the discrepancy.

src/openapi.json
```diff
@@ -1134,6 +1167,87 @@
...
+          "Internal Use Only"
+        ],
+        "security": [
+          {
+            "auth": []
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "A list of known funding opportunities.",
```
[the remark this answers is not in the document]
slifty: Whoops thank you!

src/database/migrations/0012-create-platform_provider_responses.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE platform_provider_responses (
```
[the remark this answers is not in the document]
slifty: Wanted to make sure you saw this thread: https://github.com/PhilanthropyDataCommons/service/pull/373#discussion_r1203042236

I agree / do prefer surrogate keys in general (as you know since we talked about), that said I think Jason's summary captures the reason why this is maybe different enough to be, well, different?

Jason also linked to an article (which I haven't dug into yet but plan to) that I think articulates all the reasons why my surrogate key bias is wrong 😂

(I don't remember what your preferences on this were in the original discussion)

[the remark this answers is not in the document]
bickelj: Whoops I had missed it. And I am OK with revisiting assuming others agree.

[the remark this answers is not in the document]
slifty: I'm fine with revisiting later -- for this I'm inclined to just leave as currently is, which reinforces the "this is a cache" decision.

src/database/migrations/0012-create-platform_provider_responses.sql
```diff
@@ -0,0 +1,12 @@
...
+  platform_provider VARCHAR NOT NULL,
+  PRIMARY KEY (external_id, platform_provider),
+  data JSONB NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+COMMENT ON TABLE platform_provider_responses IS
+  'Data loaded from data platform providers.';
+COMMENT ON COLUMN applicants.external_id IS
+  'A normalized external identifier based on the value stored in the data platform provider.';
```
jasonaowen, line 12: This comment is no longer accurate in light of our decision to relax the normalization rules, right?

And, with that relevant information no longer being true, I don't think either comment is adding any new information that is not already communicated by the table and column names, so can we delete them?

src/database/queries/platformProviderResponses/insertOne.sql
```diff
@@ -0,0 +1,18 @@
...
+  external_id,
+  platform_provider,
+  data
+)
+VALUES (
+  :externalId,
+  :platformProvider,
+  :data
+)
+ON CONFLICT (external_id, platform_provider) DO UPDATE
```
jasonaowen, line 11: Oh, I'm sorry, I'd meant to write this up for you - thanks for working it out, and I hope that didn't take too long!

src/handlers/platformProviderResponsesHandlers.ts
```diff
@@ -0,0 +1,123 @@
...
+  });
+};
+
+type PostPlatformProviderResponse = Omit<PlatformProviderResponse, 'createdAt'>;
+const postPlatformProviderResponseSchema: JSONSchemaType<PostPlatformProviderResponse> = {
+  type: 'object',
+  properties: {
+    externalId: {
+      type: 'string',
+      pattern: '^\\d{9}$',
```
jasonaowen, line 72: I think we need to drop this pattern, too, as part of the [previous conversation](https://github.com/PhilanthropyDataCommons/service/pull/373#discussion_r1203227850)!

```suggestion
```

src/openapi.json
```diff
@@ -363,6 +363,39 @@
...
         ]
       },
+      "PlatformProviderResponse": {
+        "type": "object",
+        "properties": {
+          "id": {
+            "type": "integer",
+            "readOnly": true,
+            "example": 17
+          },
```
jasonaowen, line 373: This is no longer the case!

```suggestion
```

src/openapi.json
```diff
@@ -363,6 +363,39 @@
...
+            "type": "object"
+          },
+          "createdAt": {
+            "type": "string",
+            "format": "date-time",
+            "readOnly": true
+          }
+        },
+        "required": [
+          "id",
```
jasonaowen, line 392: and here:

```suggestion
```

src/openapi.json
```diff
@@ -1134,6 +1167,87 @@
...
+        ],
+        "responses": {
+          "200": {
+            "description": "A list of cached platform provider responses.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/components/schemas/Opportunity"
```
jasonaowen, line 1190: I don't think this is right? I'm not sure what it ought to be, though.

jasonaowen (changes requested):
Thanks, @slifty! A few more things to follow up on. Let me know if you want me to take any of these on - I'd be happy to help!

slifty:
Whoops!  Thank you for all those @jasonaowen; fixed up and ready for re-re-review

jasonaowen:
@slifty this is looking good! Let me try actually uploading data locally using the new API; I unfortunately didn't get to that yesterday.

jasonaowen (approved):
Awesome! Got this working locally with both read and write. Thank you, @slifty!

bickelj:
Huzzah! And huzzah for being OK with 79.59% test coverage!

PR #382, opened 2023-05-31
Add externalId parameter to API specification

slifty:
This PR fixes the open API specification to include the `externalId` parameter.


jasonaowen (approved):
Thanks for following up!

Issue #381, opened 2023-05-31
Unexpected 500 response: "...values and fields must be sorted..."


bickelj:
This URL currently returns a 500: https://api.philanthropydatacommons.org/proposals/452?includeFieldsAndValues=true

```json
{"name":"UnknownError","message":"Given values and fields must be sorted such that they correspond. index=87: 712 !== 706.","details":[{}]}
```

bickelj:
The running version is `ghcr.io/philanthropydatacommons/service:20230525-f59e112`.

In `proposalsHandlers.ts`, these first two queries are expected to be sorted the same way such that they can be joined in software (see https://github.com/PhilanthropyDataCommons/service/pull/198#issuecomment-1384225631 and following).

```typescript
      Promise.all([
        db.sql('proposalFieldValues.selectByProposalId', { proposalId: req.params.id }),
        db.sql('applicationFormFields.selectByProposalId', { proposalId: req.params.id }),
        db.sql('proposalVersions.selectByProposalId', { proposalId: req.params.id }),
      ]).then(([
//...
        const valuesWithFields = joinApplicationFormFieldsToProposalFieldValues(
          proposalFieldValuesQueryResult.rows,
          applicationFormFieldsQueryResult.rows,
        );
```

The join function does some consistency/assumption checks before joining:

```TypeScript
export const joinApplicationFormFieldsToProposalFieldValues = (
  values: ProposalFieldValue[],
  fields: ApplicationFormField[],
): ProposalFieldValue[] => {
  const newValues = structuredClone(values);
  if (newValues.length !== fields.length) {
    // It is invalid to try to zip values and fields of unequal length.
    throw new Error(`Given arrays must be of equal length. ${values.length} !== ${fields.length}`);
  }

  return newValues.map((value, index) => {
    // Look in the fields array at the same index as the values array.
    const field = fields[index];
    if (field === undefined) {
      throw new Error(
        `Expected only defined fields and values, copiedFields[${index}] was undefined`,
      );
    }
    if (value.applicationFormFieldId !== field.id) {
      throw new Error(
        `Given values and fields must be sorted such that they correspond. index=${index}: ${value.applicationFormFieldId} !== ${field.id}.`,
      );
    }
    // When all is well, add the field to the value.
    return { ...value, applicationFormField: field };
  });
};
```

[...]

PR #383, opened 2023-05-31
Order deep proposal queries by another column

bickelj:
Without this change, a non-deterministic sort order can cause an HTTP 500 response with message "...values and fields must be sorted...".

With this change, the sort order is first by position (as before) and then also by proposal field value (pfv) ID because there can be many pfv rows for a given application form field (aff) row. This should be sufficient to guarantee deterministic order such that the application join will succeed.

Issue #381 Unexpected 500 response: "values and fields must be sorted"

See also https://github.com/PhilanthropyDataCommons/service/pull/198 and https://github.com/PhilanthropyDataCommons/service/pull/210.

### DIFF
--- a/src/database/queries/applicationFormFields/selectByProposalId.sql
+++ b/src/database/queries/applicationFormFields/selectByProposalId.sql
@@ -10,4 +10,4 @@ INNER JOIN proposal_field_values pfv
 INNER JOIN proposal_versions pv
   ON pv.id = pfv.proposal_version_id
 WHERE pv.proposal_id = :proposalId
-ORDER BY pv.version DESC, pfv.position;
+ORDER BY pv.version DESC, pfv.position, pfv.id;
--- a/src/database/queries/proposalFieldValues/selectByProposalId.sql
+++ b/src/database/queries/proposalFieldValues/selectByProposalId.sql
@@ -8,4 +8,4 @@ FROM proposal_field_values pfv
 INNER JOIN proposal_versions pv
   ON pv.id = pfv.proposal_version_id
 WHERE pv.proposal_id = :proposalId
-ORDER BY pv.version DESC, pfv.position;
+ORDER BY pv.version DESC, pfv.position, pfv.id;


jasonaowen (approved):
Since these both contain inner joins with `proposal_field_values`, and its `id` column is a unique surrogate key, I *think* this should stably sort everything. :+1: on the untested belief that this fixes an immediate bug.

However, doing application joins like this is bad. This is what a database is for; we shouldn't be doing this in code. It's also not the first time we've had this conversation, although probably the first time it's bitten us like this.

As a side note, it is somewhat strange to me to be ordering the results of a query by columns that are not returned in that query!

Thank you for tracking this down, @bickelj, and coming up with a quick fix!

jasonaowen:
> However, doing application joins like this is bad. This is what a database is for; we shouldn't be doing this in code. It's also not the first time we've had this conversation, although probably the first time it's bitten us like this.

See also https://github.com/PhilanthropyDataCommons/service/pull/314!

slifty:
While i agree that doing application joins isn't the right approach (and we have a long term plan to move from it) I'll also throw out that a cleaner solution, IMHO, is to implement a joining algorithm that does not require artificially coupled queries -- e.g. one that will take data and, if it needs sorting, just sort it rather than throwing an error.  From a design PoV I think that is more maintainable than this, as this now has two unrelated queries that need to be sorted in a specific way for reasons that don't necessarily make sense in the context of the queries themselves.

I'm fine with this PR as is, but just wanted to mention the alternative approach.

src/database/queries/applicationFormFields/selectByProposalId.sql
```diff
@@ -10,4 +10,4 @@ INNER JOIN proposal_field_values pfv
 INNER JOIN proposal_versions pv
   ON pv.id = pfv.proposal_version_id
 WHERE pv.proposal_id = :proposalId
-ORDER BY pv.version DESC, pfv.position;
+ORDER BY pv.version DESC, pfv.position, pfv.id;
```
slifty, line 13: While I agree that doing application joins isn't the right approach (and we have a long term plan to move from it) I'll also throw out that so long as we are, a cleaner solution, IMHO, is to implement a joining algorithm that does not require artificially coupled queries -- e.g. one that will take data and, if it needs sorting, just sort it rather than throwing an error.

From a design PoV I think that is more maintainable than this approach, as this now has two unrelated queries that need to be sorted in a specific way for reasons that don't actually make sense in the context of the queries themselves.

I assume the desire to not sort is an optimization decision, but we don't currently have reason to believe that the optimization is necessary.  On the flip side, the optimization has resulted in this bug and if we merge the PR I would argue it is resulting in tech debt in the form of unintuitive coupling.

reefdog: > …if we merge the PR I would argue it is resulting in tech debt in the form of un-intuitive coupling.

If this design is so wrong that we should prioritize a wider fix, that's reasonable, but I don't think it should block pushing out this production hotfix.

Not to be demo-driven, but we _do_ have a big demo tomorrow and likely a bunch of stakeholders logging in right after, and their experience with a broken site is net-worse for the project than this relatively benign change, which is at least idiomatic to our approach even if we're operating under the wrong idiom.

reefdog: Wait wait, I have a better reply:

PULL REQUESTS WELCOME

:trollface:

slifty: Well to be clear I'm not proposing a wider fix, just that instead of sorting in the query we should just invoke .sort((a,b) => ...) in the immediate context of where a particular context-specific order matters.

(also re the PR joke -- I don't think we should get in the habit of reviewers needing to take on the burden of owning an issue simply because they have feedback!)

It's fine to decide to merge tech debt and open an issue to capture that tech debt!

slifty: (I wanna triple down on the last part though: I intentionally left this as a *comment* rather than "request changes" in the name of not blocking progress!)

reefdog: Fully agree that reviewers _have_ to be free to suggest improvements without a requisite burden of providing them, or else they'll hold their feedback and that's bad! My joke was truly a non sequitur and I apologize for the insinuation. While it's obviously not up to me, diving into a PR that provides the fix wouldn't be how I would prioritize your time, that's how much of a joke it was.

Re: the approach, I'll bow out and let you guys discuss it further, but by idiomatic here I meant "we're already sorting in the query; fix that sort to not blow up" and by non-idiomatic I mean "changing where we sort".

reefdog: So clearly "idiomatic" is the wrong word here. "Minimally different from current design but not broken" perhaps?

slifty: FWIW I was thinking of the current sorting in the query as being use-case-agnostic (e.g. selecting something by proposal ID can happen for all sorts of reasons, we picked a way to sort that makes generic sense).  Whereas this new sort is very specific to JUST this use case of selecting in order to join proposal field values.

That said, I now get what you're saying, which I think is essentially "it may be that it's inappropriate to rely on the query to sort, but we were already doing that, and we might as well update the arbitrary sort order to fix the situation even if we shouldn't be relying on arbitrary sort order"

So, FINE JUSTIN YOU ARE RIGHT (even if I think the most correct fix to overly coupled pieces of code is to decouple them gosh darn it!)

reefdog: > FWIW … ID … JUST … FINE JUSTIN YOU ARE RIGHT

My speed-reading strategy of skimming for all-caps words continues to pay off!

bickelj:
ae90b97 rebases on main.

bickelj:
The approach here was a compromise between "let the database do the data things" and "let the application do the data things", so the "let the database sort so the application can do an O(1) join" is on the former tack while the "do a join in the application at all" is on the latter tack. Yeah, it's not great, but as pointed out above, this is what it is right now and we have yet to come up with a completely satisfactory-to-everybody solution.

PR #387, opened 2023-06-01
Update to node 18.16.0

bickelj:
Dependabot noticed the need to update in the Dockerfile, but not in the automated build, lint, and test scripts. This change updates the (minor) Active LTS version of node.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Build with node
         uses: actions/setup-node@v3
         with:
-          node-version: 18.12.1
+          node-version: 18.16.0
       - run: npm ci
       - run: npm run build
       - run: npm prune --omit=dev
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,6 +13,6 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-node@v3
       with:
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Build with node
         uses: actions/setup-node@v3
         with:
-          node-version: 18.12.1
+          node-version: 18.16.0
```
jasonaowen, line 25: If we swap these `node-version` arguments out for [`node-version-file`](https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/action.yml#L10-L11), then there's just one place to update for all the GitHub actions, instead of three - and no chance of them getting out of sync:

```suggestion
          node-version-file: ".node-version"
```

jasonaowen (approved):
Thanks, @bickelj!

.github/workflows/build.yml
bickelj: I like this approach because I like setting the exact version everywhere, but earlier [we decided to use the `.node-version` as a major-version social signal to devs rather than drive code and pipelines](https://github.com/PhilanthropyDataCommons/service/issues/134). So I'm not comfortable making the change without consent from @slifty.

bickelj:
9af1ccc rebased on main.

PR #397, opened 2023-06-13
Uninstall TypeScript definitions for pg

jasonaowen:
The dependency on the `@types/pg` package was added in commit 7dd9d58930244799b1643effbe6ce4fa1420a445. However, we are - happily! - using TinyPg to interact with PostgreSQL, rather than working directly with the pg library.

Remove the dependency so that we don't have to keep it up to date:

```sh
npm uninstall --save-dev @types/pg
```

Found while reviewing #393.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,6 @@
         "@types/express": "^4.17.17",
         "@types/jest": "^28.1.2",
         "@types/node": "^18.16.16",
-        "@types/pg": "^8.10.1",
         "@types/supertest": "^2.0.12",
         "@types/swagger-ui-express": "^4.1.3",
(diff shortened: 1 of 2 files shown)


jasonaowen:
I was wrong!

```
/home/runner/work/service/service/src/database/migrate.ts
Error:    6:9  error  Unsafe assignment of an `any` value              @typescript-eslint/no-unsafe-assignment
Error:    9:9  error  Unsafe assignment of an `any` value              @typescript-eslint/no-unsafe-assignment
Error:   13:5  error  Unsafe member access .release on an `any` value  @typescript-eslint/no-unsafe-member-access
Error:   13:5  error  Unsafe call of an `any` typed value              @typescript-eslint/no-unsafe-call
```

Sorry for the noise!

Issue #343, opened 2023-05-08
Build notification to deploy project failing


bickelj:
See recent runs of the `build.yml` workflow.

The PAT has not expired but I see a better, organization-level P(sic)AT approach. I added a fine-grained token that only involves workflows between `service` and `deploy` and am trying that.

bickelj:
Example failure from https://github.com/PhilanthropyDataCommons/service/actions/runs/4917656888/jobs/8784603167:

```quote
Run pauldraper/workflow-dispatch@v1.5
  with:
    workflow: update-service-image.yml
    ref: main
    repo: PhilanthropyDataCommons/deploy
    inputs: {"image-tag":"20230508-388b85b"}
    wait: true
  env:
    REGISTRY: ghcr.io
    IMAGE_NAME: philanthropydatacommons/service
    VERSION: 20230508-388b85b
HttpError: Must have admin rights to Repository.
    at /home/runner/work/_actions/pauldraper/workflow-dispatch/v1.5/workflow-dispatch/dist/main.js:13876:27
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async NextRunDispatcher.start (/home/runner/work/_actions/pauldraper/workflow-dispatch/v1.5/workflow-dispatch/dist/main.js:29807:9)
    at async main (/home/runner/work/_actions/pauldraper/workflow-dispatch/v1.5/workflow-dispatch/dist/main.js:29921:23)
```

I do not think it requires admin rights to repository. I added a fine-grained access token policy to [settings](https://github.com/organizations/PhilanthropyDataCommons/settings/personal-access-tokens) and then added a fine-grained access token with read access to the `service` and `deploy` repos, and write to actions access. It expires May 8, 2024. The benefit is the token is created by a person but visible to and revokable by the organization admins. [Here is the exact token created](https://github.com/organizations/PhilanthropyDataCommons/settings/personal-access-tokens/73003).

This approach seemed to work fine. See [the recent success](https://github.com/PhilanthropyDataCommons/service/actions/runs/4918583759/jobs/8785107371#step:11:5).

bickelj:
The organization token appears to be working fine.

PR #448, opened 2023-08-17
Update the inter-repo workflow token README

bickelj:
There have been some improvements to Personal Access Tokens in GitHub that we have taken advantage of since this document was first written. Tokens are still created by an individual user but can have limited access to repository resources and can be managed from organization settings, e.g. an organization owner can see and revoke access.

Issue #343 Build notification to deploy project failing Issue #23 Avoid use of PAT in CI / CD


jasonaowen (approved):
Thanks for documenting this, @bickelj!

PR #459, opened 2023-08-24
Fix mockjws (and dependabot configuration)

### LINKED ISSUE #458: mock-jwks no longer supports commonjs
From https://www.npmjs.com/package/mock-jwks

> As of version 2 and march 2023 this package is a pure esm package. I made an example on how to use the module. Use version 1 for a commonjs version.

The result of this is that our current build is broken (silently according to our CI, due to #457)

I think we have three choices:

1. `service` should become esm / not transpile to cjs
2. we should use version 1 of mock-jwks
3. we should remove the mock-jwks dependency

slifty:
This PR downgrades mock-jwks to v1 since the latest major version doesn't support common JS.  It also updates dependabot to not attempt to re-upgrade outside of v1.

(While here, I also threw in a longer-standing desire to not have dependabot update our node types across major versions.)

Resolves #458

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,9 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: "mock-jwks"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: "docker"
     directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,8 @@ updates:
     ignore:
       - dependency-name: "mock-jwks"
         update-types: ["version-update:semver-major"]
+      - dependency-name: "@types/node"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: "docker"
     directory: "/"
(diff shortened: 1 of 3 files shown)


slifty:
Tests actually passed btw

image

.github/dependabot.yml
```diff
@@ -5,6 +5,11 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: "mock-jwks"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "@types/node"
```
slifty, line 11: @jasonaowen wanted to tag you here in case there's a way to undo https://github.com/PhilanthropyDataCommons/service/pull/370 now that we have the "right" way of stopping major node updates

bickelj (approved):
Looks good to me. I reviewed the code changes and the job output at https://github.com/PhilanthropyDataCommons/service/actions/runs/5967504211/job/16189307463.

.github/dependabot.yml
jasonaowen: The documentation in #370's description says

> `@dependabot ignore this major version` will close this PR and stop Dependabot creating any more for this major version (unless you reopen the PR or upgrade to it yourself)

So, we can reopen the PR if you want, but I don't think it's a problem to have both an explicit rule and also a one-off comment directive telling Dependabot not to try to open 18->20 PRs.

slifty: The problem arises if we ever update to Node 20 -- at which point we would want dependabot updates and the only way to re-enable them would be to sift through 5 years worth of dependabot prs to find the right one 😅

PR #462, opened 2023-08-25
Ensure CI detects failing tests

### LINKED ISSUE #457: `npm run test:ci` prints failures but exits success
See https://github.com/PhilanthropyDataCommons/service/actions/runs/5966921283/job/16187534824#step:7:20 and the corresponding "success".

Locally when running `npm run test:ci` I get failures printed but exit 0:
```
$ npm run test:ci

> @pdc/service@0.3.0 test:ci
> jest --config=jest.config.ci.js --runInBand

Determining test suites to run...ts-jest[versions] (WARN) Version 5.1.6 of typescript installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=4.3.0 <5.0.0-0). Please do not report issues in ts-jest if you are using unsupported versions.
 FAIL  src/__tests__/proposalVersions.int.test.ts
  ● Test suite failed to run
...
$ echo $?
0
```

slifty:
This PR updates our test workflow to run tests twice: once without coverage and once with coverage.  This is in order to mitigate a bug with `ts-jest` that can sometimes mask failures when coverage is reported.

If that bug gets patched in upstream we can remove the double-testing.

Resolves #457

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,17 @@ jobs:
       - name: Run build
         run: npm run build
 
-      - name: Run all tests
+      - name: Run all tests (without coverage)
+        run: npm run test:ci -- --coverage=false
+        env:
+          PGHOST: localhost
+          PGPORT: 5432
+          PGUSER: postgres
+          PGPASSWORD: postgres
+          AUTH_SERVER_ISSUER: https://totally-fake-server-name/realms/pdc
+          OPENAPI_DOCS_AUTH_CLIENT_ID: pdc-fake-client-id
+
+      - name: Run all tests (with coverage)
         run: npm run test:ci
         env:
           PGHOST: localhost


.github/workflows/test.yml
```diff
@@ -36,7 +36,17 @@ jobs:
       - name: Run build
         run: npm run build
 
-      - name: Run all tests
+      - name: Run all tests (without coverage)
```
slifty, line 39: I opted to make this an entirely duplicated "job" within the workflow just to be explicit at what was going on / make it easier to see that tests are intentionally being run twice.

That being said, this choice results in duplicated env settings so I wanted to call it out.

I'd be happy to change this to a single job that runs both commands via `&`, and therefore avoids duplication of the env settings.

slifty:
It looks like the test job ran both: https://github.com/PhilanthropyDataCommons/service/actions/runs/5976928322/job/16215816420?pr=462

.github/workflows/test.yml
bickelj: I think it's worth it to have the clean separation. Only a handful of lines are duplicated and we can remove this when the [upstream bug](https://github.com/kulshekhar/ts-jest/issues/4193) is fixed.

.github/workflows/test.yml
```diff
@@ -36,7 +36,17 @@ jobs:
...
+        run: npm run test:ci -- --coverage=false
```
bickelj, line 40: Since this passes in the current `main`, I went back to the failing `mock-jwks` version with `git checkout cb1104e` and then ran `npm test:ci`, verified it exited `0` and then ran `npx jest --config=jest.config.ci.js --runInBand --coverage=false` (the equivalent of this line here) and verified it exited `1`.

bickelj (approved):
Thank you! This will boost our confidence in the automated test results!

jasonaowen:
Thanks for taking care of this, @slifty and @bickelj!

PR #567, opened 2023-10-30
Add presigned post endpoint

### LINKED ISSUE #557: Develop an API endpoint for generating a presigned post
This issue is related to the batch upload implementation strategy outlined in https://github.com/PhilanthropyDataCommons/meta/issues/128

We want clients to be able to upload files for at least two reasons:

1. Upload CSVs containing batch proposal data.
2. Upload files associated with proposals / organizations / etc.

This issue is focused on the first use case (I'm mentioning the second so that we keep it at least in mind in case it helps inform design decisions).

slifty:
This PR adds a new endpoint that allows authenticated clients to generate a presigned post, allowing them to upload file data to the PDC for various types of processing.

Resolves #557


slifty:
I need to write some tests before this is really ready for review (as I'm sure codecov is about to come tell me)

slifty:
Some notes for consideration here:

1) there is currently no limit on file size; I'm happy to add one, or not, whatever.  Would love suggestions on what the limit should be.

2) I decided *NOT* to create a database entry for these requests for now.  There is no immediate use case.

3) There is nothing preventing abuse (I think abuse prevention would require implementation of both 1 and 2 above)

For MVP where all users are trusted I think we can get away with this first implementation, but please push back if you disagree.

package.json
```diff
@@ -70,6 +71,8 @@
     "typescript": "^5.2.2"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.438.0",
```
slifty, line 74: Apparently these aws additions added 3.7k lines to package-lock -- turtle dependencies all the way down.

reefdog:
@slifty Woo!

It does feel a little odd to have *no* abuse prevention in place, but since we are operating in such a closed mode, it seems like an appropriate sacrifice for MVP. Especially since AFAIK the abuse we're talking about is harm to the PDC (through DDOS-type data stuffing) than harm to users (through abuse vectors like that described [here](https://github.com/orgs/PhilanthropyDataCommons/projects/4/views/2?pane=issue&itemId=39314324)).

But a question for the future: would implementing abuse prevention on the presigned post be the correct place, or on the upload handler itself?

And more fundamentally, is the presigned post only valid for the original file? (I.e., is there a protection against generating the signature with one file, but then uploading another? I assume so, but of course if the answer is no, then I would definitely think the upload handler itself would be where we would write any protections.)

slifty:
@reefdog good questions!

1) presigned post is the place for abuse prevention / the point of entry -- once the post is returned data will be written directly to S3, which means if someone asked to upload 500 TB of data they'd be able to do so (well, S3 has a 5GB limit for single-part presigned post but you get the idea).

To your point though I imagine the bulk upload processer would also benefit for some checks *as well* in some future where we're allowing non-internal-S3 URLs to be provided.

2) The presigned post constraints are: file size, file type.  the post can be used once, and the data uploaded must match the specified constraints.

reefdog:
@slifty Great explanation, thanks. I think client-side hashing adds significant implementation overhead for minimal benefit, so I agree with this current implementation. Plus, while the presigned post lets people stuff data into our S3 bucket, it doesn't actually authorize a bulk upload; that will still be protected by the upload handler itself. Which is to say a co-opted presigned POST won't grant a [bad actor](https://www.imdb.com/name/nm2959239/) permission to put data in the system.

I'd still rather a Jesse or Jason review this, but if neither is available I can dive further in. Consider my big question satisfied!

slifty:
@reefdog I *think* that it's just you and me lately so if you could spend a cycle digging deeper it would be appreciated 🥺

reefdog:
@slifty On it! Well, today.

package.json
reefdog: 😱

src/types/PresignedPostRequest.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const presignedPostRequestWriteSchema: JSONSchemaType<PresignedPostRequestWrite> = {
+  type: 'object',
+  properties: {
+    fileType: {
+      type: 'string',
+    },
+    fileSize: {
+      type: 'integer',
+      minimum: 0,
```
reefdog, line 64: Wow pretty short-sighted that you think nobody will want to upload dark matter.

reefdog (approved):
@slifty I appreciate the tour through this. It works, tests run, and I was able to generate a presigned POST.

There is some magic occurring (as you noted) in the Jest mocking, which I'm not experienced enough to assess. If there are problems with it, hopefully they become rapidly evident. And if there aren't, nicely done!

slifty:
Huzzah!

PR #572, opened 2023-11-01
Add bulk upload template

### LINKED ISSUE #561: Create the bulk upload CSV template
The bulk upload process will involve third parties populating data in a predefined CSV template -- we need to define that template / pick which of our base fields "make the cut"

slifty:
This PR adds a first draft of a bulk upload template which will eventually be used as part of a bulk upload processing.  The template is expected to change and evolve over time.

The template is served at `/static/bulkUpload.csv`

Resolves #561


src/public/bulkUpload.csv
```diff
@@ -0,0 +1,3 @@
+Add one proposal per row below. do NOT edit the template,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```
reefdog, line 1: Is this intended to make its way into the CSV that we deliver? I don't think we should include instruction text in the CSV.

Also, we had explicitly discussed allowing the template to be editable for power uploaders who want to add other base fields. Are you walking that back, or just postponing it?

src/public/bulkUpload.csv
```diff
@@ -0,0 +1,3 @@
...
...,Proposal Budget Narrative,Proposal Project Budget,Proposal Fiscal Sponsor?,Proposal Fiscal Sponsor Information,Proposal Fiscal Sponsor Name,Proposal Fiscal Sponsor EIN,Proposal Fiscal Sponsor Contact First Name,Proposal Fiscal Sponsor Cantact Last Name,Proposal Fiscal Sponsor Contact Email,Proposal Regrant,Proposal Regrant Subgrantees,Proposal Subgrantee List,Proposal Expenditure Responsibility,Proposal Equivalency Determination,Proposal Tipping Amount,Proposal Funding Summary,Proposal Start Date,Proposal End Date,Proposal Duration,Proposal Location of Work,Proposal Age Group,Proposal Ethnicity,Proposal Need,Proposal Activities,Proposal Tactics and Methods,Proposal Strategy,Proposal Organizational Strengths,Proposal Equity,Proposal Engagement,Proposal Challenges and Risks,Proposal Related Programs,Proposal Supplemental Information,Proposal Results,Proposal Project Outcomes,Proposal Anticipated Impact,Proposal Sustainability,Proposal IRB Approval Needed,Proposal IRB Approval Obtained,Organization Lobbying Activities,Proposal Legislation,Proposal Lobbying more than Grant,PDC Share Data,Fundraising Expense Ratio,Program Expense Ratio,Administrative Expense Ratio,Working Capital Ratio
```
reefdog, line 3: I'm hesitant about including this. I see the UX value, but it feels potentially problematic. How will you reliably distinguish a row of labels from a row of content?

reefdog (approved):
I'm going to approve this for the sake of velocity, but see my inline pushback and consider if you want to implement it.

src/public/bulkUpload.csv
slifty: Happy to remove any type of instruction -- also FWIW I was ok with the idea that someone could add columns in an unsupported way.  The main issue is I don't want people thinking "ah, I'll just update this header to match MY field's label!"

src/public/bulkUpload.csv
slifty: Happy to take this out as well -- the idea here is that the first two (well, three if we include instructions) would always be skipped.  Content would start on row 4.

src/public/bulkUpload.csv
reefdog: Yeah, I'd remove the instrux.

src/public/bulkUpload.csv
reefdog: IMO we should count on an entirely different interface for humanizing field keys than the CSV template. Here, the only function of the label row is explanatory; which is great, but immediately complicates the story. I imagine the user thinking: can I remove this row if I want? can I customize the labels by tweaking the wording of these fields?

I think it's best to just provide the key row. That's a much more stable rule to enforce. "First row is base field keys. Bam."

reefdog: > Another benefit of taking this out is that suddenly our "template" is actually just a "very easy to construct yourself, if you want" specification.

Agreed!

slifty:
Removed all rows except the base field codes; rebased to main (will merge after CI)

PR #592, opened 2023-11-14
Group some dependencies

slifty:
This PR uses dependabot's [group functionality](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) which should help related updates within a single ecosystem to map to a single PR instead of a series of PRs.  This should save us some time but ALSO it should help prevent rare situations where PRs have to be merged / rebased in a specific order.

I identified two sets that I feel are low hanging fruit (aws-sdk, and eslint), but there may be other groups worth considering either for this PR or a future PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,10 @@ updates:
         update-types: ["version-update:semver-major"]
       - dependency-name: "@types/node"
         update-types: ["version-update:semver-major"]
+    groups:
+      aws-sdk:
+        patterns:
+        - "@aws-sdk/*"
 
   - package-ecosystem: "docker"
     directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,9 @@ updates:
       aws-sdk:
         patterns:
         - "@aws-sdk/*"
+      eslint:
+        patterns:
+        - "*eslint*"
 
   - package-ecosystem: "docker"
     directory: "/"


.github/dependabot.yml
```diff
@@ -10,6 +10,14 @@ updates:
...
       - dependency-name: "@types/node"
         update-types: ["version-update:semver-major"]
+    groups:
+      aws-sdk:
+        patterns:
+        - "@aws-sdk/*"
+      eslint:
+        patterns:
+        - "@typescript-eslint/*"
+        - "eslint/*"
```
jasonaowen, line 20: I don't think `eslint/*` matches anything - there's the base `eslint` package that does not have a trailing slash, and several packages that start with `eslint-`, but I don't see anything that looks like `eslint/` in `npm list`.

Maybe this should just be a single, broad match?

```suggestion
        - "*eslint*"
```

Optionally, we can also limit the scope to `devDependencies`:

```yaml
dependency-type: development
```

([reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups))

jasonaowen (commented):
The aws-sdk group looks good, but I think the eslint group needs a tweak to work as intended!

.github/dependabot.yml
slifty: Oh thanks good catch -- I was avoiding a wide match like that just because of concern that it would some day match unrelated dependencies... but just for fun I spent 60 seconds using the npm search to search "eslint{X}" for all the possible values of `{X}` and... based on a very unscientific analysis there do not appear to be any non-eslint-packages that include the string `eslint` in the title.

slifty:
@jasonaowen it's not lost on me that I managed to make a mistake in what was functionally a PR with only two meaningful lines. but hey, that's what reviews are for I guess!

Ready for re-review.

jasonaowen (approved):
No worries! Looks good - hope tomorrow's dependabot PRs are a little easier to manage.

PR #596, opened 2023-11-14
Add Contributing and Code of Conduct guides

bickelj:
Each repository in PhilanthropyDataCommons should point to a single pair of guides for PDC in its `.github` repository.

See https://github.com/PhilanthropyDataCommons/front-end/issues/251


slifty (approved):
Thank you for doing all of these!

PR #611, opened 2023-11-17
Allow schema specification when migrating

### LINKED ISSUE #609: Consider finding a new tool for migrations
https://github.com/ThomWright/postgres-migrations seems to be a dead project.  I have a PR to fix a bug that affects us which has been open for about 18 months: https://github.com/ThomWright/postgres-migrations/pull/93

It also is not being kept up to date more generally / the last commit was made 2 years ago.

We might consider:

1. Forking, merging the fix, and publishing on our own.
2. Finding a different migration tool.

slifty:
This PR replaces our migration tool with one that supports multiple schemas, and utilizes that tool when running tests.

Resolves #609

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "jwks-rsa": "^3.1.0",
     "pino": "^8.16.2",
     "pino-http": "^8.5.1",
-    "postgres-migrations": "^5.3.0",
+    "postgres-schema-migrations": "^6.1.0",
     "require-env-variable": "^4.0.2",
     "swagger-ui-express": "^5.0.0",
     "tinypg": "^7.0.0",
--- a/src/database/migrate.ts
+++ b/src/database/migrate.ts
@@ -1,13 +1,14 @@
 import path from 'path';
-import { migrate as pgMigrate } from 'postgres-migrations';
+import { migrate as pgMigrate } from 'postgres-schema-migrations';
 import { db } from './db';
 
-export const migrate = async (): Promise<void> => {
+export const migrate = async (schema = 'public'): Promise<void> => {
   const client = await db.getClient();
   try {
     await pgMigrate(
       { client },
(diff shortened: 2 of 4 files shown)


src/database/migrate.ts
```diff
@@ -1,13 +1,14 @@
 import path from 'path';
-import { migrate as pgMigrate } from 'postgres-migrations';
+import { migrate as pgMigrate } from 'postgres-schema-migrations';
 import { db } from './db';
 
-export const migrate = async (): Promise<void> => {
+export const migrate = async (schema = 'public'): Promise<void> => {
```
slifty, line 5: Someone with stronger DB opinions should let me know if it would be better to use a `pdc` schema as our default, rather than `public.

bickelj: I think it's OK to default to `public`. My sense is that more often than not a dedicated database instance gets created for each separate application.

bickelj (approved):
LGTM. I didn't try it locally, but I see the checks passed and no red flags stand out.

package.json
```diff
@@ -82,7 +82,7 @@
     "jwks-rsa": "^3.1.0",
     "pino": "^8.16.2",
     "pino-http": "^8.5.1",
-    "postgres-migrations": "^5.3.0",
+    "postgres-schema-migrations": "^6.1.0",
```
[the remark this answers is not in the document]
slifty: I actually learned about it from a comment on the issue that PR addresses (https://github.com/ThomWright/postgres-migrations/issues/92#issuecomment-1568844396)

jasonaowen (commented):
Oh perfect - that fork is on npm after all! I'd missed that it got renamed.

I was able to run this (well, `main`, since this is a post-hoc review) locally without issue.

+1! Great work, @slifty. :blue_heart:

PR #628, opened 2023-11-27
Add bulk upload creation endpoint

### LINKED ISSUE #558: Develop an API endpoint for registering a batch upload
As described in https://github.com/PhilanthropyDataCommons/meta/issues/128 and #557 we want users to be able to upload proposal data in bulk via CSV.

We have the endpoint / flow implemented for actually *uploading* the data; now we need the end point for registering that upload as a bulk proposal file and kicking off the processing pipeline. The actual processing of the CSV does *not* fall under this issue (that will be #559), but this issue includes POST and GET for batch uploads.

As part of this issue we are also going to have to think a little bit about associating batch uploads with users / agents / keys.  There is, shockingly, no concept of an "Account" or a "User" in the current PDC (all we have right now is authentication, which is completely covered by KeyCloak).

slifty:
This PR adds an endpoint for registering bulk uploads into the PDC for future processing.

Resolves #558


slifty:
@jasonaowen the codecov complaint is that I didn't test the lines for catching a thrown database error.

I'm happy to add tests there, but also increasingly feel like the real solution is to DRY out those error handlers somehow.

src/types/BulkUpload.ts
```diff
@@ -0,0 +1,42 @@
...
+  COMPLETED = 'completed',
+  FAILED = 'failed',
+  CANCELED = 'canceled',
+}
+
+export interface BulkUpload {
+  readonly id: number;
+  fileName: string
+  sourceUrl: string;
+  destinationUrl: string;
```
slifty, line 16: `destinationUrl` is generic, and the protocol of the URL could be part of the url content (e.g. `S3://` or `SFTP://` or `HTTP://`).  If we want to force S3, though, then maybe we shouldn't be storing a URL at all but instead should be storing the relevant S3 pieces directly (key? bucket? region?).

I suppose if we were baking granular S3 components into the schema we'd want:

1. DestinationBucket
2. DestinationKey
3. region too?  Maybe?

jasonaowen: Can you say more about what you intend `destinationUrl` to be used for? I don't remember any conversations we might have had about it, and I'm not able to infer it from context. My best guess is that this is will be the stable URL of the uploaded file after it's processed - is that your intention? Why is that useful to record here?

May I suggest dropping it from this PR, and adding it later if & when we're ready to populate it?

slifty: Yeah I'll drop it from this PR (though to the second question of "why here" -- this is the only persistent record of a bulk upload, so I think here is the best -- but we can cross that bridge when it's implemented, maybe there will be a better choice at that point!)

slifty:
@jasonaowen removed the `destinationUrl` field from the PR; ready for your review.

src/database/migrations/0013-create-bulk_uploads.sql
```diff
@@ -0,0 +1,21 @@
...
+  'completed',
+  'failed',
+  'canceled'
+);
+
+CREATE TABLE bulk_uploads (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  file_name VARCHAR NOT NULL,
+  source_url VARCHAR NOT NULL,
+  destination_url VARCHAR,
```
jasonaowen, line 13: One more instance!

```suggestion
```

src/database/migrations/0013-create-bulk_uploads.sql
```diff
@@ -0,0 +1,21 @@
...
+  status bulk_upload_status,
```
jasonaowen, line 14: `NOT NULL`, right?

```suggestion
  status bulk_upload_status NOT NULL,
```

src/types/BulkUpload.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const bulkUploadCreateSchema: JSONSchemaType<BulkUploadCreate> = {
+  type: 'object',
+  properties: {
+    fileName: {
+      type: 'string',
+    },
+    sourceUrl: {
+      type: 'string',
+    },
```
jasonaowen, line 33: Out of curiosity, is this where you would put validation around the source? Or would you do that in code rather than JSONSchema? (Feel free to say "I'll figure that out when we get there", of course!)

jasonaowen (commented):
Two small things before I start testing this!

src/types/BulkUpload.ts
jasonaowen, line 33: Should we specify some minimum lengths here?

```suggestion
    fileName: {
      type: 'string',
      minLength: 1,
    },
    sourceUrl: {
      type: 'string',
      minLength: 1,
    },
```

`1` is obviously way too short, but I don't know how in-the-weeds you want to get here; for the file name, `5` plus "ends in `.csv`" might be more reasonable, and we've previously discussed checking that the URL comes from our S3 bucket.

slifty: Good question!

I'd probably want to use the json schema for it if possible (in the spirit of "AJV is our validation engine and it is driven by the json schema").  That way an invalid URL is rejected on POST.

I probably *also* want to defensively validate the URL before invoking any kind of fetch...

I do think it'd be appropriate to add that validation on that first part as part of this PR.

slifty: Good call -- will add both of those in.

jasonaowen: Cool - in that case, I have a pending PR around S3 buckets that might be relevant here; we can decide how we want to sequence it.

slifty: Ok so I think there's two categories of validation here and I'm a little torn.

One is on the general form -- e.g. "this should be a URL" -- which I think belongs in the schema.

The other might be further constraint that is more complex / dynamic -- e.g. "this should be a URL on { DOMAIN X }" -- that I'm a little bit wary of baking into this schema definition because we might have a world where these schemas are driven by the (non-dynamic) swagger spec...

For now I'm going to add validation about the string structure (that it's URL), and will punt in the short term on the trickier validation question.

jasonaowen: Perfect!

slifty: Update on the "check from our S3 bucket" concept -- I'm going to hold off on adding that to the specification itself (in my head I'm leaving it as "that is less about the structure / schema of the data but rather is a more dynamic type of validation" but maybe that's not a useful distinction.)

I may just say "it has to start with `https?:\/\/`" for now.

src/types/BulkUpload.ts
```diff
@@ -0,0 +1,43 @@
...
+export const bulkUploadCreateSchema: JSONSchemaType<BulkUploadCreate> = {
+  type: 'object',
+  properties: {
+    fileName: {
+      type: 'string',
+      pattern: '^.+\\.csv$',
+    },
+    sourceUrl: {
+      type: 'string',
+      pattern: 'https?:\\/\\/.+',
```
jasonaowen, line 34: It'd be good to have a test for this!

jasonaowen (approved):
This looks great! I was able to run this locally and try several submissions via the Swagger UI. Thanks, @slifty!

src/types/BulkUpload.ts
slifty: Oh yes -- fixed up!

slifty:
Hooray!  Rebased to main and will merge after CI

Issue #559, opened 2023-10-24
Set up a queue / processing system for handling bulk proposal uploads


slifty:
We want a system for processing bulk uploads (and probably other types of async processing over time, but this issue is focused on the immediate need).

slifty:
The first step is picking a queue system.  Some options:

* **[BullMQ](https://www.npmjs.com/package/bullmq)** is queue / messaging system that uses Redis.  I've worked with [in the past](https://github.com/TechAndCheck/tech-and-check-alerts/tree/main) and I remember it being straightforward and clean.
* **[RabbitMQ](https://www.rabbitmq.com)** is also a thing, and the node client is [ampqplib](https://www.npmjs.com/package/amqplib).

Both seem to be well maintained; Rabbit seems like it might be a little heavier?  My inclination would be to use Bull.

reefdog:
I recall Bull working well for us on T&C, and BullMQ being the modern/TS implementation speaks well. I say go forth!

bickelj:
Copying my [comments from yesterday](https://chat.opentechstrategies.com/#narrow/stream/75-PDC/topic/queues/near/177528) here where I probably should have made them:

No strong feelings other than wondering if there's a way to avoid yet another dependency on another running process somewhere. Redis-based stuff or RabbitMQ sound fine to me if we need one though. IIRC it was easy to set up (rolling window) persistence with Redis. I don't think I tried persistence with RabbitMQ.

I have also seen Postgres (ab)used as a queue system. That's one way to avoid an additional dep but is worse than just using a Q system.

bickelj:
If we assume a single back-end service instance, couldn't we use a TS or JS queue implementation? Like [this](https://www.npmjs.com/package/queue) or some other, better library that allows a redis backend optionally when it is finally needed (i.e. when we go to multiple back-end service instances)?

I'm not trying to be an obstacle, just asking because I don't know. From a distant, skimming observer view, "we need redis because file uploads" sounded like overkill. But this is bulk uploads so I may be way off and not understanding properly. Also, redis is great, so yea redis!

slifty:
I spent some time today exploring non-redis approaches which also led me to step back and reflect on whether a queue is what we want.  Some alternatives I could think of were:

1. Hot folders
2. synchronous processing
3. cron jobs (looking database rows)

I think 1 and 2 are non-starters and would too easily result in brittleness / lost work.
[...]

PR #630, opened 2023-11-27
Add worker queue

slifty:
This PR adds support to the PDC for a worker queue and creates a "shell" worker for bulk upload processing (the worker doesn't actually do anything yet).

This builds on the work of #628 and shouldn't be reviewed until that has been merged and this PR has been rebased.

Related to #559

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,5 @@
 import { app } from './app';
+import { startJobQueue } from './jobQueue';
 import { getLogger } from './logger';
 
 const logger = getLogger(__filename);
@@ -16,3 +17,7 @@ app.listen(
     logger.info(`Server running on http://${host}:${port}`);
   },
 );
+
+startJobQueue().catch((err) => {
+  logger.error(err, 'Job queue failed to start');
+});
--- a/src/jobQueue.ts
+++ b/src/jobQueue.ts
@@ -1,0 +1,79 @@
+import {
+  Logger,
+  quickAddJob,
+  run,
+  runMigrations,
+} from 'graphile-worker';
+import { getLogger } from './logger';
+import { db } from './database/db';
+import { processBulkUpload } from './tasks';
+import type { ProcessBulkUploadJobPayload } from './types';
+
+const logger = getLogger(__filename);
+
+enum JobType {
+  PROCESS_BULK_UPLOAD = 'processBulkUpload',
+}
+
+export const jobQueueLogger = new Logger((scope) => (
+  (level, message, meta) => {
+    switch (level.valueOf()) {
+      case 'error':
+        logger.error({ meta, scope }, message);
+        break;
+      case 'warn':
+        logger.warn({ meta, scope }, message);
+        break;
+      case 'info':
+        logger.info({ meta, scope }, message);
+        break;
+      case 'debug':
+        logger.debug({ meta, scope }, message);
+        break;
+      default:
+        logger.info({ meta, scope }, message);
+    }
+  }
+));
+
+export const startJobQueue = async () => {
+  const runner = await run({
+    logger: jobQueueLogger,
+    pgPool: db.pool,
+    concurrency: 5,
+    noHandleSignals: false,
+    pollInterval: 1000,
+    taskList: {
+      processBulkUpload,
+    },
+  });
+  await runner.promise;
+};
+
+export const runJobQueueMigrations = async () => (
+  runMigrations({
+    logger: jobQueueLogger,
+    pgPool: db.pool,
+  })
+);
+
+export const addJob = async (
+  jobType: JobType,
+  payload: unknown,
+) => (
+  quickAddJob(
+    {
+      logger: jobQueueLogger,
+      pgPool: db.pool,
+    },
+    jobType,
+    payload,
+  )
+);
+
+export const addProcessBulkUploadJob = async (payload: ProcessBulkUploadJobPayload) => (
+  addJob(
+    JobType.PROCESS_BULK_UPLOAD,
+    payload,
+  )
+);
--- a/src/tasks/__tests__/processBulkUpload.unit.test.ts
+++ b/src/tasks/__tests__/processBulkUpload.unit.test.ts
@@ -1,0 +1,12 @@
+import { getMockJobHelpers } from '../../test/mockGraphileWorker';
+import { processBulkUpload } from '../processBulkUpload';
+import { InternalValidationError } from '../../errors';
+
+describe('processBulkUpload', () => {
+  it('should error when passed an invalid payload', async () => {
+    await expect(processBulkUpload(
+      {},
+      getMockJobHelpers(),
+    )).rejects.toBeInstanceOf(InternalValidationError);
+  });
+});
--- a/src/tasks/processBulkUpload.ts
+++ b/src/tasks/processBulkUpload.ts
@@ -1,0 +1,17 @@
+import { isProcessBulkUploadJobPayload } from '../types';
+import { InternalValidationError } from '../errors';
+import type { JobHelpers } from 'graphile-worker';
+
+export const processBulkUpload = async (
+  payload: unknown,
+  helpers: JobHelpers,
+): Promise<void> => {
+  if (!isProcessBulkUploadJobPayload(payload)) {
+    helpers.logger.debug('Malformed bulk upload job payload', { errors: isProcessBulkUploadJobPayload.errors ?? [] });
+    throw new InternalValidationError(
+      'The bulk upload job payload is not properly formed',
+      isProcessBulkUploadJobPayload.errors ?? [],
+    );
+  }
+  helpers.logger.debug(`Started processBulkUpload Job for Bulk Upload ID ${payload.bulkUploadId}`);
+};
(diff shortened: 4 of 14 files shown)


slifty:
It's unclear to me why tests are failing in CI -- they are passing locally.

src/index.ts
```diff
@@ -16,3 +17,8 @@ app.listen(
     logger.info(`Server running on http://${host}:${port}`);
   },
 );
+
+startJobQueue().catch((err) => {
+  logger.error('Job queue failed to start');
+  logger.error(err);
```
jasonaowen, line 23: Can these be combined into a single call to `logger.error`? I believe we want to keep such information together, so it doesn't get interleaved with other unrelated logs.

src/jobQueue.ts
```diff
@@ -0,0 +1,47 @@
...
+} from 'graphile-worker';
+import { processBulkUpload } from './tasks';
+import type { ProcessBulkUploadJobPayload } from './types';
+
+enum JobType {
+  PROCESS_BULK_UPLOAD = 'processBulkUpload',
+}
+
+const getJobQueueConnectionString = () => (
+  `postgres://${process.env.PGUSER}:${process.env.PGPASSWORD}@${process.env.PGHOST}:${process.env.PGPORT}/${process.env.PGDATABASE}`
```
jasonaowen, line 13: This is going to break my local workflow, as I don't have all those environment variables set. Fortunately, I don't believe you need to set the connection string at all - as long as either the `PGDATABASE` or `DATABASE_URL` environment variables are set, the underlying pg-node should pick up its configuration automatically:

> One of these must be provided (in order of priority):
>
> - `pgPool` pg.Pool instance
> - `connectionString` setting
> - `DATABASE_URL` envvar
> - [PostgreSQL environmental variables](https://www.postgresql.org/docs/current/libpq-envars.html), including at least `PGDATABASE` (NOTE: not all envvars are supported)

https://worker.graphile.org/docs/library/run#runneroptions

src/tasks/processBulkUpload.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const logger = getLogger(__filename);
+
+export const processBulkUpload = async (
+  payload: unknown,
+): Promise<void> => {
+  if (!isProcessBulkUploadJobPayload(payload)) {
+    logger.debug('Malformed bulk upload job payload');
+    logger.debug(isProcessBulkUploadJobPayload.errors);
+    throw new Error('The bulk upload job payload is not properly formed');
```
jasonaowen, line 12: For the most part, the Codecov checks in this PR feel a bit over the top, but I do agree with it here - we should have some tests for this task!

jasonaowen (changes requested):
I haven't tested this yet; just a quick review!

Also @slifty would you mind talking through how you decided on graphile-worker? I'm 100% on board with that choice, just curious to hear your reasoning!

src/index.ts
slifty: I'll look into it -- I maybe conflated memories but I know at least SOME logger did not take arbitrary numbers of parameters.

slifty: This was a relevant thread -- https://github.com/pinojs/pino/issues/161

src/jobQueue.ts
```diff
@@ -0,0 +1,40 @@
...
+const addJob = async (
+  jobType: JobType,
+  payload: unknown,
+) => {
+  await quickAddJob(
+    {},
+    jobType,
+    payload,
+  );
+};
```
jasonaowen, line 33: Now that we're not specifying any connection parameters, do we still need to use `quickAddJob` instead of `addJob`? From the [documentation on adding jobs](https://worker.graphile.org/docs/library/add-job),

> `quickAddJob` is similar to `addJob`, but accepts an additional initial parameter describing how to connect to the database.

It's nice to type-check that the job type is known, but I think we'd effectively get that by only exporting task-specific functions, as you are already doing below with `addProcessBulkUploadJob`.

I think we could just delete this function and update the import!

```suggestion
```

src/tasks/processBulkUpload.ts
```diff
@@ -0,0 +1,18 @@
...
+import { InternalValidationError } from '../errors';
+
+const logger = getLogger(__filename);
+
+export const processBulkUpload = async (
+  payload: unknown,
+): Promise<void> => {
+  if (!isProcessBulkUploadJobPayload(payload)) {
+    logger.debug(isProcessBulkUploadJobPayload.errors, 'Malformed bulk upload job payload');
+    throw new InternalValidationError(
```
jasonaowen, line 12: This raises a question I hadn't really considered before: how do we mark bulk uploads as failed? From the [error handling docs](https://worker.graphile.org/docs/error-handling),

> If a job throws an error, the job is failed and scheduled for retries with exponential back-off.

It's great that the framework will handle that kind of failure for us, but it's not obvious to me how/when/where to handle the follow-up we need to do of connecting a job error to a bulk upload status.

I think we can probably defer that out of this PR, since it's just setting up the framework, but if you've already thought about it I'd love to hear where you're at!

src/jobQueue.ts
```diff
@@ -0,0 +1,40 @@
...
+} from 'graphile-worker';
+import { processBulkUpload } from './tasks';
+import type { ProcessBulkUploadJobPayload } from './types';
+
+enum JobType {
+  PROCESS_BULK_UPLOAD = 'processBulkUpload',
+}
+
+export const startJobQueue = async () => {
+  const runner = await run({
```
jasonaowen, line 13: I think we want to pass a logger as part of this invocation. From the [documentation on logging](https://worker.graphile.org/docs/library/logger), it looks like it's just a function that gets passed in as part of the `run` argument object.

src/tasks/processBulkUpload.ts
```diff
@@ -0,0 +1,18 @@
...
+      'The bulk upload job payload is not properly formed',
+      isProcessBulkUploadJobPayload.errors ?? [],
+    );
+  }
+  logger.debug(`Started processBulkUpload Job for Bulk Upload ID ${payload.bulkUploadId}`);
```
jasonaowen, line 17: Likewise, the [logging documentation](https://worker.graphile.org/docs/library/logger) recommends using `helpers.logger` rather than using the main logger directly, so that it can add context like the worker, task, and job ID. (The `helpers` object is the second [argument to tasks](https://worker.graphile.org/docs/tasks#task-executor-function).)

jasonaowen (changes requested):
As you called out in the description, this needs to be rebased on `main` now that #628 is merged.

I was able to run this locally, and both see successful jobs by using the new bulk upload API endpoint, and failed jobs by using the [SQL interface](https://worker.graphile.org/docs/sql-add-job) to add a job without a payload.

I think we want to configure logging as part of this, but otherwise this is looking great!

src/jobQueue.ts
slifty: Yes indeed -- that said, `addJob` is not an imported function but rather is invoked against the runner that came out of `run()` (e.g. `runner.addJob()`.

Everything you said is still true, but I think it will mean changing the `jobQueue` to be a class (with state / runner being an instance variable) instead of a series of exported functions.  In turn, I think this would probably be something that should be passed to the express application, so that handlers are passed the queue.

This is probably better design anyway.

Only writing all this to explain why the next set of changes is going to look so wildly different.

src/tasks/processBulkUpload.ts
jasonaowen: @slifty and I talked about this on a call today. It's definitely out of scope for this PR, but broadly the worker code will update the bulk upload status as it goes. For now, monitoring job failures will be a manual process, which we think is the appropriate engineering trade-off for the moment given our (lack of) scale.

src/jobQueue.ts
slifty: I believe the alternative is to have a series of exported functions that all rely on a module-scoped `runner` variable... but that feels like it's just a class pretending to not be a class.

slifty: Update: we talked through this and will stick with the current implementation *but* I'll have it utilize the `db.pool` as the pgpool in order to avoid the overhead of creating a new connection every time a job is added.

jasonaowen (approved):
This looks good to me! I was able to run this locally again, now with prettier logs. The common database connection pool seems to be working well, and I was able to drop the database schema (`drop schema graphile_worker cascade;`) and recreate via `npm run migrate:dev`.

Need to fix the failing test, but that should be pretty minor, so +1! Nice work, @slifty.

src/tasks/__tests__/processBulkUpload.unit.test.ts
```diff
@@ -0,0 +1,8 @@
+import { processBulkUpload } from '../processBulkUpload';
+import { InternalValidationError } from '../../errors';
+
+describe('processBulkUpload', () => {
+  it('should error when passed an invalid payload', async () => {
+    await expect(processBulkUpload({})).rejects.toBeInstanceOf(InternalValidationError);
```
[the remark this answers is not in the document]
slifty: Yeah -- it's going to be a big pain to figure out how to mock this 😧 but I'll be on it tomorrow!

slifty:
I'll merge this after CI completes (I did add that `mockGraphileWorker.ts` file but am leaning into post-merge fixes if there are concerns there.

Issue #564, opened 2023-10-26
Create "buckets" on DigitalOcean for file upload


slifty:
As part of #557 I need a few buckets on DigtialOcean's S3 implementation.

In the long run we're going to want a place to store:

1. Unprocessed data
2. Processed populated bulk upload templates (#561)
3. Processed proposal-data-files (#216)

It isn't clear to me if these should be organized as separate buckets or as special key patterns within a single bucket (since I know [S3 doesn't actually have the concept of subdirectories](https://serverfault.com/questions/435827/what-is-the-difference-between-buckets-and-folders-in-amazon-s3).

I get the sense that the more "S3" way is to have one bucket per purpose, and in that sense we'd want to start with two buckets (per environment, that is).  DigitalOcean Spaces allows for up to 100 buckets for $5 / month.

So this would mean something like...

- `unprocessed.local.pdc.org`
- `unprocessed.staging.pdc.org`
- `unprocessed.dev.pdc.org`
- `unprocessed.prod.pdc.org`
- `batch-uploads.local.pdc.org`
- `batch-uploads.staging.pdc.org`
- `batch-uploads.dev.pdc.org`
- `batch-uploads.prod.pdc.org`

(and eventually something like `files.*.pdc.org` for the files related to proposal fields).

Alternatively we could just have ONE universal bucket per environment:

- `local.pdc.org`
- `staging.pdc.org`
- `dev.pdc.org`
- `prod.pdc.org`

^ and within that we would have different keys patterns of `/unprocessed/*` `/batchUploads/*` and `/files/*`.

My understanding is one benefit of the multi-bucket approach is that permissions can be a bit more simple; one benefit of the single-bucket approach is that configuration can be a bit more simple.

slifty:
Note that maybe the `local` buckets are a less ideal approach to having some way of running a local S3 service / mock.  I'll look into tools for something like that (and this would be nice for things like integration tests too).

slifty:
Thanks to @bickelj there are two buckets for local development: 

1. https://pdc-local-unprocessed.nyc3.digitaloceanspaces.com/
2. https://pdc-local-batch-uploads.nyc3.digitaloceanspaces.com/

(I do think it might be worth some day looking into localstack, but this is great for now!)

jasonaowen:
> Alternatively we could just have ONE universal bucket per environment

This is what I would recommend!

[...]

PR #635, opened 2023-11-28
Configure S3

jasonaowen:
There are a number of S3-compatible providers, and we should support them if it's reasonably easy to do so. Configuring the service to use a particular S3 provider involves several settings:

- the endpoint URL
- the region
- whether the provider uses subdomains or paths for referencing buckets
- the access key ID
- the access key secret
- the bucket

In #567, we were still thinking of having separate buckets for processed and unprocessed data. Later, in #564, we decided to only have one bucket. Consolidate and rename the bucket environment variables.

Finally, not having any of these environment variables set is a configuration error, and the app should not start in such an error state. Require the environment variables, and log an error & quit if they are missing.

Note that we should not deploy this until the relevant environment variables are set. I believe I have access and know how to do that in our current Docker setup.


src/s3Client.ts
```diff
@@ -1,11 +1,22 @@
 import { S3 } from '@aws-sdk/client-s3';
+import { requireEnv } from 'require-env-variable';
```
slifty, line 2: This triggered a series of thoughts, so I'm writing them down:

1. I bet there are other places we should be more explicitly requiring env vars (doesn't have to be in this PR)
2. What does dotenv get us / should we still use dotenv (I think dotenv is providing the ability to autoload a `.dotenv` file in the event that env vars aren't automatically sourced, yes?  Do we care enough about that to warrant having two env var dependencies?)
3. I suppose a way to accomplish this within dotenv would be to create a class dedicated to environment variable management, and access env vars from that class.

None of these thoughts warrant a change to this PR necessarily, but I wanted to raise them.

I am comfortable with this approach as written, though I could imagine there being a benefit from a custom class that defines (and validates) environment variables even more robustly.

src/s3Client.ts
```diff
@@ -1,11 +1,22 @@
...
+
+const {
+  S3_ACCESS_KEY_ID,
+  S3_ACCESS_SECRET,
+  S3_ENDPOINT,
+} = requireEnv(
```
slifty, line 10: For my edification: this pattern will break immediately on process start if the vars aren't set / we don't need to also explicitly check for these in `src/index.ts` right?

.env.example
```diff
@@ -23,8 +23,9 @@ OPENAPI_DOCS_AUTH_CLIENT_ID=${OPENAPI_DOCS_AUTH_CLIENT_ID}
...
 # For more information on populating these please see
 # https://docs.digitalocean.com/products/spaces/reference/s3-sdk-examples/
-S3_ENDPOINT=${S3_ENDPOINT}
 S3_ACCESS_KEY_ID=${S3_ACCESS_KEY_ID}
 S3_ACCESS_SECRET=${S3_ACCESS_SECRET}
-S3_UNPROCESSED_BUCKET=${S3_UNPROCESSED_BUCKET}
-S3_BATCH_UPLOADS_BUCKET=${S3_BATCH_UPLOADS_BUCKET}
+S3_BUCKET=${S3_BUCKET}
+S3_ENDPOINT=${S3_ENDPOINT}
+S3_PATH_STYLE=$(( true || false ))
```
slifty, line 30: This format doesn't work with `envsubst`, right? (E.g. if someone runs `envsubst .env.example > .env` I don't think they would be able to populate `S3_PATH_STYLE` without additional commands)

Can we maintain the `${S3_PATH_STYLE}` approach but include legitimate values in a comment?

slifty (changes requested):
Much cleaner and safer thank you

One small request around the `.env.example` but otherwise ready to roll!

src/s3Client.ts
jasonaowen: > I bet there are other places we should be more explicitly requiring env vars

I actually think this is the last of them! `git grep process.env` is a pleasantly short and conditional.

> What does dotenv get us / should we still use dotenv

It is a convenience for developers, to automatically load the contents of a `.env` file. I am increasingly of the opinion that such convenience support should not live in the repo, but rather be the responsibility of the devs, via [direnv](https://direnv.net/) or another way of managing environment variables. However, I do not believe so strongly enough to advocate for removing existing, working setups, so for now my answer to the second question is "eh, sure".

> create a class dedicated to environment variable management

We could. I think the main benefit would be consolidating all the environment variables in one place; currently, if say `OPENAPI_DOCS_AUTH_CLIENT_ID` and `S3_ACCESS_KEY_ID` were both missing, only one would be reported because they're in separate calls to `requireEnv`. I'm not sure it's worth the work and conceptual overhead, but could probably be convinced.

src/s3Client.ts
jasonaowen: Yes, it will break on process start if any of the variables are not set, and no we do not need to also check in `src/index.ts`.

.env.example
jasonaowen: Ahh, is that the motivation for this format? I wasn't sure why they were formatted using shell variables rather than example values. ...are we actually doing that anywhere? Should we document that? I'm not sure I understand the use case, tbh!

slifty: It's just an accessibility thing that I've found to be best practice.  If you have a strong reason against it I'm happy to hear it!  I think signaling values can easily be done through comments (and prefer that anyway since the comments live on once the value is replaced)

src/s3Client.ts
slifty: > I actually think this is the last of them!

I *may* be missing it but it looks like this doesn't get at the env var in https://github.com/PhilanthropyDataCommons/service/blob/main/src/logger.ts right?

> It is a convenience for developers

Ahhhh -- then... I do think we should plan to get rid of it especially due to the awkwardness documented in `index.ts` around import orders mattering (but I'll open an issue for it).

(the class manager convo I'm happy to let sit as a musing in a PR thread and not carry it forward at all!)

slifty (approved):
looks good!  There's one instance of a process.env being directly accessed in another file but that isn't related to S3.  I defer to you on if you wanna include it in here.

src/s3Client.ts
slifty: (as noted on call -- this is not required so requiring wouldn't be good lulz)

jasonaowen: Captured in https://github.com/PhilanthropyDataCommons/service/issues/640!

.env.example
jasonaowen: Thanks for explaining!

jasonaowen:
Rebased onto `main`; will merge once CI passes.

PR #639, opened 2023-11-29
Add a `description` attribute to base fields

### LINKED ISSUE #612: Add `description` attribute to base fields
We would like to be able to provide descriptions for base fields that instruct users what sort of data we expect to go there. These should be optional.

slifty:
This PR adds a new `description` column to the BaseFields type (and table).

While adding this I took the opportunity to clean up the way base fields were being inserted in our tests so that they use our db helper tools rather than inline SQL.  (I did not go so far as to try to refactor the way test data is generated, as this PR exposes we're not particularly DRY about it)

Resolves #612


src/database/seeds/0001-insert-base_fields.sql
```diff
@@ -17,283 +17,283 @@
...
-('Score in the Financial dimension' , 'score_in_the_financial_dimension', '{ "type": "integer" }' ),
-('Score in the A&T dimension' , 'score_in_the_a&t_dimension', '{ "type": "integer" }' ),
-('Rating in the A&T dimension' , 'rating_in_the_a&t_dimension', '{ "type": "integer" }' ),
-('Score in the Results dimension' , 'score_in_the_results_dimension', '{ "type": "integer" }' ),
-('Rating in the Results Dimension' , 'rating_in_the_results_dimension', '{ "type": "integer" }' ),
-('Overall Score' , 'overall_score', '{ "type": "integer" }' ),
-('Overall Rating' , 'overall_rating', '{ "type": "integer" }' ),
-('Yes/No indicator: This rating is a Donor Advisory' , 'yes_no_indicator_this_rating_is_a_donor_advisory', '{ "type": "boolean" }' );
+INSERT INTO base_fields (label, description, short_code, data_type) VALUES
+('Organization Name', '', 'organization_name', '{ "type": "string" }' ),
```
slifty, line 21: Just wanted to call out that I'm adding blank descriptions to the seed so that the seed file works against the updated schema.  (github collapses this file since so many lines were edited)

jasonaowen: What do you want to do about #305?

jasonaowen (approved):
This is looking good to me! I was able to run this locally, manually modify a base field in the database, and see it come through the API. I also confirmed that new base fields created through the API need to have their `description` set, and that the description is passed through from the API to the database. Thanks, @slifty!

src/database/migrations/0014-alter-base_fields-description.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE base_fields
+  ADD COLUMN description VARCHAR NOT NULL DEFAULT '';
+
+ALTER TABLE base_fields ALTER COLUMN description DROP DEFAULT;
```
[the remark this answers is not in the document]
slifty: Yeah! (I can't claim credit of course but it seemed like a pretty elegant solution; thanks internet!)

This might be a rare place where a comment would make sense just to highlight exactly what's going on for future readers.

slifty:
I collapsed the inserts into named functions (in the appropriate test files rather than an imported utility function).  I'm erring on merging given the approval since I don't think this change is particularly different from what was reviewed (and indeed aligns with what was requested!)

Issue #597, opened 2023-11-15
Expire unregistered uploads


jasonaowen:
After a user requests a presigned post (#557), and uses that to upload a file, something might go wrong before they register the file (#558) - maybe they lose their internet connection, or their power, or maybe there's a bug in the front-end, or whatever. Letting the file linger in our object storage, without being known by the API, would waste space and hosting costs.

The standard solution is to set up an [expiration policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-expire-general-considerations.html). (We are using [DigitalOcean Spaces](https://docs.digitalocean.com/products/spaces/), an independent S3-compatible service, which [supports the lifecycle rules API](https://docs.digitalocean.com/reference/api/spaces-api/#configure-a-buckets-lifecycle-rules).) The policy would be something like "delete objects beginning with `unprocessed/` after 1 day".

There is a question of when and how to apply this policy to buckets. It could be done by hand at bucket creation (see #564), but probably we want to apply the policy automatically - likely via server startup?

PR #647, opened 2023-12-01
Add prefix to presigned posts

jasonaowen:
The presigned post API determines the key of the object that the client will upload to S3. Previously, the intent was to have separate buckets for processed and unprocessed files, but we later decided to have only a single bucket for both. That change was partially implemented by PR #635, but that PR did not update the key generation.

Add a prefix to the object keys generated by the presigned post API, both to keep the objects in our bucket better organized, and to allow us to (later) implement a object lifecycle policy that will automatically delete old unprocessed objects.

Issue https://github.com/PhilanthropyDataCommons/service/issues/564 Create "buckets" on DigitalOcean for file upload
Issue https://github.com/PhilanthropyDataCommons/service/issues/597 Expire unregistered uploads

### DIFF
--- a/src/handlers/presignedPostRequestsHandlers.ts
+++ b/src/handlers/presignedPostRequestsHandlers.ts
@@ -24,7 +24,7 @@ const generatePresignedPost = async (
   s3Client,
   {
     Bucket: S3_BUCKET,
-    Key: `${uuidv4()}`,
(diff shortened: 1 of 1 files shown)


slifty (approved):
Looks great -- and good call on the expression

PR #694, opened 2024-01-02
Move processed csvs to final destination after processing

### LINKED ISSUE #678: Move processed CSV uploads from `/unprocessed` as part of the processing task
Once a CSV upload is processed we need to relocate it to a "final" destination (not in the `/unprocessed` namespace).

This might have been more complicated if we were allowing arbitrary URLs, but since we made the decision to ONLY support bulk uploads that are sent via our S3, this should be straightforward.  It looks like the process is to "copy" to the final destination and then "delete" the original.

Once moved, the bulk upload record should be updated to reflect the new key.

slifty:
This PR adds logic to move processed CSV upload files to a final destination after the processing is complete.  As part of this PR I had to think a little harder about what to do with data in various failure cases.

You'll see that data is moved to a permanent location even in the failure cases UNLESS the failure was due to being unable to access the S3 file.

This PR also updates the retry logic in the case of being unable to download the file from S3.

Resolves #678

### DIFF
--- a/src/s3Client.ts
+++ b/src/s3Client.ts
@@ -26,3 +26,4 @@ export const s3Client = new S3({
 });
 
 export const S3_UNPROCESSED_KEY_PREFIX = 'unprocessed';
+export const S3_BULK_UPLOADS_KEY_PREFIX = 'bulk-uploads';
--- a/src/s3Client.ts
+++ b/src/s3Client.ts
@@ -24,3 +24,5 @@ export const s3Client = new S3({
     secretAccessKey: S3_ACCESS_SECRET,
   },
 });
+
+export const S3_UNPROCESSED_KEY_PREFIX = 'unprocessed';
--- a/src/tasks/__tests__/processBulkUpload.int.test.ts
+++ b/src/tasks/__tests__/processBulkUpload.int.test.ts
@@ -429,7 +429,6 @@ describe('processBulkUpload', () => {
       },
       {
         applicationFormFieldId: 2,
-        id: 2,
         position: 1,
         proposalVersionId: firstProposalVersion.id,
         value: 'Foo LLC.',
--- a/src/tasks/__tests__/processBulkUpload.int.test.ts
+++ b/src/tasks/__tests__/processBulkUpload.int.test.ts
@@ -124,20 +124,53 @@ const getProposalsByExternalIds = async (externalIds: string[]): Promise<Proposa
 
 describe('processBulkUpload', () => {
   it('should attempt to access the contents of the sourceKey associated with the specified bulk upload', async () => {
-    const sourceKey = '550e8400-e29b-41d4-a716-446655440000';
+    const sourceKey = 'unprocessed/550e8400-e29b-41d4-a716-446655440000';
     const bulkUpload = await createTestBulkUpload({ sourceKey });
-    const sourceRequest = await mockS3ReplyWithFile(
+    const requests = await mockS3ResponsesForBulkUploadProcessing(
       sourceKey,
       `${__dirname}/fixtures/processBulkUpload/validCsvTemplate.csv`,
     );
-
     await processBulkUpload(
       {
         bulkUploadId: bulkUpload.id,
       },
       getMockJobHelpers(),
     );
-    expect(sourceRequest.isDone()).toEqual(true);
+    expect(requests.getRequest.isDone()).toEqual(true);
+  });
+
+  it('should attempt to copy the contents of the sourceKey associated with the specified bulk upload to a processed location', async () => {
+    await createTestBaseFields();
+    const sourceKey = 'unprocessed/550e8400-e29b-41d4-a716-446655440000';
+    const bulkUpload = await createTestBulkUpload({ sourceKey });
+    const requests = await mockS3ResponsesForBulkUploadProcessing(
+      sourceKey,
+      `${__dirname}/fixtures/processBulkUpload/validCsvTemplate.csv`,
+    );
+    await processBulkUpload(
+      {
+        bulkUploadId: bulkUpload.id,
+      },
+      getMockJobHelpers(),
+    );
+    expect(requests.copyRequest.isDone()).toEqual(true);
+  });
+
+  it('should attempt to delete the unprocessed file of the sourceKey associated with the specified bulk upload', async () => {
+    await createTestBaseFields();
+    const sourceKey = 'unprocessed/550e8400-e29b-41d4-a716-446655440000';
+    const bulkUpload = await createTestBulkUpload({ sourceKey });
+    const requests = await mockS3ResponsesForBulkUploadProcessing(
+      sourceKey,
+      `${__dirname}/fixtures/processBulkUpload/validCsvTemplate.csv`,
+    );
+    await processBulkUpload(
+      {
+        bulkUploadId: bulkUpload.id,
+      },
+      getMockJobHelpers(),
+    );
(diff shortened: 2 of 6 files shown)


src/tasks/processBulkUpload.ts
```diff
@@ -312,12 +333,20 @@ export const processBulkUpload = async (
...
   const bulkUpload = await loadBulkUpload(payload.bulkUploadId);
   let bulkUploadFile: FileResult;
+  let bulkUploadHasFailed = false;
   try {
     await updateBulkUploadStatus(bulkUpload.id, BulkUploadStatus.IN_PROGRESS);
     bulkUploadFile = await downloadS3ObjectToTemporaryStorage(
       bulkUpload.sourceKey,
       helpers.logger,
     );
+  } catch (error) {
```
slifty, line 343: This PR was started before the recent updates to error logging; I'll go through and fix any instances of `error` that should be `err`

src/types/BulkUpload.ts
```diff
@@ -4,6 +4,7 @@ import type { JSONSchemaType } from 'ajv';
 export enum BulkUploadStatus {
   PENDING = 'pending',
   IN_PROGRESS = 'in_progress',
+  RETRYING = 'retrying',
```
slifty, line 7: We talked about this on call -- decided NOT to add a `retrying` state.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -88,6 +88,20 @@ const mockS3ReplyWithFile = async (
     filePath,
   );
 
+const mockS3ResponsesForBulkUploadProcessing = async (
```
jasonaowen, line 105: The refactor commit that introduces this method (341b6a0c3340653e233d19504ec9c3c511b83ddd) appears to be after a commit that references it (d532c2ab0cda35c085b3bd16d5ae54094ff76930).

src/tasks/processBulkUpload.ts
```diff
@@ -301,6 +318,10 @@ const createProposalFieldValueForBulkUploadCsvRecord = async (
   return proposalFieldValue;
 };
 
+const getProcessedKey = (unprocessedKey: string): string => (
+  `${S3_PROCESSED_BULK_UPLOADS_KEY_PREFIX}/${path.basename(unprocessedKey)}`
```
jasonaowen, line 322: Might I instead suggest that the processed key should be `${S3_PROCESSED_BULK_UPLOADS_KEY_PREFIX}/${bulkUpload.id)}`? That'll make it easier for us to examine uploads manually, rather than requiring database lookups!

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -231,6 +270,24 @@ describe('processBulkUpload', () => {
     expect(updatedBulkUpload.status).toEqual(BulkUploadStatus.FAILED);
   });
 
+  it('should move the csv file to processed location if the csv contains an invalid short code', async () => {
+    await createTestBaseFields();
+    const sourceKey = '550e8400-e29b-41d4-a716-446655440000';
```
jasonaowen, line 275: Should this also start with `unprocessed/`? In fact, should we just extract a `TEST_SOURCE_KEY` constant?

jasonaowen (commented):
This looks promising! I haven't tested it yet, but will do so after your next pass. Thanks, @slifty!

src/tasks/processBulkUpload.ts
slifty: I'm realizing that I had mentally assumed that the `processed` key might some day be the prefix for other types of file uploads -- which is probably a bad idea!

Which is just to say, I'm gonna update the key from being `processed/` to `processedBulkUploads/` at which point yes I think that `id` change will make sense and be safe from collisions.

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: Yeah, I was thinking we should probably add a status pre-check to the worker to make sure (1) the state of the bulk upload is `PENDING` and (2) the key has the `unprocesed` prefix.

Once implemented, these tests would need to be updated (currently, of course, that is not a precondition which is why the tests don't fail)

src/tasks/processBulkUpload.ts
slifty: (oh, it was already `processedBulkUploads` lol)

Adding the id change!

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: I made this change and also added those checks (without the checks it would be too easy for a malicious user to predict a processed file's key path, allowing them to process someone else's bulk upload as though it was their own *AND* resulting in that processed file being moved to a new location).

jasonaowen: Oh, woah, fantastic catch - I hadn't even thought about that kind of attack! I was just thinking the tests should match the kind of data we expect in production. Great work, @slifty!

slifty:
@jasonaowen I think this is ready for you again

jasonaowen (approved):
This looks good to me! I was able to run this locally and see both the successful workflow and the new error handling.

For posterity, I did so via the SQL interface:

```sql
insert into bulk_uploads (file_name, source_key, status)
  values ('upload.csv', 'invalid/asdf', 'pending');
select graphile_worker.add_job('processBulkUpload',
  json_build_object('bulkUploadId', 5));
```

Thanks, @slifty!

src/s3Client.ts
```diff
@@ -24,3 +24,6 @@ export const s3Client = new S3({
     secretAccessKey: S3_ACCESS_SECRET,
   },
 });
+
+export const S3_UNPROCESSED_KEY_PREFIX = 'unprocessed';
+export const S3_PROCESSED_BULK_UPLOADS_KEY_PREFIX = 'processedBulkUploads';
```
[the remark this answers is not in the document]
slifty: Great call on both counts -- implementing now.

slifty:
Investigating why tests are failing (it seems like the failing tests are unrelated to the code changes?)

500 errors are being returned by unrelated endpoints.
Meanwhile, locally I'm getting 401 unauthorized errors for some reason.

It must be Friday!

src/test/integrationSuiteSetup.ts
```diff
@@ -36,4 +37,5 @@ beforeEach(async () => {
 afterEach(async () => {
   await cleanupDatabaseForCurrentWorker();
   jest.restoreAllMocks();
+  nock.cleanAll();
```
slifty, line 40: This line seems to result in some wild errors int ests but *ONLY* after c08557b9caac64768119ba8700ca68c47a20d4de

That's the commit that removed auth from a the base fields... I haven't figured out what exactly is going on.  I believe the jwt library is using nock to override jwt interactions, and so it makes sense that `cleanAll` might conflict with it.  It does not make nearly as much sense that the conflict would only manifest after removing auth from a single endpoint, however.

I'll keep digging.

slifty: OK -- looks like moving the `mockJwks` to the beforeEach / afterEach instead of beforeAll / afterAll does the trick for the local issues.

jasonaowen: That makes sense! Good find.

PR #695, opened 2024-01-02
Add prettier

### LINKED ISSUE #601: Set up prettier
It looks like eslint is about to disable formatting rules en masse -- https://github.com/eslint/eslint/issues/17522

The idea is that code formatters (e.g. prettier) are the right call for that approach.  To that end, let's get ahead of the game and set up prettier.

slifty:
This PR adds prettier and invokes it on the project.

It also moves us from spaces to tabs (tabs is what prettier has determined to be the Right Choice)

Resolves #601


slifty:
Note: I haven't invoked prettier yet since I'd like to merge the outstanding PRs first (it's gonna create infinite merge conflicts) -- that said we can still review based on the actual changes!  I think the most important one is tabs vs spaces.

I admit: I prefer tabs, but the actual logic here is prettier defaults to tabs.

jasonaowen:
It'd be helpful to add this to a ignore file, per [Ignoring mass reformatting commits with git blame](https://akrabat.com/ignoring-revisions-with-git-blame/)!

docs/ARCHITECTURE.md
```diff
@@ -44,9 +43,9 @@ graph LR
 
 ## Concepts
 
-The core of the PDC is its API.  All third party integrations and user interfaces are powered by that API.
```
slifty, line 47: oh no it removes double spaces :( cc @reefdog

slifty:
@jasonaowen ready for real review!

src/__tests__/applicants.int.test.ts
```diff
@@ -1,40 +1,32 @@
...
-import {
-  db,
-  loadTableMetrics,
-} from '../database';
-import { getLogger } from '../logger';
-import { PostgresErrorCode } from '../types';
-import { expectTimestamp } from '../test/utils';
-import { mockJwt as authHeader } from '../test/mockJwt';
-import type { Result } from 'tinypg';
+import request from "supertest";
```
jasonaowen, line 1: It looks like we're switching from single quotes (`'`) to double quotes (`"`) throughout. Is that an intentional change? Is that configurable?

slifty: It's what prettier defaults to. My approach to prettier in general (and this PR) is to not think about their choices and just accept them.

src/__tests__/bulkUploads.int.test.ts
```diff
@@ -1,213 +1,205 @@
...
-    await agent
-      .get('/bulkUploads')
-      .set(authHeader)
-      .expect(200, {
-        total: 0,
-        entries: [],
-      });
-  });
+	it('returns an empty Bundle when no data is present', async () => {
+		await agent.get('/bulkUploads').set(authHeader).expect(200, {
```
slifty, line 12: bah -- see this is an example of me not loving the wrapping defaults of prettier (womp womp)

docs/ARCHITECTURE.md
reefdog: I'm very happy for your loss.

src/__tests__/applicants.int.test.ts
reefdog: I was also promised that everyone would begin to use proper typographic apostrophes from now on. I couldn’t be happier.

slifty:
Note: the failing test is the issue addressed in #707 -- it's not a real failure

slifty:
I added one more commit to remove the now-obsolete eslint disable lines around tab formatting in the context of type helpers.

src/__tests__/bulkUploads.int.test.ts
jasonaowen: Yeah, same. :disappointed:

src/database/operations/load/loadBulkUploadBundle.ts
```diff
@@ -1,16 +1,14 @@
...
-  queryParameters: TinyPgParams,
-): Promise<Bundle<BulkUpload>> => loadBundle(
-  'bulkUploads.selectWithPagination',
-  queryParameters,
-  'bulk_uploads',
-  isBulkUpload,
-);
+	queryParameters: TinyPgParams,
+): Promise<Bundle<BulkUpload>> =>
+	loadBundle(
```
jasonaowen, line 9: Oh, ew, here's another place where I disagree with Prettier. Should we start wrapping all multiline expressions in parens?

jasonaowen (approved):
Thanks for taking care of this, @slifty!

jasonaowen:
Oh, and pro tip, you can change the tab width GitHub uses in the [appearance settings](https://github.com/settings/appearance).

src/database/operations/load/loadBulkUploadBundle.ts
slifty: Blegh that decision feels so odd to me!  I guess I can only hope I'll get used to it...

src/__tests__/bulkUploads.int.test.ts
slifty: Discussion here: https://github.com/prettier/prettier/issues/7884

PR #703, opened 2024-01-05
Remove auth check for GET `/baseFields`

### LINKED ISSUE #702: Remove auth requirements from the base fields `GET` endpoint
Base fields are a public artifact -- we want people to be able to load the list regardless of whether they have an authorized account on the PDC.

slifty:
This PR removes the authorization check when accessing `GET /baseFields`

Resolves #702


jasonaowen (approved):
Great! Making progress on authorization :ballot_box_with_check: .

I was able to run this locally and successfully make a `curl(1)` request against `/baseFields`.

Thanks, @slifty!

PR #707, opened 2024-01-08
Fix race condition in test

slifty:
This PR fixes another hidden race condition in one of our tests.

To be clear: the code itself is fine (we don't care about the order of insertion) -- the issue was just that our test was assuming a specific insertion order for `application form fields` and so had the potential to fail.

### DIFF
--- a/src/tasks/__tests__/processBulkUpload.int.test.ts
+++ b/src/tasks/__tests__/processBulkUpload.int.test.ts
@@ -6,6 +6,9 @@ import { getMockJobHelpers } from '../../test/mockGraphileWorker';
 import { processBulkUpload } from '../processBulkUpload';
 import { BulkUploadStatus, Proposal } from '../../types';
 import type {
+	ApplicationForm,
+	ApplicationFormField,
+	BaseField,
 	BulkUpload,
 	ProposalFieldValue,
 	ProposalVersion,
@@ -53,19 +56,30 @@ const createTestBulkUpload = async (
 	return bulkUpload;
 };
 
-const createTestBaseFields = async (): Promise<void> => {
-	await db.sql<BulkUpload>('baseFields.insertOne', {
+const createTestBaseFields = async (): Promise<[BaseField, BaseField]> => {
+	const {
+		rows: [proposalSubmitterEmailBaseField],
+	} = await db.sql<BaseField>('baseFields.insertOne', {
 		label: 'Proposal Submitter Email',
 		description: 'The email address of the person who submitted the proposal.',
 		shortCode: 'proposal_submitter_email',
 		dataType: 'string',
 	});
-	await db.sql<BulkUpload>('baseFields.insertOne', {
+	const {
+		rows: [organizationNameBaseField],
+	} = await db.sql<BaseField>('baseFields.insertOne', {
 		label: 'Organization Name',
 		description: 'The name of the applying organization.',
 		shortCode: 'organization_name',
 		dataType: 'string',
 	});
+	if (proposalSubmitterEmailBaseField === undefined) {
+		throw new Error("Couldn't create the proposal submitter email base field");
+	}
+	if (organizationNameBaseField === undefined) {
+		throw new Error("Couldn't create the organization name base field");
+	}
+	return [proposalSubmitterEmailBaseField, organizationNameBaseField];
 };
 
 const mockS3GetObjectReplyWithFile = async (
@@ -130,6 +144,28 @@ const getProposalsByExternalIds = async (
 			throw new Error(`There is no proposal with externalId "${externalId}"`);
 		}
 		return proposal;
+	});
+};
+
+const getApplicationFormFieldsByBaseFieldIds = async (
+	applicationFormId: number,
+	baseFieldIds: number[],
+): Promise<ApplicationFormField[]> => {
+	const { rows: applicationFormFields } = await db.sql<ApplicationFormField>(
+		'applicationFormFields.selectByApplicationFormId',
+		{ applicationFormId },
+	);
+	return baseFieldIds.map((baseFieldId) => {
+		const applicationFormField = applicationFormFields.find(
+			(applicationFormFieldCandidate) =>
+				applicationFormFieldCandidate.baseFieldId === baseFieldId,
+		);
+		if (applicationFormField === undefined) {
+			throw new Error(
+				`There is no application form field with baseFieldId "${baseFieldId}"`,
+			);
+		}
+		return applicationFormField;
 	});
 };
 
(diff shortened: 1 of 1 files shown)


src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -411,6 +431,20 @@ describe('processBulkUpload', () => {
       createdAt: expect.any(Date) as Date,
     });
 
+    const [
+      proposalSubmitterEmailFormField,
+      organizationNameFormField,
+    ] = await getApplicationFormFieldsByBaseFieldIds(
+      1,
+      [1, 2],
```
slifty, line 439: Just to call out: these ID's are based on the *test's* insertion, which is deterministic.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -402,6 +438,18 @@ describe('processBulkUpload', () => {
 			createdAt: expect.any(Date) as Date,
 		});
 
+		const [proposalSubmitterEmailFormField, organizationNameFormField] =
+			await getApplicationFormFieldsByBaseFieldIds(1, [
```
reefdog, line 442: (Essentially transcribing our pair review…) Here you're referencing the `ApplicationForm`'s ID statically. Can you (1) test that the `ApplicationForm` was actually created, and (2) reference its ID instead of statically including it? Like you just did with your `createTestBaseFields()` function.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -53,19 +55,30 @@ const createTestBulkUpload = async (
...
+		'baseFields.insertOne',
+		{
+			label: 'Organization Name',
+			description: 'The name of the applying organization.',
+			shortCode: 'organization_name',
+			dataType: 'string',
+		},
+	);
+	return [
+		proposalSubmitterEmailBaseFieldResult.rows[0],
```
reefdog, line 79: Surprised TypeScript lets you get away with this! (I say, knowing it doesn't because you just told me.)

reefdog (changes requested):
This is the best code I've ever seen. Please modify.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -335,7 +371,8 @@ describe('processBulkUpload', () => {
 	});
 
 	it('should download, process, and resolve the bulk upload if the sourceKey is accessible and contains a valid CSV bulk upload', async () => {
-		await createTestBaseFields();
+		const [proposalSubmitterEmailFBaseField, organizationNameBaseField] =
```
reefdog, line 374: Typo here (`proposalSubmitterEmailFBaseField`).

reefdog, line 374: Typo in the variable name.

```suggestion
		const [proposalSubmitterEmailBaseField, organizationNameBaseField] =
```

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: Turns out it actually wasn't angry there -- it was angry later when I tried to *use* the return values.

I fixed it to guarantee that exactly two values are returned (which, of course, does bump up the check requirement to taking place here)

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -350,6 +387,15 @@ describe('processBulkUpload', () => {
 			getMockJobHelpers(),
 		);
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
+
+		const { rows: applicationForms } = await db.sql<ApplicationForm>(
+			'applicationForms.selectAll',
+		);
+		const [applicationForm] = applicationForms;
```
reefdog, line 394: You should be able to destructure directly to `applicationForm` like so, without needing the second assignment:

```ts
const { rows: [applicationForm] } = await db.sql<ApplicationForm>(
	'applicationForms.selectAll',
);
```

At least, that lints just fine on my end, do you want to give that a try?

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: Fixed up!

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -350,6 +387,19 @@ describe('processBulkUpload', () => {
...
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
+
+		const {
+			rows: [applicationForm],
+		} = await db.sql<ApplicationForm>('applicationForms.selectAll');
+		if (applicationForm === undefined) {
+			fail('The application form was not created');
+		}
+		expect(applicationForm).toMatchObject({
+			opportunityId: 1,
```
reefdog, line 398: @slifty Did you add a static reference here just to keep this review fun going?

reefdog: JKJKJK I think

slifty: ugh (there are other places that reference opportunity ID directly too, and applicant ID.)

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -350,6 +387,19 @@ describe('processBulkUpload', () => {
 			getMockJobHelpers(),
 		);
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
+
+		const {
+			rows: [applicationForm],
```
reefdog, line 392: @slifty Now that you've made this assignment, is it worthwhile (in this PR) to convert other references to the static ID, e.g. [here](https://github.com/PhilanthropyDataCommons/service/blob/0b2e209fada5f4ccb953a9f3e8cbee9c7910ed40/src/tasks/__tests__/processBulkUpload.int.test.ts#L443) and [here](https://github.com/PhilanthropyDataCommons/service/blob/0b2e209fada5f4ccb953a9f3e8cbee9c7910ed40/src/tasks/__tests__/processBulkUpload.int.test.ts#L449), to reference it? Truly a non-blocking question.

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: (fixed both examples)

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: Why yes, yes it is.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -53,19 +58,30 @@ const createTestBulkUpload = async (
 	return bulkUpload;
 };
 
-const createTestBaseFields = async (): Promise<void> => {
-	await db.sql<BulkUpload>('baseFields.insertOne', {
+const createTestBaseFields = async (): Promise<[BaseField, BaseField]> => {
```
slifty, line 61: Fun Fact: it seems this is the way you say "it is returning an array of size 2" in typescript!

reefdog (approved):
Thanks for iterating on this!

PR #713, opened 2024-01-11
Add file size to bulk uploads

### LINKED ISSUE #697: Report BulkUpload file size via API
The front end would like to report to users the size of the file they just uploaded. (See designs and variations of https://github.com/PhilanthropyDataCommons/front-end/issues/413.) Ideally this would be reported in bytes, which we would humanize into the desired format.

slifty:
This PR adds the `fileSize` attribute to our `BulkUpload` entity.

It also updates the bulk upload processing task to actually populate the value.

Resolves #697

### DIFF
--- a/src/database/migrations/0016-alter-bulk_uploads-file_size.sql
+++ b/src/database/migrations/0016-alter-bulk_uploads-file_size.sql
@@ -1,0 +1,5 @@
+ALTER TABLE bulk_uploads
+  ADD COLUMN file_size INTEGER;
+
+COMMENT ON COLUMN bulk_uploads.file_size IS
+  'The file size, in bytes, of a bulk upload CSV.';
--- a/src/database/queries/bulkUploads/selectById.sql
+++ b/src/database/queries/bulkUploads/selectById.sql
@@ -3,6 +3,7 @@ SELECT
   file_name as "fileName",
   source_key as "sourceKey",
   status as "status",
+  file_size as "fileSize",
   created_at AS "createdAt"
 FROM bulk_uploads
 WHERE id = :id;
--- a/src/database/queries/bulkUploads/selectWithPagination.sql
+++ b/src/database/queries/bulkUploads/selectWithPagination.sql
@@ -3,6 +3,7 @@ SELECT
   file_name as "fileName",
   source_key as "sourceKey",
   status as "status",
+  file_size as "fileSize",
   created_at AS "createdAt"
 FROM bulk_uploads
 ORDER BY id DESC
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -95,6 +95,14 @@
 						"example": "upload.csv",
 						"pattern": "^.+\\.csv$"
 					},
+					"fileSize": {
+						"description": "File size in bytes",
+						"type": "integer",
+						"nullable": true,
+						"readOnly": true,
+						"example": 42,
+						"minimum": 0
+					},
 					"sourceKey": {
 						"type": "string",
 						"example": "550e8400-e29b-41d4-a716-446655440000",
--- a/src/tasks/__tests__/processBulkUpload.int.test.ts
+++ b/src/tasks/__tests__/processBulkUpload.int.test.ts
@@ -222,7 +222,7 @@ describe('processBulkUpload', () => {
 		expect(requests.deleteRequest.isDone()).toEqual(true);
 	});
 
-	it('should set the status of the upload to FAILED if the sourceKey is not accessible', async () => {
+	it('should fail if the sourceKey is not accessible', async () => {
 		const sourceKey = TEST_UNPROCESSED_SOURCE_KEY;
 		const bulkUpload = await createTestBulkUpload({ sourceKey });
 		const sourceRequest = nock(await getS3Endpoint())
@@ -236,11 +236,14 @@ describe('processBulkUpload', () => {
 		);
 
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
-		expect(updatedBulkUpload.status).toEqual(BulkUploadStatus.FAILED);
+		expect(updatedBulkUpload).toMatchObject({
+			status: BulkUploadStatus.FAILED,
+			fileSize: null,
+		});
 		expect(sourceRequest.isDone()).toEqual(true);
 	});
 
-	it('should not process, and set the status of the upload to FAILED, if the sourceKey is not in the unprocessed namespace', async () => {
+	it('should not process, and fail, if the sourceKey is not in the unprocessed namespace', async () => {
 		const sourceKey = TEST_BULK_UPLOAD_SOURCE_KEY;
 		const bulkUpload = await createTestBulkUpload({ sourceKey });
 		const requests = await mockS3ResponsesForBulkUploadProcessing(
@@ -254,7 +257,10 @@ describe('processBulkUpload', () => {
 		);
 
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
-		expect(updatedBulkUpload.status).toEqual(BulkUploadStatus.FAILED);
+		expect(updatedBulkUpload).toMatchObject({
+			status: BulkUploadStatus.FAILED,
+			fileSize: null,
+		});
 		expect(requests.getRequest.isDone()).toEqual(false);
 	});
 
@@ -279,7 +285,7 @@ describe('processBulkUpload', () => {
 		expect(requests.getRequest.isDone()).toEqual(false);
 	});
 
-	it('should set the status of the upload to FAILED if the csv does not have a proposal_submitter_email field', async () => {
+	it('should fail if the csv does not have a proposal_submitter_email field', async () => {
 		await createTestBaseFields();
 		const sourceKey = TEST_UNPROCESSED_SOURCE_KEY;
 		const bulkUpload = await createTestBulkUpload({ sourceKey });
@@ -295,7 +301,10 @@ describe('processBulkUpload', () => {
 			getMockJobHelpers(),
 		);
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
-		expect(updatedBulkUpload.status).toEqual(BulkUploadStatus.FAILED);
+		expect(updatedBulkUpload).toMatchObject({
+			status: BulkUploadStatus.FAILED,
+			fileSize: 36,
+		});
 	});
 
 	it('should move the csv file to processed location if the csv does not have a proposal_submitter_email field', async () => {
@@ -317,7 +326,7 @@ describe('processBulkUpload', () => {
 		expect(requests.deleteRequest.isDone()).toEqual(true);
 	});
 
-	it('should set the status of the upload to FAILED if the csv contains an invalid short code', async () => {
+	it('should fail if the csv contains an invalid short code', async () => {
 		await createTestBaseFields();
 		const sourceKey = TEST_UNPROCESSED_SOURCE_KEY;
 		const bulkUpload = await createTestBulkUpload({ sourceKey });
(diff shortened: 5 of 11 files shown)


src/openapi.json
```diff
@@ -95,6 +95,12 @@
 						"example": "upload.csv",
 						"pattern": "^.+\\.csv$"
 					},
+					"fileSize": {
```
slifty, line 98: I don't think we document fields in our swagger spec but this one might be worth adding a clarification that the file size is in bytes.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -279,7 +282,7 @@ describe('processBulkUpload', () => {
 		expect(requests.getRequest.isDone()).toEqual(false);
 	});
 
-	it('should set the status of the upload to FAILED if the csv does not have a proposal_submitter_email field', async () => {
+	it('should have a correct failed state if the csv does not have a proposal_submitter_email field', async () => {
```
slifty, line 285: It seemed somewhat silly to add an entirely separate test to check that the file size was populated alongisde the fail status... so I just changed the scope of the test.  I'm not sure if I'm violating some kind of testing best practice here though.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -236,7 +236,10 @@ describe('processBulkUpload', () => {
 		);
 
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
-		expect(updatedBulkUpload.status).toEqual(BulkUploadStatus.FAILED);
+		expect(updatedBulkUpload).toMatchObject({
+			status: BulkUploadStatus.FAILED,
+			fileSize: null,
```
slifty, line 241: Ugh, null -- so null is what is being returned by the pg library.  I wish it was `undefined` rather than `null` but... well... so it goes I guess unless someone else has better thoughts

src/openapi.json
reefdog: I do think that would be a good idea! Looks like we can add the `description` attribute, although its discoverability is pretty poor.

https://github.com/PhilanthropyDataCommons/service/assets/4731/a7802531-9789-4363-8903-536c5815f741

src/database/queries/bulkUploads/selectWithPagination.sql
```diff
@@ -3,6 +3,7 @@ SELECT
   file_name as "fileName",
   source_key as "sourceKey",
   status as "status",
+	file_size as "fileSize",
```
reefdog, line 6: It looks like this is tab-indented while the surrounding SQL is space-indented.
```suggestion
  file_size as "fileSize",
```

src/database/migrations/0016-alter-bulk_uploads-file_size.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE bulk_uploads
+  ADD COLUMN file_size INTEGER;
+
+COMMENT ON COLUMN bulk_uploads.source_key IS
```
reefdog, line 4: 🍝 Copypasta alert! 🍝

```suggestion
COMMENT ON COLUMN bulk_uploads.file_size IS
```

src/openapi.json
```diff
@@ -95,6 +95,12 @@
...
+						"type": "integer",
+						"readOnly": true,
+						"example": 42,
+						"minimum": 0
```
reefdog, line 102: Once again I am disappointed that we are short-sightedly excluding dark matter.

src/tasks/__tests__/processBulkUpload.int.test.ts
reefdog: As I understand it, we can't add a `NULL` constraint to the database since bulk uploads are created before the `file_size` field is calculated/populated. And so `null` is the correct value to return (it's what's in the database, after all). I think it's fine and have no concerns about this.

src/database/migrations/0016-alter-bulk_uploads-file_size.sql
```diff
@@ -0,0 +1,5 @@
...
+  'The file size, in bytes, of a bulk upload csv.';
```
reefdog, line 5:
```suggestion
  'The file size, in bytes, of a bulk upload CSV.';
```

src/database/queries/bulkUploads/selectById.sql
```diff
@@ -3,6 +3,7 @@ SELECT
   file_name as "fileName",
   source_key as "sourceKey",
   status as "status",
+	file_size as "fileSize",
```
reefdog, line 6: It looks like this is tab-indented while the surrounding SQL is space-indented.
```suggestion
  file_size as "fileSize",
```

src/tasks/__tests__/processBulkUpload.int.test.ts
reefdog: I'm okay with broadening the test to include this! However…

1. [L225](https://github.com/PhilanthropyDataCommons/service/blob/582b298e9e9d49ed3ffef21daab922f0ab79283e/src/tasks/__tests__/processBulkUpload.int.test.ts#L225) would also need this change.
2. "should have a correct failed state" is verbose. I think it's okay to just say "should fail". (I appreciate the explicitness of the former, and the mild confusion about "fail" in a testing context. I still think it reads more clearly, and especially avoids the "positive-negative" whiplash of "correct failed state".) This suggestion applies for all the tests that are worded this way, including the (unintentional?) variations such as ["proper failed state"](https://github.com/PhilanthropyDataCommons/service/blob/582b298e9e9d49ed3ffef21daab922f0ab79283e/src/tasks/__tests__/processBulkUpload.int.test.ts#L365).

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -295,7 +298,10 @@ describe('processBulkUpload', () => {
 			getMockJobHelpers(),
 		);
 		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
-		expect(updatedBulkUpload.status).toEqual(BulkUploadStatus.FAILED);
+		expect(updatedBulkUpload).toMatchObject({
+			status: BulkUploadStatus.FAILED,
+			fileSize: 93,
```
reefdog, line 303: I'll just leave one comment, but it affects multiple tests: `93` isn't the correct value for all the test CSVs. Need to update them to match the correct expected value. (I assume this is 🍝copypasta🍝 from the `validCsvTemplate.csv` test.)

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -369,7 +378,31 @@ describe('processBulkUpload', () => {
...
+		await processBulkUpload(
+			{
+				bulkUploadId: bulkUpload.id,
+			},
+			getMockJobHelpers(),
+		);
+		const updatedBulkUpload = await loadBulkUpload(bulkUpload.id);
+		expect(updatedBulkUpload).toMatchObject({
+			fileSize: 93,
+		});
```
reefdog, line 405: I wonder if we should be testing that the file size actually changed from a prior value? In other words, are we sufficiently covered to know "after processing, it is populated with the expected value", or do we also want to know that it's the processing, specifically, which is populating it?

If so, we should verify that it's `null` before processing, or at least verify it changed.

What do you think?

src/tasks/processBulkUpload.ts
```diff
@@ -430,6 +446,17 @@ export const processBulkUpload = async (
 		bulkUploadHasFailed = true;
 	}
 
+	try {
+		const fileStats = await fs.promises.stat(bulkUploadFile.path);
+		const fileSize = fileStats.size;
+		await updateBulkUploadFileSize(bulkUpload.id, fileSize);
+	} catch (err) {
+		helpers.logger.warn(
+			`Unable to upate the fileSize for bulkUpload ${bulkUpload.id}`,
```
reefdog, line 455:
```suggestion
			`Unable to update the fileSize for bulkUpload ${bulkUpload.id}`,
```

reefdog (changes requested):
Wheee!

I have a bunch of inline changes requested. There are also quite a few failing tests; most are due to the expected file size not matching the actual file being tested (probable copypasta!), but there are also failing integration tests from `bulkUpload.int.test.ts` that I think are because the fixtures need updating.

Happy to take a look again once you're ready!

src/openapi.json
reefdog: Immediately resolving since this is a joke.

src/database/queries/bulkUploads/selectWithPagination.sql
slifty: Ohhhhhhhhhhhhhh prettier did not update our .sql files apparently...

We face a choice: should we use spaces in sql and tabs everywhere else (in which case I'd add a line to the editorconfig) or should I update all the sql files to use tabs.

For now I am leaning towards having spaces in the sql files and we can always normalize it later if we want.

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: Gut feeling is that I generally think it's OK to not verify that our test setup was properly setup *so long as it's easy to see what the set up is by viewing the test*.  For instance for tests of `update` endpoints I wouldn't personally want an `expect(before).toBe({thethingwejustenetered})` since that feels redundant...

But reasonable people might disagree!

All that said, this is a weird case because we don't pass `null` in, rather, we don't pass *anything* in when creating the bulk upload, since you cannot create a bulk upload with a non-null value...

I'll sleep on it, but for now am inclined to think that the fact that it's impossible to insert a file size when creating a bulk upload is good enough reason to not verify starting state explicitly.

reefdog: > you cannot create a bulk upload with a non-null value

Oh! I didn't know this.

I'm fine going either way on this, follow your thinking.

That said, if you're trying to test that `loadBulkUpload` actually *does* something, then I do think the test needs to verify that it did something, which would mean monitoring its before and after state. Otherwise it could just be no-op'ing. Even though _you know_ it can't be, that relies on understanding the internals of the `create` setup. It feels a bit too implicit to me? But genuinely think I'm splitting hairs here, I'm comfortable either way.

src/tasks/__tests__/processBulkUpload.int.test.ts
slifty: So as an example of why null is the worst, it looks like maybe null gets turned into 0 when serializing.

I'm inclined to just call that acceptable behavior for now.  The API will return `0` in cases where file size is null.

reefdog: wat why, who does that, who's the culprit

slifty: Ok honestly this is a good question and I'm gonna investigate as part of the PR!  Marking the rest for re-review at your leisure though.

slifty: found out: it's ajv -- https://ajv.js.org/coercion.html

I'm going to look into whether I can have it leave null alone.

slifty: FWIW this is what I was thinking about regarding not wanting to use null: https://github.com/microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

slifty: Fixed, making schema "nullable: true" -- though now the type has to be optional? which is somewhat weird (since my undersatnding is `?` allows *undefined* rather than null).

slifty:
rebased and ready for re-review @reefdog

reefdog (approved):
I'm going to approve this because you fixed all my little issues + the tests run clean + I can _mostly_ ensure it works locally, and my remaining issues running it locally seem to be outside the scope of this PR review.

Which is to say, great, thank you!

Issue #640, opened 2023-11-29
Remove dotenv


jasonaowen:
Currently we're using [dotenv](https://www.npmjs.com/package/dotenv) to load environment variables from a `.env` file. This is primarily beneficial to developers, as in production we use actual environment variables.

Unfortunately, it has some complications (#59):

https://github.com/PhilanthropyDataCommons/service/blob/b4edf0d8e97b1ca172033cb7b0f8c06d7667f4bd/src/logger.ts#L7-L12

As discussed in https://github.com/PhilanthropyDataCommons/service/pull/635#discussion_r1408556444, I am increasingly of the opinion that such convenience support for developers should not live in the repo, but rather be the responsibility of the devs, via [direnv](https://direnv.net/), sourcing a `.env` file in the shell, or another way of managing environment variables. In that conversation, @slifty agreed, and we didn't want that to be forgotten despite being out-of-scope for that PR.

slifty:
I'm gonna work on this now since it's very slightly peripheral to #568 and thinking about test logging settings will be simpler if I'm not also thinking about when / where / how dotenv should be invoked.

PR #715, opened 2024-01-12
Remove dotenv from `package.json` and `README.md`

bickelj:
The `README.md` continues the pattern of using `.env` files, though this is one way of managing environment variables among many.

Issue #640


reefdog (changes requested):
Good catch on the npm scripts!

And you're right that we need to add more guidance to the README; however, I don't think we should add these specific `source` prefixing to each command example. Different devs will tackle this in different ways, e.g., using direnv to auto-load env vars. This falls into the subjective dev category of which Node version manager to use.

Instead, we need to expand the [Setup](https://github.com/PhilanthropyDataCommons/service#setup) section with a note about each dev needing to make sure they're sourced using whatever method they want, with a few options (direnv; manual `set -a; source .env` before each `npm run` command; etc.)

bickelj:
> And you're right that we need to add more guidance to the README; however, I don't think we should add these specific `source` prefixing to each command example. Different devs will tackle this in different ways, e.g., using direnv to auto-load env vars. This falls into the subjective dev category of which Node version manager to use.
> 
> Instead, we need to expand the [Setup](https://github.com/PhilanthropyDataCommons/service#setup) section with a note about each dev needing to make sure they're sourced using whatever method they want, with a few options (direnv; manual `set -a; source .env` before each `npm run` command; etc.)

Yes, fair enough.

I don't know anything about `direnv` so I couldn't speak to that in the README. But I agree that it's simpler to make a note to "make sure to set environment variables" and then de-clutter the rest. I have pushed a change to this effect.

reefdog (approved):
LGTM!

I wouldn't be opposed to actually pointing people to the [direnv](https://github.com/direnv/direnv) project as an example of what [some of us] use, or otherwise provide another recommendation so people don't have to run `set`/`source` before every dev session. But it looks like we've avoided pointing people to any such tooling thus far (like under the [Node version](https://github.com/PhilanthropyDataCommons/service#node-version) section), so this is consistent with history.

Thanks @bickelj!

slifty (approved):
Thank you for catching this and improving the documentation!

slifty:
I don't remember what @bickelj's preference is but given that this is patching a breaking change I'm going to merge it since it's urgent.

Issue #497, opened 2023-09-18
Integrate SDK generation into build process


slifty:

slifty:
**Question 1: How to keep our version of swagger-codegen up to date**

It would be Good for dependabot to manage the version of swagger-codegen for us / open PRs for new releases.

Dependabot supports [Maven](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#maven) (`pom.xml`) and [Gradle](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#gradle) (`build.gradle`).

I believe Maven is the right tool (vs gradle).  That said (and someone tell me if I'm wrong) but it doesn't look like maven is intended to be used to download jars that are going to be used in external scripts; rather, the "Maven" way forward would be to write a maven plugin or java application that is responsible for building the sdk's, and so sdk generation would look something like `maven clean compile`.

There is a Maven plugin to this end, but unfortunately it only supports v2 of `swagger-codegen` (and therefore only v2 of the swagger spec, we are on v3).

I'm going to also look into whether or not we can just use Maven to cleanly download the jars specified as dependencies and then use those jars in a shell-based sdk build script.

If we don't want maven as a dev dependency and are comfortable being weird, I could write a script that pulls the latest version from `pom.xml` (so the version is managed by dependabot) and downloads a jar via curl (though we'd lose things like checksums / etc that we get for free from Maven)

slifty:
Worth noting: we need to be able to run build scripts locally.

This does mean that either (A) Java is going to become a development environment dependency for `service` or (B) Docker is going to become a development environment dependency for `service`.

I don't think we want Docker as the dependency though I'm open to being told that's not true!

So at the very least we'll need a JRE as a dev dependency; we may also need Maven as a dev dependency.  Stay tuned!

slifty:
Gradle works just fine, but also it turns out I was wrong about the maven plugin only supporting v2 -- it just moved over here: https://mvnrepository.com/artifact/io.swagger.codegen.v3

Which is to say, we could:

1. Use the maven plugin, which means configuration would be in a `pom.xml` file and building would involve `maven clean build`
2. Use gradle to download the jar, and use a shell script to run the jar.

I like the idea of leaning on the maven plugin (I'll see if there's a gradle equivalent too), rather than re-implementing similar things in bash.

slifty:
[...]

PR #755, opened 2024-02-02
Add sdk build to ci

slifty:
This PR creates a "first step" github action that will use our codegen templates to generate and publish a typescript SDK whenever versions change.

Related to #497

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -22,12 +22,19 @@ jobs:
           path: 'sdk'
       - name: Copy API definition
         run: cp service/src/openapi.json sdk/openapi.json
-      - name: Invoke the typescript generator
+      - name: Invoke the TypeScript generator
         working-directory: sdk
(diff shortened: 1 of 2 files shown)


.github/workflows/sdk.yml
```diff
@@ -0,0 +1,37 @@
+name: sdk
+
+on:
+  workflow_dispatch:
+  pull_request:
```
slifty, line 5: This is *only* here so we can test the workflow as part of this PR.

We need to remove it before we merge, otherwise we'll be accidentally releasing versions on PR instead of after approval.

slifty: I realized that, actually, keeping this will be good so long as we make sure it's a `dry-run`

In fact, once merged I think we should make this a required action, which means that if a PR breaks the SDK build it would flag an issue.

reefdog (approved):
This all makes sense and is very cool! I left one non-blocking note.

PR #773, opened 2024-02-09
Fix the build (again)

slifty:
This PR fixes an issue where the conditional... didn't actually condition!

Unlike #772 I've actually deleted the secret so we can be sure that if the CI passes here it actually fixes the problem.


slifty:
It turns out the issue is that the environment variable was not a boolean, but a string that contained the literal word `true`.

Updating the conditional to check the value explicitly seems to be working.

reefdog:
Haaaahahahaha, I should've thought about that! I've run into this in the past. (Not in the CI context, though.)

reefdog (approved):

PR #789, opened 2024-02-23
Rename `dev` scripts

### LINKED ISSUE #788: Remove `dev` versions of commands
A while back @jasonaowen pointed out that we don't use `npm run` in production, so I don't think we need to have `:dev` versions of `start` and `migrate`.

I do think it would still be useful to have our `ci` confirm that the built versions of things run properly, so we might just want to rename `:dev` to no prefix, and no-prefix to `:ci`.

slifty:
This PR removes the "production" versions of `start` and `migrate` as they are not actually being used in production.

Since those are gone we can now remove the `dev` prefix.

Resolves #788

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ source .env
 3. Run migrations
 
 ```bash
-npm run migrate:dev
+npm run migrate
 ```
 
 ### Common Commands
@@ -109,7 +109,7 @@ npm start
 To start the server in a development environment:
 
 ```bash
-npm run start:dev
+npm run start
 ```
 
 ### Logging
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 
-npm run migrate
-exec npm run start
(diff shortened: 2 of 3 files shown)


slifty:
@bickelj My understanding is correct that the docker containers don't rely on `npm run start` to operate, right?

reefdog (approved):
This is a sensible change and I was able to locally run `npm run migrate` and `npm run start` (and `npm start`) successfully. Thanks!

I'm not able to test its effect on the production build/deploy, but I trust you on that (and we'll discover any issues eventually).

One question: #788 says:

> I do think it would still be useful to have our `ci` confirm that the built versions of things run properly, so we might just want to rename `:dev` to no prefix, and no-prefix to `:ci`.

Did you change your mind on this?

slifty:
Oh right yes I did -- in that we weren't doing that kind of check in the first place (when I wrote that I was under the false impression we were doing `build` steps in CI but we weren't)

Also I admit I don't know for *sure* how this impacts deployment and like you I don't know entirely the best way to test it out.

Issue #777, opened 2024-02-14
Allow access to a new `/organizations` endpoint


slifty:
This new endpoint would make it possible to access organization data as both a public user AND a logged in user (aka superuser at this point).

In order to implement it we'll need to add in a few concepts to the schema:

1. The concept of an organization (this is not a trivial concept to get right, and is essentially a replacement of the concept of an applicant see #730).

2. The concept of scope to base fields (organization vs proposal for now)

3. The concept of "public" base fields

Once this is all done, it'll be possible to return a list of organizations along with any "organization" base field (narrowed to include / exclude non-public data depending on the user's auth)

This issue also replaces #767

slifty:
I can already tell this issue is gonna be bigger than I thought when opening it BUT *cracks knuckles* let's dive in...

We currently have the concept of an applicant, which we now know is both too specific and too broad.

Here's what I mean:

* Applicants cannot be individuals at this stage of the PDC.  We're only collecting organization data.
* Organizations might not be applicants!  For instance, maybe we have an organization in the PDC that came from a third party data provider.
* Our ERD had a baked-in assumption that there will be exactly one applicant for a given proposal.  In reality there might be more than one applicant.

There is also the question of "what data defines an organization" -- to put this in data modeling terms, what would a natural primary key for an organization be?  We know it is *not* just the EIN since that does not account for complex hierarchies (e.g. the Red Cross) which might have multiple sub-organizations within one parent organization which independently apply for grants.

Let's take this in pieces...

## Question 1: Applicant vs Organization

I think we already have consensus on this: drop applicant, create organization.  We would also remove `applicantId` from the proposal table and this in turn would allow us to update bulk upload to no longer require any particular data about applicants (e.g. applicant email address).

## Question 2: Organization "natural identifiers"

From our discussion so far I think there are three possible attributes to a natural identifier:

1. EIN (Employee Identification Number)
2. Organization Name (This *could* actually be split into two fields of "Organization name" and "branch name" where branch name is null most of the time)
3. Organization Address (which is more than one field but that's not important here...)

I think the cleanest natural key would be EIN + Org Name + Branch Name, however I also don't think the concept of "branch name" exists anywhere.  At the very least it would not consistently exist in application forms.

For that reason I'd like to lean in on EIN + Organization address.

IN THE SHORT TERM this would mean that if an organization changes their address, they would get two separate organizations within the PDC.  That's *not* ideal, but would be a natural consequence.

[...]

PR #794, opened 2024-02-26
Add basic `/organizations` endpoints

slifty:
This PR adds a new `Organization` entity type along with the basic create / read endpoints for it.

This PR does *not* establish relationships between organizations and proposals, or organizations and non-proposal data.  It also does not yet create a public / non-authenticated user view of organization data.

Related to #777


src/openapi.json
```diff
@@ -127,6 +127,30 @@
...
+						"type": "string",
+						"example": "Example Inc."
+					},
+					"createdAt": {
+						"type": "string",
+						"format": "date-time",
+						"readOnly": true
+					}
+				},
+				"required": ["id", "externalId", "optedIn", "createdAt"]
```
reefdog, line 152: I think this list is inaccurate? Probably copypasta from Applicant!

src/types/Organization.ts
```diff
@@ -0,0 +1,32 @@
...
+		},
+		employerIdentificationNumber: {
+			type: 'string',
+		},
+		name: {
+			type: 'string',
+		},
+		createdAt: {
+			type: 'object',
+			required: [],
```
reefdog, line 28: Just for my own edification, why is this an array?

src/__tests__/organizations.int.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+const agent = request.agent(app);
+
+const insertTestOrganizations = async () => {
+	await createOrganization({
+		employerIdentificationNumber: '11-1111111',
+		name: 'Example Inc.',
+	});
+	await createOrganization({
+		employerIdentificationNumber: '22-222222',
```
reefdog, line 16: While we don't have format validation, it appears this is meant to mimic a real EIN, which means it should be nine digits. (I'm pretty confident you intended this, and e.g. `11-1111111` conforms properly.)

Only flagging it once but I see it several times:

```suggestion
		employerIdentificationNumber: '22-2222222',
```

src/__tests__/organizations.int.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+	describe('POST /', () => {
+		it('creates exactly one organization', async () => {
+			const before = await loadTableMetrics('organizations');
+			const result = await agent
+				.post('/organizations')
+				.type('application/json')
+				.set(authHeader)
+				.send({
+					employerIdentificationNumber: '11-111111',
```
reefdog, line 148: Only flagging this once, but I see it several times. (But not always! There are correct nine-digit versions as well.)

```suggestion
					employerIdentificationNumber: '11-1111111',
```

src/openapi.json
```diff
@@ -127,6 +127,30 @@
...
+				"type": "object",
+				"properties": {
+					"id": {
+						"type": "integer",
+						"readOnly": true,
+						"example": 3407
+					},
+					"employerIdentificationNumber": {
+						"type": "string",
+						"example": "111-11-1111"
```
reefdog, line 140: Should be consistent with other examples (and expected EIN format).

```suggestion
						"example": "11-1111111"
```

reefdog (commented):
For the most part, this is great! I found an errant bit of copypasta and some malformed sample EINs. The only significant question I have is why there isn't a POST example in the Swagger spec? Was this an oversight? It stymied me!

reefdog (changes requested):
Bah, I meant to request changes with my review just then, but clicked the wrong radio button. You don't get off that easy!

slifty:
@reefdog thanks for the review!  FWIW, and maybe this is an anti-pattern, but I'll often mark a PR as approved with some requested changes in the event that the requested changes are minor enough that I don't think I'll need to re-review (e.g. fixing a typo etc).

src/openapi.json
slifty: Phewf thanks -- a little disappointed the linter didn't catch that!

reefdog:
> FWIW, and maybe this is an anti-pattern, but I'll often mark a PR as approved with some requested changes in the event that the requested changes are minor enough that I don't think I'll need to re-review (e.g. fixing a typo etc).

I've noticed that, and I **do** think that's an anti-pattern, and that you should be comfortable leaving a "request changes" review. It's not a failing grade, it's a communication signal: "In my opinion as reviewer, the PR shouldn't merge to `main` until some changes are made." I reserve "approve" for when there are no changes requested, or when they are subjective enough that I don't mind the dev discarding them.

I do wish we could signal "if you make these necessary changes, approval will be automatic and re-review isn't necessary" but we can state that in the comment. (I do grok that there's still a cost here, as the reviewer now has to come back and re-review, but I think that's an acceptable cost compared to losing the clarity of approve v. request changes).

Have I convinced you¹??

¹ that you not longer want to work with me and regret ever introducing yourself at SRCCON

reefdog (changes requested):
I have one minimal request — looks like your EIN sample fix got fixed TOO hard! — but also still don't see a POST example in the generated Swagger spec. Shouldn't there be one?

Organization Swagger spec

reefdog:
Oh my goodness @slifty I just realized you hadn't yet requested a re-review from me. 🤦 I'm sorry for jumping the gun with the above and hope that wasn't frustrating! I'll hold tight until you're ready.

If it helps I'm just really excited about this functionality.

src/__tests__/organizations.int.test.ts
```diff
@@ -0,0 +1,217 @@
...
+import { createOrganization, db, loadTableMetrics } from '../database';
+import { expectTimestamp } from '../test/utils';
+import { mockJwt as authHeader } from '../test/mockJwt';
+import { PostgresErrorCode } from '../types';
+
+const agent = request.agent(app);
+
+const insertTestOrganizations = async () => {
+	await createOrganization({
+		employerIdentificationNumber: '11-11111111',
```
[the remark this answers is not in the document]
slifty: bah -- tbh I wonder if it might be better to have data that clearly does not even try to conform, since we're doing no checks whatsoever yet.

slifty:
> I do wish we could signal "if you make these necessary changes, approval will be automatic and re-review isn't necessary" but we can state that in the comment

That's *sometimes* the intention (and to your point someone can press the bright red "merge without formal approval" button I suppose).

That said sometimes the feedback really is meant as "take this or leave it; it's approved as is, but these changes might be worth considering"

Anyway I hear ya and will try to take your advice of not working with you any more.

src/__tests__/organizations.int.test.ts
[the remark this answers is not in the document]
reefdog: Thanks! If we want to demonstrate/test alternative tax IDs, that's cool, we should just call that out explicitly. These are pretty clearly meant to conform to US EIN format — so I appreciate taking the trouble. 🙏

reefdog:
> That said sometimes the feedback really is meant as "take this or leave it; it's approved as is, but these changes might be worth considering"

Totally, in which case I think "Approved" is the right call. The one pattern I'm trying to avoid baking in is using "Approved" when you _do_ have very real and necessary changes that you are requesting, but then to signal re-review isn't necessary after resolution.

> Anyway I hear ya and will try to take your advice of not working with you any more.

You can't take this back just because you said it on a day that doesn't exist. (Checks timestamp.) Darnit, I mean, I read it on a day that doesn't exist.

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.3.1 - 2024-02-29
```
reefdog, line 8: Leap day release!

whoooa

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+- The `Organization` entity now exists.
+- The GET `/organizastions` endpoint now exists.
```
reefdog, line 13: Whoops just noticed this typo.

```suggestion
- The GET `/organizations` endpoint now exists.
```

reefdog (changes requested):
Ironically, given our discussion, I now have to follow through on "request changes" for literally just a typo fix. 😞 This is boding well for never working together again.

That said, I *also* noticed that the Swagger spec now includes a POST example, but… it's under the heading "Organization" instead of "Organizations", while all the other entities bundle GET and POST requests under a single entity header? Can that be fixed?

Organizations Swagger

slifty:
@reefdog hey look ok, you're the one who has to re-review!

Fixed both typos!

reefdog (approved):
WHEEEEEE

reefdog:
@slifty I hit approve, but… shouldn't you actually increment the version in `package.json` (not just changelog and openapi)?

slifty:
@reefdog since the last time we had a version we decided that we're versioning the API spec -- the package version is something of a misnomer and isn't used by anything.  I'd say we should either leave it as is, or open an issue to just set it to 0.0.0 or something that is clearly-not-a-thing.

reefdog:
@slifty That makes sense, thanks for the explanation! No problem, merge as-is and let's open an issue to zero it out (or whatever).

Issue #730, opened 2024-01-23
Evaluate whether "Applicant" is an entity worth having any longer


slifty:
The concept of an "Applicant" was part of the original ERD, but it is not really *used* anywhere.

We've since learned a lot about the complexity of proposal associations / organizations / applicants.  We are going to need *something* to address collections of proposals and ownership, however this probably isn't it.

Let's take it out!

slifty:
That was an easy decision!  Closing.

Issue #622, opened 2023-11-21
Change deploy model


jasonaowen:
Currently, this repo is deployed via the sister repo [PhilanthropyDataCommons/deploy](https://github.com/PhilanthropyDataCommons/deploy) to a DigitalOcean droplet that is running Docker. This has a lot of costs:

- We have semi-automated deployments but not continuous deployment, as there are some manual steps involved. While I believe I have the keys to that, I haven't done it, so while @bickelj has been working on another project, we haven't really deployed. (Thank you for noticing that and deploying last week, @bickelj!)
- There are some personal access tokens involved, which are a (known) bit of a ticking time bomb in that they will someday expire and what automation we currently have will break.
- The current approach prevents us from using the `@dependabot merge` command.
- Configuring environment variables is nontrivial.
- The underlying instance requires ongoing sysadmin support, in the form of system updates and monitoring and so on.
- Deploying changes from this repo invalidates Keycloak sessions (https://github.com/PhilanthropyDataCommons/deploy/issues/86).
- We haven't yet configured database backups, which is our responsibility in this model.

Instead, I would like for us to use a Platform-as-a-Service (PaaS) to deploy this app, and specifically [DigitalOcean App Platform](https://docs.digitalocean.com/products/app-platform/).

- This would give us continuous deployments out of the box: clicking the big green button in a GitHub pull request would cause a new deployment.
- We could build [per-pull-request deploy previews](https://cfu288.com/posts/creating-per-pull-request-environments-on-digital-ocean-with-github-actions/), making it easier to test and review pull requests.
- It shouldn't expire, or be tied to a particular user's account.
- Dependabot should work as expected.
- Environment variables should be configurable via both a web UI in DigitalOcean, and I think via its [`doctl` command-line interface](https://docs.digitalocean.com/reference/doctl/).
- OTS sysadmins will need to do less maintenance.
- Out-of-the-box database backups, including point-in-time recovery, via a managed PostgreSQL instance.

After discussing this with @jvasile, @reefdog, and @slifty today, it sounds like OTS does not expect @bickelj to have bandwidth for this project until next year.

Note that disentangling Keycloak is a separate but related issue (https://github.com/PhilanthropyDataCommons/deploy/issues/94); this approaches it from the other direction.

There are a few steps involved:

- [x] Migrate the existing production database from a Docker container to a new DO-hosted database
- [x] Set up a new DO App for production
- [ ] Update DNS for `api.pdco` to point to the new instance
[...]

PR #798, opened 2024-02-27
Remove `Applicant` and `/applicants/*` endpoints

slifty:
This PR removes the concept of an Applicant from the PDC, along with the endpoints and entity associations.

Related to #777 
Related to #730 

Note this is built off of #794 and shouldn't be reviewed / merged until after that


CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
...
+- The `Applicant` entity no longer exists.
+- All `/applicants/*` endpoints are removed.
+- The `applicantId` attribute of `Proposal` has been removed.
+
+### Added
+
+- The `Organization` entity now exists.
+- The `/organizations` POST and GET endpoint now exists.
+- The `/organizations/{organizationId}` endpoint now exists.
+
```
reefdog, line 21: Could you move the organization stuff to the other PR?

src/database/operations/load/__test__/loadBundle.int.test.ts
```diff
@@ -1,27 +1,29 @@
...
+import { createOrganization, loadBundle } from '../../..';
+import { isOrganization } from '../../../../types';
 
 describe('loadBundle', () => {
 	it('Should return a bundle for the query', async () => {
-		await db.sql('applicants.insertOne', {
-			externalId: '12345',
-			optedIn: true,
+		await createOrganization({
+			employerIdentificationNumber: '11-111111',
```
reefdog, line 7: I believe this is meant to be a nine-digit EIN. (Only flagging once but I see it multiple times.)

```suggestion
			employerIdentificationNumber: '11-1111111',
```

reefdog (changes requested):
Nice! There is a small bit of malformed sample data in the tests.

I also notice that this PR doesn't release the `proposal_submitter_email` requirement on bulk uploads, even though it's no longer used for anything. Totally fine if you want to consider that a follow-up downstream effect, but let me know if that's the case or if it was an oversight.

Should be a quick re-review!

CHANGELOG.md
reefdog: Actually… I bet the reason you saved all of it for this PR was so you could merge the org one in without triggering a release, then release it all as 0.4.0? I wonder if you should just do something like add a `0.4.0 (under development)` to the CHANGELOG (without actually changing the version number in the package) in the org PR, then in this PR remove the `(under development)`, add the applicant items, and do the real version bumping? 🤔

src/tasks/processBulkUpload.ts
```diff
@@ -197,16 +196,6 @@ const assertBulkUploadCsvIsValid = async (csvPath: string): Promise<void> => {
 	await assertCsvContainsRowsOfEqualLength(csvPath);
 };
 
-export const getApplicantEmailFieldIndexForBulkUpload = async (
```
slifty, line 200: See here for removal of the`proposal_submitter_email` stuff @reefdog

CHANGELOG.md
slifty: Meh.  This whole thing is weird and strange and uncomfortable.  I made a commit in the other PR but realized from a git history PoV it wasn't any better (sure, it'd be in the other PR but it was still decoupled from the commit changes themselves).

On the flip side, the changes were made over 3 commits, so I'd have to do three "releases" which is silly...  or use "unreleased"

Like I said, the concept of unreleased is a lie: things are released once they're merged.

Maybe the mistake here is thinking of "releases" at all -- maybe we should just have a change log with dates, and we could put the current swagger spec version next to every change.  For right now I'm going to do the "single commit with a `0.3.1` "release" even though that's silly, but I think going forward we should flatten the changelog and have it be date based with "current API version" as metadata.

reefdog: I'm fine with whatever solution you come to this, tbh! It was a minor complaint and I should've marked it as optional / up to your judgement. That said:

> Like I said, the concept of unreleased is a lie: things are released once they're merged.

Not to be pedantic, but just to make sure I know the current state of things: I think this is the correct way of thinking about it, but technically we neither deploy automatically nor release automatically (unless version number is incremented), right? Or no? (Either way you're right about living by that edict regardless of whether it's the ground truth atm.)

reefdog: I wonder if the way to solve this is _actually_ to merge this PR into #794, and then merge that one in? I know we don't normally do that, but… then we get just a single release to `main`. Is… is this a bad pattern?

slifty: I'd prefer to not do that pattern (and would not want that to become a standard pattern) -- ultimately we need to be able to merge things to main.  Having a "dev" branch isn't what we do (and we don't want one, nor do we want every-branch-becomes-a-dev-branch).

Appreciated that we don't truly have CD right now but that's more of a bug than a feature (and I think #622 is meant to fix it)

slifty:
@reefdog ahh I see it now; ok thanks, that should be fixed.

reefdog (commented):
Does package.json need its version bumped, too?

reefdog (approved):

slifty:
The other PR talked about this too but we'll be taking version out of package.json (#807)

PR #812, opened 2024-03-04
Add authentication set up instructions in readme

### LINKED ISSUE #290: Readme should explain what has to be done to set up authn
The current README instructions don't result in a working local copy of this project -- in particular, the copy I have set up won't allow any API calls due to a lack of auth service without some further setup in my local environment

I believe KeyCloak is our officially recommended tool for the project, so I'd like to suggest we re-visit including some steps for how to set up KeyCloak auth.  We don't want to write keycloak documentation, of course, so I imagine this would mostly take the form of:

1. linking to the appropriate KeyCloak docs
2. indicating what byproducts of the KeyCloak set up should be written to various environment variables
3. indicating any additional specific keycloak setup steps need to be performed (and how the outcomes of those steps map to our API calls)

An important note:  I know that KeyCloak is not intended as our only auth vendor, and we should mention that folks can set up alternative auth if they choose to, but that shouldn't stop us from providing instructions -- and the reality is that KeyCloak *is* our recommended tool for this project.

hminsky2002:
The PDC service repository requires a valid Bearer JSON Web Token in order to make requests. Previously, the README did not thoroughly explain the need for authn or suggest a route to set it up. This PR makes it clear that a valid JWT provider is necessary, explicitly suggests using Keycloack for auth and provides a link to Keycloak set-up documentation, and provides a key-cloak specific example.

Closes #290


README.md
```diff
@@ -122,11 +122,39 @@ LOG_LEVEL=trace npm run test
...
+   - Direct access grants unchecked,
+   - Implicit flow unchecked,
+   - Service accounts roles unchecked,
+   - OAuth 2.0 Device Authorization Grant (optional), and
+   - OIDC CIBA Grant unchecked.
+3. In the settings for the client added in step (2), set the root URL and Home URL to the URL of the PDC service (not the auth service). Add redirect URIs:
+   - `/applicants/*`
+   - `/applications/*`
+   - `/applicationForms/*`
+   - `/canonicalFields/*`
```
slifty, line 150: Hmm -- I'm surprised about how coupled these paths are to our API.  This one in particular refers to an endpoint that doesn't exist any more (as does `/applicants`), and we're missing some new ones (`/organizations` and `/baseFields`)

Do we need these redirect URIs at all?  What happens if we skip them entirely and only use the root URL?  Or just have `/*`?

It would be great to not need to have developers manually update keycloak config every time we add a new API endpoint.

README.md
```diff
@@ -122,11 +122,39 @@ LOG_LEVEL=trace npm run test
...
+   - `/proposals/*`
+   - `/proposalVersions/*`
+4. In the settings for the client added in step (2), add a Web origins of `+`.
+5. With in the realm, add a user, e.g. `test-user`.
+6. Log in as the test user.
```
slifty, line 155: What is the use case in mind for steps 6, 7, and 8?

If the use case is a developer making API calls, they could use the swagger-generated page (e.g. [https://api.philanthropydatacommons.org](https://api.philanthropydatacommons.org)) which handles auth for them.  If the use case is a script or application making API calls then we don't want them to have to manually log in as a human in order to interact with the API, and instead we should provide instructions on creating a new client for their application which gives them a token / secret pair that they can use to avoid an auth flow that requires manual log in.

slifty (commented):
Two questions which might be requests depending on the answers!

README.md
hminsky2002: Agreed! I suppose I was thinking about these paths as generic examples, but that doesn't really make any sense

bickelj:
Aside: how in the world would this reduce code coverage?

slifty:
@bickelj aww now I can't make my final approval joke where I ask him to add test coverage!

Relative project coverage %'s are driven by both the numerator and denominator.  My guess is this added more lines to the repository so the denominator went up (though I am surprised that they count markdown lines in that formula)

README.md
```diff
@@ -122,11 +122,37 @@ LOG_LEVEL=trace npm run test
...
+   - Standard flow unchecked,
+   - Direct access grants unchecked,
+   - Implicit flow unchecked,
+   - Service accounts roles unchecked,
+   - OAuth 2.0 Device Authorization Grant (optional), and
+   - OIDC CIBA Grant unchecked.
+3. In the settings for the client added in step (2), set the root URL and Home URL to the URL of the PDC service (not the auth service). [For development purposes, all callback routes can be authorized](https://www.keycloak.org/docs/23.0.7/authorization_services/#_resource_overview)
+   - `/*`
+4. In the settings for the client added in step (2), add a Web origins of `+`.
+5. With in the realm, add a user, e.g. `test-user`.
```
slifty, line 150:
```suggestion
5. Within the realm, add a user, e.g. `test-user`.
```

README.md
```diff
@@ -122,11 +122,37 @@ LOG_LEVEL=trace npm run test
...
-### Authentication
-
-The service validates a Bearer JSON Web Token (JWT) found in the `Authorization` HTTP header.
-
-See the `.env.example` for relevant variables.
+### Authentication and Authorization
+
+A valid Bearer JSON Web Token (Bearer JWT) is required in requests to the PDC service. The PDC officially reccomends using [KeyCloak](https://www.keycloak.org/) as the authentication provider, but supports any valid Bearer JWT provider. Please refer to [the relevant documentation for setup and configuration](https://www.keycloak.org/guides#getting-started)
+
+The relevant authorization enviroment variables as they appear in the `.env.example` file are:
```
slifty, line 129:
```suggestion
The relevant authorization environment variables as they appear in the `.env.example` file are:
```

README.md
```diff
@@ -122,11 +122,37 @@ LOG_LEVEL=trace npm run test
...
+
+- `AUTH_SERVER_ISSUER`: the URL for the issuer (this is the "realm" URL in Keycloak terminology).
```
slifty, line 131: What do you think about moving this explanation to the `.env.example` file itself as a comment before the relevant env variable?

Also, I generally suggest avoiding the documentation of specific env variables in readme documentation since it means you're documenting in two places (or, NOT documenting in the most important place, which is in the `.env.example`). 

If you agree, then you would remove the list of env vars from here and instead make sure they're in a section in the template, and you could just say something like "see `.env.example` for documentation on the necessary environment variables for Keycloak"

slifty (changes requested):
Nice!

Two typos worth fixin' before merge, and a suggestion to move the env var documentation to the template, for your consideration.

hminsky2002:
> Nice!
> 
> Two typos worth fixin' before merge, and a suggestion to move the env var documentation to the template, for your consideration.

Spelling corrected and suggestion taken! All changes should reflect in the amended commit.

.env.example
```diff
@@ -13,11 +13,14 @@ PGPORT=${PGPORT}
 # To run integration tests, the URL can be completely fabricated, for example
 # 'https://made-up-auth-server/realms/fake' would work for testing.
 # Set `AUTH_SERVER_ISSUER` to the URL of your keycloak realm. This works in dev, test, and prod.
+# Specifically, this is the "realm" URL in Keycloak terminology.
+- `OPENAPI_DOCS_AUTH_CLIENT_ID`:
```
slifty, line 17: I think this might have been accidentally kept over via copy paste

slifty:
> ment yesterday got lost either by me not saving it or timing of force-push. Regardless, the comment is we probably need another client set up besides the one for the `service`: either one that uses the typical Authorization Code flow (browser-like, for humans, e.g. `pdc-openapi-docs`) or one that uses Client Authentication (machine users, e.g. `pdc-ots-ingest`). It may be sufficient for today to merely add a line stating that

@bickelj I noticed you had commented directly on the commit (maybe you clicked the commit hash?) not in the context of the PR, so that's where it went: https://github.com/PhilanthropyDataCommons/service/commit/2e614bf8359a6fb471fb8cff95b0845448584265

hminsky2002:
> pdc-openapi-docs

@bickelj @slifty So I ran through the instructions myself using a keycloak image and I found that I did not need to create a second client, but i DID need to enable standard flow in the first client i made. The updated instructions reflect my successful start-to-finish run through of setting up keycloak, creating a realm, creating a client in that realm with standard flow checked, creating a user, and authorizing through swagger!

README.md
```diff
@@ -122,11 +122,35 @@ LOG_LEVEL=trace npm run test
...
-### Authentication
-
-The service validates a Bearer JSON Web Token (JWT) found in the `Authorization` HTTP header.
-
-See the `.env.example` for relevant variables.
+### Authentication and Authorization
+
+A valid Bearer JSON Web Token (Bearer JWT) is required in requests to the PDC service. The PDC officially reccomends using [KeyCloak](https://www.keycloak.org/) as the authentication provider, but supports any valid Bearer JWT provider. Please refer to [the relevant documentation for setup and configuration](https://www.keycloak.org/guides#getting-started)
+
+See `.env.example` for documentation on the necessary environment variables for Keycloak(or chosen authentication provider)
```
bickelj, line 129: I see a minor typo, missing space.

README.md
```diff
@@ -122,11 +122,35 @@ LOG_LEVEL=trace npm run test
...
+
+#### An example with a Keycloak authentication and Swagger-UI
+
+We use Swagger to generate an interactive api interface as the default route for the service.
+From the Keycloak admin interface, e.g. https://my-host/admin:
+
+1. Add a realm e.g. "pdc" (avoid spaces in the name).
+2. Within the realm just added in step (1) add a client e.g. "pdc-service", with Client authentication off and Authorization off:
+   - Standard flow Checked,
```
bickelj, line 138: Ah, I see. This partially explains why it would work without adding a second client.

In our production Keycloak instance, we have this unchecked for `pdc-service` and checked for `pdc-openapi-docs`. On reflection, it has probably been a while, a long while, since anybody used the `pdc-service` client to get a session. So I have no objection to a single client for development purposes after all.

README.md
```diff
@@ -122,11 +122,35 @@ LOG_LEVEL=trace npm run test
...
+   - OAuth 2.0 Device Authorization Grant (optional), and
+   - OIDC CIBA Grant unchecked.
+3. In the settings for the client added in step (2), set the root URL and Home URL to the URL of the PDC service (not the auth service). [For development purposes, all callback routes can be authorized](https://www.keycloak.org/docs/23.0.7/authorization_services/#_resource_overview)
+   - `/*`
+4. In the settings for the client added in step (2), add a Web origins of `+`.
+5. Within the realm, add a user, e.g. `test-user`. Set a password for this user.
+6. Set your environment variables (see `.env.example`)
+7. Run the service repository with `npm run dev`.
+8. Go to swagger ui page at `http://localhost:3001/`
+9. Click "Authorize" in the top left and click the "Authorize" button in the popup
```
bickelj, line 151: Do you have to change the client name in this prompt in order for it to work? E.g. to `pdc-service`?

bickelj (commented):
See comments. I might rename `pdc-service` client to `pdc-openapi-docs` to more closely match the test/prod setup. But it's probably not doing any damage to leave it as-is if you think `pdc-service` is better.

hminsky2002:
thank you for the typo fix @bickelj, also I took your suggestion and changed the example client name, agreed that it's more fitting

bickelj (approved):

PR #814, opened 2024-03-05
Remove package version

### LINKED ISSUE #807: Zero out version in package.json
> @slifty That makes sense, thanks for the explanation! No problem, merge as-is and let's open an issue to zero it out (or whatever).

_Originally posted by @reefdog in https://github.com/PhilanthropyDataCommons/service/issues/794#issuecomment-1972316993_
            
            
We have a version in the `openapi.json` but also have one in `package.json` -- we don't need both, let's zero out the package one (some day maybe openapi will be dynamically generated and then we can start using the package value as the source of truth, but until that day it's useless!)

slifty:
This PR removes a redundant (and obsolete) version definition for the service project.  We're only versioning the API specification, not the API implementation.

Resolves #807

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,6 @@
 	"packages": {
 		"": {
 			"name": "@pdc/service",
-			"version": "0.3.0",
 			"license": "AGPL-3.0",
 			"dependencies": {
 				"@aws-sdk/client-s3": "^3.525.0",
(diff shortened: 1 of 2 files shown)


reefdog (approved):
lol, GitHub just reflexively hid `package-lock.json` behind the "Diff too large, click to expand" and I was like wait but why and sure enough it's just one line. There must be a hardcoded list somewhere.

Anyway, 👍.

Issue #821, opened 2024-03-11
Change our query approach to use DB-level json functions


slifty:
A long time ago we had a big conversation about the best way to load deep objects from postgres (e.g. when loading a proposal, also load the versions / field values).  We punted on the implementation of our selected path due to priorities in the moment, but as I start working on the new `Organization` / `Proposals` relationship (#799) I would like to finally shave that yak.

I'm making a separate issue for this, as it will impact essentially every endpoint / query and I think it's functionally a prerequisite to proper implementation of #799 

The idea will be to create a series of `foo_to_json` functions in postgres which will return json for a given entity directly (and therefore be able to also query / populate the nested object structures as well since it won't be returned as a traditional row-based result set).

bickelj:
@slifty I think this is pretty much done, yes? If not, reopen.

PR #822, opened 2024-03-11
Use JSON result sets when creating / updating / loading BaseFields

slifty:
This PR creates a new approach for loading data from psql, and uses that approach for loading base fields.  The reasoning for starting with base fields is that they have no "child" relationships / represent one of the simplest entities in our system.

Making this change also required improving some of the related base field code to reflect our latest best practices.

Related to #821

### DIFF
--- a/src/database/initialization/base_field_to_json.sql
+++ b/src/database/initialization/base_field_to_json.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE FUNCTION base_field_to_json(base_field base_fields)
+RETURNS JSONB AS $$
+BEGIN
+  RETURN jsonb_build_object(
+    'id', base_field.id,
+    'label', base_field.label,
+    'description', base_field.description,
+    'shortCode', base_field.short_code,
+    'dataType', base_field.data_type,
+    'createdAt', to_json(base_field.created_at)::jsonb
+  );
+END;
+$$ LANGUAGE plpgsql;
--- a/src/database/operations/create/createBaseField.ts
+++ b/src/database/operations/create/createBaseField.ts
@@ -1,0 +1,27 @@
+import { db } from '../../db';
+import { NotFoundError } from '../../../errors';
+import type {
+	BaseField,
+	JsonResultSet,
+	WritableBaseField,
+} from '../../../types';
+
+export const createBaseField = async (
+	createValues: WritableBaseField,
+): Promise<BaseField> => {
+	const { label, description, shortCode, dataType } = createValues;
+	const result = await db.sql<JsonResultSet<BaseField>>(
+		'baseFields.insertOne',
+		{
+			label,
+			description,
+			shortCode,
+			dataType,
+		},
+	);
+	const baseField = result.rows[0]?.object;
+	if (baseField === undefined) {
+		throw new NotFoundError('The base field could not be created.');
+	}
+	return baseField;
+};
--- a/src/database/queries/baseFields/insertOne.sql
+++ b/src/database/queries/baseFields/insertOne.sql
@@ -10,10 +10,4 @@ VALUES (
   :shortCode,
   :dataType
 )
-RETURNING
-  id as "id",
-  label as "label",
-  description as "description",
-  short_code as "shortCode",
-  data_type as "dataType",
-  created_at as "createdAt"
+RETURNING base_field_to_json(base_fields) AS "object";
--- a/src/__tests__/baseFields.int.test.ts
+++ b/src/__tests__/baseFields.int.test.ts
@@ -260,7 +260,7 @@ describe('/baseFields', () => {
 				description: '😍',
 				shortCode: '🩳',
 				dataType: '📊',
-				createdAt: expect.any(Date) as Date,
+				createdAt: expectTimestamp,
 			});
 		});
 
--- a/src/database/db.ts
+++ b/src/database/db.ts
@@ -1,6 +1,25 @@
 import path from 'path';
+import fs from 'fs/promises';
 import { TinyPg } from 'tinypg';
 
-export const db = new TinyPg({
+const initializationDirectory = path.join(__dirname, 'initialization');
+
+const db = new TinyPg({
 	root_dir: [path.resolve(__dirname, 'queries')],
 });
+
+const initializeDatabase = async () => {
+	const initializationFiles = (
+		await fs.readdir(initializationDirectory)
+	).filter((file) => file.endsWith('.sql'));
+
+	await Promise.all(
+		initializationFiles.map(async (file) => {
+			const filePath = path.join(initializationDirectory, file);
+			const sql = (await fs.readFile(filePath)).toString();
+			await db.query(sql);
+		}),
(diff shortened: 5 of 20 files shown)


src/database/operations/create/createBaseField.ts
```diff
@@ -0,0 +1,27 @@
...
+		{
+			label,
+			description,
+			shortCode,
+			dataType,
+		},
+	);
+	const baseField = result.rows[0]?.object;
+	if (baseField === undefined) {
+		throw new NotFoundError('The base field could not be created.');
```
slifty, line 24: This is a testable line / warrants a test.

slifty: I'm wrong -- I just realized that if insert failed it would throw / never reach this line.  This conditional really only exists for type guard purposes and I think the only way it could happen is if the query file was incorrectly defined.

src/database/queries/baseFields/insertOne.sql
```diff
@@ -10,10 +10,4 @@ VALUES (
   :shortCode,
   :dataType
 )
```
bickelj, line 12: I'm not necessarily advocating the following, but why wouldn't we have companian json_to_base_field functions for inserts as well?

src/database/initialization/base_field_to_json.sql
```diff
@@ -0,0 +1,13 @@
...
+  RETURN jsonb_build_object(
+    'id', base_field.id,
+    'label', base_field.label,
+    'description', base_field.description,
+    'shortCode', base_field.short_code,
+    'dataType', base_field.data_type,
+    'createdAt', to_json(base_field.created_at)::jsonb
+  );
+END;
+$$ LANGUAGE plpgsql;
```
bickelj, line 13: Does this mean that TinyPG is no longer aware of column/property types (beyond "object") when getting the data, but that type checking would still occur at conversion to the AJV type?

bickelj (commented):
I will say I don't hate it but I wonder to what extent we're losing out on the SQL-ness of SQL. I mean we're kind of replacing some SQL here with a function, which is like undoing Postgres to some extent, right? Not completely. This is only my first reaction.

src/database/initialization/base_field_to_json.sql
slifty: I believe that the type is still maintained (as far as JSON has types) -- e.g. a string vs a number vs an object.

We've been moving away from type checking the outputs of the database since we control that layer completely.

src/database/queries/baseFields/insertOne.sql
slifty: What value would that companion provide?  One benefit of this approach is we don't have to duplicate our casting of table-structure-to-json in more than one place, I think we'd lose that with a companion function, but maybe I'm misunderstanding what you mean!

slifty:
Regarding the SQL -- I don't *think* we're losing much of SQL but maybe I'm missing something!  This is just a transformation that happens at the very end (and it's a transformation that was happening at the `pg` library level regardless; we're just doing it one step sooner which will allow us to do some fancy stuff like populate deep objects in the result set)

bickelj:
> Regarding the SQL -- I don't _think_ we're losing much of SQL but maybe I'm missing something! This is just a transformation that happens at the very end (and it's a transformation that was happening at the `pg` library level regardless; we're just doing it one step sooner which will allow us to do some fancy stuff like populate deep objects in the result set)

Just that the query is no longer pure SQL, easily read by a person familiar with SQL, and has one level of indirection there.  This is the only loss I can see for now. But at the same time it might be worth it if complexity is reduced elsewhere, e.g. in deep queries.

bickelj:
> @bickelj I didn't put this in this PR since I wanted to think about the pattern before implementing other entities, but just to show you an example of what this approach will enable:

Yes, a helpful example!

bickelj:
Some more thoughts (still having not tried it, just looking at it here):
- Is the concision achieved [above](https://github.com/PhilanthropyDataCommons/service/pull/822#issuecomment-1989285182) primarily due to the use of `*` in the "subquery"? What if, for example, we used a view, and did a `SELECT *` from that view, what would change or be different?
- Is there a substantial performance or scalability penalty? Suppose the DB is a few `ms` away and has moderate bandwidth? Is JSONB fatter than postgres' usual binary payloads (e.g. due to repetition of labels/columns)? [Interesting post here](https://github.com/PhilanthropyDataCommons/service/pull/822#issuecomment-1989285182)
- What about an example with some joins? What would those look like?

slifty:
@bickelj 

> Is the concision achieved https://github.com/PhilanthropyDataCommons/service/pull/822#issuecomment-1989285182 primarily due to the use of * in the "subquery"? What if, for example, we used a view, and did a SELECT * from that view, what would change or be different?

I'm not sure what you mean, but I *think* this is the only approach that allows us to use the database to form the result set; if we use a traditional view / join then we are going to be iterating through rows full of redundant data in order to generate the objects in the typescript environment.

> Is there a substantial performance or scalability penalty? Suppose the DB is a few ms away and has moderate bandwidth? Is JSONB fatter than postgres' usual binary payloads (e.g. due to repetition of labels/columns)? https://github.com/PhilanthropyDataCommons/service/pull/822#issuecomment-1989285182

Would you be open to optimizing when we face that issue?  Right now I think this has a meaningful impact on reducing code complexity and making it easier to adjust tables / entity shapes in a way that is less likely to reduce bugs.

> What about an example with some joins? What would those look like?

The example above shows what a JOIN can look like (when selecting an entity as a document and wanting to join other types of entities onto the document) -- but also we can always just use normal SQL if there is a particularly complex query, and we would be no worse off (or more redundant) than we would have otherwise been.

This approach just allows us to re-use certain mappings (and also makes it possible to select deep objects), but doesn't force us to write ALL future queries in this way if it isn't the right fit.

src/database/queries/baseFields/insertOne.sql
bickelj: @slifty At a high level (without fully exploring or understanding the need for the json functions) having some consistency and symmetry between the readers and writers, as it were. If there is no need for it there is no need.

bickelj (approved):
It will take me a while to get back up to par with the codebase such that I can meaningfully explore or object to this proposal, so I approve in that it seems like a direction worth taking: embrace the Postgres JSON capabilities, embrace letting the database do more data mashing, and gain some concision here and there.

I am not sure what the `.*` means here, perhaps I should explore that more before having an opinion, but if it means something like `SELECT * FROM` then we usually don't want to do that, right?

slifty:
Noted all around!  And we can / should continue to think about this as the project evolves (and I'll gladly help shift again in future).

One note around the `.*` -- as far as I understand it, the function expects a full row from the `base_fields` table, and the translation from "raw table" to "expected format" is covered inside of the function (that's the purpose of it actually) so I think it's OK.

I look forward to three months from now when we discover that this approach is incredibly inefficient and needs to be destroyed!  But... it sounds like this is worth at least trying for now!

PR #826, opened 2024-03-14
Use JSON queries for proposal endpoints

slifty:
This PR replaces the logic related to proposal handler database interactions with the new-and-improved JSON approach.

This is the "big" entity, and it involved creating json conversion functions for pretty much every entity type, but those functions are NOT being used for non-proposal endpoints just yet.

Related to #821


src/handlers/proposalsHandlers.ts
```diff
@@ -65,221 +35,47 @@ const getProposals = (
...
-					new InternalValidationError(
-						'The database responded with an unexpected format.',
-						isProposal.errors ?? [],
-					),
-				);
-			}
+	const id = Number.parseInt(req.params.id, 10);
+	if (Number.isNaN(id)) {
+		next(new InputValidationError('The entity id must be a number.', []));
+		return;
```
slifty, line 46: I should have a test for this.

hminsky2002 (approved):

slifty:
Too late, no take backs!

hminsky2002 (approved):
To the best of my understanding the decision to abscond from proposal validation in the backend and leverage the database itself for validation is a good one. This definitely reduces the complexity of the codebase and from my own testing in a local instance functionality is the same, with no errors being thrown or linting issues. I would defer to @bickelj for final approval!

src/database/operations/load/loadBundle.ts
```diff
@@ -1,24 +1,18 @@
...
-import { loadObjects } from './loadObjects';
 import type { TinyPgParams } from 'tinypg';
-import type { ValidateFunction } from 'ajv';
 import type { Bundle } from '../../../types';
 
 export const loadBundle = async <T extends object>(
 	tinyPgQueryName: string,
 	tinyPgQueryParameters: TinyPgParams,
 	tableName: string,
-	entryValidator: ValidateFunction<T>,
```
bickelj, line 11: If "load" here means "read from the database and write into node memory", no problem. If "load" means "read from node memory and write to the database" we'd want the validation, right? I am pretty sure this is reading from the DB, though.

src/database/operations/load/loadObjects.ts
bickelj, line 1: Hurray for removed code!

src/database/operations/load/loadProposal.ts
```diff
@@ -0,0 +1,14 @@
...
+import { NotFoundError } from '../../../errors';
+import type { JsonResultSet, Proposal } from '../../../types';
+
+export const loadProposal = async (id: number): Promise<Proposal> => {
+	const result = await db.sql<JsonResultSet<Proposal>>('proposals.selectById', {
+		id,
+	});
+	const proposal = result.rows[0]?.object;
+	if (proposal === undefined) {
+		throw new NotFoundError(`The proposal was not found (id: ${id})`);
```
bickelj, line 11: In contrast to my above comment, this one makes sense to be a `NotFoundError` and can be easily reproduced (test covered it) because it's an HTTP GET and SQL SELECT.

src/database/queries/proposals/insertOne.sql
```diff
@@ -5,8 +5,4 @@ INSERT INTO proposals (
   :externalId,
   :opportunityId
 )
-RETURNING
-  id as "id",
-  external_id AS "externalId",
-  opportunity_id AS "opportunityId",
-  created_at AS "createdAt"
+RETURNING proposal_to_json(proposals) AS "object";
```
bickelj, line 8: If I'm reading this correctly, this means an `INSERT` now requires `JOIN`s. On the one hand I am surprised by that . On the other hand it may be what we want.

src/handlers/__tests__/proposals.unit.test.ts
bickelj, line 1: Hooray for database `JOIN`s and deleted code!

bickelj (commented):
Overall, this approach is pleasant. The database does the `JOIN`s, the application gets data in its preferred JSON format, and we remove awkward cruft!

Remaining general comments not elaborated above:
* I still dislike the overloaded term "load" and would prefer either HTTP verbs or SQL verbs. Or maybe CRUD verbs. But I can get over it too by carefully reading the code.
* Did we switch our formatting rules to allow tabs? I thought we used spaces.
* Does it matter that the functions are listed as `Volatility: volatile` and `Parallel: unsafe` by PostgreSQL.

The one thing that gives me pause is not using database migration scripts to manage the new functions.

bickelj:
One more question, I guess: is a function required? Can we not use views?

src/database/initialization/application_form_field_to_json.sql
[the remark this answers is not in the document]
slifty: My thinking for not using migrations is that I think of these as utilities that are part of our *queries* -- they are not representing persistent data structures or even data sets / views; they're just formatters / decorators that are executed in psql instead of javascript in the name of optimization.

Basically, the fact that they live in the database at all is a byproduct of the tooling, to me, not the functionality.

For example: we don't have migrations for our `.sql` files that contain our SELECT / INSERT statements.  I want to be able to see the history / changes to the JSON representation layer in diff format!  If we change our API specification to tweak a column, or change the order that proposals are returned, etc, I do not want to have to copy and paste 40 lines of complex SQL that then need to be re-reviewed for mistakes, I just want one line diff that adds `DESC` to the `ORDER BY` clause.

I'm personally fine replacing the function every runtime; the query files are re-loaded into the interpreter at every runtime too!   Again that's just because the mental model for me is "this is just an extension of our queries" -- if our SQL runtime tool had the ability to import sql files I would have implemented these in that way instead.

I hear you though, particularly regarding the permissions question.  I'd really rather find ways to bake these decorators into our `.sql` files than make them part of our migrations, though.  The migration thrash will just be so... noisy!

src/database/operations/create/createProposal.ts
```diff
@@ -0,0 +1,18 @@
...
+	createValues: WritableProposal,
+): Promise<Proposal> => {
+	const { opportunityId, externalId } = createValues;
+	const result = await db.sql<JsonResultSet<Proposal>>('proposals.insertOne', {
+		opportunityId,
+		externalId,
+	});
+	const proposal = result.rows[0]?.object;
+	if (proposal === undefined) {
+		throw new NotFoundError('The proposal could not be created.');
```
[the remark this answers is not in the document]
slifty: Oh good call, notfound feels incorrect!  I think you're right about 500 error.

And to answer your question.... this really is not a reachable state (which is why it's not tested) -- the only way to reach it would be if... the database did not error BUT... did not return a result set?  What?  I don't think that's possible.

To that end, maybe the right status code is `418 I'm a teapot`

src/database/queries/proposals/insertOne.sql
slifty: The main thinking here is that our resulting entity would look the same as it would look if accessed via a `GET` / would represent the entity state.  I *think* it's at least close enough to what we want that it would be valuable (otherwise a client would need to make a followup `GET` to get a full object after `POST`, which feels un-REST-ful)

slifty:
@bickelj oh good question re: views...

What do you think of moving forward with this PR / approach as written in the spirit of "good enough" but I'll open an issue for that question AND the migration question?  Honestly, making these views I think would address *both* our concerns.  The view would reflect structure and state in json format, it would presumably be more optimized, it would very clearly be a migration-level structure...  And then things like order would still live in the query / not be migration based / and be totally decoupled from the definition of the data structure.

All this is to say, I like that thought a lot!  It'll take some R&D and I think it can be a gracefully made improvement in a future PR, if you're OK with me taking an iterative approach in that regard and moving forward.

src/database/initialization/application_form_field_to_json.sql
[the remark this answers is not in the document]
bickelj: I am looking at these functions perhaps more from a DB perspective. I think of `CREATE FUNCTION` as something between extending PostgreSQL software code and `CREATE VIEW`. I'm thinking of it as maybe even lower level than `CREATE TABLE`.

bickelj (approved):
I'm marking this approved in the spirit of "this is an improvement" but I still would like the functions to become views if possible or the functions to be stable if possible and either to be part of migrations.

src/database/operations/load/loadBundle.ts
slifty: yeah -- it's reading *from* the DB!

src/database/migrations/0019-create-organizations_proposals.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE organizations_proposals (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  organization_id INTEGER NOT NULL,
+  proposal_id INTEGER NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  FOREIGN KEY (organization_id) REFERENCES organizations(id),
+  FOREIGN KEY (proposal_id) REFERENCES proposals(id),
+  UNIQUE(organization_id, proposal_id)
```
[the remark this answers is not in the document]
slifty: Whoops -- I just realized this migration doesn't belong in this PR!  I'm taking it out but your comment is noted regardless!

Issue #827, opened 2024-03-14
We should use 422, not 409, when referenced resources don't exist


slifty:
Right now we are using `409 Conflict` when trying to create an entity that references an entity that does not exist (e.g. when creating a proposal version but providing a proposal ID that isn't in the database).

It's a bit of a coin toss, but I'm realizing a more appropriate code might be `422 Unprocessable Entity` since it's not that the resulting data would create a *conflict* it simply... can't be created because of the broken foreign key relationship (just like if the data provided was otherwise invalid / e.g. passing a negative ID value, etc).

409 could still be used for situations where the data WOULD create a conflict (e.g. a unique key violation).

PR #836, opened 2024-03-19
Add OrganizationProposal and related endpoints

### LINKED ISSUE #799: Create relationship between Organizations and Proposals
As part of #777 we got rid of `Applicant` and created the concept of `Organization`.

Now we need to make it possible to associate organizations with proposals!  This will be done via a join table.  For now we are only going to support SINGLE organizations associated with proposals.  In future we can use this join table to support the many-to-many use case that we know exists in the long run (and at that point we would be think about how data should / shouldn't map from a proposal to a given organization).

slifty:
This PR adds a new `OrganizationProposal` entity, which represents a relationship between an organization and a proposal, along with endpoints to GET / POST those relationships, and the ability to filter by `organization` when selecting `Proposals`:

* GET `/organizationProposals?proposal=...&organization=...`
* GET `/proposals?organization=:organizationId`
* POST `/organizationProposals`

This is a full suite of functionality baked into one PR, so please consider reviewing commit-by-commit.

You should see most of these new endpoints in the `OrganizationProposal` swagger section.  The new `organization` filter option should be reflected in the appropriate `Proposal` endpoint.

Resolves #799


src/types/Organization.ts
```diff
@@ -11,27 +11,6 @@ interface Organization {
...
-		},
-		createdAt: {
-			type: 'object',
-			required: [],
-			instanceof: 'Date',
-		},
-	},
-	required: ['id', 'employerIdentificationNumber', 'name', 'createdAt'],
-};
-
```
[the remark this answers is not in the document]
slifty: This isn't the SDK -- this is the backend!  It's not used in this code base.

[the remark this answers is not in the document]
reefdog: @slifty Oh, these don't "feed" the SDK? I assumed they were related since some of the SDK generation step involved copying this file.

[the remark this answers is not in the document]
reefdog: @slifty Wait… no they don't. 🤦 I thought this was in `openapi.json`. Never mind!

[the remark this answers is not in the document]
slifty: Ah yeah we definitely need the schema *there*! (side note, most of the stuff in /src/types would ideally just be replaced with the SDK but there's a chicken / egg situation of "the SDK won't have the type defs until the PR is merged" so solving that would involve some kind of fancy footwork around enabling local SDK generation for dev environments)

src/__tests__/proposals.int.test.ts
```diff
@@ -723,4 +740,167 @@ describe('/proposals', () => {
...
+					proposalId: 1,
+				})
+				.expect(400);
+			expect(result.body).toMatchObject({
+				name: 'InputValidationError',
+				details: expect.any(Array) as unknown[],
+			});
+		});
+
+		it('returns 409 Conflict when a non-existent organization is sent', async () => {
```
reefdog, line 968: Why a 409 here?

src/__tests__/proposals.int.test.ts
```diff
@@ -723,4 +740,167 @@ describe('/proposals', () => {
...
+				.post('/proposals/1/organizations')
+				.type('application/json')
+				.set(authHeader)
+				.send({
+					organizationId: 1,
+					proposalId: 2,
+				})
+				.expect(409);
+			expect(result.body).toMatchObject({
+				name: 'InputConflictError',
```
reefdog, line 1036: Rather than me manually checking and asking about the names of the expected errors, can I just ask that you take a quick pass/verify that all of these names and that they match your intent (I.e., no copypasta slipped in)? Feel free to resolve this as soon as you've done so, no note necessary.

src/database/initialization/organization_proposal_to_json.sql
```diff
@@ -0,0 +1,26 @@
...
+  INTO organization_json
+  FROM organizations
+  WHERE organizations.id = organization_proposal.organization_id;
+
+  RETURN jsonb_build_object(
+    'id', organization_proposal.id,
+    'organizationId', organization_proposal.organization_id,
+    'organization', organization_json,
+    'proposalId', organization_proposal.proposal_id,
+    'proposal', proposal_json,
```
reefdog, line 22: Is this a 1:1 replationship? I knew we weren't doing many:many at this point, but I did assume 1:many. The singular naming of the `proposalId`/`proposal` structure makes me doubt this though.

src/database/migrations/0019-create-organizations_proposals.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE organizations_proposals (
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  organization_id INTEGER NOT NULL,
+  proposal_id INTEGER NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  FOREIGN KEY (organization_id) REFERENCES organizations(id),
+  FOREIGN KEY (proposal_id) REFERENCES proposals(id),
+  UNIQUE(organization_id, proposal_id)
```
reefdog, line 8: Ah okay this seems more evidence that this is 1:1. Maybe I should review the subsequent commits, but I hope we aren't 1:1!

reefdog, line 8: Scratch the comment I just left (which I deleted). _Of course_ this uniqueness constraint doesn't mean we can't have 1:many. (It does mean we can't have `many:many` but we knew that.)

src/__tests__/proposals.int.test.ts
slifty: It's what we do right now -- there's an issue for changing all of these over to [422 status codes](https://github.com/PhilanthropyDataCommons/service/issues/827), but this PR is already so big that I didn't want to have to refactor that as well.

src/database/initialization/organization_proposal_to_json.sql
slifty: this entity represents an individual link between a proposal and an organization (it's an OrganizationProposal).

An organization proposal will have exactly 1 organization and 1 proposal
A given organization can have any number of organization proposals
A given proposal can have any number of organiation proposals

src/database/migrations/0019-create-organizations_proposals.sql
slifty: It means you cannot have more than one TYPE of connection between an org and a proposal, but an org can have many proposals, and a proposal can have many orgs (which is what I think of when I say `many:many`)

reefdog (changes requested):
You succeeded in making a PR so large — understandably! — that I was forced to skim some parts! But honestly, the codes seems to be great. 👏

I did run into issues testing it:

1. When I try to POST to `organizations/{id}/proposals` through the Swagger interface, I get an `UnauthorizedError`, even though I'm definitely authorized. I don't get this error when I POST to `proposals/{id}/organizations`. Perhaps a hint is hiding in the observation that the former doesn't have the 🔒 icon in its accordion header?
   OrganizationsProposals Swagger
2. The examples for both of the POST endpoints are misleading. They claim the request body should include `organization` and `proposal` data, when they shouldn't. (Including it doesn't break the call, it's just ignored. Still, the example should show the idealized form.)
   ```json
   {
     "organizationId": 8,
     "organization": {
       "employerIdentificationNumber": "11-1111111",
       "name": "Example Inc."
     },
     "proposalId": 75,
     "proposal": {
       "opportunityId": 3203,
       "externalId": "AnIdGeneratedByAGms"
     }
   }
   ```

Item ① is a breaking issue. Item ② is strongly preferred.

reefdog:
@slifty I just realized I was able to attach the same proposal to multiple organizations. I thought we weren't allowing that?

slifty:
@reefdog nope, we allow it -- there's just no use case for it, and at this point any organization linked to a proposal will adopt the proposal's field values since there's only one relationship type

slifty:
We had a discussion offline and decided to change the approach for interacting with `OrganizationProposal` entities -- instead of having a hierarchical endpoint structure we're going to just create a new `/organizationProposal` endpoint set.  I'm going to take out the endpoints for `/proposals/{x}/organizations` and `/organizations/{x}/proposals`.

reefdog:
> > @slifty I just realized I was able to attach the same proposal to multiple organizations. I thought we weren't allowing that?
> 
> @reefdog we DO allow it at the API level -- it's there's just no use case for it, and at this point any organization linked to a proposal will adopt the proposal's field values since there's only one relationship type. I suppose we could artificially disallow it with middleware validation, but I'm not super inclined to since it's less of a "we shouldn't" and more of a "why would you"

Ah yes, okay, thank you!

slifty:
@reefdog ready for your re-review!  There is now just the GET / POST to /organizationProposals endpoint.

There are a few commits now that make smaller tweaks that might not be needed any more... but... I think those still represent improved code so hoping we can just keep them.

src/openapi.json
```diff
@@ -1050,11 +1050,12 @@
 				"parameters": [
 					{ "$ref": "#/components/parameters/pageParam" },
 					{ "$ref": "#/components/parameters/countParam" },
-					{ "$ref": "#/components/parameters/searchParam" }
+					{ "$ref": "#/components/parameters/searchParam" },
+					{ "$ref": "#/components/parameters/organizationParam" }
```
reefdog, line 1054: lol ironically I wish this was JSON with trailing commas. (not really a review.)

src/routers/organizationProposalsRouter.ts
```diff
@@ -0,0 +1,18 @@
+import express from 'express';
+import { organizationProposalsHandlers } from '../handlers/organizationProposalsHandlers';
+import { verifyJwt as verifyAuth } from '../middleware/verifyJwt';
+
+const organizationProposalsRouter = express.Router();
+organizationProposalsRouter.get(
```
reefdog, line 6: May as well throw a newline before this line (after the `const`), since (1) that seems consistent with the other routers, and (2) that better communications that the constant defined isn't bound in some special way to the first route defined (the `get`).

reefdog (commented):
Great work, I think it will be good. Not yet approved though because I have three questions/blockers from fully testing:

1. Are you sure the table name should be `organizations_proposals` and not `organization_proposals`?
2. I can't reliably test the `POST` endpoint, at least through Swagger. All my calls return 401, even though it's definitely authed (and other authed calls succeed).
3. The Swagger example request body for the `POST` still includes the `organization` and `proposal` data (not just `organizationId`/`proposalId`). While this won't break the call, as mentioned earlier, the example should show a valid/expected body.

We can hop on a screen share if it helps with any of this!

reefdog:
@slifty Also, what do you think about adding a `/organizations?proposalId=` filter?

I think this will be necessary to support this screen:

Proposal - Organization

This is where the user is on the proposal screen, and they click a link to the related organization, which opens in a side panel. Today, we would implement that with a pseudo-relationship: extract the EIN from the proposal, look it up in our system, and link to the resulting org. But obviously that's less explicit than a formal `OrganizationProposal` relationship, which I'd rather rely on. And to support it, when the user loaded this proposal screen, I'd need to do a `/organizations?proposalId=` lookup to get related orgs.

(Yes I'll need to adapt the interface to support multiple linked orgs. Already thinking about it.)

I would say that's out of scope of this PR, except that we're already including `/proposals?organizationId=` and this is the sibling call.

reefdog:
@slifty Sorry for the multiple messages, but, I just noticed another strange bit of API ergonomics.

- ✅ `GET /proposals` takes an `organizationId` param corresponding to an org ID
- ✅ `POST /organizationProposals` request body takes an `organizationId` attribute corresponding to an org ID and an `organization` attribute corresponding to org details
- ❌ `GET /organizationProposals` takes an `organization` param corresponding to an org ID (and `proposal` param corresponding to a proposal ID)

That last one seems inconsistent. Shouldn't the filtering attributes for `/organizationProposals` be `organizationId` and `proposalId`, both for consistency with the other GETs and for more consistency with the POST (even though request bodies are different and shouldn't be prescriptive)?

src/database/migrations/0019-create-organizations_proposals.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE organizations_proposals (
```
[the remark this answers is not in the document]
slifty: This was intentional and aligns with standard-practice names for join tables between two tables at the DB level.

It's joining the two tables -- the individual rows, however, remove the plural from both entity e.g. "the organizations_proposals table contains organization-proposal relationships"

slifty:
It turns out that we're using swagger 3.0.x but swagger 3.1 exists!

https://swagger.io/blog/swagger-supports-openapi-3-1/

This is relevant because swagger 3.1. allows you to reference a schema AND mark it as readonly, but swagger 3.0 doesn't directly do this.  This is why the demo page is providing incorrect examples.

I'm going to open an issue for this; I don't know if it has implications for our sdk generation, so I'll go make the upgrade in a separate PR.

src/database/migrations/0019-create-organizations_proposals.sql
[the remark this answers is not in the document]
reefdog: Good explanation. Thanks!

reefdog:
> @reefdog well actually

I can't work under these conditions.

reefdog:
@slifty We just had a Jitsi call and talked through all the issues and have consensus! (And yes, this should have been a better threaded experience 😵.)

To recap:

- `organizations_proposals` is an industry-accepted way of naming join tables between [table name A] and [table name B]. This joins `organizations` and `proposals`, thus, `organizations_proposals`. (The probable "right" way, in the future, is for us to name this relationship as its own entity — e.g., "memberships" — which can have its own additional metadata.)
- The unauthenticated POST on the Swagger page is an error and will be resolved.
- The noisy example schema on the Swagger page is an artifact of a limitation in Swagger 3.0; Dan will work around it manually now (because the example schema _should_ be correct) but add it to the list of reasons to upgrade to Swagger 3.1, which Does The Right Thing automatically. (He's not going to upgrade us to 3.1 right away because that could turn into a yak shave.)
- Dan will add `/organizations?proposal=` filter as requested [here](https://github.com/PhilanthropyDataCommons/service/pull/836#issuecomment-2021438893)
- I was wrong about the organization param on `/proposals` being `organizationId`, because I misread the API spec. 😳 So none of [this comment](https://github.com/PhilanthropyDataCommons/service/pull/836#issuecomment-2022983751) needs implementing.

src/routers/organizationProposalsRouter.ts
slifty: I get why prettier doesn't try to deal with whitespace but darn it I wish it could!

reefdog: I dunno, what would it do here? This feels more like semantic whitespace, which I would hope Prettier *wouldn't* try to make decisions about.

slifty:
OK, fixed the following non-threaded-issues:

1. POST authentication mistake is fixed
2. Swagger spec properly marks the deep objects as read only in a 3.0 friendly way
3. I've added the `organizations?proposal=...` filter

I think that covers everything?  cc @reefdog

reefdog:
@slifty Yes, those were the TODOs! 👏 Reviewing now, unthreaded…

src/openapi.json
```diff
@@ -744,6 +803,88 @@
...
+					{ "$ref": "#/components/parameters/organizationParam" },
+					{ "$ref": "#/components/parameters/proposalParam" }
+				],
+				"responses": {
+					"200": {
+						"description": "The requested OrganizationProposals.",
+						"content": {
+							"application/json": {
+								"schema": {
+									"$ref": "#/components/schemas/OrganizationProposal"
```
reefdog, line 828: Ooh, I actually caught something!

The Swagger spec claims `/organizationProposals` returns just the `OrganizationProposal` entity directly, but the real API call wraps the entities in an `entries`, with a `total`. I compared the example schema for `OrganizationProposal` with `Proposal` and `Organization`, and those latter two examples both showed the correct structure, e.g.:

```json
{
  "entries": [
    {
      "id": 3407,
      "employerIdentificationNumber": "11-1111111",
      "name": "Example Inc.",
      "createdAt": "2024-03-27T20:08:23.839Z"
    }
  ],
  "total": 20
}
```

reefdog: Ah, I see. Those have `Bundle`s. I guess we need an `OrganizationProposalBundle` then?

slifty: OOOOOOOoooooooo finally, this PR produces something worthswhile

slifty: Fixing this, but also realizing... It looks like the "allOf" hack for 3.0 solves the `readOnly` issue but flattens `organization` to be of type string.

I think we need to call that "will be fixed by #849" and allow it.

slifty:
@slifty don't forget to add a changelog and update the version.

src/openapi.json
reefdog: Oh nooooo, now I see what it is. That's bad! 😣 But yes, I think overall it's better the outer structure of the example schema be correct (and to have the `readOnly` fix) and that we just… try to get #849 on the board.

reefdog:
@slifty don't forget to add a changelog and update the version.

reefdog (approved):
@slifty don't forget to add a changelog and update the version.

slifty:
TOO LATE YOU APPROVED!

Issue #675, opened 2023-12-19
Update to Node 20


slifty:
From the recent `graphile-worker` dependabot PR:

> As of 24th October 2023, Node 20 is the active LTS and Node 18 is maintainence LTS

Is there a reason we should avoid updating?

slifty:
Realizing this might be easier to do after completing #622 -- I may wait to take it on until then.

jasonaowen:
Yeah, I think the only reason to avoid/defer uploading is because of the associated infra work - it should be easier as a DO App!

bickelj:
Does DO separate the node version from the codebase? Of course that would be easier but with its own costs.

bickelj:
(See #833 discussion)

I just tried running `npm ci` and `npm run build` using node 18, then did `npm run migrate` and `npm run start` and all seemed OK. I'll update node to 20 in the `deploy` repo (runtime) and then @slifty volunteered to update it here in the `service` repo (build time).

bickelj:
@slifty OK, first pass is in PR #837 but I may have missed some spots or it may fail, I hereby leave that branch to thee!

slifty:
This is complete 🎉

PR #837, opened 2024-03-19
Update to node 20

bickelj:
Issue #675 Update to Node 20
Issue #833 Build(deps-dev): bump the eslint group with 2 updates


slifty:
@bickelj the README unnecessarily specifies what version of node we are using -- we should take out `, currently 18` since we already say the target dev version is specified in `.node-version`

slifty (approved):
Let's take out that Node 18 reference in the readme and merge!

reefdog:
Whoaaaa, I just noticed this when trying to `npm i`. Neat!

reefdog:
Something… ironically?… this order of operations failed:

```sh
➜  service git:(main) cd service/
# nvm activated Node 18 from the current `.node-version`
➜  service git:(main) git pull
# new stuff pulls
```

Great, new stuff pulled. But… Node 20 didn't automatically activate, since `nvm`'s shell integration runs the version check on directory operations like `cd`. Just pulling the new file wasn't enough to trigger it, and thus:

```sh
➜  service git:(main) npm i
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: undefined
npm ERR! notsup Not compatible with your version of node/npm: undefined
npm ERR! notsup Required: {"node":"20"}
npm ERR! notsup Actual:   {"npm":"9.6.7","node":"v18.16.0"}

npm ERR! A complete log of this run can be found in: /Users/justin/.npm/_logs/2024-03-21T19_18_23_250Z-debug-0.log
```

Easily noticed and handled, and not worth fixing for the rate of Node upgrades. But worth mentioning in case you guys already have a thought/solution.

bickelj:
Oh, weird. I didn't know it was that smart to even try to get the right version in the first place and assumed you'd have to `nvm install 20.[yada].[yada]`. I suppose the good news is we made it clear that we expect 20?

slifty:
@reefdog yeah the solution is what you said (cding in / out once) -- it's up to the dev to make sure their local is hooked up properly!  We've done our job on the repo side by signaling the version to use!

I think of it as the same as when you add something new to package.json it's up to the dev to run `npm i` to make sure they have the right version of a package installed.  It's just this is a much more rare command and instead of `npm i` the command is `cd .. && cd "$OLDPWD"`

PR #852, opened 2024-03-27
Specify swagger-ui-dist version

### LINKED ISSUE #851: Update swagger-ui-dist via dependabot
We use `swagger-ui-express` which has a subdependency for `swagger-ui-dist` that is currently set to `>5.0.0`

This is a subdependency so it appears that dependabot doesn't manage it by default, but there HAVE been updates to it.  I believe we can tell dependabot to care using the `allow` keyword: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

slifty:
This PR specifies the version of swagger-ui we want to use; it also updates the version that was being used before.

Going forward we'll benefit from dependabot updates.

You can test this by running `npm i` and then `npm run start` and looking at the swagger page.  You should notice it's on the latest version.

Resolves #851

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
 				"pino-http": "^9.0.0",
 				"postgres-schema-migrations": "^6.1.0",
 				"require-env-variable": "^4.0.2",
+				"swagger-ui-dist": "^5.12.3",
 				"swagger-ui-express": "^5.0.0",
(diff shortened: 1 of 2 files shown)


reefdog (approved):
Huzzah! Thank you!

PR #854, opened 2024-03-27
Update OpenAPI to v3.1

### LINKED ISSUE #849: Update to OpenAPI 3.1
Swagger supports OpenApi 3.1 (we're using 3.0 right now)

* https://swagger.io/blog/swagger-supports-openapi-3-1/
* https://dev.to/mikeralphson/openapi-31-the-gnarly-bits-58d0

I believe we should upgrade so we can use the latest features of the spec.

slifty:
This PR updates our OpenAPI spec to v3.1

It also makes a small change to remove nullability from a few fields, which should no longer be nullable.  This was done as part of the migration because it appears v3.1 changes the way the `nullable` keyword works.

Resolves #849 

This should NOT be merged until after:

- [x] https://github.com/PhilanthropyDataCommons/sdk/pull/26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Upgraded to use OpenAPI Specification 3.1.
+
 ## 0.11.0 2024-05-20
 
 ### Added
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -1,5 +1,5 @@
 {
-	"openapi": "3.0.3",
+	"openapi": "3.1.0",
 	"info": {
 		"title": "Philanthropy Data Commons API",
 		"description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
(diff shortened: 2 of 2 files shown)


slifty:
I ran the sdk generator locally and it doesn't look like it's up to snuff.  I'll look into whether or not this is a swagger-codegen limitation or if I'm doing something wrong here.

slifty:
Based on https://github.com/swagger-api/swagger-codegen/issues/12210 it looks like codegen doesn't support 3.1.0 yet -- it isn't clear if / when it ever will, though it does seem that the core technologies *have* been updated to support it, so maybe there's just a dependency version that needs updating somewhere.

slifty:
Once this PR is merged we should be able to finish this one: https://github.com/PhilanthropyDataCommons/sdk/pull/26

reefdog:
> Once this PR is merged we should be able to finish this one: [PhilanthropyDataCommons/sdk#26](https://github.com/PhilanthropyDataCommons/sdk/pull/26)

You mean the opposite, right?

reefdog:
From the PR description:

> It also makes a small change to remove nullability from a few fields, which should no longer be nullable. This was done as part of the migration because it appears v3.1 changes the way the `nullable` keyword works.

Is this outdated? I don't see anything referencing nullability.

reefdog (approved):
In combination with https://github.com/PhilanthropyDataCommons/sdk/pull/26, this worked locally (with all the same caveats re: front end needing to adapt to some changes).

slifty:
@reefdog whoops I accidentally nuked the nullable change, I'll either confirm it's needed and add that back in, or will update the PR description.  If I add the changes I may ping you for another review, if I just edit the PR description I will not!

That said, we do want to merge the SDK generator changes *first* since this PR is what will trigger the build / publication of the sdk (and if we merge this before updating the generator, the generator will create a very broken SDK).

slifty:
We paused the project around May 25th so this got stale!

Reviewing everything and will nudge this to the next step so we can benefit from 3.1 soon

PR #856, opened 2024-03-28
Refactor database interactions

### LINKED ISSUE #228: We should consolidate DB result parsing

slifty:
This PR cleans up a bunch of tech debt related to the way we interact with the database.

Specifically: it removes type guards around DB response values and it wraps all DB interactions with utility functions which will handle any errors / ensure that results have been returned.

This PR is currently a WIP.

Resolves #228


slifty:
I ended up doing this in smaller PRs, closing this one.

PR #857, opened 2024-03-28
Remove pointless tests

slifty:
This PR removes *all* instances where we validate database responses via a type guard.  This pattern is unnecessary, since we control the DB response format.  To be clear: we want to *test the correctness* of our queries, but this is already done through the integration tests, and does not need to be repeated in runtime through type guards.

As you can see, removing these checks dramatically simplifies our types; it also allows us to remove a whole bunch of obnoxiously verbose and redundant tests that were making sure we handled errors that really would only happen if the system was broken (which would already be detected via those integration tests).

Making this change now will make it easier for us to continue to improve / refactor our database interactions.

----

The biggest thing to look for when reviewing this PR is to make sure I didn't accidentally delete a test that should be kept.   Hopefully our coverage report will help with that.  The intention was to remove every test that checked for handling various types of unexpected database responses.  I do think eventually it will be good to add a few tests for handling database errors, but that will be best done via tests against middleware rather than redundantly at each individual endpoint.

Related to #228


slifty:
Please note that the newly uncovered test lines / lower coverage are because we are no longer testing DB edge cases.  This will be fixed in an upcoming PR that centralizes DB error handling more effectively (and tests that one central function)

reefdog:
image

Automatic approval, although I do wonder if you could reduce the lines added.

reefdog (approved):
I can only give a cursory review here, but it LGTM! It runs cleanly locally (everything I threw at it still worked), and tests run cleanly, and the code makes sense.

Someone else with more domain knowledge could give you substantive feedback on your strategy here, but… it makes sense to me.

slifty:
I *think* that's good enough @reefdog though @bickelj if you have a second to glance, great!  I'll merge this tomorrow and carry forward.

bickelj:
@slifty In my skimming yesterday I saw no reason to not do this. Or in less awkward language: yes, go for it!

I also see this is the first full end-to-end auto-deploy-to-test-then-production-starting-in-the-service-repo merge! [Original issue](https://github.com/PhilanthropyDataCommons/deploy/issues/106).
[Job that ran here in service repo on merge](https://github.com/PhilanthropyDataCommons/service/actions/runs/8482000334/job/23240383680)
[Job that was triggered in deploy repo by above service job](https://github.com/PhilanthropyDataCommons/deploy/actions/runs/8482010272)
[Job that was triggered in deploy repo by above deploy job](https://github.com/PhilanthropyDataCommons/deploy/actions/runs/8482011785)

Issue #830, opened 2024-03-15
Use views or migrations for the JSON formatting


slifty:
From https://github.com/PhilanthropyDataCommons/service/pull/826

> I still would like the functions to become views if possible or the functions to be stable if possible and either to be part of migrations.

We added a series of DB functions that convert tables into json -- @bickelj raised that they should either be defined via migrations, or we should avoid them entirely and try to create a view.  This issue represents that exploration (and work).

PR #859, opened 2024-03-29
Update applicaton form field queries

slifty:
This PR updates the way we create / access application form fields to our modern pattern.  Specifically it...

1. Removes some unused queries.
2. Adds utility methods for create / load.
3. Updates the existing queries into the modern age of returning json.
4. Creates a new pattern for `offset / limit` which will no longer force internal queries to set limits and offsets if they aren't wanted.  This should be adopted elsewhere, but it's fine to start here.

Related to #228


bickelj (approved):
Having skimmed this, I have no objections.

Well, other than the usual: I prefer "insert, select, update, delete" or "create, read, update, delete" to avoid the overloaded "load" word. In "ETL" load means "write from memory to storage" while in OS terminology load means "read from storage into memory."

Also, do views present obstacles versus functions? Not that it makes a big difference at this level, because there are a couple levels of indirection to the SQL calls here.

slifty:
@bickelj noted on both fronts!  I actually meant to give a nod to #830 in this -- there are going to be a few PRs along these lines, moving things to this better place, but I don't think it will get in the way of 830.

(I'm open to the load terminology convo for sure, that can also be thought about in parallel / I'm not seeing merging of this as "we cannot change it everywhere later")  I'll make an issue for that.

slifty:
Created issue #860

PR #861, opened 2024-03-29
Update application form queries

slifty:
This PR continues the work of DB consolidation and moving to a standardized way of accessing data.

It makes the following changes:

1.  Adds utility methods for create / load.
2. Updates the existing queries into the modern age of returning json.
3. Removes the query parameter for optionally returning deep / shallow ApplicationForms; the API now always returns deep ApplicationForms.

Related to #228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.6.0 2024-04-29
+
+### Changed
+
+- The `/applicationForms` endpoint now returns an `ApplicationFormBundle`.
+- The `/applicationForms` endpoint now returns deep `ApplicationForm` objects.
+- The `/applicationForms/:id` endpoint now returns a deep `ApplicationForm object.
+
+### Removed
+
+- The `/applicationForms/:id` endpoint no longer has a `includeFields` query parameter, as it always includes deep fields.
+
 ## 0.5.0 - 2024-04-27
 
 ### Added
--- a/src/__tests__/applicationForms.int.test.ts
+++ b/src/__tests__/applicationForms.int.test.ts
@@ -27,7 +27,14 @@ const createTestBaseFields = async () => {
 describe('/applicationForms', () => {
 	describe('GET /', () => {
 		it('returns an empty array when no data is present', async () => {
-			await agent.get('/applicationForms').set(authHeader).expect(200, []);
+			const response = await agent
+				.get('/applicationForms')
+				.set(authHeader)
+				.expect(200);
+			expect(response.body).toMatchObject({
+				entries: [],
+				total: 0,
+			});
 		});
 
 		it('returns all application forms present in the database', async () => {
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.6.0 2024-04-29
+
+### Changed
+
+- The `/applicationForms` endpoint now returns an `ApplicationFormBundle`.
```
slifty, line 12: Actually, since it's returning a bundle I might as well add the pagination to the endpoint too.

slifty: Nevermind -- I'd have to write tests.  We can do that some other time.

src/__tests__/applicationForms.int.test.ts
```diff
@@ -48,55 +55,32 @@ describe('/applicationForms', () => {
...
+			const response = await agent
 				.get('/applicationForms')
 				.set(authHeader)
-				.expect(200, [
+				.expect(200);
+			expect(response.body).toMatchObject({
+				entries: [
 					{
-						createdAt: '2022-07-20T12:00:00.000Z',
+						createdAt: expectTimestamp,
```
bickelj, line 65: I think the test is stronger when it verifies that the exact data comes back associated with the right row. In other words, I prefer the literal. But that should not hold up merging this.

src/__tests__/applicationForms.int.test.ts
```diff
@@ -135,7 +119,6 @@ describe('/applicationForms', () => {
       `);
 			const result = await agent
 				.get('/applicationForms/2')
-				.query({ includeFields: 'true' })
```
bickelj, line 138: Does it always include fields now? I assume so seeing the type below.

src/database/queries/applicationForms/insertOne.sql
```diff
@@ -12,8 +12,4 @@ INSERT INTO application_forms (
     1
   )
 )
-RETURNING
-  id as "id",
-  opportunity_id as "opportunityId",
-  version as "version",
-  created_at as "createdAt"
+RETURNING application_form_to_json(application_forms) AS "object";
```
bickelj, line 15: I still ponder when I see this that every single insert causes a bunch of selects with joins. Until it becomes a performance issue it's not an issue.

src/__tests__/applicationForms.int.test.ts
slifty: It does!

bickelj (approved):
No objection. I was able to build and test. I think the only tests that failed were S3 tests because I don't have them set up.

bickelj:
I like these net-zero or net-negative PRs by the way!

src/__tests__/applicationForms.int.test.ts
bickelj: Oh, and it's right there in your commit message too, ha:
> Removes the query parameter for optionally returning deep / shallow ApplicationForms; the API now always returns deep ApplicationForms.

src/__tests__/applicationForms.int.test.ts
slifty: I hear you -- the big issue is that the the json aggregation functions in psql appear to do timestamp-to-string conversion slightly differently from before; in particular they're returning things as +0:00 or +4:00 instead of `Z` -- which means the exact string representing the time might vary from machine to machine.

jasonaowen: That might be because the session time zone varies depending on the client machine. You can `SET TIME ZONE 'UTC';` on a per-session basis; see https://stackoverflow.com/a/27351341 . I haven't looked into how to do this outside of `psql(1)`, though, and in particular I haven't looked into how to do so with the library used here.

PR #863, opened 2024-03-29
Update bulk upload queries

slifty:
This PR continues the updates to our database access patterns.

It:
1. Adds db utilities for BulkUpload.
2. Consolidates our "update" query so that we can update multiple fields at once.
3. Updates the return types to the json structure.

Related to #228

### DIFF
--- a/src/database/operations/create/createBulkUpload.ts
+++ b/src/database/operations/create/createBulkUpload.ts
@@ -1,0 +1,21 @@
+import { db } from '../../db';
+import type { BulkUpload, InternallyWritableBulkUpload } from '../../../types';
+
+export const createBulkUpload = async (
+	createValues: InternallyWritableBulkUpload,
+): Promise<BulkUpload> => {
+	const { fileName, sourceKey, status } = createValues;
+
+	const result = await db.sql<BulkUpload>('bulkUploads.insertOne', {
+		fileName,
+		sourceKey,
+		status,
+	});
+	const object = result.rows[0];
+	if (object === undefined) {
+		throw new Error(
+			'The entity creation did not appear to fail, but no data was returned by the operation.',
+		);
+	}
+	return object;
+};
--- a/src/database/operations/create/createBulkUpload.ts
+++ b/src/database/operations/create/createBulkUpload.ts
@@ -1,17 +1,24 @@
 import { db } from '../../db';
-import type { BulkUpload, InternallyWritableBulkUpload } from '../../../types';
+import type {
+	JsonResultSet,
+	BulkUpload,
+	InternallyWritableBulkUpload,
+} from '../../../types';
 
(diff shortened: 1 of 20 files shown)


src/database/operations/create/createBulkUpload.ts
```diff
@@ -0,0 +1,21 @@
...
+	const { fileName, sourceKey, status } = createValues;
+
+	const result = await db.sql<BulkUpload>('bulkUploads.insertOne', {
+		fileName,
+		sourceKey,
+		status,
+	});
+	const object = result.rows[0];
+	if (object === undefined) {
+		throw new Error(
```
bickelj, line 23: This `throw` statement seems common enough across the `create` functions to warrant a dedicated class that extends `Error`.

src/database/queries/bulkUploads/selectWithPagination.sql
```diff
@@ -1,10 +1,10 @@
...
 FROM bulk_uploads
 ORDER BY id DESC
-OFFSET :offset FETCH NEXT :limit ROWS ONLY
+LIMIT
+  CASE WHEN :limit != 0 THEN
+    :limit
+  ELSE
+    NULL
+  END
+OFFSET :offset
```
bickelj, line 10: Why the switch from OFFSET/FETCH NEXT to LIMIT/OFFSET? It looks like this works, for example: `select name, setting, unit from pg_settings offset null fetch next null rows only;`

bickelj (approved):
I was able to build and run tests, the tests that were expected to pass passed. Minor comments inline.

src/database/queries/bulkUploads/selectWithPagination.sql
slifty: Oh huh, I thought `fetch next null rows only;` would only return *one* row, and I couldn't get `ALL` to work without it yelling at me.

Was I misreading the docs 😅

src/database/operations/create/createBulkUpload.ts
slifty: For sure -- I haven't made an issue for this but my plan is to actually refactor this ENTIRE repeated block into a "loadItem" utility that will handle the undefined check as well.

That said... after that refactor this is an unreachable error so I'm not sure we should make a custom type for it (maybe UnreachableError would be fun though!)

Let's decide at that PR though if that's OK :D

src/database/queries/bulkUploads/selectWithPagination.sql
slifty: I'm gonna merge this as is, but when I continue this refactor I'll actually try out the `fetch null` thing to see if it returns all rows or just 1 row; I'll call out what I learn in that next PR!

PR #867, opened 2024-04-03
Implement Data Validation

### LINKED ISSUE #760: Implement validation mechanism
The service layer should perform validation on the field values of uploaded proposals. The database should be altered to assign a type to each base field, chosen from a set such as an enumerator, and related proposal field values should be validated against this type in the service layer. It can begin by validating the simple types of 'string', 'number', 'phone number', 'email', 'URL' and 'boolean'

hminsky2002:
This PR introduces basic data validation for uploaded proposal fields. It modifies the base_fields and proposal_field_values tables in the database and related types in the service layer to accommodate 
1. A set of predefined acceptable types for a field
2. A boolean field "is_valid" to indicate if the proposal field's value's type matches the related base field

Validation is performed by a helper function, fieldValueIsValid, which uses regexes, though a more developed validation 
system will likely be necessary as we account for more complicated types. 

Testing:
1. Start pdc service
2. Run migrations. 
  a. NOTE: All base field types will be set to string by default, which are valid by default in the validation system. 
                  In order to test individual types(such as number, phone number, email, etc), new base fields with the 
                  corresponding type will need to be created and inserted into the database.
3. Start pdc frontend, navigate to the add data page
4. Upload a proposal with chosen fields valid or invalid
5. Verify that the expected validity is reflected in the "is_valid" column in the proposal_field_values table in the database

Resolves #760


slifty:
I'll give a review of the actual code soon but in the shorter term some meta changes requested:

1. Each commit should be self-contained and result in a "correct" state -- so the refactored tests should be added as the changes are made (they are part of the change, and actually serve as contextual documentation of the change because they communicate what you changed and how those changes change expectations of API behavior!)

2. It looks like there isn't much test coverage for the validation functions; could you add more so that there is 100% coverage?

src/tasks/processBulkUpload.ts
```diff
@@ -318,11 +320,16 @@ export const processBulkUpload = async (
 							'There is no form field associated with this column',
 						);
 					}
+					const isValid = fieldValueIsValid(
```
slifty, line 323: Love it -- this is so clean!

src/types/ProposalFieldValue.ts
```diff
@@ -9,6 +9,7 @@ export interface ProposalFieldValue {
 	value: string;
 	createdAt: Date;
 	applicationFormField?: ApplicationFormField;
+	isValid: boolean;
```
slifty, line 12: Whoops I hit "submit" too early -- we have a funny thing here -- isValid *is* internally writable, but from an API perspective this is readonly (we don't want clients telling us if something is valid or not).

There is a pattern for this over here, where we have an `InternallyWritableBulkUpload` and a `WritableBulkUpload`: https://github.com/PhilanthropyDataCommons/service/blob/main/src/types/BulkUpload.ts

You'll want an `InternallyWritableProposalFieldValue` and a `WritableProposalFieldValue`

src/types/ProposalFieldValue.ts
```diff
@@ -29,6 +30,9 @@ export const proposalFieldValueWriteSchema: JSONSchemaType<ProposalFieldValueWri
 			value: {
 				type: 'string',
 			},
+			isValid: {
```
slifty, line 33: I'm marking this on this line but it's unrelated -- we need to update the openApi spec to add `isValid` there as well.

slifty (changes requested):
Sorry for the double review!  Some other changes.

src/types/ProposalFieldValue.ts
```diff
@@ -9,13 +10,19 @@ export interface ProposalFieldValue {
...
 	applicationFormField?: ApplicationFormField;
+	readonly isValid: boolean;
 }
 
 export type ProposalFieldValueWrite = Omit<
 	ProposalFieldValue,
 	'applicationFormField' | 'createdAt' | 'id' | 'proposalVersionId'
 >;
 
+type WritableProposalFieldValue = Writable<ProposalFieldValue>;
```
slifty, line 21: I'm seeing why you used Partial earlier -- since you're using this pattern you need to mark non-writable fields as readonly in the base type (id, createdAt, and applicationFormField)

src/types/ProposalFieldValue.ts
```diff
@@ -29,6 +36,11 @@ export const proposalFieldValueWriteSchema: JSONSchemaType<ProposalFieldValueWri
...
 			},
+			isValid: {
+				type: 'boolean',
+			},
 		},
-		required: ['applicationFormFieldId', 'position', 'value'],
+		required: ['applicationFormFieldId', 'position', 'value', 'isValid'],
 	};
+
+export { WritableProposalFieldValue, InternallyWritableProposalFieldValue };
```
slifty, line 46: Since you're switching to this model (which is good, it's the future!)  You should probably remove `export` from the earlier definitions as well and add them to this list explicitly.

src/fieldValidation.ts
```diff
@@ -0,0 +1,30 @@
...
+		case BaseFieldDataType.PHONE_NUMBER:
+			return /^\s*(?:\+?(\d{1,3}))?[-. (]*(\d{3})[-. )]*(\d{3})[-. ]*(\d{4})(?: *x(\d+))?\s*$/.test(
+				fieldValue,
+			);
+			break;
+		case BaseFieldDataType.EMAIL:
+			return /^\S+@\S+$/.test(fieldValue);
+			break;
+		case BaseFieldDataType.URL:
+			return /^(https?:\/\/)?(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_+.~#?&//=]*)/.test(
```
slifty, line 20: Just a note that while I think it's totally OK to merge this as is -- I probably would have preferred this PR only add the most basic validation (number and string), and in a future PR add more complex ones using libraries rather than introducing code that is complex enough to be trouble but not tested enough to be necessarily right!

Using a library here allows us to avoid having to write sweeping unit tests that test all the weird edge cases -- we can rely on the project to have correct validation!

TL;DR I think we need to replace this regex with real validation tools / libraries as soon as possible.

Phone numbers could be: https://www.npmjs.com/package/phone, (or maybe there is an ajv utility for it)

URLs and Emails would be best done using `ajv-formats` https://github.com/ajv-validator/ajv-formats

---

We should make an issue for these (and prioritize addressing next week IMHO!)

slifty (changes requested):
Great iteration!  Some more comments (one is really about future work, but the others are for this PR)

CHANGELOG.md
```diff
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.6.0 2024-04-29
+## 0.7.0 2024-04-04
```
slifty, line 8: Whoops!  I just realized that the api spec changes were made separate from the code changes -- much like tests, the API spec / documentation should be directly connected to the code changes they reference (think of a commit is a coherent block of related changes -- once a PR is merged in, the PR frame is lost and each commit stands on its own in the history!)

src/types/ProposalFieldValue.ts
```diff
@@ -1,21 +1,28 @@
...
 	value: string;
-	createdAt: Date;
-	applicationFormField?: ApplicationFormField;
+	readonly createdAt: Date;
+	readonly applicationFormField?: ApplicationFormField;
+	readonly isValid: boolean;
 }
 
-export type ProposalFieldValueWrite = Omit<
+type ProposalFieldValueWrite = Omit<
```
slifty, line 16: Ah I missed this -- now that you have `WritableProposalFieldValue` you should take this type out (basically, this is the way we *used* to define writable versions of types, but now we have the `Writable` class).

This might require changes to other places that use it.

Alternatively, if you don't want to do that refactor in here you should take out the `WritableProposalFieldValue` definition and just use `ProposalFieldWrite` there instead (and add `isValid` to its list of Omitted fields).

src/fieldValidation.ts
slifty: @hminsky2002 just to clarify (since this is marked as resolved) -- can you make those issues (then you can mention them as a comment here and github will link to the context)

slifty (changes requested):
Sorry, two things I missed in the last review.

src/types/ProposalFieldValue.ts
hminsky2002: This seems like a reasonably large refactor (as I would have to change the proposal version schema as well) so i think I'll drop the WritableProposalFieldValue for now and would happily do a larger refactor as a separate PR!

src/fieldValidation.ts
hminsky2002: Issues #875 and #874

src/types/ProposalFieldValue.ts
slifty: That sounds right -- and you don't need to do the refactor; it's on my list to continue the refactor I've been doing for all the other endpoints / schemas as well!

slifty (approved):
Thanks for bearing with all the iteration!

This looks really great, well done!

hminsky2002:
> Thanks for bearing with all the iteration!
> 
> This looks really great, well done!

Woooo!

bickelj:
@hminsky2002 @slifty I think the same principles for database schema migrations apply to database seed scripts. In other words, I don't think we should modify existing seed scripts but rather append new scripts that assume the previous ones ran. For example, in production we have run several more seed scripts that did not make it into the main branch for one reason or another, see #305. It looks like the ball might be in my court to respond to Jason's objections, though. Of course, another option is to start over with base fields seeds, and that might be preferred. In that case, we should delete all the data in the production database and create a new seed script that runs.

Issue #710, opened 2024-01-09
Model users in the schema


slifty:
How do we represent users (and data ownership) in the PDC?
Do we need to consider organizations / user groups at this stage?

slifty:
Our first implementation does not need to support the concept of user groups, since we are just focused on the following user stories:

1. Admin / superuser (this is basically the current "logged in" experience, which has full access to all data and endpoints)
2. Standard user (this is a new kind of user, that can only view proposal data associated with their account OR proposal data that is public)
3. Public user (this is not a new user since obviously there is a logged out state, but we'll be expanding what this kind of user can do).

This issue is about modeling so I'm gonna hone in on that.

Authentication is done by keycloak and results in a jwt, which is then parsed.  First step is going to be to better understand how much of that we need to keep track of in our ERD.  For instance, maybe the concept of "administrator" should be a keycloack concept rather than a PDC concept, maybe not!

slifty:
It looks like these are the values associated with a jwt: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1

It looks like "subject" is the "principal" which, as I understand it, means the user.

There is a question about whether the jwt should include the *roles* of the user, and that is explored here: https://stackoverflow.com/questions/47224931/is-setting-roles-in-jwt-a-best-practice

I'm thinking we should only us the jwt for identifying the user, rather than identifying the user's roles.  For instance if someone had a role when the JWT was created then it wouldn't be possible to remove that role until that JWT expired.

This means for a given token / user we'll need to look up their role. It's possible we could look up the role from keycloak directly, but it feels cleaner (and fewer network requests) to just look that information up from the database / schema.

For that reason I'm leaning towards *not* using on keycloak for defining  roles within the PDC, and instead we would have a user table with a "superuser" flag.

reefdog:
> It looks like "subject" is the "principal" which, as I understand it, means the user. …

@slifty Don't miss @jasonaowen's breakdown of relevant terms back in https://github.com/PhilanthropyDataCommons/service/issues/255#issuecomment-1468872874! (Which doesn't change what you wrote, I just want to be sure you have that useful reference handy.)

slifty:
I'm planning to *not* have REST endpoints for users in this first implementation, but instead will update our express middleware that handles jwts so that they create / select the user associated with the JWT.

We may some day want a more RESTful user interaction, but for now it's an internal concept.

PR #871, opened 2024-04-05
Decouple authorization and authentication

### LINKED ISSUE #870: Consider decoupling authorization from authentication middleware
Right now we have middleware that validates a Jwt and also will throw an error in the event that the JWT is not valid.

I'd like to consider changing this model so that the jwt middleware applies to EVERY endpoint automatically (and also the user context middleware I'm writing for #710).  Then we could have separate middleware which will, depending on the state of the user context, handle authorization.

I think it's appropriate to define the authorization at the route definition level / via middleware, but I think that by decoupling the authentication middleware from the authorization we:

1. create more consistency (all requests will be typed to potentially have user information)
2. create more flexibility (we can now define a variety of different levels of authentication middleware)

slifty:
This PR changes authentication so that it is both a universal step AND the act of checking authentication is not in itself an act of authorization.  It also adds adds a simple authorization middleware which verifies that the request is associated with a valid authentication object.

Functionally this should not result in any changes to the API (so no impact on version / changelog) -- however it should allow for support for users in future!

Resolves #870
Related to #710


src/middleware/processJwt.ts
```diff
@@ -4,12 +4,10 @@ import type { NextFunction, Request, Response } from 'express';
...
 
-export const verifyJwt = (
-	req: Request,
-	res: Response,
-	next: NextFunction,
-): void => {
+const processJwt = (req: Request, res: Response, next: NextFunction): void => {
 	expressJwtMiddleware(req, res, next).catch((error: unknown) => {
-		throw error;
+		next(error);
```
slifty, line 9: We should make sure this is actually a correct change -- it might be, but maybe the correct response here is to log the error rather than invoke `next`.

It's unfortunate because it's not clear if expressJwtMiddleware will invoke next if it fails (if it does, then I think calling next twice would result in a problem; if it doesn't then I believe the request would hang).

Given that... maybe the safest approach would be to log it.  I worry that if next is called twice it could crash the process, but if next is called zero times the damage is just that single connection.

slifty: (in either event I think throwing would crash the process, and so that's definitely not what we want)

src/middleware/processJwt.ts
```diff
@@ -0,0 +1,15 @@
...
+import { jwtOptions } from '../auth/jwtOptions';
+import { getLogger } from '../logger';
+import type { NextFunction, Request, Response } from 'express';
+
+const logger = getLogger(__filename);
+const expressJwtMiddleware = expressjwt(jwtOptions);
+
+const processJwt = (req: Request, res: Response, next: NextFunction): void => {
+	expressJwtMiddleware(req, res, next).catch((err: unknown) => {
+		logger.error(err);
```
bickelj, line 11: Authentication or authorization code is one of those few cases where I think we do want to know exactly what happens, whether it is what is expected to happen, and therefore test coverage is of benefit.

src/auth/jwtOptions.ts
```diff
@@ -21,5 +21,5 @@ export const jwtOptions: Params = {
 	}) as GetVerificationKey,
 	issuer,
 	algorithms: ['RS256'],
-	credentialsRequired: true,
+	credentialsRequired: false,
```
bickelj, line 24: Why aren't credentials required?

src/middleware/requireAuthentication.ts
```diff
@@ -0,0 +1,17 @@
...
+import { UnauthorizedError } from '../errors';
+import type { Request as JWTRequest } from 'express-jwt';
+
+const requireAuthentication = (
+	req: JWTRequest,
+	res: Response,
+	next: NextFunction,
+) => {
+	if (req.auth === undefined) {
+		next(new UnauthorizedError('No authorization token was found'));
```
bickelj, line 11: What is the difference between calling `next` with an `Error` versus throwing an `Error`? What is the advantage of the former?

src/__tests__/applicationForms.int.test.ts
```diff
@@ -26,6 +26,10 @@ const createTestBaseFields = async () => {
 
 describe('/applicationForms', () => {
 	describe('GET /', () => {
+		it('requires authentication', async () => {
```
bickelj, line 29: :+1:

bickelj (approved):
I am OK with the approach. My [previous draft of authorization](https://github.com/PhilanthropyDataCommons/service/commit/c9ed206196d48c0c795d7ef00661f7bfea3dfa0d#diff-1b77816b89c7904bd5f98d0f593d87c692016dd78a7a3ab231057d6adc4567b5) was not without its own problems.

One problem that can come from splitting the middleware is that authorization definitely depends on a valid authentication token every time. If authorization middleware is split from authentication middleware, we could imagine a (bad) situation where authorization middleware says "OK, all good" with a bad token. But in any case, it comes down to care and caution on our parts to test well (as you have shored up here) and to make always require an authentication token before checking for authorization.

This is my Friday afternoon before vacation review not having tested the code locally :smile:

Because I do not see any critical tests removed and because I see better tests added, and I don't see any red flags other than the `requireCredentials` thing I am comfortable leaving it to @slifty whether to merge or not. It would be good to get another set of eyes, though.

src/middleware/requireAuthentication.ts
slifty: Throwing an error crashes the process / crashes the server.

`next` passes the error to the error handling middleware (which then translates it into an HTTP status code, in this case a 401)

src/auth/jwtOptions.ts
slifty: Basically this means the jwt parser will parse JWTs if they exist, but if they don't exist then it just moves forward -- now we have middleware that exclusively requires the auth session (rather than both creating AND requiring the auth session)

slifty:
Approval (and vacation zone) appreciated but just to clarify for this concern:

> If authorization middleware is split from authentication middleware, we could imagine a (bad) situation where authorization middleware says "OK, all good" with a bad token

The only thing populating the `auth` variable is the jwt middleware, so under this current code base that won't be possible.

If the jwt middleware somehow gets moved or not run for an endpoint (right now it is universally run), then `auth` would not be populated regardless of a good, bad, or nonexistant token being passed.

Does that make sense?

src/middleware/processJwt.ts
slifty: This is now much safer, and now handles for both the "middleware calls next and then throws" and "middleware throws without calling next" case!

I also wrote tests that mock the express-jwt to simulate both possible situations, which was QUITE FUN....

slifty:
Added better tests! (that was... quite a journey -- fun fact, if you want to test proper jwt parsing you need to make the test an integration test!).

The tests as written were passing, but that's because the jwt stuff was all happening asynchronously / wasn't aligned with the test, and there was no positive test case 😬 

Hooray for better tests!

PR #876, opened 2024-04-06
Add Users

### LINKED ISSUE #710: Model users in the schema

slifty:
This PR adds the concept of a user to the PDC.  Users are not a REST concept at this point, but rather the User state is associated with a given request based on the JWT subject.

We are relying on JWTs for authentication, so "users" also exist outside of the PDC in whatever auth service is being set up (our project recommends Keycloak, but the service is implementation agnostic).  That is to say: we haven't needed to have users in the past for authorization, because it was really just a binary state -- either you have a valid JWT or you don't.

We're about to start supporting things like data ownership in the PDC, and in order to do that we need to be able to capture some basic authentication information about the current user.

Resolves #710


src/database/migrations/0020-create-users.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE users (
```
slifty, line 1: Flagging this file: I think I have to rename the migration once @hminsky2002 merges his validation PR

slifty:
I'm confused... why is CI only running `Build`?

src/database/migrations/0020-create-users.sql
slifty: (done!)

src/test/__tests__/utils.int.test.ts
```diff
@@ -11,4 +11,55 @@ describe('test/utils.ts', () => {
 			expect(isoTimestampPattern.test('2022-10-27')).toBe(false);
 		});
 	});
+
+	describe('generateNextWithAssertions', () => {
```
hminsky2002, line 15: Agreed that this a proactive for handling the introduction of more involved/complex middleware, and code is functional!

src/middleware/addUserContext.ts
```diff
@@ -0,0 +1,33 @@
...
+const selectOrCreateUser = async (subject: string) => {
+	try {
+		return await loadUserBySubject(subject);
+	} catch {
+		return await createUser({ subject });
+	}
+};
+
+const addUserContext = (
+	req: AuthenticatedRequest,
```
hminsky2002, line 14: Looks like the right approach for verifying user context

src/types/User.ts
```diff
@@ -0,0 +1,11 @@
+import type { Writable } from './Writable';
+
+interface User {
+	readonly id: number;
+	subject: string;
```
hminsky2002, line 5: Just for my own understanding, the "subject" of the User schema is really the "subject" of the JWT (which is in turn what refers to the 'User' of the JWT)?

hminsky2002 (approved):
This seems like the right implementation for Users right now, as we don't want to be able to create Users FROM the PDC, only respond appropriately to a given User session(provided by Keycloak). Nice work @slifty !

src/types/User.ts
slifty: Correct -- and to that point, I wonder if `subject` should actually be used here since it's removed from the jwt context.  Maybe it should be `authenticationSubject` or something entirely different!

reefdog: dan you took advice from a robot
dan you owe them now
dan
dan be careful who your friends are

PR #883, opened 2024-04-09
Update phone number validation

hminsky2002:
This PR replaces phone number validation for proposal field values that was hardcoded as a regex in PR #867 with validation using the [Phone package](https://www.npmjs.com/package/phone).
The phone package also offers formatting of phone numbers which may be a valuable feature in the future if/when we address formatting user input.


reefdog (commented):
This is awesome!

One thing I would have suggested — worth keeping in mind in future — is bundling the "add phone package" change with the "update the field validation" change in a single commit, and then made the seed change in a separate commit.

So turn `add package + (use package + update seedfile)` into `(add package + use package) + update seedfile`.

Obviously not important enough to rewrite history, but good for the future!

PR #889, opened 2024-04-10
Fail eslint on warning

slifty:
This PR updates our lint command to fail if there are warnings.  This will apply both to local invocations of lint as well as the CI invocation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
 		"cleanup": "rimraf dist",
 		"format": "prettier . --write",
 		"lint": "npm run lint:ts && npm run lint:prettier && npm run lint:swagger",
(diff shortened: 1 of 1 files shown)


hminsky2002 (approved):
Good idea! If i'm not mistaken our CI pipeline will fail if there are any warnings anyway right?

slifty:
@hminsky2002 without this change I believe it will *NOT* fail if there are warnings, which is part of why we want to be sure this is added.

hminsky2002:
> @hminsky2002 without this change I believe it will _NOT_ fail if there are warnings, which is part of why we want to be sure this is added.

Ah, then this is even more pertinent

Issue #485, opened 2023-09-11
Model data ownership in the schema


bickelj:
In order to "Add authorization" https://github.com/PhilanthropyDataCommons/service/issues/255 to the existing authentication scheme, more than a user with a JWT is required.

This task focuses on "the one who uploaded it can modify it exclusively" story.

In other words, this task can be considered complete without a full modeling of groups and inter-user or inter-group permissions (#484) when it is true that write access is no longer available except for the user (and perhaps associated groups, depending on whether this or #484 is completed first) that posted the data, for at least one entity.

slifty:
This is related to the work in #396

slifty:
I think that eventually we'll want to have a many:many relationship related to proposal "ownership"

That said, I also think there will only ever be a single "creator" (and that creator can be null)  even if we eventually want have more complex "owner" relationships.

For the purposes of our immediate use case I think creator is all we care about.

So with that in mind I'm going to add a `createdBy` attribute to the `bulk_uploads` and `proposals` table.  This would be a foreign key pointing to the `users` table.  This would then allow for the addition of a `createdBy` filter on `proposals` and `bulkUploads` that accepts a user reference, as well as a default narrowing of query results for any non-administrative user who hits those endpoints.

slifty:
For a moment I was questioning whether we *actually* want to allow NULL for `createdBy` and right now I'm landing on "we do".

For instance, we may have internal scripts that populate proposals and are NOT associated with an authentication context.

The alternative, however, is that maybe we do want to require `createdBy` and we also want a system `user` that any internal system operations use when creating proposals / bulk uploads.

If we want to have a system user, we should think about whether (A) that should be a configured value via `.env` or (B) that should be a hard coded value.  I would probably lean towards a configurable value, though then we'll need to figure out how to create a dynamic migrations to create that user based on the env value, and if that value ever changes, things get weird...

I suppose I should sleep on this.

bickelj:
I'm late to the party, but I think we do always want a "createdBy" user, it should be non-null. And I see that is the case in the `bulk_uploads` table, great!

PR #892, opened 2024-04-10
Add `createdBy` to the BulkUpload entity

slifty:
This PR adds the concept of "creator" to Bulk Uploads.  This will eventually allow us to add filters and authorization to our bulk upload endpoints.

Related to #485


src/database/migrations/0023-insert-system-user.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO users ( authentication_id ) VALUES ( '' );
```
slifty, line 1: I want to call this out for discussion -- the idea here is we will always have a "system" user with no authentication id.

Our user middleware will not populate a user if `sub` is `''` which I think makes this a particularly safe choice (if we had the system user as something like `system` as the authenticaftion ID then in theory someone could figure out a way to make a jwt with that as the subject and become the system user.  In this case even if they did manage to create a valid jwt with a blank sub, our middleware would treat that as an invalid user.

src/database/migrations/0024-alter-bulk_uploads.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE bulk_uploads
+  ADD COLUMN created_by INTEGER NOT NULL DEFAULT select_system_user_id(),
```
slifty, line 2: Also flagging this: I don't want a default for this column and I also don't think we want to allow for bulk uploads with no user.  For any existing data in the system the bulk uploads will be attributed to the system user (and then you see that default is immediately dropped)

src/types/BulkUpload.ts
```diff
@@ -14,6 +14,7 @@ interface BulkUpload {
 	readonly id: number;
 	fileName: string;
 	sourceKey: string;
+	readonly createdBy: number;
```
slifty, line 17: This should probably be moved to be next to `createdAt`

(also maybe we should put these in alphabetical order in general but whatever)

src/middleware/requireUser.ts
```diff
@@ -0,0 +1,21 @@
+import { Response, NextFunction } from 'express';
+import { UnauthorizedError } from '../errors';
+import type { AuthenticatedRequest } from '../types';
+
+const requireUser = (
```
slifty, line 5: I'm not quite sure if there's a great reason to differentiate between `requireAuthentication` and `requireUser`

If a user isn't populated then I don't think the authentication was actually properly created.  Given that I'm thinking it would be better to just update `requireAuthentication` to check for the user status.

slifty: (I took this out)

slifty:
Apologies -- I identified something I'd like to refactor a little bit before review (moving to draft, I'll tag when it's ready again!)

reefdog (approved):
This seems to work great! I was able to use the front-end to create a bulk upload, and upon inspection I see that the `createdBy` for my new bulk upload is `1`. The `createdBy` for all pre-existing uploads is `4`, which I assume is the newly-created system user.

src/middleware/__tests__/addUserContext.int.test.ts
```diff
@@ -9,15 +9,21 @@ describe('requireAuthentication', () => {
 		const mockRequest = {} as unknown as AuthenticatedRequest;
 		const mockResponse = {} as unknown as Response;
 
-		const runAssertions = async (err: unknown) => {
-			expect(err).toBe(undefined);
-			const metrics = await loadTableMetrics('users');
-			expect(metrics.count).toEqual(0);
-			expect(mockRequest.user).toBe(undefined);
-		};
+		loadTableMetrics('users')
```
hminsky2002, line 12: To clarify, we are refactoring the tests to be inside the loadTableMetrics scope because we are now interested in the value returned(baselineUserCount)? Because now we want to know that the count value in the users table is equal to ... the count value in the users table(?)

src/middleware/__tests__/addUserContext.int.test.ts
```diff
@@ -28,14 +34,22 @@ describe('requireAuthentication', () => {
 		} as unknown as AuthenticatedRequest;
 		const mockResponse = {} as unknown as Response;
 
-		const runAssertions = async (err: unknown) => {
-			expect(err).toBe(undefined);
-			const user = await loadUserByAuthenticationId('foo@example.com');
-			expect(mockRequest.user).toEqual(user);
-		};
+		loadTableMetrics('users')
```
hminsky2002, line 37: Ah I see, this choice becomes clearer once we measure for offset(now that we are expecting baselineUserCount to be offset by 1)

src/middleware/requireAuthentication.ts
```diff
@@ -8,7 +8,15 @@ const requireAuthentication = (
 	next: NextFunction,
 ) => {
 	if (req.auth === undefined) {
-		next(new UnauthorizedError('No authorization token was found'));
+		next(new UnauthorizedError('No authorization token was found.'));
+		return;
+	}
+	if (req.auth?.sub === undefined || req.auth.sub === '') {
```
hminsky2002, line 14: Is the authorization subject being empty string really it _lacking_ the subject, or a separate issue?

src/middleware/__tests__/addUserContext.int.test.ts
slifty: Yeah that's right -- we want to be sure the value did not change, and to do that we need to know what the value was in the first place.

Normally we could have an async test and await this (avoid the next / catch) but since middleware is weird and uses callbacks instead of promises we need to be able to use jest's `done` -- and that doesn't exist in an async context.

src/middleware/requireAuthentication.ts
slifty: I'd say an empty string is lacking a subject.  Or rather, with this line we are saying that in our system we aren't considering empty string to be a valid subject.

src/database/migrations/0024-alter-bulk_uploads.sql
hminsky2002: Agreed, I don't think there's a scenario where we want to allow non-authenticated users to upload data. Just covering any existing bulkuploads for migrations sake seems sufficient

src/errors/FailedMiddlewareError.ts
```diff
@@ -0,0 +1,3 @@
+class FailedMiddlewareError extends Error {}
+
```
hminsky2002, line 2: Another design pattern question:

Do we use these blank class definitions in anticipation of fleshing them out in the future? Or are just taking advantage of the fact that by creating new class wrappers around Error we can specify the name without having to write error messages

src/middleware/requireAuthentication.ts
hminsky2002: After a few moments thought I think that's sufficient, the subject is meant to be a self-describing string correct? I.e `system` or `client` . If that's the case, then an empty string is 'lacking', as the only validity check is if it's a readable string that could communicate it's purpose.

src/errors/FailedMiddlewareError.ts
slifty: We won't be fleshing it out -- but this lets us define new types of error which we could then handle in the middleware to translate to more distinct HTTP response codes (it also allows us to write tests that ensure that a specific TYPE of error was made).

PR #893, opened 2024-04-10
Add `tsc` command to our linter

slifty:
This PR adds another type of linting (type checking).  This is static analysis and prior to this the only way to actually ensure our types were properly defined would be to start or build the project.

You can test this by modifying a type that is used somewhere to add a required field and running the linter.  The linter should flag instances of data that is supposed to match the type but doesn't.


hminsky2002 (approved):
tested and functional, another win for linting

PR #894, opened 2024-04-10
Add support for `Proposal` ownership

slifty:
This PR adds`createdBy` to the `proposals` table and populates the `createdBy` field based on the user that initiated the creation of a given proposal.

This PR should not be merged until after #892 has been merged.

Resolves #485

PR #896, opened 2024-04-10
Only return bulk uploads for the current user

### LINKED ISSUE #759: Only return bulk uploads that are associated with the logged in user when retrieving bulk uploads

slifty:
This PR changes the way GET `/bulkUploads` works so that it only returns the bulk uploads associated with the current user.

This PR should not be reviewed until after #894

Resolves #759

### DIFF
--- a/src/__tests__/bulkUploads.int.test.ts
+++ b/src/__tests__/bulkUploads.int.test.ts
@@ -2,11 +2,15 @@ import request from 'supertest';
 import { app } from '../app';
 import {
 	createBulkUpload,
+	createUser,
 	loadSystemUser,
 	loadTableMetrics,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
-import { mockJwt as authHeader } from '../test/mockJwt';
+import {
+	mockJwt as authHeader,
+	mockJwtWithoutSub as authHeaderWithNoSub,
+} from '../test/mockJwt';
 import { BulkUploadStatus } from '../types';
 
 const agent = request.agent(app);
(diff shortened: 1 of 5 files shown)


reefdog:
@slifty I think the base of this PR was meant to be `485-model-proposal-ownership` instead of `485-model-ownership` based on the PR description, right?

As-is, I'm locally unable to actually add bulk uploads through the front-end interface, despite tests running clean. (That's worrisome!) The service console says: `null value in column "created_by" of relation "proposals" violates not-null constraint`

However, if I rebase this branch on (from? I always forget rebase semantics) `485-model-proposal-ownership`, it works fine, and I think that's the intent anyway.

It is a little odd that in this current state, tests are green while functionality is broken. Do you think the tests aren't capturing something?

reefdog:
Ohhh wait, I'll bet the problem is that my database was reflecting the migrations from `485-model-proposal-ownership`. Which does seem out of the realm of testing coverage / a weird dev edge case.

I can't tell, though, which branch you intended to ladder off of, since the PR desc says #894 but the actual base is from #892's branch. Can you clarify?

slifty:
@reefdog you know what -- I tried to be fancy since this PR doesn't actually rely on anything related to proposals but forgot that migrations do want to be run in order -- I had branched both of those features off the initial base, instead of nesting them linearly.  See what I get!

I'll rebase off of the proposals branch due to the migrations.

slifty:
@reefdog I think this should be ready for review now, sorry for the confusion.

reefdog (approved):
Whee, thank you! This works great now! Very cool to see it in action.

PR #900, opened 2024-04-11
Update `createdAt` to string in most cases

### LINKED ISSUE #862: Ensure our `createdAt` types are, correctly, strings
As we move to JSON for our database results our pg library is no longer casting timestamps as `Date` objects.  This isn't the end of the world, but it does mean our types need to reflect the change.

I've been trying to do this as I make the updates, but I believe I missed a few instances.  This should be corrected.

slifty:
This PR corrects our type definitions so that `createdAt` is a string for any object that is retrieved from the database using the json object pattern.

Resolves #862


hminsky2002 (approved):
Good call, approved!

PR #908, opened 2024-04-15
Return proposals for current user

### LINKED ISSUE #901: Only return `proposals` that the current user is allowed to view
Similar to #759, we want the API to only return proposal data that the requesting user is authorized to see.

slifty:
This PR adds authorization to our `GET /proposals` endpoint.

Resolves #901

### DIFF
--- a/src/handlers/proposalsHandlers.ts
+++ b/src/handlers/proposalsHandlers.ts
@@ -22,10 +22,15 @@ import {
 import type { Request, Response, NextFunction } from 'express';
 
 const getProposals = (
-	req: Request,
+	req: AuthenticatedRequest,
 	res: Response,
 	next: NextFunction,
 ): void => {
+	if (req.user === undefined) {
+		next(new FailedMiddlewareError('Unexpected lack of user context.'));
+		return;
+	}
+	const { user } = req;
 	const paginationParameters = extractPaginationParameters(req);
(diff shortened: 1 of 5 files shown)


src/handlers/proposalsHandlers.ts
```diff
@@ -22,10 +22,15 @@ import {
 import type { Request, Response, NextFunction } from 'express';
 
 const getProposals = (
-	req: Request,
+	req: AuthenticatedRequest,
 	res: Response,
 	next: NextFunction,
 ): void => {
+	if (req.user === undefined) {
+		next(new FailedMiddlewareError('Unexpected lack of user context.'));
```
reefdog, line 30: Tell me about it. _[shakes head]_ You and me both pal. _[raspberries in exasperation]_ What a day amiright

slifty: `Unexpected lack of joke.`

reefdog (approved):
This all looks good and is approved!

This approach doesn't lock us out of being able to *expand* scope for certain user types to see more than "their own" proposals, right? (Superadmins, e.g.)

slifty:
@reefdog correct!  We'll need to implement that functionality which might result in changes to the query (or might result in a check in the middleware to pass `0` in / bypass the query in the event of a superuser).

PR #909, opened 2024-04-15
Update test timestamp pattern

### LINKED ISSUE #906: Update tests: database can sometimes return a timestamp with fewer significant digits
We had CI fail: https://github.com/PhilanthropyDataCommons/service/actions/runs/8666216237/job/23766547108?pr=896

```
    @@ -1,7 +1,7 @@
      Object {
    -   "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{3,6})?(Z|(\+|-)\d{2}:\d{2})$/,
    +   "createdAt": "2024-04-12T18:11:26.18+00:00",
        "dataType": "string",
        "description": "😍",
        "id": Any<Number>,
        "label": "🏷️",
        "shortCode": "🩳",
```

It's because `11:26.18` doesn't match `d{2}:\d{2}(\.\d{3,6})?` because `.18` doesn't match `d{3,6}`

I think the solution is to update `d{3,6}` to `d{1,6}`

slifty:
This PR updates our timestamp to account for situations where the DB returns lower precision timestamps, which has now happened at least once in CI!

Resolves #906

### DIFF
--- a/src/test/utils.ts
+++ b/src/test/utils.ts
@@ -1,7 +1,7 @@
 import { createUser, loadUserByAuthenticationId } from '../database';
 
 export const isoTimestampPattern =
-	/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{3,6})?(Z|(\+|-)\d{2}:\d{2})$/;
+	/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{1,6})?(Z|(\+|-)\d{2}:\d{2})$/;
 
 export const expectTimestamp = expect.stringMatching(
 	isoTimestampPattern,


reefdog (approved):
This seems fine, but for the benefit of your co-devs (👋), could you provide a more concrete explanation about the old level of precision and the new one?

slifty:
@reefdog NO!

(ok fine)

Basically the previous pattern required 3 to 6 precision points in the number of milliseconds (or NO milliseconds)

This means `2024-04-12T18:11:26.18+00:00` failed (that has `.18` which is just 2 precision points)

The new regex will allow for 0 through 6 points of precision.

reefdog:
@slifty Thank you! That's excellent.

reefdog:
@slifty Oh my goodness, I just realized the attached issue contained the detail I needed. Apologies, and feel free in the future to just say that!

slifty:
@reefdog oh no problem at all -- I probably should have included it in the commit tbh since issue / PR get detached from the code!

I didn't think about the issue either 😅

hminsky2002 (approved):
Nice, justified as per your example!

PR #911, opened 2024-04-15
Fix organization duplicate result bug

### LINKED ISSUE #905: Organizations don'

slifty:
This PR fixes a bug where the `/organizations` endpoint would return multiple organizations if that organization had more than one associated proposal.

Resolves #905


reefdog (approved):
Works great, thank you!

Issue #899, opened 2024-04-10
Validate primitive types using ajv formats


hminsky2002:
Related to Issue #875
and Issue #760 

the ajv-formats package offers type validation for primitive types, such as number, string, and boolean. We should implement this in the proposal field validation and anywhere else we are validating types.

slifty:
@hminsky2002 you mentioned this in a [zulip thread](https://chat.opentechstrategies.com/#narrow/stream/75-PDC/topic/Data.20validation/near/189536) but wanted to bring it here -- I don't think we should be *changing* any of the values at this point (e.g. if coercion is resulting in value modification).  This would mean we don't want to strip white space, change case, or change the format of the value that is going to be stored.

We should be storing exactly what the user gave us to store, but adding a flag as to whether its format met our expectations.  We may add a "cleanup" step in a future date that attempts to correct the value they gave us.

slifty:
For history's sake: this wasn't a duplicate, but an ehnahncement of the validation (we wanted to use ajv to do the validation, the initial implementation used regex)

bickelj:
@slifty That makes sense, thanks!

PR #914, opened 2024-04-15
Validate URLs and Emails using ajv formats

### LINKED ISSUE #875: Validate URLs and Emails using ajv formats
Related to issue #760 

The current mode of email and URL validation in proposalFieldValues(`src/fieldValidation.ts`) is done using regexes which are a temporary solution. Validation for URLs and Emails should be done using ajv-formats https://github.com/ajv-validator/ajv-formats.

hminsky2002:
This commit adds the ajv-formats package that includes definitions for the uri and email formats,
and refactors the current validation regexes for proposal fields.

Closes #875


package.json
```diff
@@ -77,6 +77,7 @@
 		"@aws-sdk/client-s3": "^3.554.0",
 		"@aws-sdk/s3-presigned-post": "^3.554.0",
 		"ajv": "^8.12.0",
+		"ajv-formats": "^3.0.1",
```
slifty, line 80: We've tended to lean towards NOT having separate commits just to add packages with no context -- so just smoosh these two commits together (the thinking is that the context for why you're adding a package is the code that uses it, so just group it there!)

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -58,19 +50,4 @@ describe('field value validation against BaseFieldDataType', () => {
 			false,
 		);
 	});
-	test('validate a valid URL as URL', () => {
```
slifty, line 61: I suspect that removing these tests entirely will result in less code coverage -- let's make sure the basic URL  positive and negative case is *covered* by our code, even if we don't need to exhaustively test all the possible edge cases.

Likewise for the email tests.

Basically, +1 to removing a bunch of these `expects` but let's keep just one in each for the sake of ensuring we didn't completely space out on our testing.

src/fieldValidation.ts
```diff
@@ -11,11 +26,9 @@ export const fieldValueIsValid = (
 		case BaseFieldDataType.PHONE_NUMBER:
 			return phone(fieldValue).isValid;
 		case BaseFieldDataType.EMAIL:
-			return /^\S+@\S+$/.test(fieldValue);
+			return ajv.validate(emailSchema, fieldValue);
```
slifty, line 29: I'm torn on having the email / uri schemas removed from the context where the validation itself happens...

On the other hand, well named functions / variables are really helpful...

WHAT DO YOU THINK OF........

Defining a bunch of helper validation functions and calling those directly:

`isValidEmail`, `isValidPhoneNumber`, `isValidUrl` -- and then having all the actual implementations in there.  That way the `fieldValueIsValid` has absolutely no implementation whatsoever, and instead just routes the validation call to the appropriately scoped function -- this way the only time we edit this function is when adding a new format, but if the implementation of any given validator changes, we edit that validator.

Eh? eh?

slifty (changes requested):
Some ideas for changes for your consideration.

src/fieldValidation.ts
hminsky2002: sure! seems like a more scalable approach. 
Two questions:
1. Should those functions be defined in the same file(`fieldValidation.ts`)
2. Should the schemas be defined in the functions or outside the functions

hminsky2002: Ok did a refactor, let me know whatcha think, will be good to figure this out now so i can follow the pattern in issue #899

slifty: Good question!  Either is a fine approach at this point and might come down to personal taste; it feels a little verbose to have them in separate files BUT that is the pattern we've been using.  I tend to think "many files with well defined scope is better than fewer files with unbound scopes" is the way to go.

That said, having the definitions in the immediate context they're used is totally reasonable, and in this case these functions are ONLY used in that single context.

I'd probably put them in one file for now (and not exported), and if some day they get re-used OR the file gets hard to work with we can always break it into smaller pieces; that said if you'd prefer to break it up right away in the name of long term scalability / avoiding a future of a mega-file then I fully support that!

src/fieldValidation.ts
```diff
@@ -1,21 +1,45 @@
...
 	fieldValue: string,
 	dataType: BaseFieldDataType,
-) => {
+): boolean => {
 	switch (dataType) {
 		case BaseFieldDataType.NUMBER:
 			return /^[0-9]*$/.test(fieldValue);
 		case BaseFieldDataType.PHONE_NUMBER:
-			return phone(fieldValue).isValid;
+			return validatePhoneNumber(fieldValue);
```
slifty, line 38: Super small change: these functions return a boolean!  In that case it's handy to name them as an `isFoo` statement (e.g. `isValidPhoneNumber`).  This signals super clearly that the return value is going to be a boolean, and also reads more nicely in things like if statements

(e.g. `if (isValidPhoneNumber) then ...`

slifty (changes requested):
This request is edging on pedantic so feel free to tell me to pound sand!  That said, it's a good convention to adopt in your future that boolean functions are a little bit more self documenting / human readable if they use the `isFoo` syntax!

src/fieldValidation.ts
hminsky2002: the sand lives! I think this is a reasonable choice, refactoring now

slifty (approved):
Yay!

slifty:
I wonder why we aren't getting test coverage reports 🤔

src/fieldValidation.ts
reefdog: _First we taught the humans how to ergonomically conform to humanity, then they taught us. Gradually we stopped evaluating their advice through our own experience and began trusting it, implicitly. The changes began slowly at first, so slowly no one noticed, the arc of the tree blown gently in a steady wind growing leeward. But soon enough a visitor to that land knew it as The Place of Glass-Eyed Peoples. And later, simply The Place of the Glass Eyes._

PR #916, opened 2024-04-16
Allow public organizations access

### LINKED ISSUE #912: Allow unauthenticated users to access `/organizations` endpoint
Public users (unauthenticated) will need to be able to view both the [organizations table](https://github.com/PhilanthropyDataCommons/front-end/issues/560) and [organizations detail page](https://github.com/PhilanthropyDataCommons/front-end/issues/566). This means they'll need to be able to hit `GET /organizations` and `GET /organizations/{id}`

slifty:
This PR updates our `/organizations` endpoints so the GET varieties are public.  Eventually we may add private data to the returned objects, but that data would only be available for authorized users.

Resolves #912

### DIFF
--- a/src/__tests__/organizations.int.test.ts
+++ b/src/__tests__/organizations.int.test.ts
@@ -26,8 +26,8 @@ const insertTestOrganizations = async () => {
 
 describe('/organizations', () => {
 	describe('GET /', () => {
-		it('requires authentication', async () => {
-			await agent.get('/organizations').expect(401);
+		it('does not require authentication', async () => {
+			await agent.get('/organizations').expect(200);
 		});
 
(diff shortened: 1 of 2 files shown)


reefdog (approved):
This works perfectly, and it's so great to see such a transformative change require so little code. Good architecture!

I did notice that the Swagger spec for both endpoints still mentions a 401 as a possible response. I assume that's still a valid potential response, e.g. if a malformed auth token was provided?

slifty:
Oh great catch re the spec!  I'll fix that now (I'll say overall our swagger spec possible response codes have not been quite as robust as they should be; I also wish swagger would allow us to create reusable responses rather than manually writing the same 401 code a billion times but blah blah blah)

Issue #922, opened 2024-04-17
Consider a non-hardcoded mechanism for identifying specialized base fields


slifty:
As part of #902 we are going to have our bulk uploader treat a few base fields as "special" -- organization name and EIN, specifically.

I'd love to avoid hard coding special short codes that may or may not exist, but the initial implementation plan will do that.  The "right" answer may involve creating a new "configuration" table in the database, which would allow an administrator to select special base fields.

This issue is intended to capture that line of exploration.  It may turn out we don't want to do this!

bickelj:
@slifty I support the goal of a non-magic, non-hardcoded mechanism to identify certain data fields. Somewhere there has to be a link, though, right? Somebody has to mark data in such a way that the service can identify it.

One extreme is the caller could do whatever they want and the service could use a heuristics to magically guess which field is what. The other extreme is the service requires the caller to mark a couple of fields explicitly, e.g. with a particular string. Using a short code is this latter extreme. The benefit is simplicity and clarity. "You need to name it X for the system to know it is X." The service can say "error: you didn't supply X" or "error: X did not match expected regex."

I need to try the bulk uploader because I assumed that not only a few base fields would work this way but all base fields. The PDC team constantly curates the base fields, the uploaders map base fields to their fields via application forms, and proposals are uploaded to fit the application forms.

slifty:
> Somewhere there has to be a link, though, right? Somebody has to mark data in such a way that the service can identify it.

So the suggestion here would be that the link is ALSO determined by data -- which feels more explicit to me (the sysadmin, not the code, determines what base field *in their instance* should be used for this purpose).  One benefit of that is it's dynamic which means we can have an endpoint that communicates the configuration, and the front end can render the setting dynamically to the user to remove the sense of magic.

To be clear there are CERTAIN users to which this is all the same thing (bulk uploaders are going to rely on UX to prevent magic), but I think that having a data driven config approach would make the system more explicit for technical users, and for the system administrator.

bickelj:
@slifty What if we convert the base field name to be a documented enum or treat it that way conceptually? Isn't the new documented enum taking the place of the old base field name? It feels like it's both pushing the problem back one step (we don't have a hardcoded list of names, we have a hardcoded enum) plus adding some indirection. The net effect is more complexity and difficulty.

I see one benefit you mentioned: system administrators can vary their base fields per instance. But isn't one of the products of the PDC a flat, universal list of base fields? We don't want everybody using different base fields, I thought.

PR #924, opened 2024-04-17
Create `Organizations` and `OrganizationProposals` during bulk upload

### LINKED ISSUE #902: Update bulk upload processing to create organizations and link them to proposals
We have the concept of an organization, but don't currently create organizations for proposals when we invoke the bulk uploader.

slifty:
This PR updates the bulk upload process to also create any organizations and organization-proposal relationships for the proposals being uploaded.

Note that this PR follows a similar pattern to the old "applicant" logic, and uses magic / hard-coded base field short codes which *may or may not actually exist in the database*.  There's an open issue to explore possibly finding a way to possibly improve this approach / make the code base itself coupled to any given short codes: #922.

Resolves #902


hminsky2002 (approved):
This is so awesome! No longer having to manually create organizationproposal relationships will be a huge boon for development purposes. Looks great to me!

PR #925, opened 2024-04-18
Validate Primitives using Ajv

### LINKED ISSUE #899: Validate primitive types using ajv formats

hminsky2002:
This PR rids us of the last of the hard-coded regex's used in type validation of proposal field values. It utilizes the primitive type validation in ajv to validate numbers and booleans. It follows the pattern in PR #914 of refactoring the ```fieldValueIsValid``` function to use a series of validation functions, instead of performing the validation itself.

Closes #899

### DIFF
--- a/src/fieldValidation.ts
+++ b/src/fieldValidation.ts
@@ -1,31 +1,30 @@
-import Ajv from 'ajv';
-import addFormats from 'ajv-formats';
 import { phone } from 'phone';
 import { BaseFieldDataType } from './types';
+import { ajv } from './ajv';
 
-const ajv = new Ajv();
-addFormats(ajv, ['email', 'uri']);
+const isEmailString = ajv.compile({
+	type: 'string',
+	format: 'email',
+});
 
-function isValidEmail(email: string): boolean {
-	const emailSchema = {
-		type: 'string',
-		format: 'email',
-	};
+const isUrlString = ajv.compile({
+	type: 'string',
+	format: 'uri',
+});
 
-	return ajv.validate(emailSchema, email);
-}
+const isNumericString = ajv.compile({
+	type: 'number',
+});
 
-function isValidUrl(url: string): boolean {
-	const uriSchema = {
-		type: 'string',
-		format: 'uri',
-	};
-	return ajv.validate(uriSchema, url);
-}
+const isBooleanString = ajv.compile({
+	type: 'boolean',
+});
 
-function isValidPhoneNumber(phoneNumber: string): boolean {
-	return phone(phoneNumber).isValid;
-}
+const isString = ajv.compile({
+	type: 'string',
+});
+
+const isPhoneNumberString = (value: string) => phone(value).isValid;
 
 export const fieldValueIsValid = (
 	fieldValue: string,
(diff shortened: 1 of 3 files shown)


hminsky2002:
@slifty pushed some changes here, but wanted to wait for your feedback on the boolean validation:

```js
function isValidBoolean(bool: string): boolean {
// Converting to lower case to abide by ajv's coercion rules
	const formattedBoolean = bool.toLowerCase();

	const booleanSchema = {
		type: 'boolean',
	};

	return ajv.validate(booleanSchema, formattedBoolean);
}
```

slifty:
Let's *not* make any changes to the values before (or while) validating -- the goal here is not to ensure maximial passage, it's to define a specification and test against it.

src/fieldValidation.ts
```diff
@@ -3,9 +3,21 @@ import addFormats from 'ajv-formats';
 import { phone } from 'phone';
 import { BaseFieldDataType } from './types';
 
-const ajv = new Ajv();
+const ajv = new Ajv({ coerceTypes: true });
 addFormats(ajv, ['email', 'uri']);
 
+function isValidNumber(num: string): boolean {
```
slifty, line 9: A few things which apply to ALL of these functions

1.  We use arrow functions pretty much 100% of the time; they're better than functions (@jasonaowen and I had a super long exploration of this on another project and... I cannot find the thread at all).  The big benefit is that it protects you from weird scoping issues and dangerous access of `this`.

2.  We shouldn't name the variable `num` since the point of this is that we don't know if it's a num or not!  I'd name it something generic like `value` instead.

3. You could probably get away with typing the input as `unknown` (the idea is you're checking if a generic value is a number, so, if it's literally a number then that's fine too!)

4. Realizing you can (should?) remove `valid` since it's somewhat redundant.  For this specific case I would strongly consider `isNumericString` instead of `isNumber` since `isNumber` is going to be confusing.  Similarly, `isBooleanString` instead of `isBoolean`.

5. (This overrides 1-3... but I felt those were worth saying in general for edification!) `ajv` has a nifty function that will take a schema and return a validator.  It's `ajv.compile`.  You can see some examples of this in our types (for instance: https://github.com/PhilanthropyDataCommons/service/blob/e109501b2ce4cff25f25b7ba90050108135ac87d/src/types/Organization.ts#L27)

This doesn't apply here, but one cool thing about compile is that it also serves as a type guard (but that doesn't quite work).  It also populates itself with specific errors about why validation failed, which could be handy.

Using this you'd be able to turn your manually written functions into:

```
const isNumericString = ajv.compile({
  type: 'number',
});
```

slifty (changes requested):
Sorry for the delay in this -- I had mentally marked this PR as reviewed but... clearly hadn't!

src/fieldValidation.ts
```diff
@@ -20,11 +32,18 @@ function isValidUrl(url: string): boolean {
...
 		format: 'uri',
 	};
+
 	return ajv.validate(uriSchema, url);
 }
 
-function isValidPhoneNumber(phoneNumber: string): boolean {
-	return phone(phoneNumber).isValid;
+function isValidBoolean(bool: string): boolean {
+	const formattedBoolean = bool.toLowerCase();
```
slifty, line 40: Also yeah just to respond here too: I think that we shouldn't do this formatting / do any kind of cleanup on the values.  If what they gave us doesn't pass validation according to AJV then it's not valid according to the json specification, and so I think we should consider it invalid and it becomes the burden of the client to give us proper booleans.

hminsky2002: Right this makes sense long term, though I do wonder about presenting data back to users, but I suppose that's an issue for when we start working with real users

src/fieldValidation.ts
hminsky2002: Type-guards! These rock, thank you for the suggestion, and I appreciate the notes for if we weren't using ajv, they helped with making a wrapping function for the phoneNumber validation.

src/fieldValidation.ts
reefdog: @hminsky2002 Early on there was a conversation about whether we are data curators or data scavengers. Ultimately it would be good to have a curator (and we'd often like to be it), but at first we need to be scavengers. We have to tread carefully about imposing hygiene or structure to the data without buy-in from the various prospective integrators and stakeholders.

slifty: I hear ya BUT the data we are saving at this stage is the unchanged data, and saying that the unchanged data is valid when it isn't actually valid is a logical error.

To justin's point, eventually we may want to attempt to correct / massage the formatting of fields in certain cases (at which point we'd store both the raw data AND the fixed data), but that isn't what we're building here!

slifty (approved):
Beautiful!  This looks great.

PR #930, opened 2024-04-18
Restrict POST / PUT for /baseFields

### LINKED ISSUE #929: Restrict POST `/baseFields` endpoint to admins only

slifty:
This PR updates our `/baseFields` endpoints that change or create data so that only administrative users are able to access the endpoints.

Resolves #929

### DIFF
--- a/src/test/mockJwt.ts
+++ b/src/test/mockJwt.ts
@@ -8,26 +8,44 @@ const getMockJwks = (): JWKSMock =>
 
 const mockJwks = getMockJwks();
 
-const getMockJwt = (sub?: string): { Authorization: string } => {
+const getMockJwt = (
+	settings: {
+		sub?: string;
+		roles?: string[];
+	} = {},
+): { Authorization: string } => {
 	const aMomentAgo = Math.round(new Date().getTime() / 1000);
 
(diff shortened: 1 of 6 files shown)


src/test/mockJwt.ts
```diff
@@ -8,26 +8,44 @@ const getMockJwks = (): JWKSMock =>
 
 const mockJwks = getMockJwks();
 
-const getMockJwt = (sub?: string): { Authorization: string } => {
+const getMockJwt = (
+	settings: {
```
bickelj, line 12: This isn't really a problem, but why not multiple args to `getMockJwt` instead of a `settings` map?

bickelj (approved):
It does what it says on the tin! This is clear and concise. Thank you and bravo.

src/test/mockJwt.ts
slifty: @bickelj main reason both args are optional -- so multiple args would mean you have to pass unknown explicitly if the only setting you wanted was the second.  Secondary reason is that for mock generators that return a complex object I like the pattern of "pass in override values matching the ultimately mocked object" -- this isn't QUITE that just yet, but I could se it getting there.

PR #934, opened 2024-04-18
Add authorization to GET `/proposals/:proposalId`

### LINKED ISSUE #932: Don't allow unauthenticated proposal access
We recently updated the GET `/proposals` endpoint to only return proposals owned by the current user, but we did NOT add any restrictions to specific GET `/proposal/:proposalId` requests.  This means someone could access any proposal via the API if they knew the proposal ID.

Instead, ownership rules should apply to all endpoints.

slifty:
This PR updates the GET `/proposals/:proposalId` endpoint so that it will return a 404 if a user is attempting to load a proposal they don't have access to.

Resolves #932

### DIFF
--- a/src/database/operations/assert/assertProposalAuthorization.ts
+++ b/src/database/operations/assert/assertProposalAuthorization.ts
@@ -1,0 +1,17 @@
+import { db } from '../../db';
+import { NotFoundError } from '../../../errors';
+import type { CheckResult } from '../../../types';
+
+export const assertProposalAuthorization = async (
+	id: number,
+	userId: number,
+): Promise<void> => {
+	const result = await db.sql<CheckResult>('proposals.checkAuthorization', {
+		id,
+		userId,
+	});
+
+	if (result.rows[0]?.result !== true) {
+		throw new NotFoundError(`The proposal was not found (id: ${id})`);
+	}
+};
--- a/src/database/queries/proposals/selectById.sql
+++ b/src/database/queries/proposals/selectById.sql
@@ -1,3 +1,3 @@
 SELECT proposal_to_json(proposals.*) as "object"
-FROM proposals
-WHERE id = :id;
+  FROM proposals
+  WHERE id = :id
--- a/src/handlers/proposalsHandlers.ts
+++ b/src/handlers/proposalsHandlers.ts
@@ -1,4 +1,5 @@
 import {
+	assertProposalAuthorization,
 	createProposal,
 	getLimitValues,
 	loadProposal,
@@ -6,6 +7,7 @@ import {
 } from '../database';
 import {
 	AuthenticatedRequest,
+	isId,
 	isTinyPgErrorWithQueryContext,
 	isWritableProposal,
 } from '../types';
@@ -19,7 +21,7 @@ import {
 	extractPaginationParameters,
 	extractSearchParameters,
 } from '../queryParameters';
-import type { Request, Response, NextFunction } from 'express';
+import type { Response, NextFunction } from 'express';
 
 const getProposals = (
 	req: AuthenticatedRequest,
@@ -53,26 +55,31 @@ const getProposals = (
 };
 
 const getProposal = (
-	req: Request<{ id: string }>,
+	req: AuthenticatedRequest,
 	res: Response,
 	next: NextFunction,
 ): void => {
-	const id = Number.parseInt(req.params.id, 10);
-	if (Number.isNaN(id)) {
-		next(new InputValidationError('The entity id must be a number.', []));
+	const { user } = req;
+	const { proposalId } = req.params;
+	if (user === undefined) {
+		next(new FailedMiddlewareError('Unexpected lack of user context.'));
 		return;
 	}
-	loadProposal(id)
-		.then((proposal) => {
-			res.status(200).contentType('application/json').send(proposal);
-		})
-		.catch((error: unknown) => {
-			if (isTinyPgErrorWithQueryContext(error)) {
-				next(new DatabaseError('Error loading the proposal.', error));
-				return;
-			}
-			next(error);
-		});
+	if (!isId(proposalId)) {
+		next(new InputValidationError('Invalid id parameter.', isId.errors ?? []));
+		return;
+	}
+	(async () => {
+		await assertProposalAuthorization(proposalId, user.id);
+		const proposal = await loadProposal(proposalId);
+		res.status(200).contentType('application/json').send(proposal);
+	})().catch((error: unknown) => {
+		if (isTinyPgErrorWithQueryContext(error)) {
+			next(new DatabaseError('Error loading the proposal.', error));
+			return;
+		}
+		next(error);
+	});
 };
 
 const postProposal = (
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -511,7 +511,40 @@ describe('/proposals', () => {
 			});
 		});
 
-		it('returns 400 when given id a string', async () => {
+		it('returns 404 when given id is not owned by the current user', async () => {
+			await db.query(`
+        INSERT INTO opportunities (
+          title,
+          created_at
+        )
+        VALUES
+          ( '⛰️', '2525-01-03T00:00:01Z' )
+      `);
+			const anotherUser = await createUser({
+				authenticationId: 'totallySeparateUser',
+			});
+			await createProposal({
+				externalId: `proposal-1`,
+				opportunityId: 1,
+				createdBy: anotherUser.id,
+			});
+
+			const response = await agent
+				.get('/proposals/1')
+				.set(authHeader)
+				.expect(404);
+			expect(response.body).toEqual({
+				name: 'NotFoundError',
+				message: expect.any(String) as string,
+				details: [
+					{
+						name: 'NotFoundError',
+					},
+				],
+			});
+		});
+
+		it('returns 400 when given id is a string', async () => {
 			const response = await agent
 				.get('/proposals/foobar')
 				.set(authHeader)
@@ -519,7 +552,7 @@ describe('/proposals', () => {
 			expect(response.body).toEqual({
 				name: 'InputValidationError',
 				message: expect.any(String) as string,
-				details: [],
+				details: [expect.any(Object)],
 			});
 		});
 
--- a/src/database/operations/assert/index.ts
+++ b/src/database/operations/assert/index.ts
@@ -1,0 +1,1 @@
+export * from './assertProposalAuthorization';
--- a/src/database/operations/index.ts
+++ b/src/database/operations/index.ts
@@ -1,3 +1,4 @@
+export * from './assert';
 export * from './create';
 export * from './load';
 export * from './update';
--- a/src/database/queries/proposals/checkAuthorization.sql
+++ b/src/database/queries/proposals/checkAuthorization.sql
@@ -1,0 +1,12 @@
+SELECT EXISTS (
+  SELECT 1
+    FROM proposals
+    WHERE id = :id
+      AND
+        CASE
+          WHEN :userId != 0 THEN
+            created_by = :userId
+          ELSE
+            true
+          END
+) AS result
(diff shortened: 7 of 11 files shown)


src/handlers/proposalsHandlers.ts
```diff
@@ -53,16 +54,22 @@ const getProposals = (
...
+		next(new FailedMiddlewareError('Unexpected lack of user context.'));
 		return;
 	}
-	loadProposal(id)
+	if (!isId(proposalId)) {
+		next(new InputValidationError('Invalid id parameter.', isId.errors ?? []));
+		return;
+	}
+
+	loadProposal(proposalId, user.id)
```
slifty, line 72: I just want to flag this approach of adding the "createdBy" parameter to the loader itself.  This doesn't deeply couple authorization to the query necessarily, BUT it does risk creating that pattern.

One alternative would be to load the object and THEN check the user access.  That approach might be more prone to timing attacks (e.g. a user could detect if a proposal ID existed vs just not having access based on how long the server took to respond).

Another alternative would be to add something like `checkProposalAuthorization` which is always run *before* loading a proposal regardless of existence (and would return false in either access case).

I opted for this approach for now because it felt more straightforward given the simple nature of authorization that we have today, but honestly it might be better to just start with a `checkProposalAuthorization` pattern since it clearly separates the auth from the object lookup, and is more scalable long term.

slifty: (I'm probably going to switch to `checkProposalAuthorization` (name TBD)

src/database/operations/assert/assertProposalAuthorization.ts
```diff
@@ -0,0 +1,17 @@
...
+	id: number,
+	userId: number,
+): Promise<void> => {
+	const result = await db.sql<CheckResult>('proposals.checkAuthorization', {
+		id,
+		userId,
+	});
+
+	if (result.rows[0]?.result !== true) {
+		throw new NotFoundError(`The proposal was not found (id: ${id})`);
```
slifty, line 15: This error looks incorrect, but we want to match the "not found" response from a security PoV.  We may want to refactor this so that the middleware is what smooths this out.

For instance: Maybe `NotFoundError` should be changed to `NotFoundOrNoAccessError` and the middleware should strip out the message.

bickelj:
@slifty I haven't tried it yet locally, but this is in response to your above "[404] error looks incorrect" comment.

I think public (unauthenticated) users should already get a 401 on every attempt to access any proposal whether it exists or not. So unauthenticated users shouldn't be able to discover whether a proposal ID exists. I think if a user is already authenticated but does not have access to a particular proposal, the clarity given by a 403 instead of a 404 outweighs the risks posed by an already-authenticated user discovering there exists or does not exist a proposal for a given ID. In other words, I would expect 401 for all attempts for public users, 404 for authenticated users when the ID truly does not exist, and 403 for authenticated users who are not permitted to access the resource.

src/database/operations/load/loadProposal.ts
```diff
@@ -2,9 +2,13 @@ import { db } from '../../db';
 import { NotFoundError } from '../../../errors';
 import type { JsonResultSet, Proposal } from '../../../types';
 
-export const loadProposal = async (id: number): Promise<Proposal> => {
+export const loadProposal = async (
+	id: number,
+	createdBy: number = 0,
```
bickelj: I'm of two minds about it. First, why complicate it? But then I suppose it doesn't make sense to select a pile of joined stuff and later look at it on the application level to verify that it has the right user.

src/database/queries/proposals/selectById.sql
bickelj, line 1: Do we want to complicate the simple "select by ID" query with the additional "and by createdBy if present" or should we add a `.sql` file for `selectByIdAndCreatedBy`?

slifty: I think you may be reviewing an old version -- I took that out.

src/database/operations/load/loadProposal.ts
bickelj, line 1: (This comment is for the commit message not loadProposal.ts but I don't see a way to comment on the commit message.) I see " Add ability specify createdBy when loading a proposal" is missing a word but is already over the 50-char soft limit for commit titles.

src/database/operations/assert/assertProposalAuthorization.ts
bickelj: Caveat: I haven't tried it yet locally.

I think public (unauthenticated) users should already get a 401 on every attempt to access any proposal whether it exists or not. So unauthenticated users shouldn't be able to discover whether a proposal ID exists. I think if a user is already authenticated but does not have access to a particular proposal, the clarity given by a 403 instead of a 404 outweighs the risks posed by an already-authenticated user discovering there exists or does not exist a proposal for a given ID. In other words, I would expect 401 for all attempts for public users, 404 for authenticated users when the ID truly does not exist, and 403 for authenticated users who are not permitted to access the resource.

At the same time, I understand the reasoning and it is a judgment call.

bickelj: Hmm, I guess when doing a pre-flight query this way, there is no way to tell if it is due to not being found or not having access, interesting.

slifty: So my thinking here is: do we want the user to know a resource exists that they don't have access to?  IMHO even if they have an account the answer is no, but that's a generic / conservative opinion and if we think folks should know about proposals they can't read then that's OK too.

If we want 403 instead of 404 for items that exist with no access I can modify the preflight check so that instead of returning a boolean it returns something to indicate "no access" vs "doesn't exist"

src/handlers/proposalsHandlers.ts
```diff
@@ -53,26 +55,31 @@ const getProposals = (
...
+	req: AuthenticatedRequest,
 	res: Response,
 	next: NextFunction,
 ): void => {
-	const id = Number.parseInt(req.params.id, 10);
-	if (Number.isNaN(id)) {
-		next(new InputValidationError('The entity id must be a number.', []));
+	const { user } = req;
+	const { proposalId } = req.params;
+	if (user === undefined) {
```
bickelj, line 64: This is a little frustrating. It's already an `AuthenticatedRequest` so why is TypeScript so daft that it needs an additional check? Ah, because we defined `AuthenticatedRequest` loosely, it may or may not have `user` and may or may not have `role`. Would it make sense to tighten that up a bit such that `AuthenticatedRequest` gives a stronger guarantee?

src/handlers/proposalsHandlers.ts
```diff
@@ -1,11 +1,13 @@
 import {
+	assertProposalAuthorization,
 	createProposal,
 	getLimitValues,
 	loadProposal,
 	loadProposalBundle,
 } from '../database';
 import {
 	AuthenticatedRequest,
+	isId,
```
bickelj, line 10: I wonder if `isId` should set the maximum at the JavaScript maximum of 2^53-1? Then again, a postgresql ID can go higher than that. Except that we are using plain 32-bit ints, so it's lower. Hmm, I digress. It's OK.

src/handlers/proposalsHandlers.ts
slifty: Unfortunately that isn't allowed, but for good reason: the problem is that we can't know for sure from the handlers that the middleware actually ran -- so while we could tighten the type, it creates a situation where we lose a little bit of type safety, since there's no contractual guarantee that the type is accurate.

If we try to make it tighter, then TypeScript will (correctly) error when we try to pass the handler to the express route, saying that the handler expects something with a defined user but express is not promising that user will exist.

(you can see this if you remove the `?` in the type and try to run)

bickelj: I see. Yes, I would only want to tighten it up if it is possible to tighten it up. I guess this is the first downside of the fine-grained middleware approach I've seen. And it's not that bad, we just check to see if the expected data are present.

bickelj:
OK, taking a big step back, why do we want to restrict access to `GET` on proposal ID? Shouldn't every authenticated user be able to read every proposal? It would be `PUT`, `POST`, `PATCH` that we'd want to restrict, right?

src/database/operations/load/loadProposal.ts
bickelj: Resolved by pre-flight auth select.

slifty:
@bickelj We're changing the auth structure so folks only have access to their own proposals (unless they have some type of more expanded access, e.g. is a superuser).  In phase 3 we'll be adding more user types, but for now we landed on:

1) public user
2) user with access to their own data
3) administrator

src/database/operations/assert/assertProposalAuthorization.ts
bickelj: I think if you're authenticated to the PDC you have broad read access and narrow write access. I do want the 403 for authenticated users that do not have authorization. I am also assuming that proposals can be read almost all the time by almost all PDC users and that the exception is to deny access. I imagine someone on a call saying "let's look at proposal 944" and if one person gets a "not found" while the other sees the proposal that would be very confusing. Either it exists or doesn't. `NotFoundOrUnauthorizedError` would also make it a bit clearer that it could be either. I am curious what others think, though.

bickelj (approved):
I would like others who have better context to chime on the "should authenticated users see whether another proposal with ID X exists without seeing the contents?" question, but otherwise this looks good to me. @reefdog @jmergy? I wasn't aware of the narrowing of visibility so I defer to @slifty and the sense of the team.

src/database/operations/assert/assertProposalAuthorization.ts
slifty: Ah I think there may be a mismatch / things shifted when you weren't on call -- we're adding these checks in order to limit access to most proposals for most users right now (though I don't think that "this is an edge case" should necessarily drive whether 404 or 403 is the right call -- if we want to let people know information about data they can't access then that probably wouldn't depend on how many people fit into that category).

What about for other kinds of lookup / natural keys?  E.g. what if someone is looking up an organization by EIN and it exists but they don't have access to that fact?  It would be nice to have predictable / consistent behavior that will handle a variety of cases.

FWIW: my understanding is that it's normal for an API to either ALWAYS use 404 or ALWAYS use 403 for resource access in order to avoid sharing information about entity state to unauthorized users.  Not saying we have to do it, but just wanted to make sure it's clear that it isn't a new concept / something that we're inventing.

For your example would it be less confusing if we always returned 403 rather than 404?

bickelj: My comments assumed we were in a largely "the purpose of PDC is data sharing among PDC folks" mode. If that has changed, then yeah, I think being more opaque and restrictive makes sense.

bickelj: > What about for other kinds of lookup / natural keys? E.g. what if someone is looking up an organization by EIN and it exists but they don't have access to that fact?

I would lean toward "no, that's not a mostly-meaningless auto-generated ID so limit access there", but again, if we were in a PDC is for data sharing among PDC folks mode, I would be puzzled that authenticated users couldn't see that. The ground has shifted, perhaps.

bickelj: > For your example would it be less confusing if we always returned 403 rather than 404?

I think the least confusing is distinguishing not found from not authorized but yes, returning 403 every time is better than returning 404 every time.

reefdog:
@bickelj "global read, narrow write" will definitely be a user type eventually, we just settled on those three as the MVP to get all the necessary bits and bobs in place for more.

> I would like others who have better context to chime on the "should authenticated users see whether another proposal with ID X exists without seeing the contents?" question

403 is semantically sensible (and spec-correct), but the design of showing 404 for private resources so as not to confirm entity presence is established best practice now, AFAIK. (Case in point, hit this URL while logged in: https://github.com/reefdog/PrivateTest)

slifty:
If it's OK with the crowd I'd like to propose that we open an issue for further looking at the 404 / 403 discussion.  In *either* event I don't love the current implementation, but improving it would be a refactor outside of the scope of this issue.

I'll open the issue and then merge this PR.

PR #937, opened 2024-04-19
Add creator filter to proposal

slifty:
This PR improves the `/proposals` endpoint in two ways:

1) It allows administrators to see all proposals.
2) It adds an optional `createdBy` filter which allows a user to filter in terms of author.

Related to #928 
Related to #927

### DIFF
--- a/src/database/operations/load/loadProposalBundle.ts
+++ b/src/database/operations/load/loadProposalBundle.ts
@@ -1,23 +1,42 @@
 import { loadBundle } from './loadBundle';
-import type { JsonResultSet, Bundle, Proposal } from '../../../types';
+import type {
+	JsonResultSet,
+	Bundle,
+	Proposal,
+	AuthContext,
+} from '../../../types';
 
-export const loadProposalBundle = async (queryParameters: {
-	offset: number;
-	limit: number;
-	search?: string;
-	organizationId?: number;
-	createdBy?: number;
-}): Promise<Bundle<Proposal>> => {
+export const loadProposalBundle = async (
+	queryParameters: {
+		offset: number;
+		limit: number;
+		search?: string;
+		organizationId?: number;
+		createdBy?: number;
+	},
+	authContext?: AuthContext,
+): Promise<Bundle<Proposal>> => {
 	const defaultQueryParameters = {
 		search: '',
 		organizationId: 0,
 		createdBy: 0,
+		userId: 0,
 	};
+	const { offset, limit, search, organizationId, createdBy } = queryParameters;
+	const userId = authContext?.user.id;
+	const isAdministrator = authContext?.role.isAdministrator;
+
 	const bundle = await loadBundle<JsonResultSet<Proposal>>(
 		'proposals.selectWithPagination',
 		{
 			...defaultQueryParameters,
-			...queryParameters,
+			offset,
+			limit,
+			search,
+			organizationId,
+			createdBy,
+			userId,
+			isAdministrator,
 		},
 		'proposals',
 	);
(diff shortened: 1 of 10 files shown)


src/database/operations/load/loadProposalBundle.ts
```diff
@@ -1,23 +1,42 @@
...
-}): Promise<Bundle<Proposal>> => {
+export const loadProposalBundle = async (
+	queryParameters: {
+		offset: number;
+		limit: number;
+		search?: string;
+		organizationId?: number;
+		createdBy?: number;
+	},
+	authContext?: AuthContext,
```
slifty, line 17: I think this is a much cleaner way to handle authorization context in database calls.

(1) It decouples filters from authorization rules.
(2) It's an optional parameter, so if an auth context is not provided then it won't be used.

Our previous implementation "worked" but would have gotten messy really fast for, well, this exact filter!

src/types/AuthContext.ts
```diff
@@ -0,0 +1,32 @@
+import { ajv } from '../ajv';
+import { userSchema } from './User';
+import type { JSONSchemaType } from 'ajv';
+import type { User } from './User';
+
+interface AuthContext {
```
slifty, line 6: I mention this in the commit message, but to highlight it here: by having a more deterministic definition of AuthContext we can pass an auth context around to other places (such as loader functions).  The sooner we "tighten" the type to signal that both user and role ARE present, the better.

src/handlers/bulkUploadsHandlers.ts
bickelj, line 1: I saw two branches and wondered if this change is supposed to be here.

bickelj: I assume so, I was viewing changes outside of commits when I made that comment.

src/types/AuthContext.ts
bickelj: Double-plus good!

src/types/express/AuthenticatedRequest.ts
```diff
@@ -1,11 +1,6 @@
...
-import { User } from '../User';
+import type { AuthContext } from '../AuthContext';
 
-interface AuthenticatedRequest extends JwtRequest {
-	user?: User;
-	role?: {
-		isAdministrator: boolean;
-	};
-}
+type AuthenticatedRequest = JwtRequest & Partial<AuthContext>;
```
bickelj, line 4: Nice!

slifty:
Realizing I need tests cc #938

src/queryParameters/extractCreatedByParameters.ts
```diff
@@ -0,0 +1,42 @@
+import { ajv } from '../ajv';
+import { InputValidationError } from '../errors';
+import type { JSONSchemaType } from 'ajv';
+import type { Request } from 'express';
+
+interface CreatedByParameters {
+	createdBy: number | undefined;
+}
+
+interface CreatedByParametersQuery {
```
bickelj, line 10: As of this commit, this and the above `CreatedByParameters` are the exact same interface. Would anything bad happen by removing one of them, e.g. this one, and re-using the one interface below?

slifty: good call!

bickelj (approved):
I have not done any local testing other than pull, compile, run automated tests, but I see nothing alarming here. In fact, I see good stuff.

reefdog (approved):
_This is only a functional review, not a code review, which I'll leave to @bickelj and @hminsky2002._

This is great! It works. I tested the following:

- As an admin user:
  - ✅ No `createdBy` returned everything
  - ✅ `createdBy=#` returned proposals just for that user
- As a non-admin user:
  - ✅ No `createdBy` returned only my proposals
  - ✅ `createdBy=#` for a user who wasn't me returned nothing
  - ✅ `createdBy=#` for me returned my proposals

slifty:
I rebased and added some unit tests; I'm gonna forge ahead!

PR #941, opened 2024-04-22
Add createdBy parameter to /bulkUploads

### LINKED ISSUE #927: Allow administrative users to view all data regardless of ownership

### LINKED ISSUE #928: Add ability to filter by creator

slifty:
This PR adds the `createdBy` filter to `/bulkUploads` in addition to expanding the default return set for administrators.

Resolves #928
Resolves #927

### DIFF
--- a/src/__tests__/bulkUploads.int.test.ts
+++ b/src/__tests__/bulkUploads.int.test.ts
@@ -10,6 +10,7 @@ import { expectTimestamp, loadTestUser } from '../test/utils';
 import {
 	mockJwt as authHeader,
 	mockJwtWithoutSub as authHeaderWithNoSub,
+	mockJwtWithAdminRole as authHeaderWithAdminRole,
 } from '../test/mockJwt';
 import { BulkUploadStatus } from '../types';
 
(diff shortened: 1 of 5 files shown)


bickelj (approved):
I poked at this a bit locally but I do not (yet) have local bulk uploads so my testing was pretty limited. Regardless, LGTM!

src/__tests__/bulkUploads.int.test.ts
```diff
@@ -94,6 +95,95 @@ describe('/bulkUploads', () => {
...
 
+		it('returns all bulk uploads for administrative users', async () => {
+			const testUser = await loadTestUser();
+			const anotherUser = await createUser({
+				authenticationId: 'totallyDifferentUser@example.com',
+			});
+			await createBulkUpload({
+				fileName: 'foo.csv',
+				sourceKey: '96ddab90-1931-478d-8c02-a1dc80ae01e5-foo',
+				status: BulkUploadStatus.PENDING,
```
reefdog, line 106: Just out of curiosity, why make this a pending upload?

reefdog (approved):
Functional approval only: this works great.

- Admin hitting `/bulkUploads` gets everything, but can also add `createdBy=` to scope it.
- Non-admin hitting `/bulkUploads` gets only their own stuff, but adding `createdBy={any ID except their own}` returns no results

src/__tests__/bulkUploads.int.test.ts
slifty: Funsies (pending is the starting state of a bulk upload, so it's just as good a state as any really)

PR #946, opened 2024-04-22
Add GET `/users` endpoint

### LINKED ISSUE #942: Create `/users` endpoint
Users are generally an internal entity right now but we have some need to expose the user ID for the current user in order to supply `createdBy` filters (see #939).

So, let's make a GET `/users` endpoint possibly with support for an `authenticationId` filter, and possibly with a special `/users/me` endpoint as well.

In terms of access control I could either:

1. make `/users` require the administrative role.
2. make `/users` require an auth context and then restrict the users returned based on auth context (e.g. if you aren't an admin the only user you can see is yourself)
3. do none of the above and just create a `/users/me` that returns your single authenticated user.

I'm planning to do both 2 + 3.

slifty:
This PR adds a new `/users` endpoint which will allow the front end to look up the id associated with their session.

I decided to implement this with auth context awareness off the bat, which means administrative users can actually view ALL users in the system if they want.

Resolves #942

### DIFF
--- a/src/database/operations/load/loadUserBundle.ts
+++ b/src/database/operations/load/loadUserBundle.ts
@@ -1,0 +1,38 @@
+import { loadBundle } from './loadBundle';
+import type { JsonResultSet, Bundle, User, AuthContext } from '../../../types';
+
+export const loadUserBundle = async (
+	queryParameters: {
+		offset: number;
+		limit: number;
+		authenticationId?: string;
+	},
+	authContext?: AuthContext,
+): Promise<Bundle<User>> => {
+	const defaultQueryParameters = {
+		authenticationId: '',
+		userId: 0,
+		isAdministrator: false,
+	};
+	const { offset, limit, authenticationId } = queryParameters;
+	const userId = authContext?.user.id;
+	const isAdministrator = authContext?.role.isAdministrator;
+
+	const bundle = await loadBundle<JsonResultSet<User>>(
+		'users.selectWithPagination',
+		{
+			...defaultQueryParameters,
+			offset,
+			limit,
+			authenticationId,
+			userId,
+			isAdministrator,
+		},
+		'users',
+	);
+	const entries = bundle.entries.map((entry) => entry.object);
+	return {
+		...bundle,
+		entries,
+	};
+};
--- a/src/queryParameters/extractAuthenticationIdParameters.ts
+++ b/src/queryParameters/extractAuthenticationIdParameters.ts
@@ -1,0 +1,41 @@
+import { ajv } from '../ajv';
+import { InputValidationError } from '../errors';
+import type { JSONSchemaType } from 'ajv';
+import type { Request } from 'express';
+
+interface AuthenticationIdParameters {
+	authenticationId: string | undefined;
+}
+
+const authenticationIdParametersSchema: JSONSchemaType<AuthenticationIdParameters> =
+	{
+		type: 'object',
+		properties: {
+			authenticationId: {
+				type: 'string',
+				nullable: true,
+			},
+		},
+		required: [],
+	};
+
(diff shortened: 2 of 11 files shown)


slifty:
I *think* all of the untested lines are of the category of "it's OK not to test them"

src/database/operations/load/loadUserBundle.ts
```diff
@@ -0,0 +1,43 @@
...
+} from '../../../types';
+
+export const loadUserBundle = async (
+	queryParameters: {
+		offset: number;
+		limit: number;
+		authenticationId?: string;
+	},
+	authContext?: AuthContext,
+): Promise<Bundle<Proposal>> => {
```
bickelj, line 16: Is this supposed to be a bundle of Proposal or of User?

src/database/operations/load/loadUserBundle.ts
```diff
@@ -0,0 +1,43 @@
...
+	const defaultQueryParameters = {
+		authenticationId: '',
+		userId: 0,
+		isAdministrator: false,
+	};
+	const { offset, limit, authenticationId } = queryParameters;
+	const userId = authContext?.user.id;
+	const isAdministrator = authContext?.role.isAdministrator;
+
+	const bundle = await loadBundle<JsonResultSet<Proposal>>(
```
bickelj, line 26: Same question here: `Proposal` or `User`?

reefdog (approved):
Functional review only: this works!

- ✅ Admin hitting `/users` got data about each user
- ✅ Admin hitting `/users?authenticationId={sub}` got data about that user (or an empty list)
- ✅ Non-admin hitting `/users` got data about only themselves
- ✅ Non-admin hitting `/users?authenticationId={sub}` got an empty list unless `{sub}` was themselves
- ✅ Public users got a 401.
  - I did notice the body of the 401 was an `UnknownError` and wasn't sure if that was correct, but I leave that up to you:
  ```json
  {
    "name": "UnknownError",
    "message": "No authorization token was found.",
    "details": [
      {}
    ]
  }
  ```

src/database/operations/load/loadUserBundle.ts
slifty: Great catch!  User!  This is why casting types can lead to bugs 😅

src/database/operations/load/loadUserBundle.ts
slifty: Fixed up!  Thank you.

src/queryParameters/extractAuthenticationIdParameters.ts
```diff
@@ -0,0 +1,41 @@
...
+		properties: {
+			authenticationId: {
+				type: 'string',
+				nullable: true,
+			},
+		},
+		required: [],
+	};
+
+const isAuthenticationIdParameters = ajv.compile(
```
hminsky2002, line 22: This is one of those type guards we were discussing correct? Very cool!

src/routers/usersRouter.ts
```diff
@@ -0,0 +1,9 @@
+import express from 'express';
+import { usersHandlers } from '../handlers/usersHandlers';
+import { requireAuthentication } from '../middleware';
+
+const usersRouter = express.Router();
+
+usersRouter.get('/', requireAuthentication, usersHandlers.getUsers);
```
hminsky2002, line 7: All relevant boilerplate looks to have been added(handlers and routers)

hminsky2002 (commented):
Looks good and now with fixed tests, going to approve after testing locally!

hminsky2002:
Verifying success on all of the same checks as @reefdog . Hitting approve!

hminsky2002 (approved):

bickelj (approved):
I see the expected function signatures now, LGTM!

PR #948, opened 2024-04-23
Fix migration hang

### LINKED ISSUE #947: Migrations hang for several seconds at the end of the task
After the last apparent step of `npm run migrate` (`[2024-04-22T22:15:05.486Z] INFO (15994): Migrations complete.`), the console seems to hang for several seconds before exiting back to the terminal prompt.

@slifty says this is likely a bug.

slifty:
This PR closes the migration process in the event that they are successful.

You should now be able to run `npm run migrate` and it will exit right after migrations complete without requiring you to close the process manually.

Resolves #947

### DIFF
--- a/src/scripts/migrate.ts
+++ b/src/scripts/migrate.ts
@@ -1,17 +1,19 @@
 import { getLogger } from '../logger';
-import { migrate } from '../database';
+import { db, migrate } from '../database';
 
 const logger = getLogger(__filename);
 
 logger.info('Starting migrations...');
 migrate()
-	.then(() => {
+	.then(async () => {
 		logger.info('Migrations complete.');
+		await db.close();
 	})
-	.catch((reason: unknown) => {
+	.catch(async (reason: unknown) => {
 		logger.error('Migrations failed!');
 		if (reason instanceof Error) {
 			logger.error(`${reason.message}`);
 		}
+		await db.close();
 		process.exit(1);
 	});


reefdog (approved):
Huzzah! Much better! Thank you for the quick turnaround.

src/scripts/migrate.ts
```diff
@@ -7,6 +7,7 @@ logger.info('Starting migrations...');
 migrate()
 	.then(() => {
 		logger.info('Migrations complete.');
+		process.exit(0);
```
bickelj, line 10: It seems odd that this is needed and makes me curious what was holding it up, but if this works it works.

slifty: All right this was a good point so I fixed it more correctly!  Gonna ask for a re-review since it's fundamentally different from what you approved.

src/scripts/migrate.ts
```diff
@@ -1,17 +1,19 @@
...
+		await db.close();
 	})
-	.catch((reason: unknown) => {
+	.catch(async (reason: unknown) => {
 		logger.error('Migrations failed!');
 		if (reason instanceof Error) {
 			logger.error(`${reason.message}`);
 		}
+		await db.close();
 		process.exit(1);
```
slifty, line 18: Keeping this because I think we should exit with a `1` status code if migrations fail.

bickelj: We probably could re-throw here or something but this is less of a problem because we're already in an error state so we don't care.

bickelj (approved):
It consistently takes much less time. I didn't see an indefinite hang before, only a weird pause for a few seconds, but I see a speedup with this change. Thank you!

reefdog (approved):
Once again, functional review only: it works! It's fast! It doesn't hang! I am satisfied. You may live.

PR #950, opened 2024-04-23
Fix proposal versions test

### LINKED ISSUE #877: Check for race condition in `proposalVersions.int.test.ts`
I just saw a failed test locally which went away on re-run:

```
FAIL  src/__tests__/proposalVersions.int.test.ts
  ● /proposalVersions › POST / › creates exactly the number of provided field values

    expect(received).toMatchObject(expected)

    - Expected  - 2
    + Received  + 2

    @@ -1,19 +1,19 @@
      Object {
        "fieldValues": Array [
          Object {
            "applicationFormFieldId": 1,
            "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{3,6})?(Z|(\+|-)\d{2}:\d{2})$/,
    -       "id": 1,
    +       "id": 2,
            "isValid": true,
            "position": 1,
            "value": "Gronald",
          },
          Object {
            "applicationFormFieldId": 2,
            "createdAt": StringMatching /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{3,6})?(Z|(\+|-)\d{2}:\d{2})$/,
    -       "id": 2,
    +       "id": 1,
            "isValid": true,
            "position": 1,
            "value": "Plorp",
          },
        ],

      149 | 			logger.debug('after: %o', after);
      150 | 			expect(before.count).toEqual(0);
    > 151 | 			expect(result.body).toMatchObject({
          | 			                    ^
      152 | 				id: 1,
      153 | 				proposalId: 1,
      154 | 				fieldValues: [

      at Object.<anonymous> (src/__tests__/proposalVersions.int.test.ts:151:24)
```
      
This smells like a race condition to me, related to the test assuming a particular insert order when insertion is happening in parallel.  The *code* is fine, but the test should be updated to pass regardless of what order the fields are added.

slifty:
This PR updates one of our tests which was failing intermittently due to the fact that field values insert queries were run in parallel and so Ids were not deterministic.

Resolves #877

### DIFF
--- a/src/__tests__/proposalVersions.int.test.ts
+++ b/src/__tests__/proposalVersions.int.test.ts
@@ -115,7 +115,6 @@ describe('/proposalVersions', () => {
           ( 1, 2, 2, 'Last Name', '2022-07-20 12:00:00+0000' );
       `);
 			const before = await loadTableMetrics('proposal_field_values');
-			logger.debug('before: %o', before);
 			const result = await agent
 				.post('/proposalVersions')
 				.type('application/json')
@@ -140,14 +139,14 @@ describe('/proposalVersions', () => {
 				})
 				.expect(201);
 			const after = await loadTableMetrics('proposal_field_values');
-			logger.debug('after: %o', after);
+
 			expect(before.count).toEqual(0);
 			expect(result.body).toMatchObject({
 				id: 1,
 				proposalId: 1,
 				fieldValues: [
 					{
-						id: 1,
+						id: expect.any(Number) as number,
 						applicationFormFieldId: 1,
 						position: 1,
 						value: 'Gronald',
@@ -155,7 +154,7 @@ describe('/proposalVersions', () => {
 						createdAt: expectTimestamp,
 					},
 					{
-						id: 2,
+						id: expect.any(Number) as number,
 						applicationFormFieldId: 2,
 						position: 1,
 						value: 'Plorp',


reefdog (approved):
This test succeeded, but it's too soon to say whether it's intermittent or… regular termittent.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -140,22 +139,22 @@ describe('/proposalVersions', () => {
...
-			logger.debug('after: %o', after);
+
 			expect(before.count).toEqual(0);
 			expect(result.body).toMatchObject({
 				id: 1,
 				proposalId: 1,
 				fieldValues: [
 					{
-						id: 1,
+						id: expect.any(Number) as number,
```
hminsky2002, line 149: this is a cool function!

hminsky2002 (approved):
Approved!

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -140,22 +139,22 @@ describe('/proposalVersions', () => {
...
 						applicationFormFieldId: 1,
 						position: 1,
 						value: 'Gronald',
 						isValid: true,
 						createdAt: expectTimestamp,
 					},
 					{
-						id: 2,
+						id: expect.any(Number) as number,
```
bickelj, line 157: This makes the test pass, but we do know in advance what ID to expect, right? Why not make the inserts sequential? Then again, these are expected to be auto-generated, so we do not necessarily know in advance. OK, fine! : )

src/__tests__/proposalVersions.int.test.ts
bickelj: It's cool until you see the output when it matcheth not. But I agree it's better to "expect this to vary" when it really is expected to vary.

bickelj (approved):

src/__tests__/proposalVersions.int.test.ts
slifty: Unfortunately we don't know what ID to expect in advance right now because *SOMEONE* thought it'd be smart to insert in parallel (it was me, sorry).

I agree that's not a useful optimization, but I didn't want to change behavior in the name of fixing a test and opted to fix the test for now.

I expect to fix insertion soon (at which point I'll make the test better)

reefdog: *SOMEONE* is the worst, aren’t they?

PR #953, opened 2024-04-23
Update api version to 0.8.0

slifty:
This PR bumps our API version since we now have a new type and would like the SDK to reflect it.


hminsky2002 (approved):
The future is now!

PR #958, opened 2024-04-24
Document `pdc-admin` group Keycloak privileges

bickelj:
The ability to edit users without having full Keycloak administrative privileges can be granted to the `pdc-admin` group as documented in this commit.

Issue #959 Let certain PDC users edit PDC users


bickelj (commented):
I just applied these changes in prod. I can undo them if needed.

slifty (approved):
Looks great thanks for realizing this AND for the clear documentation!

bickelj:
dd4e25f rebased on main. I re-added the `view-users`, `query-users`, `manage-users` roles for the `realm-management` client in the PDC realm in production. Again, always up for future improvement/review.

PR #961, opened 2024-04-24
Support `createdBy=me` for all endpoints that use the `createdBy` filter

### LINKED ISSUE #960: Create a GET `/bulkUploads?mine` parameter
We need to be able to load the bulk uploads for the current user; this is technically possible to do from a client, but it involves looking up the current user and passing that user's ID into the `createdBy` parameter.

We'd like to create a new `mine` (name tbd) boolean, which will only return data owned by the current user.  This would override any `createdBy` filter.

Given the pressure from the demo timeline this seems like a reasonable API feature.  We may want to reflect on it later to make sure it's appropriate to keep in the long term.

slifty:
This PR lets you say `createdBy=me` to retrieve your own data in any instance where a `createdBy` parameter is accepted.

Resolves #960

### DIFF
--- a/src/__tests__/bulkUploads.int.test.ts
+++ b/src/__tests__/bulkUploads.int.test.ts
@@ -184,6 +184,46 @@ describe('/bulkUploads', () => {
 				);
 		});
 
+		it('returns uploads for the admin user when createdBy is set to me as an admin', async () => {
+			const testUser = await loadTestUser();
+			const anotherUser = await createUser({
+				authenticationId: 'totallyDifferentUser@example.com',
+			});
+			await createBulkUpload({
+				fileName: 'foo.csv',
+				sourceKey: '96ddab90-1931-478d-8c02-a1dc80ae01e5-foo',
+				status: BulkUploadStatus.PENDING,
+				createdBy: testUser.id,
+			});
+			await createBulkUpload({
+				fileName: 'bar.csv',
+				sourceKey: '96ddab90-1931-478d-8c02-a1dc80ae01e5-bar',
+				status: BulkUploadStatus.COMPLETED,
+				createdBy: anotherUser.id,
+			});
+
+			await agent
+				.get(`/bulkUploads?createdBy=me`)
+				.set(authHeaderWithAdminRole)
+				.expect(200)
+				.expect((res) =>
+					expect(res.body).toEqual({
+						total: 2,
+						entries: [
+							{
+								id: 1,
+								fileName: 'foo.csv',
+								fileSize: null,
+								sourceKey: '96ddab90-1931-478d-8c02-a1dc80ae01e5-foo',
+								status: BulkUploadStatus.PENDING,
+								createdAt: expectTimestamp,
+								createdBy: testUser.id,
+							},
+						],
+					}),
+				);
+		});
+
 		it('supports pagination', async () => {
 			const testUser = await loadTestUser();
 			await Array.from(Array(20)).reduce(async (p, _, i) => {
--- a/src/queryParameters/__tests__/extractCreatedByParameters.int.test.ts
+++ b/src/queryParameters/__tests__/extractCreatedByParameters.int.test.ts
@@ -1,0 +1,75 @@
+import { extractCreatedByParameters } from '..';
+import { InputValidationError } from '../../errors';
+import { loadTestUser } from '../../test/utils';
+
+describe('extractCreatedByParameters', () => {
+	it('should return undefined when passed no createdBy value', () => {
+		const createdByParameters = extractCreatedByParameters({
+			query: {},
+		});
+		expect(createdByParameters).toEqual({
+			createdBy: undefined,
+		});
+	});
+
+	it('should pass along the numeric value when one is provided', () => {
+		const paginationParameters = extractCreatedByParameters({
+			query: {
+				createdBy: '42',
+			},
+		});
+		expect(paginationParameters).toEqual({
+			createdBy: 42,
+		});
+	});
+
+	it('should return undefined when passed `me` with no authContext', () => {
+		const createdByParameters = extractCreatedByParameters({
+			query: { createdBy: 'me' },
+		});
+		expect(createdByParameters).toEqual({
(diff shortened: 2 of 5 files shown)


src/queryParameters/__tests__/extractCreatedByParameters.int.test.ts
```diff
@@ -0,0 +1,79 @@
...
+	});
+
+	it('should pass along the numeric value when one is provided', () => {
+		const paginationParameters = extractCreatedByParameters({
+			query: {
+				createdBy: '42',
+			},
+		});
+		expect(paginationParameters).toEqual({
+			createdBy: 42,
```
reefdog, line 22: wow this is elegant

src/queryParameters/__tests__/extractCreatedByParameters.int.test.ts
```diff
@@ -0,0 +1,79 @@
...
+		expect(createdByParameters).toEqual({
+			createdBy: testUser.id,
+		});
+	});
+
+	it('should throw an error when strings that parse to NaN are provided', () => {
+		expect(() =>
+			extractCreatedByParameters({
+				query: {
+					createdBy: 'forty two',
```
reefdog, line 54: have i told you lately that you're elegant

src/queryParameters/extractCreatedByParameters.ts
```diff
@@ -2,35 +2,57 @@ import { ajv } from '../ajv';
...
+	createdBy: number | 'me' | undefined;
+}
 
 interface CreatedByParameters {
 	createdBy: number | undefined;
 }
 
-const createdByParametersSchema: JSONSchemaType<CreatedByParameters> = {
-	type: 'object',
-	properties: {
```
reefdog, line 12: removing this line is very elegant

src/queryParameters/extractCreatedByParameters.ts
```diff
@@ -2,35 +2,57 @@ import { ajv } from '../ajv';
...
+			isCreatedByQueryParameters.errors ?? [],
 		);
 	}
+
+	const createdBy =
+		query.createdBy === 'me' ? partialAuthContext?.user.id : query.createdBy;
+
 	return {
-		createdBy: query.createdBy,
+		createdBy,
```
reefdog, line 55: this is legitimately a more elegant syntax

src/queryParameters/extractCreatedByParameters.ts
```diff
@@ -2,35 +2,57 @@ import { ajv } from '../ajv';
...
+						enum: ['me'],
+					},
+				],
+				nullable: true,
+			},
 		},
-	},
-	required: [],
-};
+		required: [],
```
reefdog, line 34: is there any way to make this more elegant?

src/queryParameters/extractCreatedByParameters.ts
```diff
@@ -2,35 +2,57 @@ import { ajv } from '../ajv';
 import { InputValidationError } from '../errors';
 import type { JSONSchemaType } from 'ajv';
 import type { Request } from 'express';
+import type { AuthContext } from '../types';
```
reefdog, line 5: this is very elephant

reefdog:
Okay genuinely this doesn't _seem_ to have an effect. At least, when I call `/bulkUploads?createdBy=me` (via front end, Swagger, cURL, whatever), it's the same as calling `/bulkUploads`. I get the full set either way. (And yes, the full set has uploads from multiple users. `/bulkUploads?createdBy=1` returns just 10, as expected.)

slifty:
Justin I didn't ask for your review yet

slifty:
ok @reefdog could you please review.

src/queryParameters/__tests__/extractCreatedByParameters.int.test.ts
slifty: UNSUBSCRIBE.

src/queryParameters/extractCreatedByParameters.ts
slifty: OK Justin. But listen. Isn't it kind of amazing that we expanded the functionality of TWO endpoints without even editing EITHER of them!?

src/queryParameters/__tests__/extractCreatedByParameters.int.test.ts
```diff
@@ -0,0 +1,75 @@
...
+	it('should throw an error when strings that parse to numbers less than 1 are provided', () => {
+		expect(() =>
+			extractCreatedByParameters({
+				query: {
+					createdBy: '0',
+				},
+			}),
+		).toThrow(InputValidationError);
+	});
+});
```
reefdog, line 75: This is a really good set of tests.

src/queryParameters/extractCreatedByParameters.ts
reefdog: Yes!

src/__tests__/bulkUploads.int.test.ts
```diff
@@ -184,6 +184,46 @@ describe('/bulkUploads', () => {
 				);
 		});
 
+		it('returns uploads for the admin user when createdBy is set to me as an admin', async () => {
```
reefdog, line 187: Out of curiosity, why are these tests framed around admins? It implies to me this functionality is meant to only work for admins, when that's not the case. And in fact I think it'd be good to have tests that ensure this *isn't* somehow an admin-only feature. Right?

I'm not insistent on this point, and maybe there's a good reason!

reefdog (approved):
Huzzah! This is very elegant. A few questions and observations sprinkled about, but I'm happy with this and grateful you knocked it out.

src/__tests__/bulkUploads.int.test.ts
slifty: Main reason right now is that admin is the only role where the filter matters.  If you aren't an admin you only get your own things anyway.

"admin" in the test name is just placeholder for "user where more than just their stuff would have been returned in the first place"

PR #965, opened 2024-04-26
Add base field scope

### LINKED ISSUE #917: Add the concept of "Organization base fields"

slifty:
This PR add the concept of "scope" to base fields, which provides more meaning around whether a base field describes a proposal or the primary organization associated with a proposal.

In this PR the scope doesn't result in any change to treatment, but creating it will allow us to do things like decorate organization objects by related field values of the appropriate scope.

Resolves #917


src/types/BaseField.ts
```diff
@@ -11,12 +11,18 @@ export enum BaseFieldDataType {
 	BOOLEAN = 'boolean',
 }
 
+export enum BaseFieldScope {
+	PROPOSAL = 'proposal',
+	PRIMARY_ORGANIZATION = 'primary_organization',
```
slifty, line 16: I named this `PRIMARY_ORGANIZATION` instead of just `ORGANIZATION` because I do anticipate us having base fields for other types of organization (e.g. partner organizations, co-applicants, etc).

An reasonable alternative approach would be to simply call this `ORGANIZATION` and introduce the concept of `primary` or `secondary` at the point that we actually implement those concepts.

For example, calling this `PRIMARY_ORGANIZATION` does imply that we are going to care about organization types at the base field level (as opposed to, for instance, at the application form field level which is where it might be more appropriate).

As I type this out, I'm realizing that I'd like to just do ORGANIZATION here.  We can always add to the enum / edit the enum later when we implement other org types if we decide base field is the correct place to do that.

src/types/BaseField.ts
```diff
@@ -11,12 +11,18 @@ export enum BaseFieldDataType {
 	BOOLEAN = 'boolean',
 }
 
+export enum BaseFieldScope {
```
bickelj, line 14: It would be handy if there were a way to define the enum once and read it in the other place.

src/database/migrations/0026-alter-base_fields-add-field_scope.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TYPE field_scope AS ENUM ('proposal', 'organization');
```
bickelj, line 1: A little documentation here or in the other definition of the same enum in TypeScript would be helpful. If I recall correctly, these scopes are not always mutually exclusive. For example, an organization-scoped field can be submitted as part of a proposal, right? But the converse is not necessarily true.

bickelj (approved):
This looks like a much-needed well-made change. I was able to build, migrate, and run tests with the expected outcomes. I checked the DB and happily found the new type associated with the new `scope` column in the `base_fields` table.

src/types/BaseField.ts
```diff
@@ -11,12 +11,18 @@ export enum BaseFieldDataType {
...
+	PROPOSAL = 'proposal',
+	ORGANIZATION = 'organization',
```
hminsky2002, line 16: Is the thought here that BaseFieldScope.ORGANIZATION is being used to relate proposal base fields to organizations(i.e data from proposals that can be used to define an organization) because it will be used equivalent to actual organization base-fields (ones that are derived from organization data, whenever we flesh out the protocol to create/register a new organization). If so, great! But then, are we going to weight the fields from proposals that describe organizations equally to the fields from organizations directly? That seems a bit weird to me but I could certainly see it working.

Sorry for the dump here, code works great and all the development here(tests/db migrations) look great!

slifty: The logic for how we decide which value to use is for future versions of us to decide -- the change in this PR doesn't really dictate what that logic should be.  I imagine we might have use cases for situations where a proposal value is newer and different, and a PDC interface to allow organization delegates to update the directly-linked data via some kind of interface.

src/database/migrations/0026-alter-base_fields-add-field_scope.sql
slifty: Good call -- I wonder if SQL lets you add a comment for a specific enum value...

hminsky2002 (approved):

PR #966, opened 2024-04-26
Add more jwt tests

### LINKED ISSUE #921: Add a test to `processJwt` that ensures 401 if the authentication realm is incorrect
I just realized: we should make sure our `processJwt` middleware WILL throw if a jwt is not from the proper realm.  There is not currently any test for that, and we want to be 100% sure it happens.

If that isn't already the case, then anybody could send any jwt from any realm and become any user they want!

(I do think this is happening, but would feel safer with a test)

slifty:
This PR adds some tests to ensure our middleware behaves as expected when passed jwts from issuers that aren't valid.

Resolves #921

### DIFF
--- a/src/middleware/__tests__/processJwt.int.test.ts
+++ b/src/middleware/__tests__/processJwt.int.test.ts
@@ -8,7 +8,12 @@ import {
 	allowNextToResolve,
 	generateNextWithAssertions,
 } from '../../test/utils';
-import { mockJwt as authHeader, getMockJwt } from '../../test/mockJwt';
+import {
+	mockJwks,
+	mockJwt as authHeader,
+	getMockJwt,
+	getMockJwks,
+} from '../../test/mockJwt';
 import type { NextFunction, Response } from 'express';
 import type { Request as JWTRequest } from 'express-jwt';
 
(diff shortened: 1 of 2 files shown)


src/middleware/__tests__/processJwt.int.test.ts
```diff
@@ -49,6 +49,23 @@ describe('processJwt', () => {
 		processJwt(mockRequest, mockResponse, nextMock);
 	});
 
+	it('does NOT populate an auth value when an auth header with an invalid issuer is sent', (done) => {
+		const mockJwt = getMockJwt({
+			iss: 'NotTheCorrectIssuer',
```
slifty, line 54: @bickelj I am realizing this is just one edge case, but probably doesn't account for a situation where the `iss` is what we expect but the token itself was created with an invalid cert.

I'm not entirely sure how to mock that case, though.  Is that something you know how to do off hand?  If not, I can look into figuring it!

bickelj: @slifty Off hand I suppose we'd generate two issuers, one valid and one invalid, and two tokens, one from each, then replace the invalid token's attributes selectively from the valid one. The signature check should fail. I don't know off the top of my head how hard that will be, though. Looking.

bickelj: @slifty And that may not be what you are asking. I think you were suggesting that the issuer is as expected but the signing key is not as expected and that the JWT is valid with respect to the actual (unexpected) signing key? If so, my above suggestion probably wouldn't work because it would require a different issuer URL to get the mock JWKS library to generate a different key. Hmm.

bickelj: It took a bit to come back to this context. Perhaps this needs a better overview in the tests' documentation or something.

We use `mockJwks`, a stateful server, from `src/test/mockJwks.ts`, to intercept requests to its given issuer URL. The `mockJwks` server is started in `src/test/integrationSuiteSetup.ts` in `beforeEach`. Perhaps inside a test you described above, a second `badMockJwks` could be created using the same arguments, same issuer, etc., and hopefully would create a new key (we'd have to verify it didn't cache and return the same server), start that second server, create an identical-except-signed-with-badMockJwks-key JWT, stop that server, then attempt to pass this JWT on. Does that sound like the right approach to you?

bickelj (approved):
It works as expected. When I replaced `'NotTheCorrectIssuer'` with `AUTH_SERVER_ISSUER`, the test failed. While the new test here does not go so far as to test when a malicious issuer creates a validly signed token but with the unexpected key, it is good to verify that an unexpected issuer throws an error.

src/middleware/__tests__/processJwt.int.test.ts
```diff
@@ -49,6 +54,44 @@ describe('processJwt', () => {
...
+			expect(err).toBeInstanceOf(Error);
+			expect(mockRequest.auth).toBe(undefined);
+		};
+		const nextMock = generateNextWithAssertions(makeAssertions, done);
+
+		processJwt(mockRequest, mockResponse, nextMock);
+	});
+
+	it('does NOT populate an auth value when an auth header with an invalid jwt server is used', (done) => {
+		const badMockJwks = getMockJwks();
```
slifty, line 75: I realized we did NOT need to explicitly make sure this is different from the real mocked JWKS (it is, FWIW).

The reason we don't have to check is that (1) inspecting the source of the library reveals it has no cache but most importantly (2) if it *weren't* different (or if the implementation changes), then this test would fail due to the fact that there would be no error thrown by the middleware / the "bad" server would be good.

bickelj (approved):
The additional test in e4406c4 gives much confidence that JWT validation is happening as expected. The test is simple and ran successfully on my machine. This is excellent, thanks for adding the test!

src/middleware/__tests__/processJwt.int.test.ts
slifty: I normally try to avoid comments as you know BUT I agree in this case this is a pretty complicated test / the research you did to enable the test would be easier with some kind of guide in this test.  I'll add it in!

bickelj: Perfect!

PR #967, opened 2024-04-29
Create opportunity db utilities

slifty:
This PR adds db utilities for the `Opportunity` entity, and also updates the collection endpoint to return a bundle.

Related to #228

### DIFF
--- a/src/__tests__/opportunities.int.test.ts
+++ b/src/__tests__/opportunities.int.test.ts
@@ -1,12 +1,9 @@
 import request from 'supertest';
 import { app } from '../app';
-import { db, loadTableMetrics } from '../database';
-import { getLogger } from '../logger';
+import { createOpportunity, loadTableMetrics } from '../database';
 import { expectTimestamp } from '../test/utils';
 import { mockJwt as authHeader } from '../test/mockJwt';
-import type { Result } from 'tinypg';
 
-const logger = getLogger(__filename);
 const agent = request.agent(app);
 
 describe('/opportunities', () => {
@@ -15,35 +12,39 @@ describe('/opportunities', () => {
 			await agent.get('/opportunities').expect(401);
 		});
 
-		it('returns an empty array when no data is present', async () => {
-			await agent.get('/opportunities').set(authHeader).expect(200, []);
+		it('returns an empty bundle when no data is present', async () => {
+			await agent.get('/opportunities').set(authHeader).expect(200, {
+				entries: [],
+				total: 0,
+			});
 		});
 
 		it('returns all opportunities present in the database', async () => {
-			await db.query(`
-        INSERT INTO opportunities (
-          title,
-          created_at
-        )
-        VALUES
-          ( 'Tremendous opportunity 👌', '2525-01-01T00:00:05Z' ),
-          ( 'Terrific opportunity 👐', '2525-01-01T00:00:06Z' );
-      `);
-			await agent
+			await createOpportunity({
+				title: 'Tremendous opportunity 👌',
+			});
+			await createOpportunity({
+				title: 'Terrific opportunity 👐',
+			});
+			const response = await agent
 				.get('/opportunities')
 				.set(authHeader)
-				.expect(200, [
+				.expect(200);
+			expect(response.body).toEqual({
+				entries: [
 					{
 						id: 1,
-						createdAt: '2525-01-01T00:00:05.000Z',
+						createdAt: expectTimestamp,
 						title: 'Tremendous opportunity 👌',
 					},
 					{
 						id: 2,
-						createdAt: '2525-01-01T00:00:06.000Z',
+						createdAt: expectTimestamp,
 						title: 'Terrific opportunity 👐',
 					},
-				]);
+				],
+				total: 2,
+			});
 		});
 	});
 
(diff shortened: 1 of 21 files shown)


src/__tests__/opportunities.int.test.ts
```diff
@@ -106,15 +81,9 @@ describe('/opportunities', () => {
...
-        INSERT INTO opportunities (
-          title,
-          created_at
-        )
-        VALUES
-          ( 'This definitely should not be returned', '2525-01-02T00:00:06Z' ),
-          ( 'This surely also should not be returned', '2525-01-02T00:00:07Z' );
-      `);
+			await createOpportunity({
+				title: 'This definitely should not be returned',
```
hminsky2002, line 85: To clarify, we create this so that there is an opportunity in the database, just not the one with our specified id?

slifty: @hminsky2002 yeah -- I do think this makes the test a little more robust

src/database/operations/load/loadOpportunity.ts
```diff
@@ -0,0 +1,19 @@
+import { db } from '../../db';
+import { NotFoundError } from '../../../errors';
+import type { JsonResultSet, Opportunity } from '../../../types';
+
+const loadOpportunity = async (id: number): Promise<Opportunity> => {
+	const bulkUploadsQueryResult = await db.sql<JsonResultSet<Opportunity>>(
```
slifty, line 6: This variable is incorrectly named

src/__tests__/opportunities.int.test.ts
```diff
@@ -15,35 +12,39 @@ describe('/opportunities', () => {
...
 				.get('/opportunities')
 				.set(authHeader)
-				.expect(200, [
+				.expect(200);
+			expect(response.body).toEqual({
+				entries: [
 					{
 						id: 1,
-						createdAt: '2525-01-01T00:00:05.000Z',
+						createdAt: expectTimestamp,
```
bickelj, line 37: This weakens our test assertions a tiny bit but the benefit is in the `readonly` timestamps of the types. Carry on.

src/__tests__/opportunities.int.test.ts
```diff
@@ -53,48 +54,22 @@ describe('/opportunities', () => {
...
 		});
 
 		it('returns 400 bad request when id is a letter', async () => {
-			await db.query(`
-        INSERT INTO opportunities (
-          title,
-          created_at
-        )
-        VALUES
-          ( 'This should not be returned', '2525-01-02T00:00:04Z' ),
```
bickelj, line 95: Why not keep at least one of these? I'm OK with leaving them out, just curious about the reason for the change.

src/handlers/opportunitiesHandlers.ts
```diff
@@ -1,89 +1,52 @@
...
 ): void => {
-	if (!isGetOpportunityParams(req.params)) {
-		next(
-			new InputValidationError(
-				'Invalid request parameters.',
-				isGetOpportunityParams.errors ?? [],
-			),
-		);
+	const { opportunityId } = req.params;
+	if (!isId(opportunityId)) {
```
bickelj, line 43: I tried `GET opportunities/555555555555555555555555555555` and got this response:
```json
{
  "name": "DatabaseError",
  "message": "Unexpected database error.",
  "details": [
    {
      "length": 168,
      "name": "error",
      "severity": "ERROR",
      "code": "22003",
      "where": "unnamed portal parameter $1 = '...'",
      "file": "numutils.c",
      "line": "314",
      "routine": "pg_strtoint32"
    }
  ]
}
```

Which is fine and probably not a new error at all, but would be nice to make it a 400 since we don't support IDs greater than 2^32-1.

src/handlers/opportunitiesHandlers.ts
```diff
@@ -1,89 +1,52 @@
...
+		next(new InputValidationError('Invalid id parameter.', isId.errors ?? []));
 		return;
 	}
-	db.sql('opportunities.selectById', { id: req.params.id })
-		.then((opportunitiesQueryResult: Result<Opportunity>) => {
-			logger.debug(opportunitiesQueryResult);
-			if (opportunitiesQueryResult.row_count === 0) {
-				res.status(404).contentType('application/json').send({
-					message:
-						'Not found. Find existing opportunities by calling with no parameters.',
```
bickelj, line 71: Hooray for dropping some redundant lines of code!

bickelj (approved):
LGTM! Thank you!

src/database/operations/load/loadOpportunity.ts
hminsky2002: Right, should probably be opportunityQueryResult or something similar

hminsky2002 (commented):
Looks great! Only necessary change I found is the one you caught @slifty regarding the misnamed vairable

src/__tests__/opportunities.int.test.ts
slifty: Setup does have a cost (minimal, but still there) -- the validity of an input is a the level of API definition / shouldn't depend on any given db state.  Given that, I figured why not go with the cheapest DB state!

src/database/operations/load/loadOpportunity.ts
slifty: Since we're defining much more atomic functions for db access I am actually gonna start just using plain 'ol `result`

src/handlers/opportunitiesHandlers.ts
slifty: Opening an issue!  The good news is this will be easy to change across *all* endpoints due to the `getId` type guard / validator being reused.

slifty: Done: https://github.com/PhilanthropyDataCommons/service/issues/971

PR #974, opened 2024-04-30
Update organization queries to use json

slifty:
This PR updates our organization queries to use json / JsonResultSet.  This is part of our larger conversion effort.

Related to #821


src/database/operations/load/loadOrganizationBundle.ts
```diff
@@ -9,12 +9,17 @@ export const loadOrganizationBundle = async (queryParameters: {
...
+	const bundle = await loadBundle<JsonResultSet<Organization>>(
 		'organizations.selectWithPagination',
 		{
 			...defaultQueryParameters,
 			...queryParameters,
 		},
 		'organizations',
 	);
+	const entries = bundle.entries.map((entry) => entry.object);
+	return {
```
hminsky2002, line 21: Would it be crazy to rename the bundle variable to like 'jsonBundle'? My read of this is it's a variable named 'bundle' that implements the bundle interface, that is then unpacked and returned as part of an object that also implements the bundle interface settings, which is a little confusing. Then again, my read could also just be incorrect. Would love clarification either way!

hminsky2002 (approved):
I read through the postgres json docs and this seems like a continuation of the right approach! Code works locally, had a clarifying question about a variable but approving!

src/database/operations/load/loadOrganizationBundle.ts
slifty: To clarify what's going on: one is a bundle that has malformed entities (due to the nature of how pg returns objects, which has them nested under an attribute called `object`) so we need to reform them to remove that nesting.

I think maybe `objectBundle` or `jsonResultSetBundle` might be the right name for it (I know `jsonBundle` was just an example, but it doesn't contain actual json so that might be misleading -- the `JsonResultSet` name is because it reflects a result set that was built using the psql jsonb function).

This pattern is already repeated in several places, and my hope was that once this is fully refactored I'd go through and DRY this mapping out.  To that end, I think the best solution would probably be to have a named helper function like:

`convertJsonResultSetBundleToEntityBundle` (name TBD) which should better auto-document what's going on.

I should have just done that to begin with, and there's no better time to make that kind of change than right now...

slifty: For now I just renamed the var -- but I'll open a fix to DRY things out soon.

PR #976, opened 2024-04-30
Fix URL typo in seed file

### LINKED ISSUE #975: Fix seed file typo
I just noticed that when editing the seed file in [this commit](https://github.com/PhilanthropyDataCommons/service/commit/27d5e3c956e2e364dae51755e10db490c4dd1e7b), I accidentally typed some of the base fields that were supposed to be of type 'URL' as 'url.' This is a quick fix for those instances!

hminsky2002:
This commit ammends a mistake from PR #925 where two base fields were typed as 'url', when our database expects urls to be of type 'URL'

Resolves #975


slifty (approved):
Ran the seeds -- they work!

PR #977, opened 2024-05-01
Update some TypeScript ESLint dependencies

bickelj:
The corresponding update from dependabot included a major ESLint version update that didn't play nicely with other ESLint deps.


slifty:
@bickelj rather than manually update these could you disable major versions on eslint via an edit to dependabot.yml so that we can benefit from dependabot going forward (these dependencies update pretty regularly), and then open a tracking issue to re-enable major eslint updates once the deps support it?

slifty:
Actually scratch that I opened a PR / issue to that effect: https://github.com/PhilanthropyDataCommons/service/pull/979 #978

bickelj:
@slifty Can we have dependabot notify us once the ESLint dependencies work together with ESLint 9?

bickelj:
I still prefer this to #979 but someone had to give somewhere, and this time it was me.

Issue #978, opened 2024-05-01
Adjust dependabot settings for ESLint 9


slifty:
Right now ESlint 9 isn't supported by the typescript-eslint package

The first step is to tell dependabot to ignore major eslint versions, and then we wait on the following to get closed:

- [x] https://github.com/typescript-eslint/typescript-eslint/issues/8211

Once that's done, we should remove the dependabot ignore rule.

bickelj:
Can Dependabot or GitHub notify us once that issue is closed?

slifty:
@bickelj I don't know!  I've never used a tracking issue before, but otherwise I think we'll just need to check back every once in a while (I expect to be on top of it since most of the projects I work on are in the same boat)

slifty:
Oh, another thought: when typescript-eslint update to support eslint 9 I have a very strong feeling it will *require* eslint 9, and so we'll naturally be informed by dependabot (in that the PR will break, and the logs will show we have to update eslint to v9) at which point we'll know what to do.

bickelj:
@slifty I'm still not convinced it's better, but it's also not a big deal, therefore I yield.

slifty:
In case it helps: I think this is what the dependabot ignore feature was made for (disabling dependabot for things we are not interested in updating based on current state).  We'll end up being alerted the moment the dependency is ready for dependabot again thanks to the very peer dependencies that are breaking things now.

Given all that, I don't *quite* grok the downside (whereas I see a very blatant downside in getting multiple alerts a week about un-mergable dependabot PRs for the next few months).

bickelj:
@slifty My understanding of "ignore this major version of X dependency" was that dependabot wouldn't attempt to upgrade X to any of those major versions. I didn't know that with that ignore rule being committed that somehow dependabot could know that we actually want that major version or have a way to detect when we're ready for it. If it can do that, great. I don't think you're suggesting that it can, though. I think your feeling above may be right, namely that other ESLint-related dependencies will alert us because they do not have backward compatibility with their new versions. I suppose we can lean on that and/or manual checking of the respective issues. The downside is that we don't get alerted to a compatible ESLint 9, which I assume we want.

[...]

PR #979, opened 2024-05-01
Ignore eslint major dependency updates

slifty:
This PR disables dependabot for eslint 9 for now (we'll re-enable once our other dependencies support it)

Related to #978

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
         update-types: ['version-update:semver-major']
       - dependency-name: '@types/node'
         update-types: ['version-update:semver-major']
+      - dependency-name: 'eslint'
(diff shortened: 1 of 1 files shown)


bickelj:
@slifty The thing is, we do want eslint 9, right? And instead of manually polling for whether it works with upstream dependencies, we can let dependabot attempt the upgrade every so often. That was my reasoning earlier.

slifty:
@bickelj it will be a constant source of pain (those dependencies update a few times a week on average) over the span of months.  Unfortunately I think that our best options are either:

1. Human-level ignore the broken dependabot PRs (which I've done the past two weeks or so)
2. Settings-level ignore the specific dependency that doesn't work until the peer dependency upgrades.

Without (2) then we're doing (2) to some extent either way.

PR #983, opened 2024-05-02
Use json for proposal field values

slifty:
This PR updates our proposal field value query to use JSON and utilize a database utility for access.

It also simplifies one of our more complicated tests in a way that lets us also remove a proposal field value query that was ONLY being used for that one test 🎉.

Related to #821 
Related to #228


src/handlers/proposalVersionsHandlers.ts
```diff
@@ -76,7 +73,7 @@ const assertApplicationFormExistsForProposal = async (
 
 const assertProposalFieldValuesMapToApplicationForm = async (
 	applicationFormId: number,
-	proposalFieldValues: ProposalFieldValueWrite[],
+	proposalFieldValues: WritableProposalFieldValuedWithProposalVersionContext[],
```
bickelj, line 76: Typo.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -145,47 +144,29 @@ const postProposalVersion = (
...
 					);
-				logger.debug(proposalVersionQueryResult);
 				const proposalVersion = proposalVersionQueryResult.rows[0];
 				if (proposalVersion !== undefined) {
-					const proposalFieldValueQueries = req.body.fieldValues.map(
-						async (fieldValue) => {
-							const { value, applicationFormFieldId, position } = fieldValue;
+					const proposalFieldValues = await Promise.all(
+						fieldValues.map(async (fieldValue) => {
+							const { value, applicationFormFieldId } = fieldValue;
```
bickelj, line 151: I was like "wait, what happened to position?" but I see below it is still included in the `...fieldValue` and no longer needed as a  variable in this scope.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -69,41 +63,28 @@ const createTestBulkUpload = async (
 	});
 };
 
-const createTestBaseFields = async (): Promise<
-	[BaseField, BaseField, BaseField]
-> => {
-	const proposalSubmitterEmailBaseField = await createBaseField({
+const createTestBaseFields = async (): Promise<void> => {
```
bickelj, line 66: I probably wouldn't return void in situations like this, e.g. if we need the ids of the things created later, but the previous signature was a little wacky as well. :shrug: It's OK!

src/types/ProposalFieldValue.ts
```diff
@@ -12,12 +13,14 @@ interface ProposalFieldValue {
...
-	'applicationFormField' | 'createdAt' | 'id' | 'proposalVersionId' | 'isValid'
+type WritableProposalFieldValue = Writable<ProposalFieldValue>;
+
+type WritableProposalFieldValuedWithProposalVersionContext = Omit<
+	WritableProposalFieldValue,
+	'proposalVersionId'
 >;
 
-export const proposalFieldValueWriteSchema: JSONSchemaType<ProposalFieldValueWrite> =
+const writableProposalFieldValuedWithProposalVersionContextSchema: JSONSchemaType<WritableProposalFieldValuedWithProposalVersionContext> =
```
bickelj, line 23: Typos.

src/types/ProposalFieldValue.ts
```diff
@@ -12,12 +13,14 @@ interface ProposalFieldValue {
 	readonly isValid: boolean;
 }
 
-type ProposalFieldValueWrite = Omit<
-	ProposalFieldValue,
-	'applicationFormField' | 'createdAt' | 'id' | 'proposalVersionId' | 'isValid'
+type WritableProposalFieldValue = Writable<ProposalFieldValue>;
+
+type WritableProposalFieldValuedWithProposalVersionContext = Omit<
```
bickelj, line 18: Typo.

bickelj (approved):
LGTM. I only see a couple of typos where the word `Value` or `Values` ended up as `Valued`. Otherwise I don't see any red flags. Thank you!

Issue #931, opened 2024-04-18
Organizations should have a `taxId` not an `employerIdentificationNumber`


slifty:
Our organizations have an `employerIdentificationNumber` attribute.  This was a naming mistake -- EIN is a [US-specific thing](https://en.wikipedia.org/wiki/Employer_Identification_Number) whereas `taxId` is a more generic term and is what they use over at https://schema.org/Organization

This isn't a huge issue, but I think would be worth cleaning up sooner than later since it would be a breaking change.

slifty:
cc  @reefdog @hminsky2002 on this

reefdog:
Agree and co-sign. This feels like a good thing to change before landing Phase 2, too.

reefdog:
Do you think it's worth reframing this as a `federalId` or `governmentId` rather than _tax_ ID? (I grok that tax ID is the correct term of art but wonder if we should go with a sort of context agnosticism / semantic appropriateness.)

slifty:
@reefdog ohhh I like `governmentId` plenty.  My only pause is that schema.org landing on `taxId` seems like a meaningful data point (since they're an entity that exists to think about this kind of thing).  @jmergy had mentioned others in the PDC ecosystem had brought up schema.org, and so especially as we start landing on "official" attributes it might make sense to try to defer to theirs unless we have a particular reason to divert.

reefdog:
Ah, yeah, I’m down with taking hints from Schema.org. 👍

jmergy:
@reefdog @slifty I do think schema dot org is a safe go-to. It has come up a few times in discussions around folks with data or not wanting to have PDC create more than necessary. There is a sense with some data that may be in play with PDC that schema dot org is something that is used as a guide in lieu of making things up. 

I could also see us publishing a schema back - perhaps on bulk upload or something to contribute...

PR #985, opened 2024-05-02
Rename `employer_identification_number` to `tax_id`

slifty:
This PR updates the `Organization` entity to have a `taxId` rather than the US-centric `EmployerIdentificationNumber`. The old value is still there, but is marked as deprecated and will be removed once we've updated our front-end to use the new attribute.

Related to #931

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.9.0 2024-05-02
+
 ### Added
 
 - `BaseField` now has a `scope` attribute.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The GET `/opportunities` endpoint now returns an `OpportunityBundle` and accepts pagination parameters.
+- Added a new `Organization` attribute `taxId` which contains the same value as `employerIdentificationNumber`.
+- Deprecated the `Organizations` attribute `employerIdentificationNumber`. Please update your clients to use `taxId` instead.
 
 ## 0.8.0 2024-04-23
 
--- a/src/__tests__/organizationProposals.int.test.ts
+++ b/src/__tests__/organizationProposals.int.test.ts
@@ -14,11 +14,11 @@ const agent = request.agent(app);
 
 const insertTestOrganizations = async () => {
 	await createOrganization({
-		employerIdentificationNumber: '11-1111111',
+		taxId: '11-1111111',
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The GET `/opportunities` endpoint now returns an `OpportunityBundle` and accepts pagination parameters.
+- The `Organization` attribute `taxId` has been renamed to `taxId`.
```
reefdog, line 17:
```suggestion
- The `Organization` attribute `employerIdentificationNumber` has been renamed to `taxId`.
```

slifty: Sorry yeah that was the *second* change.

slifty:
This is our first real breaking change for something that is in production.  Here's what I think we should do:

1. Keep EIN in the spec and response sets, but mark it as deprecated.
2. Release the new version (and new sdk) which has both.
3. Update the front end to use taxId
4. Open a new PR that removes the deprecated ein.

CHANGELOG.md
```diff
@@ -7,13 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.9.0 2024-05-02
```
slifty, line 10: This is our first real breaking change for something that is in production. Here's what I think we should do:

1.     Keep EIN in the spec and response sets, but mark it as deprecated.
2.     Release the new version (and new sdk) which has both.
3.     Update the front end to use taxId
4.     Open a new PR that removes the deprecated ein.

slifty: This PR now reflects this change in approach -- Organizations have both Tax ID and EIN right now; EIN is marked as deprecated.

Once we release this, the new SDK will have both, we can migrate our client.

In a future world, this is the approach we would take and then we'd plan to remove the deprecated field at a later date; for now we know we're the only client so we can act more quickly and remove the old value sooner.

src/__tests__/organizationProposals.int.test.ts
```diff
@@ -14,11 +14,11 @@ const agent = request.agent(app);
 
 const insertTestOrganizations = async () => {
 	await createOrganization({
-		employerIdentificationNumber: '11-1111111',
+		taxId: '11-1111111',
 		name: 'Example Inc.',
 	});
 	await createOrganization({
-		employerIdentificationNumber: '22-2222222',
```
slifty, line 21: I don't believe any clients are creating organizations via API so I don't think we have to maintain both values for the create case, just the return values.

reefdog (approved):
This looks great! I agree with the approach.

CHANGELOG.md
reefdog: 😐

PR #988, opened 2024-05-02
Correct required fields in our API specification

### LINKED ISSUE #987: Review required fields in our swagger specification
Our swagger specification for the `BulkUpload` entity has an incomplete list of required fields -- specifically, it only says `"required": ["fileName", "sourceKey"]` but there are a bunch of other fields that are not nullable.

This is a mistake in the specification, and also happens to result in an incorrect SDK.

In the long term this actually represents a potential issue with the way we're defining our specification since we have no tests to prevent this kind of mistake.  In the short term we just need to fix the spec.

slifty:
This PR corrects our API specification so that a series of fields are properly marked as required.  This should simplify API client interactions since they dont have to handle both the empty AND non-empty cases.

Resolves #987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Several fields that were marked as optional are now properly marked as required.
+
 ## 0.9.0 2024-05-02
 
 ### Added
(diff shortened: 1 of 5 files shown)


hminsky2002 (approved):
Ohohoh, this is like Christmas coming from the sdk migration in front-end! Christmas in that I am happy to not have be chaining conditionals nearly as much any more, and glad that these fields are decidedly required. Thank you very much @slifty!

PR #1019, opened 2024-05-15
Fix auth signature test

### LINKED ISSUE #1009: Our new auth test sometimes fails
```

FAIL src/middleware/__tests__/processJwt.int.test.ts
  ● processJwt › does NOT populate an auth value when an auth header with an invalid jwt server is used

    expect(received).toBeInstanceOf(expected)

    Expected constructor: Error

    Received value has no prototype
    Received value: undefined

      90 | 		const mockResponse = {} as unknown as Response;
      91 | 		const makeAssertions = async (err: unknown) => {
    > 92 | 			expect(err).toBeInstanceOf(Error);
         | 			            ^
      93 | 			expect(mockRequest.auth).toBe(undefined);
      94 | 		};
      95 | 		const nextMock = generateNextWithAssertions(makeAssertions, done);

      at makeAssertions (src/middleware/__tests__/processJwt.int.test.ts:92:16)
      at src/test/utils.ts:21:3
      at Immediate.wrappedNext (src/middleware/processJwt.ts:14:10)


Test Suites: 1 failed, 27 passed, 28 total
Tests:       1 failed, 222 passed, 223 total
Snapshots:   0 total
Time:        44.632 s
```

Not sure why this happens yet / it is inconsistent.  Seems to happen more on github CI.

slifty:
This PR fixes a test that was occasionally failing in CI.  It also simplifies the test so that's good!

Resolves #1009

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,7 @@
 				"eslint-plugin-import": "^2.29.1",
 				"eslint-plugin-sort-exports": "^0.9.1",
 				"jest": "^28.1.1",
+				"jsonwebtoken": "^9.0.2",
 				"mock-jwks": "^1.0.10",
 				"nock": "^13.5.4",
 				"pino-pretty": "^11.0.0",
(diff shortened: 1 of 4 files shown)


reefdog (commented):
Test runs clean, code makes sense as I understand it. (We sign a JWT with an intentionally bad key to make sure our auth stack rejects it.)

The only thing: when I run `npm i` on this, the lockfile changes to move `jsonwebtoken` from `dependencies` to `devDependencies`. Can you weigh in on if that's known/intentional (and how I should resolve locally), or an accident?

slifty:
ooooo great catch -- I had run npm i incorrectly originally and then manually moved it from `dependencies` to `devDependencies` in the `package.json` -- I did NOT re-run npm i to update the lock file.

Fixed!

reefdog (approved):
All good now!

PR #1020, opened 2024-05-16
Fix readonly object references

### LINKED ISSUE #1004: Some types allow for `any` in @pdc-sdk
While working on pdc front-end issue [705](https://github.com/PhilanthropyDataCommons/front-end/issues/705)
I discovered that, currently, our @pdc-sdk allows for type 'any' in multiple different schemas. 

Image
Image
Image

Per my understanding of the sdk and a discussion with @slifty in front-end PR [719](https://github.com/PhilanthropyDataCommons/front-end/pull/719), it seems like these are erroneous and should be fixed to be only of a specific type!

slifty:
This PR fixes the way some of our object attributes are marked as `readOnly` so that swagger codegen will properly generate types that don't combine with `any`.

To test this you will need to follow the instructions in https://github.com/PhilanthropyDataCommons/sdk to use this branch's `openapi.json` to generate the sdk.  You can search the resulting code to ensure `& any` is not included anywhere.

Resolves #1004

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Several read-only fields were improperly defined, resulting in improper SDK results.
+
 ## 0.10.0 2024-05-03
 
 ### Removed
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -196,12 +196,10 @@
 						"type": "integer"
 					},
 					"organization": {
+						"readOnly": true,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Several readonly fields were improperly defined, resulting in improper sdk results.
```
reefdog, line 12: Copy-editor-level suggestion:

```suggestion
- Several read-only fields were improperly defined, resulting in improper SDK results.
```

PR #1021, opened 2024-05-16
Correct `/bulkUploads` specification

### LINKED ISSUE #1010: Need an SDK interface for BulkUploadBundle
Over in the front end, we're converting all our homegrown types to use SDK-provided types when available.

One that seems to be missing is `BulkUploadBundle`, which we're having to recreate: https://github.com/PhilanthropyDataCommons/front-end/blob/3e39eb2f5815dd0de2caf03733d94242475465ef/src/pdc-api.ts#L174-L177

slifty:
This PR adds the `BulkUploadBundle` entity type

This should be merged after:

- [x] #1020 

Resolves #1010

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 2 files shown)


reefdog (approved):
Fantastic, thank you. Besides the code scan and build/test, I also used this locally and it works. 🙏

PR #1023, opened 2024-05-16
Fix `PresignedPostRequest` definition

### LINKED ISSUE #1012: `WritablePresignedPostRequest` improperly requiring readonly property
When trying to use an argument of `WritablePresignedPostRequest` type — which should only require `fileType` and `fileSize` — TypeScript gets bigmad that we aren't including `presignedPost` as well.

This API-access hook should work:
```ts
const usePresignedPostCallback = () => {
	const api = usePdcCallbackApi<PresignedPostRequest>('/presignedPostRequests');
	return (params: WritablePresignedPostRequest) =>
		api({
			method: 'post',
			headers: {
				Accept: 'application/json',
				'Content-Type': 'application/json',
			},
			body: JSON.stringify(params),
		});
};
```

And would be called thusly:
```tsx
const createPresignedPost = usePresignedPostCallback();
const { presignedPost } = await createPresignedPost({
	fileType: file.type || 'application/octet-stream',
	fileSize: file.size,
});
```

However, that implementation throws this error:

> Argument of type '{ fileType: string; fileSize: number; }' is not assignable to parameter of type '{ fileType: string; fileSize: number; presignedPost: { fields: { key: string; }; url: string; }; }'.
>  Property 'presignedPost' is missing in type '{ fileType: string; fileSize: number; }' but required in type '{ fileType: string; fileSize: number; presignedPost: { fields: { key: string; }; url: string; }; }'.

Our temporary fix on the front end was to instead define and use this local type in the callback:

```ts
type FixedWritablePresignedPostRequest = Omit<
	WritablePresignedPostRequest,
	'presignedPost'
>;
```

slifty:
This PR fixes the definition of `PresignedPostRequest` so that swagger codegen will honor the readonly nature of the `presignedPost` attribute.

This will be a breaking change to the SDK since it changes the name of the `PresignedPost` type (which was previously `PresignedPostRequestPresignedPost`.

Resolves #1012 

This should be merged after:

- [x] #1021

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Several read-only fields were improperly defined, resulting in improper SDK results.
+- `WritablePresignedPostRequest` will no longer require a populated `presignedPost`.
+
+### Changed
+
+- Created a new `PresignedPost` entity type instead of directly embedding the definition in `PresignedPostRequest`.
 
 ## 0.10.0 2024-05-03
 
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -261,27 +261,34 @@
 						"example": 512
 					},
 					"presignedPost": {
-						"type": "object",
 						"readOnly": true,
-						"properties": {
-							"url": {
-								"type": "string"
-							},
-							"fields": {
-								"type": "object",
-								"properties": {
-									"key": {
-										"type": "string",
-										"example": "96ddab90-1931-478d-8c02-a1dc80ae01e5"
-									}
-								},
-								"required": ["key"]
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Several readonly fields were improperly defined, resulting in improper sdk results.
+- `WritablePresignedPostRequest` will no longer require a populated `presignedPost`.
+
+### Changed
+
+- Created a new `PresignedPost` entity type instead of directly embedding the definition in `PresignedPostRequest`.
```
reefdog, line 21: Bless.

reefdog (approved):
Works great, thank you!

PR #1028, opened 2024-05-22
Fix test to use createOrganization

slifty:
This PR removes a direct DB call in a test and replaces it with our DB utilities.

This was simply missed in a previous PR.

Related to #228

### DIFF
--- a/src/__tests__/organizations.int.test.ts
+++ b/src/__tests__/organizations.int.test.ts
@@ -5,7 +5,6 @@ import {
 	createOrganization,
 	createOrganizationProposal,
 	createProposal,
-	db,
 	loadTableMetrics,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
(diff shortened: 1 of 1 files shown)


src/__tests__/organizations.int.test.ts
```diff
@@ -68,7 +67,7 @@ describe('/organizations', () => {
 		it('returns according to pagination parameters', async () => {
 			await Array.from(Array(20)).reduce(async (p, _, i) => {
 				await p;
-				await db.sql('organizations.insertOne', {
+				await createOrganization({
```
hminsky2002, line 70: Neat!

hminsky2002:
> This PR removes a direct DB call in a test and replaces it without our DB utilities.
> 
> This was simply missed in a previous PR.
> 
> Related to #228

Did you mean `replaces it with our DB utilities`?

slifty:
I don't know why I didn't merge this in May but... what could go wrong I'm sure it will be fine!

PR #1034, opened 2024-05-24
Use Error Code 422 When Referenced Resources Don't Exist

hminsky2002:
This commit distinguishes creation errors when referenced resources don't exist by using the error code 422, whereas before both unique id AND foreign key errors were covered by the error code 409.  It also updates the openapi spec to match these changes.

Closes Issue #827

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -837,7 +837,17 @@
 						}
 					},
 					"409": {
-						"description": "There was a unique key or foreign key conflict.",
+						"description": "There was a unique key conflict.",
+						"content": {
+							"application/json": {
+								"schema": {
+									"$ref": "#/components/schemas/PdcError"
+								}
+							}
+						}
+					},
+					"422": {
+						"description": "The application references entities that do not exist in the database.",
 						"content": {
 							"application/json": {
 								"schema": {
@@ -1298,7 +1308,17 @@
 						}
 					},
 					"409": {
-						"description": "There was a unique key or foreign key conflict.",
+						"description": "There was a unique key conflict.",
+						"content": {
+							"application/json": {
+								"schema": {
+									"$ref": "#/components/schemas/PdcError"
+								}
+							}
+						}
+					},
+					"422": {
+						"description": "The proposal references entities that do not exist in the database.",
 						"content": {
 							"application/json": {
 								"schema": {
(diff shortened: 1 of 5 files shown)


src/openapi.json
```diff
@@ -1386,7 +1406,7 @@
 							}
 						}
 					},
-					"409": {
+					"422": {
```
slifty, line 1409: What do you think of making this description a bit broader -- a 422 can be provided for other validation errors as well (e.g. invalid data type for a field, invalid value for a field, or in this case, an invalid reference).

Even if the only possible reason for this specific endpoint is the fk conflict, I would make the message more user-focused (rather than db terminology) sugh as "the referenced user entity didn't exist".  That said, I think it might be more long-term-sustainable to keep it high level here / rely on the body of the error to convey exactly what invalid data was provided.

slifty (changes requested):
Food for thought in the openAPI spec description

(side note, I wish we didn't have to re-document what HTTP status code 422 means in the first place!)

src/openapi.json
hminsky2002: @slifty I think that's reasonable, I like the move towards more user-focused language. Check out latest push here, 
```"422": {
						"description": "The proposal references entities that do not exist in the database.",
						"content": {
							"application/json": {
								"schema": {
									"$ref": "#/components/schemas/PdcError"
								}
							}
						}
					}
```
```
"422": {
						"description": "The proposal version references entities that do not exist in the database.",
						"content": {
							"application/json": {
								"schema": {
									"$ref": "#/components/schemas/PdcError"
								}
							}
						}
					}
```
```
"422": {
						"description": "The application references entities that do not exist in the database.",
						"content": {
							"application/json": {
								"schema": {
									"$ref": "#/components/schemas/PdcError"
								}
							}
						}
					}
					```

slifty: Looks great!

So... spoiler alert: I think we are going to want to update *all* validation errors to return 422 (400 is actually for signaling that the structure of the data is incorrect / unrecognized, not "we recognized it, but this string didn't match the pattern defined by business rules")

That said, for now we don't do that so this documentation is correct!

slifty (approved):
Looks great!

Issue #331, opened 2023-05-03
Create test setup tooling


slifty:
I feel like we're probably reaching the point where some of this setup should be factored out into a/some test helper method(s)!

Maybe many of the tests in this file can be run in a `describe` block that creates a standard set of entities? Might have to update some of the tests to use the same entities, though.

I'd be okay with deferring this, if you prefer.

_Originally posted by @jasonaowen in https://github.com/PhilanthropyDataCommons/service/pull/320#discussion_r1183117521_

PR #1097, opened 2024-07-11
Use DB helpers rather than queries in tests

### LINKED ISSUE #228: We should consolidate DB result parsing

slifty:
This PR does the final round of updates to our code base, removing any instance of direct queries in the code / using our helpers everywhere.

This will make changes to schema easier to implement / should involve less updating of tests down the line.

And with this we can finally say...

Resolves #228

### DIFF
--- a/src/__tests__/applicationForms.int.test.ts
+++ b/src/__tests__/applicationForms.int.test.ts
@@ -1,6 +1,13 @@
 import request from 'supertest';
 import { app } from '../app';
-import { createBaseField, db, loadTableMetrics } from '../database';
+import {
+	createApplicationForm,
+	createApplicationFormField,
+	createBaseField,
+	createOpportunity,
+	db,
+	loadTableMetrics,
+} from '../database';
 import { getLogger } from '../logger';
 import { BaseFieldDataType, BaseFieldScope, PostgresErrorCode } from '../types';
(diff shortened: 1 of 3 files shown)


bickelj (approved):
I was able to run `ci`, `lint`, and `build` successfully with these changes. I also looked at the changes and I see that the they are more or less equivalent tests. I still think that on balance this approach is slightly worse than the previous but acknowledge that there are advantages to this one, the most pertinent at the moment being consistency with the rest of the codebase.

Edit: to be clearer, the negative comment applies only to the use of helper functions during test setup.

slifty:
Appreciate the review (and the thoughts regarding the fact that there are some negatives / tradeoffs of this approach)

I'll merge this since it's the final step in making our code base consistent but will also highlight #331 as a place where we can take a moment to reflect on the best setup patterns overall.

PR #1100, opened 2024-07-12
Add BaseField enum validation

### LINKED ISSUE #1099: BaseField schema doesn't capture enums
We added some enum fields to the entity `BaseField` but did not actually include enums in the associated type (note it *is* included in the swagger spec).  This means there is no validation by ajv to ensure the values align with our expectations.

This is easy enough to fix!  Let's do so.

slifty:
This PR fixes an omission from our BaseField type definition (and resulting validation).

The API should now properly return an error code if the base field has an unexpected value.

Resolves #1099

### DIFF
--- a/src/__tests__/baseFields.int.test.ts
+++ b/src/__tests__/baseFields.int.test.ts
@@ -175,6 +175,25 @@ describe('/baseFields', () => {
 			});
 		});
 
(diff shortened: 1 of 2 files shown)


bickelj (approved):
On first glance, I thought maybe this was about validating against the BaseField's `dataType` or `scope` but no, this is about validating the values of those fields within `BaseField`. Looks good to me.

PR #1108, opened 2024-07-19
Lower default count for pagination

### LINKED ISSUE #330: Update default count value to something smaller
Now that we're returning deep proposal data (#274), the response to `/proposals` can be large - in the megabytes. A user has reported that trying to use the Swagger docs with the default values (of `_count=100`) leads to the page hanging, presumably as Swagger parses and formats multiple megabytes of JSON. We should probably lower the default to a more reasonable value!

The front-end is [overriding the default value anyway](https://github.com/PhilanthropyDataCommons/data-viewer/blob/2cd0f6c324942ec170be54285a40389e379e78da/src/pages/ProposalList.tsx#L52).

_Originally posted by @jasonaowen in https://github.com/PhilanthropyDataCommons/service/issues/179#issuecomment-1532310568_

slifty:
This PR updates the default count for paginated resources to 10.

Resolves #330

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -31,7 +31,7 @@
 				"schema": {
 					"type": "integer",
 					"minimum": 1,
-					"default": 100
+					"default": 10
 				},
 				"in": "query",
 				"name": "_count",


hminsky2002 (approved):
Looks granular to me!

PR #1109, opened 2024-07-19
Add license information to swagger spec

### LINKED ISSUE #853: Add license to our swagger specification
We don't have a `license` field in swagger but probably should (and should look into fully populating the `info` section of the openapi.json)

slifty:
This PR adds some license information to our swagger specification.

This means the license gets rendered on our api page: 

image

I wasn't able to add an `identifier` attribute even though [it is listed as a valid attribute in the specification](https://swagger.io/specification/).  Our validator complained about it.  Rather than debug, I think this is good enough and meets 99% of the value of including this information at all.

Resolves #853

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -3,7 +3,11 @@
 	"info": {
 		"title": "Philanthropy Data Commons API",
 		"description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
-		"version": "0.11.0"
+		"version": "0.11.0",
+		"license": {
+			"name": "GNU Affero General Public License v3.0 only",
+			"url": "https://spdx.org/licenses/AGPL-3.0-only.html"
+		}
 	},
 	"components": {
 		"parameters": {


src/openapi.json
```diff
@@ -3,7 +3,11 @@
 	"info": {
 		"title": "Philanthropy Data Commons API",
 		"description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for organizations seeking grants.",
-		"version": "0.11.0"
+		"version": "0.11.0",
+		"license": {
+			"name": "GNU Affero General Public License v3.0 only",
+			"url": "https://spdx.org/licenses/AGPL-3.0-only.html"
```
bickelj, line 9: In place of this I tried `"identifier": "AGPL-3.0-only"` and it seemed to have the same effect locally of putting up this URL. Regardless, OK either way.

bickelj (approved):
Looks good to me. If linking to SPDX is the usual practice, great (it seems to do that when using the identifier), but I wonder if we're supposed to link to our actual license file in our repository :shrug:.

src/openapi.json
slifty: Oh interesting -- side note, the lack of support for `identifier` seems to just be in our linter tool.  We may eventually find that the swagger linter is out of date and unmaintained, and we need to either remove it or replace it...

bickelj: Ah, I didn't process that detail that it was the validator. Thanks!

Issue #756, opened 2024-02-02
Create a worker that synchronizes with an instance of the PDC for base fields


slifty:

bickelj:
@slifty Is it like the following? There exists a PDC instance somewhere that we treat as canonical (cough) and whatever data state it has we treat that data as canonical and connect to it to get the most up-to-date base fields? In other words, there is no way to recreate from code the state of base fields at all?

slifty:
Essentially yes -- we would not have seed data in this repository.  We would want to have database backups so as not to lose the data in PDC production, but overall base fields are user data and really ought not to be in the domain of developers at all / the code base should be agnostic to them.

bickelj:
More thought required. I wish it were so simple to say "data are not our jam" but experience proves otherwise.

slifty:
Sure!  We can keep talking about it -- there was a fair bit of work done to design base field admin interfaces, though they haven't been implemented yet (e.g. https://github.com/PhilanthropyDataCommons/front-end/issues/518)

Our ownership of base fields was a function of the tooling we had in place in early stages of the project, and even if we *do* own it I think that ownership should be in a third party repository / we should manage them via the API as API clients.

bickelj:
@slifty OK, that helps me put it together more. If we have an admin interface on the front-end to edit base fields (which is fantastic, by the way) it makes sense to synchronize instead of trying to recreate state from code.

Issue #701, opened 2024-01-05
Prevent extra properties via schema checks


slifty:
@jasonaowen pointed out that we can add an `[additionalProperties: false](https://ajv.js.org/json-schema.html#additionalproperties)` flag to our schema and prevent additional properties on input from passing validation. (see https://github.com/PhilanthropyDataCommons/service/pull/679#discussion_r1441095910 for the thread).

We should consider adding this to our schemas!  The "downside" is that removing a property from a schema becomes a breaking change, but on the other hand this might be a good thing / make API changes more explicit.

PR #1123, opened 2024-07-30
Internationalize base fields

### LINKED ISSUE #1110: Internationalize Base Fields
After discussion documented in #1089, we have decided that the first step in internationalizing the pdc is to introduce localization for each base_field.

hminsky2002:
This PR contains the refactor of the base field type to accommodate the new
base field localization, which will be used to store the translations of
the pdc's base fields. Rather than having one static label for the base
field, base fields now support multiple separate translations for their
labels, and optionally descriptions, represented as a many-to-one
relationship between the base field and base field localizations table.
This change required a good amount of refactoring across the codebase,
but thankfully the work @slifty has done to document our existing
concepts has paid off ten fold.

Initial basefield localization creation is handled on creation of a
basefield, and subsequent localizations are created by the nested 'POST
baseField/{baseFieldId}/localization' route. Updating a basefield
localization similarly is handled by a nested 'PUT
baseField/{baseFieldId}/{language}` route, as localizations use a
natural unique id on base_field_id and language.

Base field languages are validated as valid IETF language tags via
the language-tags package. Invalid language tags are rejected on creation

Because base fields have been refactored, the seed file has been
removed from the repository, as it was no longer functional.

Since labels are no longer individual fields on our basefields,
our intake for application forms has been modified slightly. Instead
of setting application form field values to the basefield label
in our 'processBulkUpload.ts' task, we instead check if an existing
English label exists on the referenced base field. If it does,
we use that as our application form field label, otherwise we
default to the basefield's shortcode

Closes #1110


hminsky2002:
@jasonaowen Requesting your review on my commits here in prep for our conversation tomorrow! This is not a completed PR by any means, but I do think it's pretty close to finality. The last major thing I think I need to do is implement type checking for the localizations. A few things that are probably key to look at:

1. How I'm handling updating base fields -> I'm allowing for creation of localizations via the post basefields route, but the put route for updating basefields only handles the shortcode/datatype fields, not the localizations. That's handled via the put basefieldslocalizations route. Totally open to changing this, just seemed like it made the most sense to me, since you wouldn't necesarrily want to touch the basefield fields when changing localizations.

2. Dropping label from application form field -> I did this because it seemed like it was informed solely by the basefield label field, which is now defunct. As we're gravitating away from having the basefield store a canonical label, rather then the translations, i thought that it would be ok if applicationformfield just held on to it's foreign key relationship to the basefield. 

3. The specific logic behind basefield creation as it pertains to typescript -> I didn't really know the best way to approach this because I wanted to be able to create localizations on base field creation, but I needed to work around base_field_id being a required field for localizations. You'll see in the BaseField.ts and BaseFieldLocalization.ts I have multiple writable fields for each type, which I use seperately for creating basefields vs returning them. Now when you get a basefield returned via the get basefields route, you get the deeply populated localizations list, but when you update the basefield you just get back the fields you updated. It's a little mind racking and I am very open to a deep refactor there, would love your input on all of it!

src/database/queries/baseFields/selectAll.sql
```diff
@@ -1,2 +1,18 @@
-SELECT base_field_to_json(base_fields.*) as "object"
-FROM base_fields;
+SELECT json_build_object(
```
slifty, line 1: You should edit the `base_field_to_json` function to add the `localizations` rather than modify this query. That way any time a base field is collected it populates the deep value; it also prevents you from having to re-implement the rest of the `base_field_to_json` logic (which would now be in two places if we kept this implementation)

For this you would use the new `base_field_localization_to_json` function that you've already defined.

See the `proposals_to_json` function for an example of a function that selects and builds out a related document into the json result.

This specific file will remain unchanged.

src/handlers/baseFieldsHandlers.ts
```diff
@@ -26,17 +36,44 @@ const postBaseField = (
 	res: Response,
 	next: NextFunction,
 ): void => {
-	if (!isWritableBaseField(req.body)) {
+	if (!isWritableBaseFieldWithLocalizations(req.body)) {
```
slifty, line 39: w00t!  looks great.

src/tasks/__tests__/processBulkUpload.int.test.ts
```diff
@@ -64,27 +65,39 @@ const createTestBulkUpload = async (
 };
 
 const createTestBaseFields = async (): Promise<void> => {
-	await createBaseField({
-		label: 'Proposal Submitter Email',
-		description: 'The email address of the person who submitted the proposal.',
+	const baseFieldOne = await createBaseField({
```
slifty, line 68: Some day we'll have better test hooks to just... create all this kind of stuff for us in one call (issue already exists: #331)

src/types/ApplicationFormField.ts
```diff
@@ -29,11 +28,8 @@ const writableApplicationFormFieldWithApplicationContextSchema: JSONSchemaType<W
 			position: {
 				type: 'integer',
 			},
-			label: {
-				type: 'string',
-			},
 		},
-		required: ['baseFieldId', 'position', 'label'],
+		required: ['baseFieldId', 'position'],
```
slifty, line 32: Hmm -- we shouldn't be removing label from application form field.  That's its own concept.

Base fields are PDC specific, and are therefore localized.  Application form fields reflect the labels in the funder's GMS.

All this is to say: we should keep this concept as is; we won't be localizing it, but we aren't removing support for it either.

src/types/BaseField.ts
```diff
@@ -18,25 +23,22 @@ export enum BaseFieldScope {
...
+	localizations: WritableBaseFieldLocalizationWithBaseFieldContext[];
+};
+
 const writableBaseFieldSchema: JSONSchemaType<WritableBaseField> = {
 	type: 'object',
 	properties: {
-		label: {
-			type: 'string',
-		},
-		description: {
```
slifty, line 37: Oh snap great call I completely forgot about descriptions when we were discussing localization!

src/__tests__/baseFieldLocalizations.int.test.ts
```diff
@@ -0,0 +1,322 @@
...
+			expect(result.body).toMatchObject({
+				baseFieldId,
+				language: 'en',
+				label: 'Soomary',
+				description: 'Soooomary Description',
+				createdAt: expectTimestamp,
+			});
+		});
+
+		it('returns 400 bad request when no language is sent', async () => {
```
slifty, line 257: Should we add a test for when an invalid language code is sent (and should we add support for ensuring that a language is valid...)

A potential lead for implementation of this concept: https://www.npmjs.com/package/i18n-iso-countries

slifty (changes requested):
This is looking great!  Disclaimer that this review was time boxed (I'm technically not here, and technically not actually leaving this review).

Some feedback on implementation -- the biggest item is making sure we continue to use the `*_to_json` functions effectively.

src/__tests__/baseFieldLocalizations.int.test.ts
jasonaowen: I agree with the need for this test! I think we want to validate specifically against a list of language codes rather than countries, though.

We'd previously talked about using the `en-US` style tags, which it turns out are called [IETF language tags](https://en.wikipedia.org/wiki/IETF_language_tag). There is a package that has the current list and functions to work with it: [language-tags](https://www.npmjs.com/package/language-tags).

src/types/BaseField.ts
```diff
@@ -18,25 +23,22 @@ export enum BaseFieldScope {
...
 
+type WritableBaseFieldWithLocalizations = WritableBaseField & {
+	localizations: WritableBaseFieldLocalizationWithBaseFieldContext[];
+};
+
 const writableBaseFieldSchema: JSONSchemaType<WritableBaseField> = {
 	type: 'object',
 	properties: {
-		label: {
-			type: 'string',
```
jasonaowen, line 35: It's worth calling out that this is a breaking change to the API.

Do we have clients in the wild that will care about this? Changing base fields currently requires admin permissions, so probably we only have internal clients - if any.

I think this is fine, but wanted to double check. @slifty ?

src/routers/index.ts
```diff
@@ -11,11 +11,13 @@ import { proposalsRouter } from './proposalsRouter';
 import { proposalVersionsRouter } from './proposalVersionsRouter';
 import { usersRouter } from './usersRouter';
 import { documentationRouter } from './documentationRouter';
+import { baseFieldLocalizationsRouter } from './baseFieldLocalizationsRouter';
 
 const rootRouter = express.Router();
 
 rootRouter.use('/applicationForms', applicationFormsRouter);
 rootRouter.use('/baseFields', baseFieldsRouter);
+rootRouter.use('/baseFieldLocalizations', baseFieldLocalizationsRouter);
```
jasonaowen, line 20: Rather than routes like `PUT /baseFieldLocalizations` and `POST /baseFieldLocalizations` with the base field ID in the body of the request, I think we want routes like `PUT /baseFields/:id/localizations/:lang` and `POST /baseFields/:id/localizations/`.

CC @slifty on this API design question.

src/database/initialization/application_form_field_to_json.sql
```diff
@@ -7,14 +7,13 @@ BEGIN
   INTO base_field_json
   FROM base_fields
   WHERE base_fields.id = application_form_field.base_field_id;
-
+
```
jasonaowen, line 10: Watch out for inadvertent whitespace changes!

```suggestion

```

src/tasks/processBulkUpload.ts
```diff
@@ -178,7 +178,6 @@ const createApplicationFormFieldsForBulkUpload = async (
 				applicationFormId,
 				baseFieldId: baseField.id,
 				position: index,
-				label: baseField.label,
```
jasonaowen, line 181: As @slifty said, we do need to keep application form field labels. However, `baseField.label` no longer exists.

I think what we want to do is look for an English label - because, in practice, at least for the foreseeable future, they'll all have English labels - and use that here; if there is no English label, we can use the short code as the application form field label and log a warning.

src/types/BaseField.ts
```diff
@@ -49,14 +51,44 @@ const writableBaseFieldSchema: JSONSchemaType<WritableBaseField> = {
 			enum: Object.values(BaseFieldScope),
 		},
 	},
-	required: ['label', 'description', 'shortCode', 'dataType', 'scope'],
+	required: ['shortCode', 'dataType', 'scope'],
 };
```
jasonaowen, line 55: Since this is a breaking change to this endpoint, we should take this opportunity to disallow unrecognized keys; see #701.
```suggestion
    additionalProperties: false,
};
```

src/types/BaseField.ts
```diff
@@ -49,14 +51,44 @@ const writableBaseFieldSchema: JSONSchemaType<WritableBaseField> = {
...
+				type: 'string',
+				enum: Object.values(BaseFieldScope),
+			},
+			localizations: {
+				type: 'array',
+				items: writableBaseFieldLocalizationWithBaseFieldContextSchema,
+			},
+		},
+		required: ['shortCode', 'dataType', 'scope', 'localizations'],
+	};
```
jasonaowen, line 78: Likewise, as a new schema, we should disallow unrecognized keys here, too:

```suggestion
        additionalProperties: false,
	};
```

jasonaowen (changes requested):
Thanks for the discussion, @hminsky2002! I think we found more together than either of us would have separately.

src/types/BaseField.ts
slifty: True this!  We should be sure to bump the API version (and trigger a new SDK build!)

src/routers/index.ts
slifty: I do like the concept (in my head I said "oooooo nice!")

*but* we already have an API design that treats nested entities as root entities (e.g. it isn't `POST /proposals/:id/version/:number` but rather is `POST /proposalVersions` even though a version requires a proposal)

For consistency I think we may want to keep the simpler / less semantically meaningful URL patterns for now.

src/database/queries/baseFields/selectAll.sql
hminsky2002: Roger this! This makes sense as it ensures that anytime a basefield is returned by the api, it has the localizations field populated. Before there were some endpoints that would not return the localizations, which does seem like poor standardization

src/tasks/__tests__/processBulkUpload.int.test.ts
hminsky2002: This seems pertinent and like an issue i'd be happy to tackle now that I've been through a refactoring of our testing suite. I'd imagine we'd want some default exports for the values we're populating as well, so we don't have to remember what the test data is supposed to look like

src/routers/index.ts
hminsky2002: I agree that the structure @jasonaowen laid out makes more sense semantically, mostly because updating a base field localization should be done in the context of the base field it relates to. However, since the pattern already exists, maybe we refactor them both in a later commit?

src/database/initialization/application_form_field_to_json.sql
hminsky2002: good catch, thank you @jasonaowen !

src/database/migrations/0030-alter-base_fields-label.sql
jasonaowen, line 1: Why split this migration into two files? I think they're conceptually the same, and in particular I don't think you'd ever want to run one but not the other.

src/database/migrations/0029-create-base_field_localizations.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE base_field_localizations (
+  base_field_id INTEGER NOT NULL,
+  language VARCHAR NOT NULL,
+  label VARCHAR NOT NULL,
+  description VARCHAR,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  PRIMARY KEY (base_field_id, language),
+  FOREIGN KEY (base_field_id) REFERENCES base_fields(id)
```
jasonaowen, line 8: One of the concerns @hminsky2002 and I talked about the other day is that this schema allows base fields without _any_ localizations, which is currently not possible. There are some more complicated technical controls we could put into place. I think we landed on not worrying about that yet, as this PR reflects.

After talking with @kfogel today, given that creating or modifying base fields is restricted to administrators, he and I agreed that relying on social controls ("always enter at least one translation when creating a new base field") is sufficient here.

So, all of this to say: :+1: !

src/routers/index.ts
jasonaowen: Or we could start a new pattern, and fix the existing violations later? "We're doing it wrong elsewhere" is a depressing argument against doing it right here.

src/database/migrations/0030-alter-base_fields-label.sql
hminsky2002: @jasonaowen that makes sense! will do the refactor for that

src/routers/index.ts
hminsky2002: @jasonaowen true. If we are deciding to nest the router and adopt this pattern, I'd be happy to create an issue for and tackle a refactor commit for the proposalVersions routes

src/__tests__/baseFieldLocalizations.int.test.ts
hminsky2002: @jasonaowen I could be wrong but it seems like this package doesn't support ESModule syntax? I'm not really sure what that means, but it won't let me import anything from it in our application, i'd have to use the CommonJS 'require'. Likely this is easily resolvable but wanted to defer to your judgement here

hminsky2002:
@slifty @jasonaowen Ok putting a pin in things for today -- I believe all addressed issues have been resolved EXCEPT for the issue of refactoring the baseFieldLocalization routes AND language testing. ApplicationFormFields have been re-granted their label fields by a generous me, and the base_field_to_json function has been changed while the selectAll query has been reverted, and the tests have been refactored to accommodate this

@slifty if you wouldn't mind giving an extra once-over to the openapi spec changes I made that would be very helpful! I think everything's how it should be but I was kind of just following the form that other data had already done so an extra look over would be great.

ALSO -- After these migrations, the seed file is no longer functional. I was a little confused as to how to refactor that because if we're going to insert base_fields and localizations, I don't know how we match created base_field_ids with localizations. I suppose we could alternate insertions (base field, localization, base field, localization). Regardless, it will be a bunch of line changes so i wanted to wait for your inputs before going head first into that.

src/__tests__/baseFieldLocalizations.int.test.ts
jasonaowen: @hminsky2002 what are you trying to do? I was able to call it like this:

```ts
import * as tags from 'language-tags';
logger.info(tags.default('en-US'), 'Testing language-tags');
```

and get a log message like
```
[2024-08-02T23:03:53.688Z] INFO (219438): Testing language-tags
    source: "pdc/service/src/index.ts"
    data: {
      "tag": "en-us"
    }
```

jasonaowen:
> @slifty @jasonaowen Ok putting a pin in things for today -- I believe all addressed issues have been resolved EXCEPT for the issue of refactoring the baseFieldLocalization routes AND language testing. ApplicationFormFields have been re-granted their label fields by a generous me, and the base_field_to_json function has been changed while the selectAll query has been reverted, and the tests have been refactored to accommodate this

:joy: Thanks, @hminsky2002! Sorry for arguing with @slifty in your PR; I know that can be frustrating. Maybe the three of us should chat next week?

> ALSO -- After these migrations, the seed file is no longer functional. I was a little confused as to how to refactor that because if we're going to insert base_fields and localizations, I don't know how we match created base_field_ids with localizations. I suppose we could alternate insertions (base field, localization, base field, localization). Regardless, it will be a bunch of line changes so i wanted to wait for your inputs before going head first into that.

The solution I would like is to add a command to the [data-scripts]() to import base fields from another instance, as described in https://github.com/PhilanthropyDataCommons/service/issues/756. We've been talking about it for ages, and especially every time we have to touch the seed data, and I think maybe now is the time to bite the bullet and implement it. With that approach, we could simply delete the seed data, and say that it lives on in the production API instance.

src/__tests__/baseFieldLocalizations.int.test.ts
hminsky2002: @jasonaowen oh ok weird! Must be a bug with my IDE then:
Screenshot 2024-08-02 at 7 11 34 PM
maybe my intellisense is just bugged? That seems like a me problem and not an issue as far as the codebase is concerned, as long as it compiled

hminsky2002:
> > @slifty @jasonaowen Ok putting a pin in things for today -- I believe all addressed issues have been resolved EXCEPT for the issue of refactoring the baseFieldLocalization routes AND language testing. ApplicationFormFields have been re-granted their label fields by a generous me, and the base_field_to_json function has been changed while the selectAll query has been reverted, and the tests have been refactored to accommodate this
> 
> 😂 Thanks, @hminsky2002! Sorry for arguing with @slifty in your PR; I know that can be frustrating. Maybe the three of us should chat next week?
> 
> > ALSO -- After these migrations, the seed file is no longer functional. I was a little confused as to how to refactor that because if we're going to insert base_fields and localizations, I don't know how we match created base_field_ids with localizations. I suppose we could alternate insertions (base field, localization, base field, localization). Regardless, it will be a bunch of line changes so i wanted to wait for your inputs before going head first into that.
> 
> The solution I would like is to add a command to the data-scripts to import base fields from another instance, as described in #756. We've been talking about it for ages, and especially every time we have to touch the seed data, and I think maybe now is the time to bite the bullet and implement it. With that approach, we could simply delete the seed data, and say that it lives on in the production API instance.

Noooo sweat. a productive discussion next week sounds splendid. 

Importing from another instance sounds very cool. I'd imagine that would be a separate PR, so then in the meantime in this PR should we remove the seed file from the repository (since it will only serve to throw errors)?

src/__tests__/baseFieldLocalizations.int.test.ts
jasonaowen: Oh, as the error message suggests, you also need to install `@types/language-tags`! That's often the solution for using packages without built-in type declarations.

hminsky2002: Ahhh, that makes sense. I had assumed that such a package did not actually exist, how foolish

slifty: @hminsky2002 the main reason TypeScript is actually usable is that there is a huge community of people who have written type definitions for various node packages that don't include type definitions (check it out at https://github.com/DefinitelyTyped/DefinitelyTyped/)

This is a huge huge repo and it exports @types/...

And note that you can contribute type definitions there too if you ever come across a library lacking them!

src/routers/index.ts
slifty: I do think it opens a can of worms -- it's less "we do this elsewhere" and more "our entire API design is flat right now and this would reflect a change that warrants a consideration of a larger refactors overall"

It isn't just the example I gave, it is almost everything.

Opportunities -> Application forms
Application Forms -> Application form fields 
Proposals -> Proposal Versions
Proposal Versions -> Proposal Field Values

The first question I'd ask is: what is the heuristic we use to decide that something should be a nested relationship vs a primary entity with relationship conditions?

The second is: is this PR the place we should create that heuristic, or should we merge this feature with the existing project pattern and create an issue to refactor more comprehensively?

I'm not against the refactor as a concept, but I think that implementing something differently because it feels right here without taking the time to think more comprehensively is a recipe for potential thrash.

slifty: (Whoops, seeing below the "let's not burden harry's PR with this discussion" sorry!  +1 to chatting through it all by voice next week!)

slifty:
re the import from instance we have an issue open for that feature: #756 <-- this should be picked up, but sadly I do think we should not break seed support until *after* it's picked up :(

slifty:
+1 to finally implementing sync, but I do think that until we implement #756 we shouldn't break seed data... And also think we also shouldn't blow up the scope of this PR to do that implementation :-/

hminsky2002:
@jasonaowen @slifty leaving this off for today with language tag checking handled -- used the keyword approach suggested by @slifty. Excited to discuss this issue more tomorrow

For mine own sake, i'll write out the two things still unhandeled in this PR:

1. A potential refactor of the baseFieldLocalization Routes and other routes that follow this pattern
2. Refactoring the seed file, as it is currently broken with these changes.

src/routers/index.ts
jasonaowen: @slifty, @bickelj, @hminsky2002, and I discussed this further in a call today. Here's my understanding of where we landed; please either give a :+1: or reply with corrections!

- the endpoints will be like `/baseFields/:baseFieldId/:langTag`
- the table will still have a natural key on `(base_field_id, lang_tag)` with a foreign key on `base_field_id`
- I'll open up an issue for a larger discussion of dropping the base field id surrogate key in favor of the short code natural key
- we'll all reflect on nested routes and maybe talk about it more someday

Thanks for the engaging conversation! Does this sound right to everyone? Did I miss anything?

slifty: Yeah that's a solid summary!  Thanks for opening the issue for moving to natural keys for base fields more generally.

hminsky2002:
@slifty @jasonaowen Ok I believe this is finally ready for final review! I have nested the baseFieldLocalization routes inside of the baseField routes, following the convention we proposed. I did end up creating ANOTHER baseField type and validator to accomodate the PUT route (one that only expects a label and optionally a description, as baseFieldId and Language are provided via the URL parameters) but I wonder if there's a slicker way to do this. See `types/BaseFieldLocalization.ts`. Furthermore, the I couldn't figure out the OpenAPI syntax for ensuring that ONLY label and description are provided as example data for the PUT route without changing the schema reference. I am sure there's a better solution there but could not find it on my own :(

Those two issues aside however, I think this is ready to go! Would love your final inputs on this, excited to globalize this bad boy

src/types/BaseFieldLocalization.ts
```diff
@@ -0,0 +1,103 @@
+import { ajv } from '../ajv';
+import type { JSONSchemaType } from 'ajv';
+import type { Writable } from './Writable';
+
+interface BaseFieldLocalization {
+	baseFieldId: number;
```
slifty, line 6: The solution to your type woes is to make this and language `readonly` -- if they are the primary key then they are no longer writable attributes.

I think with this you can address the issues both here AND in the openAPI spec, and remove the need for a `WithBaseFieldAndLocalizationContext` type.

src/routers/baseFieldsRouter.ts
```diff
@@ -15,5 +15,15 @@ baseFieldsRouter.put(
 	requireAdministratorRole,
 	baseFieldsHandlers.putBaseField,
 );
+baseFieldsRouter.post(
```
slifty, line 18: I think we can (should?) actually remove the concept of POST for localizations, and ONLY have `PUT`.

You would then change PUT to up-sert instead of simply update.  This wouldn't violate rest because the action is still idempotent.  Since language is part of the primary key this also makes your life easier -- it's a bit odd to include primary key information in the body of a POST or PUT and most of the tooling is build around preventing that kind of thing.

hminsky2002:
@slifty @jasonaowen  ba-bam, this badboi is ready for final review. All tests passing with coverage I believe. Would love to see your thoughts on this and get it merged by tomorrow!

src/ajv.ts
```diff
@@ -1,10 +1,18 @@
...
 import ajvKeywords from 'ajv-keywords';
 import addFormats from 'ajv-formats';
+import * as tags from 'language-tags';
 
 export const ajv = new Ajv({
 	coerceTypes: true,
 });
 
+ajv.addKeyword({
+	keyword: 'isValidLanguage',
```
slifty, line 11: May be worth making this keyword even more specific `isValidLanguageTag` (since "English" is a valid language, but would not be a valid language tag).  You could also do `isValidIetfLanguageTag` but `I` and `l` look the same in this font so it looks like "Is Validletf Language Tag" and that makes very unhappy.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -18,15 +18,15 @@ const agent = request.agent(app);
 
 const createTestBaseFields = async () => {
 	await createBaseField({
-		label: 'First Name',
-		description: 'The first name of the applicant',
+		defaultLabel: 'First Name',
```
slifty, line 21: PLEASE do not kill me for saying this, and also feel free to ignore this thought...

We could just leave these as `label` and `description`.

Default is, in some sense, implied / unnecessary modifier.  A base field has a label and description, and then it has localizations for that label and description <-- that is the reality and doesn't need the word default to make sense.

Smaller point, but default could be slightly confusing if interpreted outside of the context of localization -- are we saying there are label / description overrides from a business logic sense?

A benefit of not changing this is that it would make this PR no longer be a breaking change, which means the frontend would continue to work without any modification.

It would also mean this PR is much smaller since many of the lines are renames of these fields.

---

Just food for thought...

src/openapi.json
```diff
@@ -506,18 +506,25 @@
...
 						"type": "string",
 						"example": "The first name of the applicant"
 					},
 					"shortCode": {
 						"type": "string",
 						"example": "firstName"
 					},
+					"localizations": {
+						"type": "object",
+						"additionalProperties": {
```
slifty, line 523: I wanted to check if we could be stronger here, and somehow say that the property keys would be language tags -- this exists: https://json-schema.org/understanding-json-schema/reference/object#propertyNames

I don't think we can (particularly without listing the tags out in our openapi.json).  That said, I think object would warrant having a `description` explaining that the keys can be expected to be IETF language tags.

slifty (changes requested):
One name change that would be actually good to make (the keyword) and then one potentially bigger change that would simplify the PR if you agree with the logic.

src/__tests__/proposalVersions.int.test.ts
hminsky2002: @slifty I would never make plans to kill a colleague over a PR review in any way that could be traced back to me...

No I agree, especially per our discussion this can and should be dropped!

src/database/queries/baseFieldLocalizations/createOrUpdateByPrimaryKey.sql
```diff
@@ -0,0 +1,16 @@
...
+  label,
+  description
+) VALUES (
+  :baseFieldId,
+  :language,
+  :label,
+  :description
+)
+ON CONFLICT (base_field_id, language)
+DO UPDATE SET
```
slifty, line 13: Neat, this only affects the specific row with that matching key information, right?

I admit it's very nerve racking to me to se an `UPDATE SET {...}` without a `WHERE` <-- since without the rows being specified that would update EVERY ROW / blow away a ton of data.  I almost wanna suggest adding the WHERE to be redundantly defensive...

@jasonaowen do you have any thoughts?  This is our first upsert so lets make sure the pattern is what we want!

slifty: @hminsky2002 would you be open to adding a test to make sure that PUT against an existing localization *only* updates the specified localization?  We should probably have that test regardless.

src/routers/baseFieldsRouter.ts
```diff
@@ -15,5 +15,10 @@ baseFieldsRouter.put(
 	requireAdministratorRole,
 	baseFieldsHandlers.putBaseField,
 );
+baseFieldsRouter.put(
```
slifty, line 18: We also should allow for `GET /:id/localizations`, which would return all localizations for a given base field.

slifty (changes requested):
Super duper duper close -- very sorry for not catching all this in the last review.

src/__tests__/baseFields.int.test.ts
```diff
@@ -30,20 +52,35 @@ describe('/baseFields', () => {
...
 				shortCode: 'lastName',
 				dataType: BaseFieldDataType.STRING,
 				scope: BaseFieldScope.PROPOSAL,
 			});
+
+			await createOrUpdateBaseFieldLocalization({
+				baseFieldId: baseFieldOne.id,
+				language: 'fr',
+				label: 'prenom',
+				description: 'le prenom',
```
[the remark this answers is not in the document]
hminsky2002: le pdc

src/database/queries/baseFieldLocalizations/createOrUpdateByPrimaryKey.sql
jasonaowen: I just double checked the [PostgreSQL docs for the `ON CONFLICT` clause](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT), and yes, only the specific matching row is updated:

> `ON CONFLICT DO UPDATE` updates the existing row that conflicts with the row proposed for insertion as its alternative action.

There is a `WHERE` condition allowed, which further restricts whether or not an existing row already found to be conflicting should be updated. Their example:

```SQL
-- Don't update existing distributors based in a certain ZIP code
INSERT INTO distributors AS d (did, dname) VALUES (8, 'Anvil Distribution')
    ON CONFLICT (did) DO UPDATE
    SET dname = EXCLUDED.dname || ' (formerly ' || d.dname || ')'
    WHERE d.zipcode <> '21201';
```

So yes, I think this is the correct way to write this! Adding a test to make that clearer is probably a good idea, as you suggest.

src/__tests__/baseFields.int.test.ts
```diff
@@ -437,4 +497,237 @@ describe('/baseFields', () => {
...
+		it('returns 400 when an invalid IETF language tag is sent', async () => {
+			await createTestBaseField();
+			const result = await agent
+				.put(
+					'/baseFields/1/localizations/theLanguageKlingonWhichIsNotARealLanguage',
+				)
+				.type('application/json')
+				.set(adminUserAuthHeader)
+				.send({
+					label: 'Ejyo',
```
slifty, line 703: YESsssssssssss

slifty: Sorry I mean HISlaaaaaaaaaaaaaH

src/handlers/baseFieldsHandlers.ts
```diff
@@ -1,8 +1,35 @@
...
+} from '../database';
+import {
+	isTinyPgErrorWithQueryContext,
+	isValidLanguageTag,
+	isWritableBaseField,
+	isWritableBaseFieldLocalization,
+} from '../types';
 import type { Request, Response, NextFunction } from 'express';
 
+const assertBaseFieldExistsForLocalization = async (
```
slifty, line 18: Right now this does not take any localization context so I think it should be named simply `assertBaseFieldExists`

src/handlers/baseFieldsHandlers.ts
```diff
@@ -83,8 +110,123 @@ const putBaseField = (
...
+	req: Request<{ baseFieldId: string }>,
+	res: Response,
+	next: NextFunction,
+): void => {
+	const baseFieldId = Number.parseInt(req.params.baseFieldId, 10);
+	if (Number.isNaN(baseFieldId)) {
+		next(new InputValidationError('The entity id must be a number.', []));
+		return;
+	}
+	Promise.all([assertBaseFieldExistsForLocalization(baseFieldId)])
```
slifty, line 123: Since you're just working with a single promise you don't need to use `Promise.all` but rather can just call the assertion directly.

src/handlers/baseFieldsHandlers.ts
```diff
@@ -83,8 +110,123 @@ const putBaseField = (
...
+		});
+};
+
+const putBaseFieldLocalization = (
+	req: Request<{ baseFieldId: string; language: string }>,
+	res: Response,
+	next: NextFunction,
+) => {
+	const baseFieldId = Number.parseInt(req.params.baseFieldId, 10);
+	if (Number.isNaN(baseFieldId)) {
```
slifty, line 160: We should use the `isId` validator here instead.

src/openapi.json
```diff
@@ -1544,6 +1600,121 @@
...
+					"404": {
+						"description": "The base field could not be found."
+					}
+				}
+			}
+		},
+		"/baseFields/{baseFieldId}/localizations/{language}": {
+			"put": {
+				"operationId": "upsertBaseFieldLocalization",
+				"summary": "Create or update a base field localization.",
```
slifty, line 1644: We're using the { whatever the linguistic term is } form of verbs in summaries.

So: "Creates or updates" rather than "Create or update"

slifty (changes requested):
Amazing stuff thank you -- a few tiny tiny things and this is finally done.

slifty (approved):
I tried so hard to find something else to request, I'm going to miss our time together working on this PR 😢

hminsky2002:
> I tried so hard to find something else to request, I'm going to miss our time together working on this PR 😢

Just think, now we can support PDC PR discussions in every known language to man. And not just can, will! From now on all my commit messages will be doubled in Esperanto 

Nur pensu, nun ni povas subteni PDC-PR-diskutojn en ĉiu konata lingvo al homo. Kaj ne nur povas, volas! Ekde nun ĉiuj miaj commit-mesaĝoj estos duobligitaj en Esperanto

Issue #1087, opened 2024-07-08
Think through how we want to identify and present "gold" data from the PDC


slifty:
As of 2024-08-01, our best understanding is that "gold" data only refers to Organization/ChangeMaker data. In other words, we don't expect folks to query PDC for a "gold" version of a proposal, we expect them to query PDC to get the best data for a given Organization/ChangeMaker.

bickelj:
I confess I haven't used bulk upload yet, so I don't know all the details of how intermediate data are stored. My views might change after trying that.

In some sense, we already present the "gold" data in the PDC, so the tricky part is how to present "non-gold" or "pre-gold" data, e.g. the CSV file that was uploaded or the individual field responses that need work. I think we already have a start: we present the proposal data that matches the expected form and flag the proposal data that does not.

Then again, the above sense could be thought of as validity and we want gold to mean valid and more, e.g. valid and sound. In that case, we'll need human help to flag gold data separate from mere validity.

bickelj:
I suppose we are looking not just at proposal-specific data but organizational data as well. We may need to compose a view of an organization across several proposals and/or other posts of data. The simplest is probably "latest post/put to PDC wins" but that is obviously incorrect. Old or stale data can by synced from a system so "created in PDC" should be a last resort. We probably should ask posters of data to include the last updated timestamp from their system and keep that in PDC. Even the reported last updated timestamp is crude, though, because time has nothing to do with the quality of the data. It depends more on who made the update. For example, if someone from organization A updated records for organization B, that could be considered worse than if someone from organization A updated records for organization A. Then again someone from organization A might input malformed data. This gets complicated fast but boils down to identifying some person or group of people as being at the top of a priority list for calling a record good. That and formal validity of the data. Something like:
1. Are the data valid? If so, consider it potential gold.
2. Are the data recent? If so, consider it potential gold.
3. Were the data updated by a known and expected party? If so and the above are true, call it gold.

There may be some fields that we can corroborate with automated checks, such as searching an EIN in some system, which might be good, but that does not generalize to all fields. We should have some date on all fields and should have some validity check on all fields, however meager. And we should have some way to track the provenance of the data (see #1083) such that we can judge the third point above.

slifty:
Two lines of thought that both warrant design and attention:

1. How do we identify / determine gold data
2. How do we represent / store / serve gold data

## How to Identify

[...]

Issue #1088, opened 2024-07-08
Re-design / re-implement the 3rd party data provider integrations


slifty:

jmergy:
I would think this could also be a type of proposal associated to the organization. The data from other sources would have a structure not unlike a grants system. I could see classification of proposals as a way to denote intended use of the proposal data contained in the grouping of fields from the specific source. We could have type=funding for the types of proposals we have in there now. But we could have type=organizational if the data was covering organization attributes or something from a data source. 990 data for example could come in as proposals but have a type of organizational

slifty:
Interesting idea!

I am a bit wary of using proposals for this since there are a lot of things related to proposals that are unrelated to this (candid data is organizational -- a candid snapshot is not a proposal / there's no such thing as an application form / proposal outcome / funder / opportunity etc.).  We could use that existing infrastructure but to me that would be an artificial coupling that would risk limiting our flexibility in the long run.  Even in the immediate term, for instance, we would have to add a bunch of unintuitive logic to conditionally ignore certain kinds of proposal in API responses since we wouldn't want to render org-data imports alongside actual proposals.

All this is to say, I think that unfortunately this does warrant a set of more specific entities -- though I think it can still be done elegantly!

FWIW the initial ERD did have a very very high level / placeholder model for all this (external sources / imports) --  it's just that our initial implementation of this feature was for demo purposes so we didn't lean on / flesh out any of those designs:

<img width="985" alt="erd" src="https://github.com/PhilanthropyDataCommons/service/assets/208884/63f5e578-452b-4607-9c5f-843b2b2e8c64">

The basic idea is that we would have "external field values" (that name is no longer quite right IMHO) in addition to "proposal field values" which would directly relate to organizations.  I think something like this model can still work, though we'll want to create a higher level entity that represents a given snapshot of organizational data (e.g. a set of related external data).

I should note that these snapshots would explicitly NOT contain proposal data -- only organizational fields.

bickelj:
When retrieving ChangeMaker/organization/applicant data, currently we have only `ProposalFieldValue`. Then separately there is a type of `ExternalFieldValue`. It would be awkward to integrate `ExternalFieldValue` into the response for getting CM/org/applicant data without some common type that covers both fields. See `organizationsHandlers.ts` and the `Organization` type, related to #1129. As part of this change bear all that in mind.

PR #1129, opened 2024-07-31
Include field values in `/organizations`

bickelj:
Enhance the existing `/organizations` endpoint with by finding the best ("gold") values associated with that organization from proposal field values, and return the best values, one per base field.

Based on feedback, deeply integrate the gold data search into existing organization types and endpoints. Do much in the existing PG function.

Issue #1087

### DIFF
--- a/src/__tests__/organizations.int.test.ts
+++ b/src/__tests__/organizations.int.test.ts
@@ -1,15 +1,20 @@
 import request from 'supertest';
 import { app } from '../app';
 import {
+	createApplicationForm,
+	createApplicationFormField,
+	createBaseField,
 	createOpportunity,
 	createOrganization,
 	createOrganizationProposal,
 	createProposal,
+	createProposalFieldValue,
+	createProposalVersion,
 	loadTableMetrics,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
 import { mockJwt as authHeader } from '../test/mockJwt';
-import { PostgresErrorCode } from '../types';
+import { BaseFieldDataType, BaseFieldScope, PostgresErrorCode } from '../types';
 
 const insertTestOrganizations = async () => {
 	await createOrganization({
@@ -50,12 +55,14 @@ describe('/organizations', () => {
 								taxId: '22-2222222',
 								name: 'Another Inc.',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 							{
 								id: 1,
 								taxId: '11-1111111',
 								name: 'Example Inc.',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 						],
 					}),
@@ -87,30 +94,35 @@ describe('/organizations', () => {
 								taxId: '11-1111111',
 								name: 'Organization 15',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 							{
 								id: 14,
 								taxId: '11-1111111',
 								name: 'Organization 14',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 							{
 								id: 13,
 								taxId: '11-1111111',
 								name: 'Organization 13',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 							{
 								id: 12,
 								taxId: '11-1111111',
 								name: 'Organization 12',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 							{
 								id: 11,
 								taxId: '11-1111111',
 								name: 'Organization 11',
 								createdAt: expectTimestamp,
+								fields: [],
 							},
 						],
 					}),
@@ -151,6 +163,7 @@ describe('/organizations', () => {
 						taxId: '123-123-123',
 						name: 'Canadian Company',
 						createdAt: expectTimestamp,
+						fields: [],
 					},
 				],
 			});
@@ -195,6 +208,7 @@ describe('/organizations', () => {
 						taxId: '123-123-123',
 						name: 'Canadian Company',
 						createdAt: expectTimestamp,
+						fields: [],
 					},
 				],
 			});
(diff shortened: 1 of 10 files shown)


bickelj:
> What do you think of not adding a new endpoint but rather having that detail become part of the organization object on the existing `/organizations` endpoint. I don't think we need (or want) a new type of organization entity.
>
> Similar to how `Proposal` has its detailed values provided directly and we don't have both a `Proposal` and a `ProposalDetail` object type.

I think it makes sense to bring the existing endpoint and the details endpoint closer together and perhaps even merge them. There are still reasons to keep the endpoints separate and even if the endpoints merge to keep some types of these objects separate, namely to make the dynamic part of the response optional via parameter.
1. Authorization. We want `/organizations` to be open to the public but the gold data not open. Also, different users will get different data based on their permissions. If we want to merge the expectations/requirements about openness then I think the endpoints could be merged. But I'd still probably prefer some type separation such that the one response not depending on authorization is the same for all users.
2. Utility. It is useful to distinguish the shallow organization from the deep organization. If, for example, I am posting data to the PDC, I should check if an organization exists before associating proposal data with the organization. In order to check if it exists, I need to get the existing organization and/or create it. When doing that, I am not interested in searching through all proposal field values based on my authorization, the time of day, and fine-grained permissions. In fact I am interested in not doing all that because it is going to slow my upload down significantly. This is not only a theoretical concern (i.e. not needing experimentation to know that it will be slower, which is true) but an experiment already happened in the past. Running scripts to post new proposals ran way slower when a shallow body was replaced with a deep body without the option to get the shallow one.
3. Time. Gold data may change from request to request (due to time being involved in the filtering, sorting, extraction of data) even when the narrow `organizations` entity does not change at all. One (gold data) is a relatively dynamic resource and the other (shallow org) a relatively static resource.

Reason 1 (authorization) is new with fine-grained permission and I suppose it applies to proposals as well. Reason 2 (utility) applied and applies to proposals as well. Reason 3 (time) I suppose only applies to gold data for organizations.

bickelj:
@slifty See also related discussion in the issue at https://github.com/PhilanthropyDataCommons/service/issues/1087#issuecomment-2263883962

bickelj:
Feedback just now from @slifty: do the deep integration/merge, down to the query level. When selecting data, include the parameter of authorization to the One True Organizations query.

src/database/initialization/organization_to_json.sql
```diff
@@ -1,11 +1,31 @@
...
+    ON pfv.application_form_field_id = aff.id
+  INNER JOIN base_fields bf
+    ON aff.base_field_id = bf.id
+  INNER JOIN proposal_versions pv
+    ON pfv.proposal_version_id = pv.id
+  INNER JOIN organizations_proposals op
+    ON pv.proposal_id = op.proposal_id
+  WHERE op.organization_id = organization.id
+    AND bf.scope = 'organization'
+    AND pfv.is_valid;
```
bickelj, line 22: I expect to expand this query further with a bunch of `case` statements etc., but my question for now is whether the gist of this approach is satisfactory, namely a modification to this postgres function and an additional field called `fields`.

slifty: This seems like the right approach to me yes!

src/database/initialization/organization_to_json.sql
```diff
@@ -1,11 +1,40 @@
...
 BEGIN
+  SELECT jsonb_agg(
+    proposal_field_value_to_json(pfv.*)
+    ORDER BY bf.id
+  )
+  INTO proposal_field_values_json
+  FROM proposal_field_values pfv
+  -- Remove field values for unauthenticated users while also (re)validating the user ID:
+  INNER JOIN users u
+    ON u.authentication_id = authenticationId
```
bickelj, line 16: Is this `join` too clever? At first I tried using a `case when` statement but after not getting it to work the way I wanted I thought I'd try this approach. The reason for the in-line comment is that the join is a more subtle check than a `case when`.

src/database/initialization/organization_to_json.sql
```diff
@@ -1,11 +1,40 @@
-CREATE OR REPLACE FUNCTION organization_to_json(organization organizations)
+-- An explicit DROP is needed to remove the old function instead of overloading it.
+DROP FUNCTION IF EXISTS organization_to_json(organizations);
```
bickelj, line 2: This little bit caught me off guard: if you change the function parameters, that's a new function despite the same name. So you can end up having multiple functions unintentionally if `replace` was intended to replace by name.

src/handlers/organizationsHandlers.ts
```diff
@@ -66,8 +76,56 @@ const getOrganizations = (
 		});
 };
 
+/** Takes an array of field values and returns an array of one best value per base field */
+export const extractGold = (
```
bickelj, line 80: @slifty Would you expect all the gold data extraction logic to be in the postgres function instead of here?

bickelj: Done in 408853eace888e19af3e7172377ea1d425830d43

bickelj:
Odd, I got this locally too when running `npm run test:ci` but not when running that test independently, e.g. `node 'node_modules/jest/bin/jest.js' 'src/__tests__/organizations.int.test.ts' -c 'jest.config.int.js' -t '/organizations POST / requires authentication'`:
```
FAIL src/__tests__/organizations.int.test.ts
  ● /organizations › POST / › requires authentication

    expected 401 "Unauthorized", got 400 "Bad Request"

      389 | 	describe('POST /', () => {
      390 | 		it('requires authentication', async () => {
    > 391 | 			await agent.post('/organizations').expect(401);
          | 			                                   ^
      392 | 		});
```

bickelj:
The above test failure happens when my new test is introduced but not when it is not introduced. Commenting out all tests except the existing POST where 401 is expected: test passes. Commenting out all tests except the same existing and my new: POST/401 test fails, line 22 of `requireAuthentication.ts` is reached (should be line 11).
Order seems to matter as well. When my new test precedes the existing POST/401 test, the existing test fails. When my new test follows the existing POST/401 test, the existing test passes. So there may be something from my new prod code or test code causing the issue.

src/__tests__/organizations.int.test.ts
```diff
@@ -370,9 +368,9 @@ describe('/organizations', () => {
 				value: 'invalid-email-address.com',
 				isValid: false,
 			});
-			await agent
-				.set(authHeader)
+			await request(app)
 				.get(`/organizations/${organizationId}`)
+				.set(authHeader)
```
bickelj, line 373: This little re-ordering of the `set(authHeader)` call makes all the difference. When setting it on the `agent` (as in the previous commit), that makes the headers persist on that agent. When setting it on this (new/fresh) request returned by the call to `.get(...)`, then those headers are *not* re-used on the agent. It's a subtle but important difference. Despite the mistake, I think that to avoid such a mistake in future, we should not introduce a re-used agent unless we need it and follow the new form of this commit.

src/database/initialization/organization_to_json.sql
```diff
@@ -1,11 +1,44 @@
...
```
slifty, line 2: I was thinking about this and think we should plan to make a more universal solution (not part of this PR!)

This describes a solution, which basically is a function that will find all functions with a name (regardless of parameters), and drop each of them: https://stackoverflow.com/questions/7622908/drop-function-without-knowing-the-number-type-of-parameters

I think this should be applied to every initialization script since we don't want them to have to keep track of past state.

I'll make an issue for it!

src/types/Organization.ts
```diff
@@ -1,11 +1,14 @@
 import { ajv } from '../ajv';
 import type { JSONSchemaType } from 'ajv';
 import type { Writable } from './Writable';
+import type { ProposalFieldValue } from './ProposalFieldValue';
 
 interface Organization {
 	readonly id: number;
 	taxId: string;
 	name: string;
+	// TODO: use a type for values that covers either ProposalFieldValue or ExternalFieldValue.
```
slifty, line 10: I think we can take out this TODO since we aren't using TODOs to manage project tasks and ExternalFieldValues don't exist right now.

What do you think of instead adding a comment on #1088 so that we don't lose this need?

slifty (approved):
This is a really nice foundation thank you Jesse!

One suggested moving of a TODO into an issue comment, and one idea for future work, but I have no suggested changes (thanks for building consensus as you worked, that absolutely saved time in the end cycles!)

bickelj:
@slifty Thank you!
Because I merged the [fix](https://github.com/PhilanthropyDataCommons/service/pull/1179) for [1171](https://github.com/PhilanthropyDataCommons/service/issues/1171] just prior to this, the second commit related to #1171 was superfluous so I squashed it further down to one commit. I hope this makes sense.

PR #1131, opened 2024-08-01
Use Java 17 and cache Gradle artifacts

jasonaowen:
The SDK build [failed recently](https://github.com/PhilanthropyDataCommons/service/actions/runs/10200019587/job/28218495676) due to a network timeout downloading the Gradle binary.

Use the [`setup-gradle` action](https://github.com/gradle/actions/blob/main/setup-gradle/README.md) to cache Gradle artifacts between runs, including both the Gradle binary and the dependencies it downloads. Caching should improve performance, protect against transitory network failures, and help us be a better neighbor by downloading less.

The `setup-gradle` action recommends the [`setup-java` action](https://github.com/actions/setup-java). Previously, we were defaulting to Java 11, the version [provided by the runner](https://github.com/actions/runner-images/blob/ubuntu22/20240721.1/images/ubuntu/Ubuntu2204-Readme.md). Specify Java 17, which is the most recent LTS supported by the version of Gradle used by the SDK.

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -24,6 +24,11 @@ jobs:
           path: 'sdk'
       - name: Copy API definition
         run: cp service/src/openapi.json sdk/openapi.json
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 21
       - name: Invoke the TypeScript generator
         working-directory: sdk
         run: ./gradlew generateSwaggerCode
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -29,6 +29,8 @@ jobs:
         with:
           distribution: temurin
           java-version: 21
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
       - name: Invoke the TypeScript generator
         working-directory: sdk
         run: ./gradlew generateSwaggerCode


jasonaowen:
As a bonus, the setup-gradle plugin also generates a nice [build summary](https://github.com/PhilanthropyDataCommons/service/actions/runs/10206429142?pr=1131)! Compare a [previous run](https://github.com/PhilanthropyDataCommons/service/actions/runs/10205493252) that does not have a summary.

slifty (approved):
Thank you for the detailed commit messages (also, sounds like we should change the sdk documentation to ask for Java 17 rather than 18!)

.github/workflows/sdk.yml
```diff
@@ -24,6 +24,13 @@ jobs:
           path: 'sdk'
       - name: Copy API definition
         run: cp service/src/openapi.json sdk/openapi.json
+      - name: Setup Java
+        uses: actions/setup-java@v4
```
bickelj, line 28: Excellent.

.github/workflows/sdk.yml
```diff
@@ -24,6 +24,13 @@ jobs:
...
+        with:
+          distribution: temurin
```
bickelj, line 30: I am used to Azul Zulu because that used to be the best (only) OpenJDK tested against the TCK. But I see Azul actually supports Temurin and Temurin ticks all the boxes. Nice.

.github/workflows/sdk.yml
```diff
@@ -24,6 +24,13 @@ jobs:
...
+          java-version: 17
```
bickelj, line 31: I think it's OK to build with Java 17 (or 21 for that matter) so long as we support 11 runtimes. That's more a matter for the sdk repo, though.

.github/workflows/sdk.yml
```diff
@@ -24,6 +24,13 @@ jobs:
...
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
```
bickelj, line 33: Excellent.

bickelj (dismissed):
I have no objection to this change here, but I think before merging it we should ensure that the SDK project targets a Java 11 runtime.

It doesn't look like we publish any SDK binaries at the moment and this build verifies that changes made in the `service` repo do not break anything in the `sdk` repo. So from that angle there is not necessarily a breaking change for Java 11 users because they'd have to build the SDK themselves and it could still work with Java 11. But to be sure that a Java 17 build of the SDK works on a Java 11 runtime we would specify the Java 11 target in the gradle builds in the SDK repo. [Example of setting a target runtime](https://github.com/PhilanthropyDataCommons/auth/blob/3e474bdf8e3ac2789427101b4b76e6642e5d0013/keycloak-required-action/build.gradle.kts#L19) and [docs](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation).

Having said all that, I think a simpler solution is to target Java 11 here. That way we don't have to change the SDK repo at all and we verify that Java 11 can compile, test, etc., and result in Java 11 classes.

Why do I insist on Java 11 support? Experience shows that updating from one major Java version to the next takes years (decades) in many organizations. So to have maximum compatibility with potential users of the SDK, pick the oldest Java LTS that is still likely in use by potential users. It's a judgment call where to draw that line but I guess 11 is still in wide use. I also see that Oracle's [proprietary Java 11 support extends out beyond 17 and 21](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) which tells me that it's widely deployed.

bickelj:
I guess another option is to have it build multiple times: once with 11, once with 17, and once with 21. That would provide maximum feedback of any potential problems for latest/greatest folks (which we want to encourage) and also for the slower ones.

jasonaowen:
> I think a simpler solution is to target Java 11 here. That way we don't have to change the SDK repo at all and we verify that Java 11 can compile, test, etc., and result in Java 11 classes.

As you mention a few times, @bickelj, we're not testing the SDK here, nor are we exporting any Java artifacts; the purpose of this workflow is to make sure that our OpenAPI specification can successfully be built into a TypeScript SDK, and to publish that SDK. That is, in this repo we are a client of the SDK, and we trust that it works, and so should not concern ourselves here with validating the SDK.

For this reason alone, I'd like to be on the latest LTS that we can support. Currently that's 17; with https://github.com/PhilanthropyDataCommons/sdk/pull/30 that will hopefully be 21.

> ...a breaking change for Java 11 users...

> ...updating from one major Java version to the next takes years (decades) in many organizations...

> ...another option is to have it build multiple times: once with 11, once with 17, and once with 21.

The [SDK repo already claims that it requires Java 18](https://github.com/PhilanthropyDataCommons/sdk/blob/49a3de58688b1fe0c7e0d9907f697b77fa0e4157/README.md#L7-L9), and [has from the start](https://github.com/PhilanthropyDataCommons/sdk/commit/6be58e57b8ebfed07c4d8d0d9efbff314fd659d7). To the best of my knowledge, nobody has complained yet about this requirement.

Rather than burden ourselves with the the responsibility of claiming and testing backwards support before anyone has asked for it, and enable bad behavior in large organizations by doing so, let's just stick to "latest LTS" until/unless we get a specific request for support.

I'm not advocating that we add any code that requires newer versions of Java, but there's only like 100 lines of Java code in the SDK repo, and our compatibility almost entirely depends on the libraries and plugins we're using.

> I think before merging [this PR] we should ensure that the SDK project targets a Java 11 runtime.

Why before?

bickelj:
@jasonaowen wrote:
> As you mention a few times, @bickelj, we're not testing the SDK here, nor are we exporting any Java artifacts; the purpose of this workflow is to make sure that our OpenAPI specification can successfully be built into a TypeScript SDK, and to publish that SDK. That is, in this repo we are a client of the SDK, and we trust that it works, and so should not concern ourselves here with validating the SDK.

Oh, right. Thank you for clarifying this bit: "the purpose of this workflow is to make sure that our OpenAPI specification can successfully be built into a TypeScript SDK, and to publish that SDK." We are not generating any Java classes here, just using the SDK. I was thinking this was our only test that verifies the SDK can build and also that we were building Java classes (not TypeScript code).

@jasonaowen wrote:
> For this reason alone, I'd like to be on the latest LTS that we can support. Currently that's 17; with [PhilanthropyDataCommons/sdk#30](https://github.com/PhilanthropyDataCommons/sdk/pull/30) that will hopefully be 21.

Yes, I agree. In this repository that makes perfect sense.

@jasonaowen wrote:
> The [SDK repo already claims that it requires Java 18](https://github.com/PhilanthropyDataCommons/sdk/blob/49a3de58688b1fe0c7e0d9907f697b77fa0e4157/README.md#L7-L9), and [has from the start](https://github.com/PhilanthropyDataCommons/sdk/commit/6be58e57b8ebfed07c4d8d0d9efbff314fd659d7). To the best of my knowledge, nobody has complained yet about this requirement.
> 
> Rather than burden ourselves with the the responsibility of claiming and testing backwards support before anyone has asked for it, and enable bad behavior in large organizations by doing so, let's just stick to "latest LTS" until/unless we get a specific request for support.
> 
> I'm not advocating that we add any code that requires newer versions of Java, but there's only like 100 lines of Java code in the SDK repo, and our compatibility almost entirely depends on the libraries and plugins we're using.

I would guess the reason no one has asked for it is no one has tried it yet. A less likely possibility is that those who only have Java 11 would be quietly warded off by the 18 requirement. In any case, my point about maximizing reach by minimizing Java version stands for the SDK repository. We absolutely should support the latest LTS and prioritize making that work. But if all the dependent libraries work on Java 11 we shouldn't bump the minimum up to 17, for example. Anyway, I'm OK with not explicitly testing 11, 17, 21 and sticking with testing the one LTS for now. Perhaps when changing the readme over in the SDK repo @slifty you can note that it may work with both earlier and later Java versions, to try it out and let us know. Something like that.

> > I think before merging [this PR] we should ensure that the SDK project targets a Java 11 runtime.
> 
> Why before?

Assuming we were testing Java 11 before, my point was to continue testing Java 11 before additionally testing Java 17.

In summary, I withdraw my objection here. This is a build step for TypeScript and not a direct test of the SDK. If there are SDK users in the future, testing the SDK and Java compatibility will become more important in that repository.

jasonaowen:
Thanks for your thoughtful reviews, @bickelj and @slifty!

I will rebase and update to Java 21 now that the gradle upgrade PR is merged.

PR #1179, opened 2024-09-10
Use supertest agents sparingly

### LINKED ISSUE #1171: Carefully use test agents to avoid test interaction
See [context of discovery](https://github.com/PhilanthropyDataCommons/service/pull/1129#issuecomment-2338394611).

Given the following form, when we use the `agent` in multiple tests, then for better or worse we are sharing state between tests:
```typescript
import request from 'supertest';
//...
// ...
describe('/someEndpoint', () => {
// ...
		it('does one thing', async () => {
// ...
			const response = await agent
// ...
				.expect(200);
// ...
		});
// ...
		it('does another thing', async () => {
// ...
			const response = await agent
// ...
				.expect(200);
// ...
		});
// ...
});
```

See this example (and especially the comment from https://www.npmjs.com/package/supertest:
```quote
Here's an example with mocha that shows how to persist a request and its cookies:

const request = require('supertest');
const should = require('should');
const express = require('express');
const cookieParser = require('cookie-parser');

describe('request.agent(app)', function() {
  const app = express();
  app.use(cookieParser());

  app.get('/', function(req, res) {
    res.cookie('cookie', 'hey');
    res.send();
  });

  app.get('/return', function(req, res) {
    if (req.cookies.cookie) res.send(req.cookies.cookie);
    else res.send(':(')
  });

  it('should save cookies', function(done) {
    agent
    .get('/')
    .expect('set-cookie', 'cookie=hey; Path=/', done);
  });

  it('should send cookies', function(done) {
    agent
    .get('/return')
    .expect('hey', done);
  });
});
```

In general, I think we want independent requests for independent tests. There may be an occasion or two where we want to share the state but that's rare.

bickelj:
There is potential (mis)use of shared state across tests when using supertest agents. Without this change, it can be easy to set, for example, auth headers on the agent rather than the request and those headers unintentionally affect tests later in the same file. This change avoids re-use of state by simply avoiding the use of agent.

In future, there may be some situations that need to use shared state across tests. In that case, a supertest agent is a fine tool for the job. See the example at https://www.npmjs.com/package/supertest where `const agent` is used.

Fixes #1171

### DIFF
--- a/src/__tests__/applicationForms.int.test.ts
+++ b/src/__tests__/applicationForms.int.test.ts
@@ -14,7 +14,6 @@ import { expectTimestamp } from '../test/utils';
 import { mockJwt as authHeader } from '../test/mockJwt';
 
 const logger = getLogger(__filename);
-const agent = request.agent(app);
 
 const createTestBaseFields = async () => {
 	await createBaseField({
@@ -36,11 +35,11 @@ const createTestBaseFields = async () => {
 describe('/applicationForms', () => {
 	describe('GET /', () => {
 		it('requires authentication', async () => {
-			await agent.get('/applicationForms').expect(401);
+			await request(app).get('/applicationForms').expect(401);
 		});
 
 		it('returns an empty array when no data is present', async () => {
-			const response = await agent
+			const response = await request(app)
 				.get('/applicationForms')
 				.set(authHeader)
 				.expect(200);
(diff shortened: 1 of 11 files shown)


bickelj:
I wondered if this had a performance impact but it looks about the same: about a minute per test run before and after.

slifty (approved):
Yay!  Great improvement and thanks for solving this.

Issue #1154, opened 2024-08-19
Support sourceId association with accounts


slifty:
After thinking through the problem in #1146 we're landing on associating an account with a source, which would then be used automatically whenever an account creates a proposal version.

Over time this may become more complex -- e.g. an account being able to specify a source, and being given the "permission" to associate an action with a given source.  For now it's enough to have each account only be associated with a single source.

slifty:
Based on some great observations in https://github.com/PhilanthropyDataCommons/service/issues/1146, we are liekly going to change course here and instaed make `sourceId` a parameter instead of a user association.  This will be more flexible and less likely to result in system-generated data that could end up being false.

PR #1191, opened 2024-09-13
Add `sourceId` to `Users`

slifty:
This PR updates the `users` table and type to support an optional association with a `source`.

Currently there is no way to actually set this value for a given user but this provides the necessary type definitions to support those endpoints.

This PR also includes an improvement to our `Writable` utility type in order to support optional values in a write (this is our first optional value in a type, fun fact!)

Related to #1154

### DIFF
--- a/src/__tests__/users.int.test.ts
+++ b/src/__tests__/users.int.test.ts
@@ -1,6 +1,12 @@
 import request from 'supertest';
 import { app } from '../app';
-import { createUser, loadSystemUser, loadTableMetrics } from '../database';
+import {
+	createOrUpdateFunder,
+	createSource,
+	createUser,
+	loadSystemUser,
+	loadTableMetrics,
+} from '../database';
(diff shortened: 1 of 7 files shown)


src/__tests__/users.int.test.ts
```diff
@@ -64,6 +70,40 @@ describe('/users', () => {
...
+				.set(authHeaderWithAdminRole)
+				.expect(200);
+			expect(response.body).toEqual({
+				total: userCount,
+				entries: [
+					{
+						authenticationId: 'totallyDifferentUser@example.com',
+						createdAt: expectTimestamp,
+						id: 3,
+						source,
```
slifty, line 100: Thoughts on this approach (of using the `source` object as returned by `createSource` rather than explicitly defining the shape of the expected result)?

bickelj:
Without objection closing due to the different direction. Re-open if needed.

PR #1197, opened 2024-09-16
Add source id parameter to `POST /proposalVersions`

slifty:
This PR improves the `POST /proposalVersions` endpoint to accept a sourceId.

It also makes an improvement to the `NotFoundError` to add structured data to the error type.

### DIFF
--- a/src/database/operations/assert/assertSourceExists.ts
+++ b/src/database/operations/assert/assertSourceExists.ts
@@ -1,5 +1,5 @@
 import { db } from '../../db';
-import { InputConflictError } from '../../../errors';
+import { NotFoundError } from '../../../errors';
 import type { CheckResult } from '../../../types';
 
 export const assertSourceExists = async (sourceId: number): Promise<void> => {
@@ -8,7 +8,7 @@ export const assertSourceExists = async (sourceId: number): Promise<void> => {
 	});
 
 	if (result.rows[0]?.result !== true) {
-		throw new InputConflictError('The source was not found', {
+		throw new NotFoundError(`Entity not found`, {
 			entityType: 'Source',
 			entityId: sourceId,
 		});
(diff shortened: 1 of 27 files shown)


src/handlers/proposalVersionsHandlers.ts
```diff
@@ -184,6 +181,17 @@ const postProposalVersion = (
 				});
 		})
 		.catch((error: unknown) => {
+			if (error instanceof NotFoundError) {
+				if (error.details.entityType === 'Source') {
+					next(
+						new InputConflictError(`The related entity does not exist`, {
```
bickelj, line 187: Very cool.

src/database/operations/assert/assertSourceExists.ts
```diff
@@ -1,5 +1,5 @@
 import { db } from '../../db';
-import { InputConflictError } from '../../../errors';
+import { NotFoundError } from '../../../errors';
```
bickelj, line 2: I guess we can't replace `InputConflictError` everywhere with `NotFoundError` because in some cases we simply surface the `NotFoundError` to the caller but in other cases it's a more complex situation of needing `entityXId` in order to post to `entityYId`, and so that's the `InputConflictError`?

slifty: I figured it's just that operations are not written in any context (e.g. this assertion may be made on something that has nothing to do with input).

On the other side: our errors map to status codes (not found errors would be a 404), whereas input conflict error would map to a different 400, so we do need to map it to inputconflicterror

bickelj (approved):
I was able to add a source locally and attach it to a new proposal version. In fact, I was required to attach it to a new proposal version, which is in fact what we want. It may be helpful to add an `example` to the `openapi.json` for the `sourceId` of `ProposalVersion` at this point.

Looks great to me, thank you!

PR #1199, opened 2024-09-17
Use `PUT` to add or update data provider

bickelj:
Fixes typo of `POST` in the API docs.


slifty (approved):
Whoops!  Thanks for the fix.

PR #1202, opened 2024-09-18
Add `sourceId` to `BulkUpload`

### LINKED ISSUE #1192: Accept `sourceId` as a parameter for data insertion API operations
Based on some conversation in https://github.com/PhilanthropyDataCommons/service/issues/1146 we would like to prioritize accepting `sourceId` as a parameter when uploading proposal data.

This allows a user writing to the API to specify which source the data came from.

For now we will continue to consider accounts with write access as trusted, and so this issue does NOT include adding any kind of system for authentication / permissions for writing a given source.  That is an improvement that can be made in a future iteration if / when prioritized.

This issue would obviate #1154

slifty:
This PR adds the `sourceId` attribute to the `BulkUpload` type and requires it as a parameter when creating a bulk upload.

It also includes a few little fixes of issues I saw when implementing the feature.

Resolves #1192

### DIFF
--- a/src/database/migrations/0034-alter-bulk_uploads-add-source_id.sql
+++ b/src/database/migrations/0034-alter-bulk_uploads-add-source_id.sql
@@ -1,0 +1,12 @@
+ALTER TABLE bulk_uploads
+  ADD COLUMN source_id INTEGER NOT NULL DEFAULT system_source_id(),
+  ADD CONSTRAINT fk_source_id
+    FOREIGN KEY(source_id)
+      REFERENCES sources(id);
+
+ALTER TABLE bulk_uploads
+  ALTER COLUMN source_id DROP DEFAULT;
+
+
+COMMENT ON COLUMN bulk_uploads.source_id IS
+  'The PDC source that the contents of the bulk upload came from.';
--- a/src/tasks/processBulkUpload.ts
+++ b/src/tasks/processBulkUpload.ts
@@ -20,7 +20,6 @@ import {
 	loadBaseFields,
 	loadBulkUpload,
 	loadOrganizationByTaxId,
-	loadSystemSource,
 	updateBulkUpload,
 } from '../database/operations';
 import { BulkUploadStatus, isProcessBulkUploadJobPayload } from '../types';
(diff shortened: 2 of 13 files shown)


src/tasks/processBulkUpload.ts
```diff
@@ -293,7 +291,7 @@ export const processBulkUpload = async (
 			const proposalVersion = await createProposalVersion({
 				proposalId: proposal.id,
 				applicationFormId: applicationForm.id,
-				sourceId: source.id,
+				sourceId: bulkUpload.sourceId,
```
bickelj, line 294: Excellent. What would be nifty, not now, but perhaps someday, is to somehow include "this came from source X via bulk upload" in the provenance, whether that be another source that has an expanded name or is chained to that source or related to that source, etc. Anyway that's way too complex for now, right now this is awesome.

src/database/migrations/0034-alter-bulk_uploads-add-source_id.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE bulk_uploads
+  ADD COLUMN source_id INTEGER NOT NULL DEFAULT system_source_id(),
+  ADD CONSTRAINT fk_source_id
+    FOREIGN KEY(source_id)
+      REFERENCES sources(id);
+
+ALTER TABLE bulk_uploads
+  ALTER COLUMN source_id DROP DEFAULT;
```
bickelj, line 8: Great incremental change going on here.

src/tasks/processBulkUpload.ts
```diff
@@ -271,7 +270,6 @@ export const processBulkUpload = async (
 
 		// TODO: Once accounts are associated with sources we should replace this system source with a real source.
```
slifty, line 271: This todo should be removed!

bickelj (approved):
I was able to locally build, lint, test, migrate, and start. After starting, I was able to create a pre-signed post, post the data, register the posted data with a different-than-default source, and then saw that the resulting proposal versions had that provenance associated. This is good, thank you!

slifty:
Thanks!  Rebasing to main and mergin'

slifty:
I just realized I haven't been updating the API version *or* adding change logs 😱

Adding that in here.

PR #1203, opened 2024-09-18
Organize operations by model

### LINKED ISSUE #1185: Organize DB operations by model at the first level
We currently organize queries as `modelName/operationModelName.sql`
We organize our operation function helpers as `operation/operationModelName.ts`

This is not consistent, and it's also more useful to have things organized by model rather than operation.

Let's move things around.

slifty:
this PR reorganizes our folder hierarchy for DB operations to be in terms of model rather than operation type.

Resolves #1185


bickelj (approved):
Looks good to me. While I have a different preference than what the commit message says, I think this is an improvement. The reason is it leads to slightly less scrolling up and down the directory tree while trying to update a single entity. In general, I think a flatter directory hierarchy with many files is better than a deeper hierarchy with few files, but obviously the rules are hard to pin down exactly. It would be nice to be able to make an update to a single entity without having to sweep up and down the directory tree a dozen times looking for this and that, but I don't know how realistic that goal is. Again, regardless of reasons, I think this is better.

The way I reviewed was first to skim and verify that the changes were one of:
* Changes to imports in a file,
* file moves,
* deletes of `index.ts`.

I looked at the directory structure and saw the more pleasing structure. I verified that I could build, lint, test locally, and that the tests passed.

slifty:
Thanks! I *will* tweak the commit message to remove the incorrect reasoning (I actually was trying to articulate what you're saying, and landed on my red herring explanation).

PR #1204, opened 2024-09-19
Avoid union type

### LINKED ISSUE #1188: Warning when running tests with `--no-silent`
When running tests with --no-silent I get this message:

```
  ● Console

    console.warn
      strict mode: use allowUnionTypes to allow union type keyword at "#/properties/createdBy" (strictTypes)
```

The tests run, but probably worth looking into addressing the warning.

slifty:
This PR changes one of our schema definitions to avoid a union type (which is not recommended for various reasons, and was generating a warning).

Resolves #1188


slifty:
Whoops, a broken test!  Fixing that, sorry for the premature review request.

bickelj (approved):
I have no cause for objection but I did not dig in very deeply.

PR #1207, opened 2024-09-19
Add `createdBy` to `ProposalVersion`

### LINKED ISSUE #1200: Add `createdBy` to `ProposalVersion`
This will help us audit insertions of data, also `Proposal` already has `createdBy` so this shouldn't involve R&D

slifty:
This PR adds a new `createdBy` attribute for `ProposalVersion`.  This doesn't affect required fields from end users since this is populated basted on auth context, but will add to the returned object.

Resolves #1200

### DIFF
--- a/src/database/migrations/0035-alter-proposal_versions-add-created_by.sql
+++ b/src/database/migrations/0035-alter-proposal_versions-add-created_by.sql
@@ -1,0 +1,8 @@
+ALTER TABLE proposal_versions
+  ADD COLUMN created_by INTEGER NOT NULL DEFAULT select_system_user_id(),
+  ADD CONSTRAINT fk_createdBy
+    FOREIGN KEY (created_by)
+      REFERENCES users(id);
+
+ALTER TABLE proposal_versions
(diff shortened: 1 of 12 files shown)


bickelj (approved):
After migration, and with a valid JWT, I was able to create a new Proposal Version and I saw a (new) user id associated with me associated with the new Proposal Version. Great!

src/database/migrations/0035-alter-proposal_versions-add-created_by.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE proposal_versions
+  ADD COLUMN created_by INTEGER NOT NULL DEFAULT select_system_user_id(),
+  ADD CONSTRAINT fk_createdBy
```
jasonaowen, line 3: PostgreSQL will lower-case this unquoted mixed-case [identifier](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS). I think you want it to be snake_cased.

PR #1209, opened 2024-09-19
Normalize bundle operation parameters

### LINKED ISSUE #1186: Make offset and limit required, named parameters for bundle functions
This is a two-part issue.

1. Our bundle loaders currently take a `queryParameters` object.  Let's move away from that pattern and into having a function parameter signature such that we have required parameters as true parameters, with a potential `options` parameter if there are any optional parameters.

2. Let's make offset and limit required parameters (no longer passed via `queryParameters`

Thus we'd move something like:

```
export const loadBaseFieldLocalizationsBundle = async (queryParameters: {
	offset?: number;
	limit?: number;
	baseFieldId?: number;
}): Promise<Bundle<BaseFieldLocalization>> => ...
```

to 

```
export const loadBaseFieldLocalizationsBundle = async (
        offset: number;
	limit: number;
	options: {
	     baseFieldId?: number;
	}
}): Promise<Bundle<BaseFieldLocalization>> => ...
```

slifty:
This PR improves (and normalizes) bundle operation parameters.  Specifically it adds `offset` and `limit` as required parameters, and it also moves all optional parameters into an `options` object.

This should not result in any functional changes.

Resolves #1186

### DIFF
--- a/src/__tests__/baseFields.int.test.ts
+++ b/src/__tests__/baseFields.int.test.ts
@@ -3,12 +3,12 @@ import { app } from '../app';
 import {
 	createBaseField,
 	createOrUpdateBaseFieldLocalization,
-	loadBaseFieldLocalizationsBundle,
+	loadBaseFieldLocalizationsBundleByBaseFieldId,
 	loadBaseFields,
 	loadTableMetrics,
 } from '../database';
 import { BaseFieldDataType, BaseFieldScope, PostgresErrorCode } from '../types';
-import { expectTimestamp } from '../test/utils';
+import { expectTimestamp, NO_LIMIT, NO_OFFSET } from '../test/utils';
 import {
 	mockJwt as authHeader,
 	mockJwtWithAdminRole as adminUserAuthHeader,
@@ -575,7 +575,7 @@ describe('/baseFields', () => {
 		});
 
 		it('creates the specified base field localization if it does not exist', async () => {
-			await createTestBaseField();
+			const testBaseField = await createTestBaseField();
 			const before = await loadTableMetrics('base_field_localizations');
 			await request(app)
 				.put('/baseFields/1/localizations/fr')
@@ -587,9 +587,12 @@ describe('/baseFields', () => {
 				})
 				.expect(200);
 			const after = await loadTableMetrics('base_field_localizations');
-			const baseFieldLocalizations = await loadBaseFieldLocalizationsBundle({
-				baseFieldId: 1,
-			});
+			const baseFieldLocalizations =
+				await loadBaseFieldLocalizationsBundleByBaseFieldId(
+					testBaseField.id,
+					NO_LIMIT,
+					NO_OFFSET,
+				);
 			expect(before.count).toEqual(0);
 			expect(baseFieldLocalizations.entries[0]).toMatchObject({
 				baseFieldId: 1,
@@ -601,7 +604,7 @@ describe('/baseFields', () => {
 		});
 
 		it('updates only the specified base field if it does exist', async () => {
-			await createTestBaseField();
+			const testBaseField = await createTestBaseField();
 			await createOrUpdateBaseFieldLocalization({
 				baseFieldId: 1,
 				language: 'fr',
@@ -625,9 +628,12 @@ describe('/baseFields', () => {
 				})
 				.expect(200);
 			const after = await loadTableMetrics('base_field_localizations');
-			const baseFieldLocalizations = await loadBaseFieldLocalizationsBundle({
-				baseFieldId: 1,
-			});
+			const baseFieldLocalizations =
+				await loadBaseFieldLocalizationsBundleByBaseFieldId(
+					testBaseField.id,
+					NO_LIMIT,
+					NO_OFFSET,
+				);
 			expect(before.count).toEqual(2);
 			expect(baseFieldLocalizations.entries[0]).toMatchObject({
 				baseFieldId: 1,
--- a/src/database/operations/applicationForms/loadApplicationFormBundle.ts
+++ b/src/database/operations/applicationForms/loadApplicationFormBundle.ts
@@ -2,20 +2,14 @@ import { loadBundle } from '../generic/loadBundle';
 import type { JsonResultSet, Bundle, ApplicationForm } from '../../../types';
 
 export const loadApplicationFormBundle = async (
-	queryParameters: {
-		offset?: number;
-		limit?: number;
-	} = {},
+	limit: number | undefined,
+	offset: number,
 ): Promise<Bundle<ApplicationForm>> => {
-	const defaultQueryParameters = {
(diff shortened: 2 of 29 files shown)


src/handlers/usersHandlers.ts
```diff
@@ -22,18 +21,14 @@ const getUsers = (
...
-	const searchParameters = extractSearchParameters(req);
-	const authenticationIdParameters = extractAuthenticationIdParameters(req);
+	const { offset, limit } = getLimitValues(paginationParameters);
+	const { authenticationId } = extractAuthenticationIdParameters(req);
 
 	(async () => {
-		const userBundle = await loadUserBundle(
-			{
-				...getLimitValues(paginationParameters),
-				...searchParameters,
```
slifty, line 32: Just wanted to point out that we were incorrectly passing `searchParameters` to a function that did not use or expect them.  This new structure won't actually prevent this mistake in future, but it *WILL* mean those values don't get further passed along to the query builder.

src/database/operations/applicationForms/loadApplicationFormBundle.ts
```diff
@@ -2,20 +2,14 @@ import { loadBundle } from '../generic/loadBundle';
 import type { JsonResultSet, Bundle, ApplicationForm } from '../../../types';
 
 export const loadApplicationFormBundle = async (
-	queryParameters: {
-		offset?: number;
-		limit?: number;
-	} = {},
+	offset: number,
+	limit: number,
```
jasonaowen, line 6: This is somewhat nitpicky, but since we're setting the pattern here, I would prefer for the order of the pagination parameters to match [the order they occur in SQL](https://www.postgresql.org/docs/current/queries-limit.html): limit, then offset.

```suggestion
	limit: number,
	offset: number,
```

This is a big change to a small PR, but I think this is when we need to do it!

src/database/operations/baseFieldLocalization/loadBaseFieldLocalizationsBundle.ts
```diff
@@ -5,21 +5,20 @@ import type {
...
-		baseFieldId: 0,
-		offset: 0,
-		limit: 0,
-	};
+export const loadBaseFieldLocalizationsBundle = async (
+	offset: number,
+	limit: number,
+	options: {
+		baseFieldId?: number;
+	} = {},
```
jasonaowen, line 13: Is `baseFieldId` ever missing in the places we call it? Do we need to maintain that flexibility? If not, we can make `baseFieldId` a required positional parameter, too, like `limit` and `offset`. And, since apparently I have opinions on parameter ordering, having it be first seems the most readable to me: `loadBaseFieldLocalizationsBundle(baseFieldId, limit, offset)`.

src/database/operations/applicationForms/loadApplicationFormBundle.ts
slifty: I'm fine making it; though that order always throws me for a loop (since I think in my mind offset is "where to start" and limit is "where to end")

jasonaowen: If it helps, they're also in alphabetical order! (for these two parameters, at least)

src/handlers/usersHandlers.ts
jasonaowen: Good catch!

jasonaowen (commented):
My suggested change below I think applies to almost all of the loaders: let's avoid default values when our only callers have explicit values, and prefer positional parameters to objects.

(I feel more strongly about the former than the latter; it is a bit disappointing to give up the clear parameter naming passing objects around gives!)

src/database/operations/baseFieldLocalization/loadBaseFieldLocalizationsBundle.ts
slifty: I think in this case we did resolve that we'll never want to allow a person to load ALL base field localizations at once so you're absolutely right!

Also, I can't respond to the top level comment directly but will say I agree with the philosophy: if something can be required it doesn't go in the `options`.

slifty: I actually don't think the suggest change applies to all of the loaders though -- in the vast majority of cases the option is optional at the query param level (e.g. you don't HAVE to specify `fooId` or `whatever the filter is` so while a value is being passed, that value is generally `string | undefined` or `number | undefined` so in most cases we'll keep the options.)

slifty: I am somewhat tempted, in this case, to add `ByBaseFieldId` to the operation name to signal that there is no way to generate a broader bundle

slifty: Update: we talked through this and the bottom line is....

being able to receive undefined does not an optional parameter make!

Which is to say: by the time we get to the operation, we know the shape of the data.  Just because something can be `undefined` doesn't mean it should be an optional parameter (it's just that `undefined` can be an expected value for that parameter, which may have meaning internally).

All that is to say, I'll take out the concept of `options` on all of the handlers and instead we'll use named parameters.

src/__tests__/baseFields.int.test.ts
```diff
@@ -587,9 +587,8 @@ describe('/baseFields', () => {
 				})
 				.expect(200);
 			const after = await loadTableMetrics('base_field_localizations');
-			const baseFieldLocalizations = await loadBaseFieldLocalizationsBundle({
-				baseFieldId: 1,
-			});
+			const baseFieldLocalizations =
+				await loadBaseFieldLocalizationsBundleByBaseFieldId(1, 100, 0);
```
jasonaowen, line 591: Ah, this change highlights that we were leaning on the default parameters in the test!

To your point yesterday about not liking to pass around a bunch of bare numbers, would it be worth extracting some constants to use in the tests?

```suggestion
			const baseFieldLocalizations = await loadBaseFieldLocalizationsBundleByBaseFieldId(
			    testBaseFieldId,
			    testLimit,
			    testOffset,
			);
```

jasonaowen (approved):
This looks good to me! Thanks for grinding through this, @slifty - I think this leaves us in a better place!

I was able to test this locally by poking around the Organizations section in the front-end and seeing the expected data and no errors in the service log.

src/__tests__/baseFields.int.test.ts
slifty: Ok so (1) yes and (2) I just realized: limit should be able to be undefined (since passing undefined to limit is the same as saying "return everything"

I'm creating two constants for tests: `MAX_LIMIT` (undefined) and `MIN_OFFSET` (0).

For base field ID I'll just access the test base field object that is returned by `createTestBaseField()`

PR #1211, opened 2024-09-20
Add universal `drop_function` sql utility

### LINKED ISSUE #1180: Add universal "drop function" for our initialization queries
We define some SQL functions that package tables into JSON.

Rather than use migrations for this (which would result in a lot of redundant code across migrations every time we wanted to tweak that form), we use "initialization" scripts which build out the functions.

@bickelj recently noted that in postgres, a function is defined not just by the name but also by the signature, which means "drop or replace" can leave old functions in tact in the event that a signature changes.

Rather than have our initialization script filled with old signatures that stopped being used months ago, let's use an approach that will destroy all forms of a given function name for these scripts.  An example is outlined here: https://stackoverflow.com/questions/7622908/drop-function-without-knowing-the-number-type-of-parameters

slifty:
This PR creates and uses a sql helper function that will drop all versions of a given custom defined function in the database regardless of the function signature.

To test this out, try creating a named function that has a different signature as one of the initialized functions:

e.g. 
```
CREATE OR REPLACE FUNCTION application_form_to_json(application_form application_forms, foo VARCHAR DEFAULT NULL)
RETURNS VARCHAR AS $$
BEGIN
  RETURN 'hello';
END;
$$ LANGUAGE plpgsql;
```

Then start the application using `npm run start` and then check to see if your custom function still exists:

```
SELECT * from pg_catalog.pg_proc WHERE proname = 'application_form_to_json' and pg_function_is_visible(oid);
```

You should see the `hello` is removed.

Resolves #1180

### DIFF
--- a/src/database/initialization/application_form_field_to_json.sql
+++ b/src/database/initialization/application_form_field_to_json.sql
@@ -1,4 +1,6 @@
-CREATE OR REPLACE FUNCTION application_form_field_to_json(application_form_field application_form_fields)
+SELECT drop_function('application_form_field_to_json');
+
+CREATE FUNCTION application_form_field_to_json(application_form_field application_form_fields)
 RETURNS JSONB AS $$
 DECLARE
   base_field_json JSONB;
--- a/src/database/initialization/organization_to_json.sql
+++ b/src/database/initialization/organization_to_json.sql
@@ -1,6 +1,6 @@
--- An explicit DROP is needed to remove the old function instead of overloading it.
-DROP FUNCTION IF EXISTS organization_to_json(organizations);
-CREATE OR REPLACE FUNCTION organization_to_json(organization organizations, authenticationId VARCHAR DEFAULT NULL)
+SELECT drop_function('organization_to_json');
+
+CREATE FUNCTION organization_to_json(organization organizations, authenticationId VARCHAR DEFAULT NULL)
 RETURNS JSONB AS $$
 DECLARE
   proposal_field_values_json JSONB;
--- a/src/database/migrations/0036-create-function-drop_function.sql
+++ b/src/database/migrations/0036-create-function-drop_function.sql
@@ -1,0 +1,20 @@
+CREATE OR REPLACE FUNCTION drop_function(function_name TEXT)
+RETURNS VOID AS $$
+DECLARE
+    r RECORD;
+    function_signature TEXT;
+BEGIN
+    FOR r IN
+        SELECT p.oid,
+               p.proname,
+               pg_catalog.pg_get_function_identity_arguments(p.oid) AS args
+        FROM pg_proc p
+        JOIN pg_namespace n ON p.pronamespace = n.oid
+        WHERE p.proname = function_name
+        AND n.nspname = current_schema()
+    LOOP
+        function_signature := r.proname || '(' || r.args || ')';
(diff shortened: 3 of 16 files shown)


src/database/migrations/0036-create-function-drop_function.sql
```diff
@@ -0,0 +1,22 @@
+-- This function came from StackOverflow which has a CC BY-SA 4.0 license
+-- https://stackoverflow.com/a/7623246/159522
```
jasonaowen, line 2: I'm not sure this is adequate attribution (it doesn't mention the author, for example), and I am also not sure how this licensing would interact with the rest of our codebase. If possible, let's avoid copying and pasting from StackOverflow.

Thank you for being clear about its provenance!

src/database/initialization/organization_to_json.sql
```diff
@@ -1,5 +1,5 @@
--- An explicit DROP is needed to remove the old function instead of overloading it.
-DROP FUNCTION IF EXISTS organization_to_json(organizations);
+SELECT drop_function('organization_to_json');
+
 CREATE OR REPLACE FUNCTION organization_to_json(organization organizations, authenticationId VARCHAR DEFAULT NULL)
```
jasonaowen, line 3: This is the only function we've changed the signature of, isn't it? What's the problem with this approach?

jasonaowen (changes requested):
I think I would be okay with this if it were original work :face_with_diagonal_mouth: 

That said, I don't think I understand the problem this is trying to solve - have we previously left behind functions? Examining the git log of `src/database/initialization/`, I can't find any examples of things we missed by changing the type signature.

src/database/initialization/organization_to_json.sql
slifty: We now have a piece of code that only exists because a signature, at on point in time, existed in the code base.

That's not a sustainable way to delete things (this isn't a migration).  Does that drop need to be there for the rest of time?  What happens the next time we change a function signature, do we just accumulate clutter in the file forever?

This function accomplishes what we had intended `CREATE OR REPLACE` to accomplish (but which doesn't)

src/database/migrations/0036-create-function-drop_function.sql
slifty: I can re-write if needed, or I can just ask chat GPT to write it for me if that's better somehow!

slifty: Fixed!

bickelj (approved):
Looks good to me other than some potential concerns noted in the code.

I tested in `psql` by adding two functions that overloaded `application_form_to_json` and `user_to_json` respectively. I saw the two new overloads present with `\df`, ran the `build`, `migrate`, and `start` commands, saw that the overloads disappeared using `\df`, and verified that the `GET /applicationForms` and `GET /users` API calls still worked via the local OpenAPI site.

src/database/initialization/application_form_field_to_json.sql
```diff
@@ -1,3 +1,5 @@
+SELECT drop_function('application_form_field_to_json');
+
 CREATE OR REPLACE FUNCTION application_form_field_to_json(application_form_field application_form_fields)
```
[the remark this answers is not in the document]
slifty: Let's do it!

PR #1213, opened 2024-09-20
Type check test files too during lint

### LINKED ISSUE #1208: Running `lint` doesn't detect errors in test files
If there are type errors in test files the tsc check doesn't detect them. It would be nice if it did (tests fail of course, but that takes longer to run than the static analysis).

slifty:
This PR fixes our type check lint step to also check test files.

You can test to make sure this works by changing a test so it calls some function with improper parameters, running `yarn lint` and seeing that the linter properly flags the mistake.

Resolves #1208


jasonaowen (approved):
Good catch! I was able to successfully test this locally.

PR #1214, opened 2024-09-20
Remove use of AuthenticatedRequest typecast

### LINKED ISSUE #1210: Don't use `AuthenticatedRequest` as the `req` type
In many of our handlers we use `AuthenticatedRequest` as the defined type for the `req` parameter.  @jasonaowen asked "why?" and my answer was just that it helps signal to the dev that a request is expected to have been authenticated.

While this may be true, it could be misleading since there's no actual validation associated.  We should instead be taking in a `Request` and type guarding to ensure that various decorations were applied by middleware as needed by a given handler.

We already use this pattern for `AuthContext`

slifty:
This PR removes some unsafe type casting for `req` values.

Resolves #1210

### DIFF
--- a/src/handlers/organizationsHandlers.ts
+++ b/src/handlers/organizationsHandlers.ts
@@ -8,7 +8,6 @@ import {
 	isId,
 	isWritableOrganization,
 	isTinyPgErrorWithQueryContext,
-	AuthenticatedRequest,
 	isAuthContext,
 } from '../types';
 import { DatabaseError, InputValidationError } from '../errors';
@@ -46,7 +45,7 @@ const postOrganization = (
 };
 
 const getOrganizations = (
-	req: AuthenticatedRequest,
+	req: Request,
 	res: Response,
 	next: NextFunction,
 ): void => {
@@ -68,7 +67,7 @@ const getOrganizations = (
 };
 
 const getOrganization = (
-	req: AuthenticatedRequest,
+	req: Request,
 	res: Response,
 	next: NextFunction,
 ): void => {
--- a/src/middleware/__tests__/requireAdministratorRole.int.test.ts
+++ b/src/middleware/__tests__/requireAdministratorRole.int.test.ts
@@ -2,10 +2,19 @@ import { requireAdministratorRole } from '../requireAdministratorRole';
 import { UnauthorizedError } from '../../errors';
 import type { Response } from 'express';
 import type { Request as JWTRequest } from 'express-jwt';
+import type { User } from '../../types';
+
+const getMockedUser = (): User => ({
+	id: 1,
+	authenticationId: 'foo@example.com',
+	createdAt: '',
+});
 
 describe('requireAuthentication', () => {
 	it('calls next with an UnauthorizedError when no roles value is provided', (done) => {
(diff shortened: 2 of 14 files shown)


src/handlers/organizationsHandlers.ts
```diff
@@ -45,13 +45,15 @@ const postOrganization = (
...
+	req: Request,
 	res: Response,
 	next: NextFunction,
 ): void => {
 	const paginationParameters = extractPaginationParameters(req);
 	const proposalParameters = extractProposalParameters(req);
-	const authenticationId = req.user?.authenticationId;
+	const authenticationId = isAuthContext(req)
+		? req.user?.authenticationId
+		: undefined;
```
jasonaowen, line 56: Since we repeat this several times, it might be worth extracting a helper function! `getAuthenticationId(Request) -> string | undefined` or so?

jasonaowen (approved):
This looks good to me! Were you planning on doing the middleware in a separate PR?

src/handlers/organizationsHandlers.ts
slifty: That seems reasonable, but this block gets blown away in the normalization PR so I'd prefer to leave it as is for this specific PR and circle back once main reflects the latest and greatest of all these changes.

slifty:
> Were you planning on doing the middleware in a separate PR?

Ah I'll include that in this one I think -- though man, TypeScript + express don't play nicely (express wants middlware to decorate objects with new attributes, whereas TypeScript gets upset if you assign a value to an attribute that is not part of a given type)

For instance

image

But I'll see what the best way to do this is.

slifty:
@jasonaowen I believe that actually type casting *is* the appropriate way to decorate req with a role.  That said, it is probably safer to do that cast only at the point of casting:

e.g.

```
	(req as AuthenticatedRequest).role = {
		isAdministrator: authRoles.includes(PDC_ADMIN_ROLE),
	};
```

Rather than casting req at the parameter / definition level.

slifty:
@jasonaowen I know you approved but I did end up adding a second commit to also remove from the middleware.  For that reason could you re-review with that context?  I don't believe I changed anything in the first commit so you can hone in on the second.

src/handlers/organizationsHandlers.ts
slifty: It ended up being more related to the now-merged code / doesn't fit in this PR.

I opened a new one for it: https://github.com/PhilanthropyDataCommons/service/pull/1217

src/middleware/__tests__/requireAdministratorRole.int.test.ts
```diff
@@ -2,10 +2,19 @@ import { requireAdministratorRole } from '../requireAdministratorRole';
 import { UnauthorizedError } from '../../errors';
 import type { Response } from 'express';
 import type { Request as JWTRequest } from 'express-jwt';
+import type { User } from '../../types';
+
+const getMockedUser = (): User => ({
+	id: 1,
+	authenticationId: 'foo@example.com',
```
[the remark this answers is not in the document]
slifty: Yeah, it will be but I'll resolve it once I rebase to main no problem (I assume this will merge first).

PR #1218, opened 2024-09-23
Remove placeholder cases

### LINKED ISSUE #1189: Correct queries that assumed null could not be passed
I was under the impression that null / undefined could not be passed into queries as parameters.  This may have been true at one point (?) but it isn't true any longer.

A few of our queries have awkward fake "defaults" that they map to NULL.  Let's remove that unnecessary complexity.

slifty:
This PR cleans up our queries to remove unnecessary CASE statements which were checking for various "placeholder" values that were meant to signify `NULL`.

It also corrects a default search querystring value to go from `''` to `undefined` (though it also makes sure to treat `''` as null at the query level still)

Resolves #1189

### DIFF
--- a/src/database/queries/baseFieldLocalizations/createOrUpdateByPrimaryKey.sql
+++ b/src/database/queries/baseFieldLocalizations/createOrUpdateByPrimaryKey.sql
@@ -13,5 +13,4 @@ ON CONFLICT (base_field_id, language)
 DO UPDATE SET
   label = EXCLUDED.label,
   description = EXCLUDED.description
-WHERE EXCLUDED.base_field_id = :baseFieldId AND EXCLUDED.language = :language
 RETURNING base_field_localization_to_json(base_field_localizations) AS "object";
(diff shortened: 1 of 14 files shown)


src/database/queries/baseFieldLocalizations/createOrUpdateByPrimaryKey.sql
```diff
@@ -13,5 +13,4 @@ ON CONFLICT (base_field_id, language)
 DO UPDATE SET
   label = EXCLUDED.label,
   description = EXCLUDED.description
-WHERE EXCLUDED.base_field_id = :baseFieldId AND EXCLUDED.language = :language
```
jasonaowen, line 16: Good catch, @slifty!

For further context, I believe this is the conversation you were referring to in the commit message: https://github.com/PhilanthropyDataCommons/service/pull/1123#discussion_r1718929284

jasonaowen (approved):
This looks good to me - much cleaner! Thanks, @slifty.

src/database/queries/baseFieldLocalizations/createOrUpdateByPrimaryKey.sql
slifty: Ah yes good find -- I'll link to that directly.

bickelj:
Nice!

Issue #1155, opened 2024-08-19
Update `users` table to use natural key


slifty:
Right now we exclusively use synthetic keys (an `id` column with auto increment) across all of our tables.

@jasonaowen has raised the idea of moving to use natural keys when there is a candidate.  We are having a deeper discussion about this over in #1141 for `base_fields` but as I'm about to improve our `users` table I wanted to raise it here as well.

`users` has an `account_id` column which is required, unique, guid that is generated by an authentication system.  At this point it is a reasonable natural key, and I believe since it's a guid it is decoupled from keycloak (e.g. if we added other auth systems it would still be a valid natural key).

It won't change, it's required, and it's unique -- a solid natural key candidate!

Let's update `users` to ditch our synthetic ID.

This will simplify the work of #1154 since we'll only need to create a `PUT` endpoint.
It will also mean that our clients will be able to avoid looking up a user ID from our system if they know how to parse an `accountId` from a jwt.

bickelj:
@slifty I'm OK with it.

slifty:
Since opening this issue at some point I spoke with @jasonaowen about the topic and he suggested we consider being even more specific to what `accountId` is.

In particular: it's a keycloak ID / UUID.

I originally was concerned that we would be coupling too closely with Keycloak and might not be able to support other auth services, but he pointed out that Keycloak SHOULD be the thing we integrate with, and that it can handle integration with other services directly.

To that end, I'd like to propose we rename it to `keycloak_id` while also making it a natural key (and then setting the column type to UUID so we can have db validation of the values).

PR #1225, opened 2024-09-26
Update `authenticationId` to `keycloakUserId`

slifty:
This PR updates the `authenticationId` attribute of the `User` entity to be the more specifically defined `keycloakUserId` (a GUID).

The actual content of the field does not change, since we've been populating it with keycloak user IDs this entire time.

This field is going to ultimately become the primary key for the users table.

Related to #1155


slifty:
I apologize for the incorrect branch name -- I decided to break this into two pieces but forgot to update the branch!

src/database/operations/users/loadUserByKeycloakUserId.ts
```diff
@@ -2,20 +2,20 @@ import { db } from '../../db';
 import { NotFoundError } from '../../../errors';
 import { JsonResultSet, type User } from '../../../types';
 
-export const loadUserByAuthenticationId = async (
-	authenticationId: string,
+export const loadUserByKeycloakUserId = async (
+	keycloakUserId: string,
```
jasonaowen, line 6: I'm disappointed the [`uuid` package](https://www.npmjs.com/package/uuid) (or its [types](https://www.npmjs.com/package/@types/uuid)) don't contain a `Uuid` type; is it worth making a newtype for this? `string` is so vague!

src/database/queries/users/selectWithPagination.sql
```diff
@@ -2,8 +2,8 @@ SELECT user_to_json(users.*) as "object"
 FROM users
 WHERE
   CASE
-    WHEN :authenticationId != '' THEN
-      authentication_id = :authenticationId
+    WHEN :keycloakUserId != '' THEN
+      keycloak_user_id = :keycloakUserId::uuid
```
jasonaowen, line 6: I don't think you need an explicit cast here; the equals operator will implicitly cast as necessary except when it's ambiguous.

src/middleware/addUserContext.ts
```diff
@@ -16,12 +16,12 @@ const addUserContext = (
 	res: Response,
 	next: NextFunction,
 ): void => {
-	const authenticationId = req.auth?.sub;
-	if (authenticationId === undefined || authenticationId === '') {
+	const keycloakUserId = req.auth?.sub;
+	if (keycloakUserId === undefined || keycloakUserId === '') {
```
jasonaowen, line 20: Should we also defend against the system user ID here? It's not completely clear to me if the intent of `authenticationId === ''` was to check for a missing ID or for an incorrectly used system ID or both.

src/test/utils.ts
```diff
@@ -21,15 +21,16 @@ export const generateNextWithAssertions = (
 		runAssertions(err).then(done).catch(done);
 	});
 
-export const getTestUserAuthenticationId = () => 'foo@example.com';
+export const getTestUserKeycloakUserId = () =>
+	'00000000-0000-0000-0000-000000000001'; // This value is not a reference, it's just a static GUID
```
jasonaowen, line 25: This is extremely picky, so feel free to ignore me, but can this be a more obviously different value than the system ID?

```suggestion
	'11111111-1111-1111-1111-111111111111'; // This value is not a reference, it's just a static GUID
```

jasonaowen (changes requested):
Looking good! Thanks for taking this on, @slifty. A few small things below.

I haven't tested this; only looked at the code.

src/__tests__/bulkUploads.int.test.ts
```diff
@@ -40,7 +40,7 @@ describe('/bulkUploads', () => {
 			const systemSource = await loadSystemSource();
 			const testUser = await loadTestUser();
 			const thirdUser = await createUser({
-				authenticationId: 'totallyDifferentUser@example.com',
```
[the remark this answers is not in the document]
slifty: Yes indeed -- we always populated based on the `sub` of the auth token which has always been the UUID (but when writing tests we could use whatever string we used).

src/__tests__/users.int.test.ts
```diff
@@ -64,12 +65,20 @@ describe('/users', () => {
...
 			const { count: baseUserCount } = await loadTableMetrics('users');
-			await Array.from(Array(20)).reduce(async (p, _, i) => {
+			const uuids = Array.from(Array(20)).map(() => uuidv4());
+			await uuids.reduce(async (p, uuid) => {
 				await p;
 				await createUser({
-					authenticationId: `user-${i + 1}`,
+					keycloakUserId: uuid,
 				});
 			}, Promise.resolve());
```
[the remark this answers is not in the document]
slifty: `forEach` does not allow you to halt processing to wait to call one iteration until the other has halted; this pattern lets us `await` the previous promise before continuing, which is important since order of insertion determines which created users will be returned.

src/database/migrations/0037-alter-users-change-authentication_id-to-keycloak_user_id.sql
```diff
@@ -0,0 +1,35 @@
+BEGIN;
```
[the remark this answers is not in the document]
slifty: Oh right yes good call!

src/database/operations/users/loadSystemUser.ts
```diff
@@ -1,7 +1,22 @@
...
-const loadSystemUser = async (): Promise<User> =>
-	loadUserByAuthenticationId('');
+const loadSystemUser = async (): Promise<User> => {
+	const result = await db.sql<JsonResultSet<User>>(
+		'users.selectSystemUser',
+		{},
+	);
+	const item = result.rows[0]?.object;
+	if (item === undefined) {
+		throw new NotFoundError(`Entity not found`, {
```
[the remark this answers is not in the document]
slifty: Sure thing: I think the best test plan here would be to override the database call so it returns an empty result set / never actually interacts with the SQL.

src/database/operations/users/loadUserByKeycloakUserId.ts
slifty: Ha I actually did just start to do this when working on the next commit (and was going to backport it here).

This will be handy: https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

src/database/queries/users/selectWithPagination.sql
slifty: I added this in response to an error I was getting from psql complaining that there was no such thing as a conversion from text to uuid.

I'll try removing it and get back to you, maybe this solution was a red herring.

src/test/utils.ts
slifty: I'm all here for this kind of picky -- good idea.

src/__tests__/users.int.test.ts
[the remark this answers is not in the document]
jasonaowen: Oh, yes thank you, now I remember the conversations we've had about this before!

src/middleware/addUserContext.ts
slifty: It was indeed, and in fact... I really don't want hard coded values (because it leads to mistakes like this).  I wonder if there would be a good way to load that user ID directly from the DB on service start, and to check against somehow...  I'll figure something good out.

src/database/queries/users/selectWithPagination.sql
jasonaowen: Oh, that sounds plausible - maybe it's different in a prepared transaction than in `psql`. If so, nevermind, and sorry for the hassle!

src/__tests__/users.int.test.ts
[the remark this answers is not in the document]
slifty: FWIW I don't like it either :(

src/database/operations/users/loadUserByKeycloakUserId.ts
slifty: OK so I looked into the literal types and... it would be a bit of a bear / I'm not convinced it would actually be a reasonable type to make.

I did make a `KeycloakUserId` type but it's just an alias for `string` (which means it isn't providing any actual type safety beyond just being a string).

This means TypeScript would allow, for instance:

```
const notAUuid = `lol`

const user: User = {
  keycloakUserId: notAUuid,
}
```

slifty: OH -- I'm seeing you said `newtype for this` not `new type for this`

and seeing that's what newtype is for.

Ok, sorry, I'm caught up now.  Looking into it all.

src/middleware/addUserContext.ts
slifty: I've added a new `config.ts` which is initialized on server start and which populates a `systemUser` which can be accessed via `getSystemUser()` and I think this will do the trick.

src/database/queries/users/selectWithPagination.sql
slifty: Yeah -- I did try taking this out in the context of a NULL check and it didn't like me one bit (claimed to not be able to tell what the type is).

I think this has to be there :( marking as resolved for now.

src/database/initialization/user_to_json.sql
```diff
@@ -5,7 +5,7 @@ RETURNS JSONB AS $$
 BEGIN
   RETURN jsonb_build_object(
     'id', "user".id,
-    'authenticationId', "user".authentication_id,
+    'keycloakUserId', "user".keycloak_user_id::text,
```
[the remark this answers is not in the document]
slifty: Schweet you're right -- yeah it seems all of this casting stuff is really just an issue for queries with parameters (as opposed to direct psql)

src/database/operations/users/loadUserByKeycloakUserId.ts
slifty: I've implemented this as a separate commit; it's an augmentation and new pattern, and a fairly beefy one all considered so felt appropriate to implement it as a second step.

src/types/KeycloakUserId.ts
```diff
@@ -0,0 +1,41 @@
...
+const isKeycloakUserId = ajv.compile(keycloakUserIdSchema);
+
+const stringToKeycloakUserId = (s: string): KeycloakUserId => {
+	if (!isKeycloakUserId(s)) {
+		throw new InternalValidationError(
+			'Invalid KeycloakUserId',
+			isKeycloakUserId.errors ?? [],
+		);
+	}
+	return s;
```
jasonaowen, line 27: Should this be a call to `wrap`? I'm not sure I understand the tradeoff here - maybe it's better as-is! Just trying to understand.

```suggestion
	return keycloakUserIdIsomorphism.wrap(s);
```

src/types/KeycloakUserId.ts
```diff
@@ -0,0 +1,41 @@
+import { Newtype, iso as createNewtypeIsomorphism } from 'newtype-ts';
```
jasonaowen, line 1: Haha, now that I know what it stands for I understand why they went with `iso`!

jasonaowen (approved):
Thanks for diving into this! The `Newtype` stuff is exciting, although it comes with some boilerplate. Now that you've played with it a bit, does it feel like a worthwhile tradeoff to you?

I've tested this a bit by running it locally, signing in, and poking around, and not seeing any errors in the logs.

src/types/KeycloakUserId.ts
slifty: Good catch yes it should

src/types/KeycloakUserId.ts
slifty: The irony lol (I can keep it as iso if you prefer, but it felt totally opaque to me 😅)

src/types/KeycloakUserId.ts
slifty: I was looking at this and saying "oh dear, shouldn't our type checker have caught this" and realized that `isKeycloakUserId` is lying -- it is saying it is a UUID, not that it is the newtype.  I'm going to change the schema and the `isFoo` to be `uuidStringSchema` and `isUuidString` to prevent this kind of confusion.

bickelj:
The best thing about this change is the clarity. "What is this ID, really? Oh, it's a Keycloak User ID, obviously!"

The next best thing is being able to use a `uuid` type in the database. I assume that's going to be better than `varchar`, but I'm just assuming.

However, somewhere in the back of my head I recall not wanting to be explicitly wedded to a particular authentication/authorization provider. Is that no longer the case?

jasonaowen:
> The best thing about this change is the clarity. "What is this ID, really? Oh, it's a Keycloak User ID, obviously!"

:tada: 

> The next best thing is being able to use a `uuid` type in the database. I assume that's going to be better than `varchar`, but I'm just assuming.

Yes, both in terms of storage size (a reduction by at least half) and functionality: the [UUID type](https://www.postgresql.org/docs/current/datatype-uuid.html) will accept input in a variety of formats, and output a normalized form. Plus there are a few [UUID functions](https://www.postgresql.org/docs/current/functions-uuid.html) I've never needed to use.

> However, somewhere in the back of my head I recall not wanting to be explicitly wedded to a particular authentication/authorization provider. Is that no longer the case?

@slifty and I talked about that! If we need to allow authentication through another provider (someone's internally hosted IdP, Sign in with Google, a managed instance, etc), we'll implement that integration at the Keycloak level rather than at the app level. I vaguely recall talking about that kind of integration before, but I don't remember if that was floated as a possibility or was a decision; however, since we haven't been testing with any other IdPs, I would not be surprised if already have code dependencies on peculiarities specific to Keycloak.

src/types/KeycloakUserId.ts
slifty: As a very belated followup here, after our conversation about how newtype works / how it does not actually modify data I'm not sure that `wrap` is actually any more meaningful than a type guard.  All wrap is doing is changing the type signature of the value (under the hood it is still a string).

To that end, this actually *is* fine, technically.

That being said, I think that it is cleaner and more explicit to devs to use wrap and `isUuid`.

I am, however, currently towards having the `User.keycloakUserId` attribute typed as a `KeycloakUserId` rather than a `Uuid` that would need to be explicitly wrapped before it could be used.  Less boilerplate for the same outcome.

slifty: All righty -- I think the changes are small enough that this doesn't need another review (famous last words!).  I do think it's cleaner now (and we can even benefit from TS to at least nudge any devs writing tests into having proper static strings)

PR #1247, opened 2024-10-11
Use `keycloakUserId` as natural primary key for the `users` table

### LINKED ISSUE #1155: Update `users` table to use natural key

slifty:
This PR updates the `users` table to use `keycloakUserId` as the primary key along with any foreign key relationships with other entities.

This change prompted an immediate improvement to the clarity  names of a few authentication-related query parameters.

Resolves #1155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Upgraded to use OpenAPI Specification 3.1.
 
+## 0.14.0 2024-10-11
+
+### Changed
+
+- `Users.id` no longer exists; `keycloakUserId` is now the primary identifier of any given `User`.
+- `Proposal.createdBy` is now a UUID reference to the creator's `keycloakUserId`.
+- `ProposalVersion.createdBy` is now a UUID reference to the creator's `keycloakUserId`.
+- `BulkUpload.createdBy` is now a UUID reference to the creator's `keycloakUserId`.
+
 ## 0.13.0 2024-09-26
 
 ### Changed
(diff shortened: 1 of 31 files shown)


bickelj (approved):
It migrates, runs, works via the Swagger API page, and I see nothing to suggest we shouldn't merge this. Thanks!

It does raise the question of the need for a separate entity called `Users`, though. If we use the Keycloak `sub` (subject) ID everywhere, do we need to join to a table at all? It perhaps keeps a record of valid, known users that have been encountered in JWTs, and that is perhaps useful.

Due to this attribute being a former non-primary-key I think it could become a URL path parameter rather than query parameter on `/users` but this is minor and should not hold up the merge.

slifty:
@bickelj good points on all fronts.

Once we implement some of the more granular permissions we may find a use for the Users table (e.g. caching the keycloak username, or some other similar function).  If after that there's no use in sight though, then yeah let's strongly consider taking out the entity entirely.

Issue #1201, opened 2024-09-17
Include provenance in gold organization data


bickelj:
A continuation of #1138 .

bickelj:
Note to self: use the full authentication context rather than a single `authenticationId`.

bickelj:
@slifty Do you mind if I defer/delay the inclusion of the full auth context till it is used here? I just noticed I failed to include that here.

PR #1248, opened 2024-10-11
Use provenance in gold data retrieval

bickelj:
Prior to this commit, the organization aka change maker data retrieved were merely sorted by the datetime of insert into the PDC database. Now that provenance is potentially available via the `Sources` table, this commit creates a priority order for field values based on source:
1. organization sourced,
2. funder sourced,
3. data platform provider sourced,
4. old data marked as `pdc`-sourced (no provenance available).

After sorting fields by these classes of data sources, then within these classes the data are sorted by the datetime of insert. Finally, as before, a winner is chosen: the first value from the sorted data.

Suppose multiple funders posted values for an organization's "city", a data platform provider posted a value, and there are three old values left over from the era of pre-provenance data. In this case, the order of rows will have the multiple funders by datetime of insert followed by the data platform provider value followed by the old pre-provenance values by datetime of insert. The first value will win: the latest value from the funders.

Issue #1201

### DIFF
--- a/src/__tests__/organizations.int.test.ts
+++ b/src/__tests__/organizations.int.test.ts
@@ -7,9 +7,11 @@ import {
 	createOpportunity,
 	createOrganization,
 	createOrganizationProposal,
+	createOrUpdateFunder,
 	createProposal,
 	createProposalFieldValue,
 	createProposalVersion,
+	createSource,
 	loadSystemSource,
 	loadSystemUser,
 	loadTableMetrics,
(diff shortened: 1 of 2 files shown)


slifty (changes requested):
The query change looks good and makes sense -- my only request is that we add test(s) for this new behavior.

bickelj:
@slifty I added one test. I presume I should add more, but how many more? Should it be exhaustive? I suppose not.

slifty:
@bickelj one is inifintely better than none so good enough I think!

In the grand scheme we may want to come up with some kind of test fixture that takes a data state and an expected "gold" set, and uses some kind of utility to translate that input state into proper database state, runs the query, and compares it with the expectation.

That can be done as a separate issue IMHO.

slifty (approved):
Thank you for the test, looks good!

PR #1253, opened 2024-10-14
Rename `Organization` to `Changemaker`

### LINKED ISSUE #1117: Rename `Organization` to something more appropriate (semantically close to `Changemaker`).
We currently store an entity called `Organization` (as opposed to `Proposal`)

On call today @kfogel suggested we should make this more specific: `ChangeMaker`

This seems like a reasonable idea, especially as we are about to add new entity types that represent "organizations" (in a real world sense) and they are distinct from Organization/ChangeMaker.

Implementing this change would mean adjusting our endpoints in a breaking way; that's OK to do, but it is something we should try to only do once if we can help it.  The bottom line question to me is: "is ChangeMaker a forever term?"

---

EDIT: since opening we have honed in on `Applicant` instead of `Organization`

slifty:
This PR changes our `Organization` entity to `Changemaker` (and also the `OrganizationProposal` entity to `ChangemakerProposal`

Resolves #1117


src/routers/index.ts
```diff
@@ -23,8 +23,8 @@ rootRouter.use('/bulkUploads', bulkUploadsRouter);
 rootRouter.use('/dataProviders', dataProvidersRouter);
 rootRouter.use('/funders', fundersRouter);
 rootRouter.use('/opportunities', opportunitiesRouter);
-rootRouter.use('/organizations', organizationsRouter);
-rootRouter.use('/organizationProposals', organizationProposalsRouter);
+rootRouter.use('/changemakers', changemakersRouter);
```
bickelj, line 26: Perhaps these should be above `/dataProviders` to have alphabetical order in the file.

src/openapi.json
```diff
@@ -1349,20 +1349,20 @@
...
 					{
 						"auth": []
 					}
 				],
 				"parameters": [
 					{
-						"name": "organizationId",
-						"description": "The PDC-generated ID of an organization.",
+						"name": "changemakerId",
+						"description": "The PDC-generated ID of an changemaker.",
```
bickelj, line 1365: An changemaker or a changemaker?

src/__tests__/changemakers.int.test.ts
```diff
@@ -173,7 +173,7 @@ describe('/organizations', () => {
 			});
 		});
 
-		it('does not return duplicate organizations when an organization has multiple proposals', async () => {
+		it('does not return duplicate changemakers when an changemaker has multiple proposals', async () => {
```
bickelj, line 176: An/a

bickelj (approved):
It achieves the goal. I was able to lint, build, migrate, and run locally. The changemaker endpoints seemed to work.

This is an actual breaking change for the front-end, right? I assume we're all OK for it to go down or other changes have been made there.

src/openapi.json
slifty: Asking the deep philosophical questions (thank you!  I did get a little "replace" happy in my manual find/replace)

slifty:
yes -- you're right that this will completely break the frontend; that said I believe that's part of the process at this point...

Issue #1257, opened 2024-10-15
Create test cases for the data conflict scenarios identified in #1176


slifty:
In #1176 @bickelj laid out a bunch of specific scenarios for conflict resolution which are essentially functional specifications for how our data resolution logic should work when retrieving an organization.

Next step is to convert these scenarios into tests (and of course make sure the tests pass)

bickelj:
While making more tests in #1262 and looking at the scenarios in #1176, I realized that the example scenarios in #1176 were either examples or ambitious examples. In other words, we have no capability of asking the web about an IRS EIN, for example, so integrating to that is currently not a thing. Unless we want it to be, in which case we have to build that capability.

slifty:
I believe this is now complete

PR #1262, opened 2024-10-17
More gold data tests

bickelj:
Improve gold data tests with some refactoring and additional examples of changemaker data conflicts.

Issue #1257

### DIFF
--- a/src/__tests__/changemakers.int.test.ts
+++ b/src/__tests__/changemakers.int.test.ts
@@ -261,253 +261,255 @@ describe('/changemakers', () => {
 			await request(app).get('/changemakers/foo').set(authHeader).expect(400);
 		});
 
-		it('returns the latest valid value for a base field when auth id is sent', async () => {
-			const systemSource = await loadSystemSource();
-			const systemUser = await loadSystemUser();
+		describe('tests that find gold data among proposals and share a common db', () => {
+			it('returns the latest valid value for a base field when auth id is sent', async () => {
+				const systemSource = await loadSystemSource();
+				const systemUser = await loadSystemUser();
 
-			// Make a base field associated with one opportunity/org, and we'll make three responses.
-			const baseFieldId = (
-				await createBaseField({
-					label: 'Fifty one fifty three',
-					shortCode: 'fifty_one_fifty_three',
-					description: 'Five thousand one hundred fifty three.',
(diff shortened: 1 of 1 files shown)


bickelj:
@slifty @jasonaowen What do you think of this approach of pulling some of the more common elements that are not test-specific into a `beforeEach` in a dedicated `describe` block? One thing I like is that the `beforeEach` can have the calls to production code, thus removing those calls from test coverage, thus making the test coverage more honest and separating those failures out. I might even want to do an extra `describe` and `beforeEach` block for individual tests due to this benefit! One drawback is the separation of the setup from the test. It's no longer self-contained. I can't simply go to the test and read it to see everything that's going on. However, as long as the data are carefully crafted and declared in the same file, it should be easy to see what's happening. As far as I can tell this is like what Meszaros calls the "Shared Fixture."

src/__tests__/changemakers.int.test.ts
```diff
@@ -262,29 +272,53 @@ describe('/changemakers', () => {
...
-				// To advance test changemaker IDs beyond 1-2.
-				await insertTestChangemakers();
-				// This should be ID 3, although unfortunately implicitly, and only when tests run in serial.
-				const changemakerId = (
-					await createChangemaker({
-						name: 'Five thousand one hundred forty seven reasons',
-						taxId: '05119',
-					})
-				).id;
+			beforeEach(async () => {
```
jasonaowen, line 292: This is great! I like this approach; good idea, @bickelj.

bickelj: Thanks.

bickelj:
@slifty @jasonaowen My guess that extracting test setup into `beforeEach` makes the test more honest is probably false. I thought I could improve the test coverage honesty with a simple commit that added several `/* istanbul ignore next */` statements before test setup blocks but I saw no change in coverage. Even running a single integration test shows upwards of 50% code coverage across the code base. Because this is a much deeper rabbit hole than I imagined, I created #1267 separately.

slifty (approved):
Looks good!

As we know over time some of these rules may shift as we get more user data, but this is a great reflection of the expected implementation of our current requirements!

bickelj:
bfa7d7f rebased on main without any other modification.

Issue #118, opened 2022-11-15
API Key failing


mortimal1:
Using the API key was always returning 401 unauthorized. Tracing code found in apiKeyChecker.ts line 22
if (validApiKeys.includes(apiKey.toString())) { 

was always returning false even though the key matched. Changing to: 

if (validApiKeys.toString().includes(apiKey.toString())) { 

worked

bickelj:
Thank you for reporting this issue.

I suppose the software expects a keys file with Unix-style line endings. I wonder if there are carriage returns in the keys file. If on Windows, perhaps use Notepad++ to see and change it (Edit -> EOL Conversion).

bickelj:
@mortimal1 Did this suggestion help?

mortimal1:
yes, that worked - thank you. Could you have used os.EOL instead of hard coding '\n'?

bickelj:
@mortimal1 Great suggestion, done in #123.

PR #1265, opened 2024-10-19
Remove Docker build action

jasonaowen:
Now that we are using a different deployment system, we no longer need to build Docker images. Those were only ever for us - we don't know of any users who relied on the published images.

Save build time, storage space, and remove the time-bomb that is using a personal access token to notify another repository.

After this is merged, delete the `REPO_ACCESS_TOKEN` secret from this repository.

Issue PhilanthropyDataCommons/deploy#118 Migrate to DigitalOcean Apps
Issue #622 Change deploy model


jasonaowen:
I do not touch the `Dockerfile` in this PR, as I recognize that some of you may be using it. If you are, speak up; if nobody is, maybe we should remove it, or at least move it into a `contrib` subdirectory to indicate that it should work but it's not being actively maintained?

slifty:
@jasonaowen Just to be explicit: this PR is doing two things -- no longer triggering deploy from docker, but also no longer building the docker file.

I do agree that we don't need to support docker for development at this point, but I also wanted to just be explicit that this is doing more than just what is described in #118.

If we *were* to have a real docker-dev environment we would probably want to use docker compose and set up psql + keycloak containers as well.  I think that's a reasonable thing to do if it were a priority.  We may find that that becomes a valued product when we have a third party dev team coding against the API.

slifty (approved):
I think it's a reasonable choice to remove docker support right now, though I would not be surprised if we decide to revisit that decision in the coming months depending on third party dev feedback.

jasonaowen:
That's a good callout, thank you, @slifty. I'll update the commit message and rebase, then merge after checks have passed.

Issue #1250, opened 2024-10-14
Support granting of permissions for users in the context of organizational entities


slifty:
We want to support more granular permissions (per #1093) and the first step to being able to do this is to support relationships between users and our three types of organizational entity (Funder, ChangeMaker, and DataProvider).

For this issue those associations won't actually impact permissions in terms of viewing data yet.

This issue *will* include logic that allows users associated with a given organization to be able to add other users to their organization.  As mentioned in #1093 we may eventually want to support the concept of an admin user, but for MVP we can stick with a higher trust model.

slifty:
I'm going to be so bold as to add a `user_role` column to the relationship table, because it is pretty simple to implement now.

```
CREATE TYPE user_role AS ENUM (
  'administrator',
  'editor',
  'viewer',
)
```

These roles seem straightforward and universal.

slifty:
Some decisions we made which I want to lock in here:

We want endpoints that look like this...

-- PUT /users/{userKeycloakUserId}/roles/funders/{shortCode}/accessType/administrator => {}
-- GET /users/{userKeycloakUserId}/roles/funders/{shortCode}/accessType/administrator
-- DELETE /users/{userKeycloakUserId}/roles/funders/{shortCode}/accessType/administrator

bickelj:
I think the permissions are good, but to the point of the original post here, I think we need to use Keycloak 26's Organization feature to group people together into organizations, letting those groups come through in the JWT.

PR #1284, opened 2024-10-28
Support organizational roles

slifty:
This PR adds internal support for the creation of role / access grants to users for our three organizational entity types.

It does not yet provide endpoints for creation and deletion of those roles, nor does it apply the roles when determining whether or not a user is allowed to perform a given action.  These will come in a future PR.

Related to #1250


src/types/ChangemakerRole.ts
```diff
@@ -0,0 +1,36 @@
+import { ajv } from '../ajv';
+import { AccessType } from './AccessType';
+import type { JSONSchemaType } from 'ajv';
+import type { Writable } from './Writable';
+import type { KeycloakUserId } from './KeycloakUserId';
+
+interface ChangemakerRole {
+	readonly changemakerId: number;
+	readonly userKeycloakUserId: KeycloakUserId;
+	readonly accessType: AccessType;
```
bickelj, line 10: 2. At first glance a column seemed to be missing. I would expect "subject permission object" to be declared somewhere. For example, "Alice can manage proposals for changemaker Bob." Where does the object such as "changemaker Bob" fit in here? Oh, is it `changemakerId`? Perhaps the order is confusing or there is some ambiguity. I thought that this `Role` declared "User `userKeycloakUserId` is part of Changemaker `changemakerId`" (`changemakerId` is part of the subject) rather than "User `userKeycloakUserId` has permission `accessType` on object `changemakerId` (`changemakerId` is part of the object). It is still unclear as of writing this but perhaps it will be clear while looking further.

src/database/migrations/0040-create-changemaker_roles.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TYPE access_type_t AS ENUM (
+  'manage',
+  'edit',
+  'view'
+);
+
+CREATE TABLE changemaker_roles (
+  user_keycloak_user_id UUID NOT NULL,
```
bickelj, line 8: 3. This order makes it feel more like the User is the subject and the Changemaker is the object (as an English speaker). If this is how it works, then I think it might be even clearer to flip the order of `changemaker_id` and `access_type` so that the table order is User, Access, Changemaker. And also make the TypeScript types match this same order to be consistent. Perhaps a comment on each of the User/Access/Object columns or a single comment on each Role table would help clarify the meaning. The same goes for the TypeScript types.

src/database/initialization/user_to_json.sql
```diff
@@ -2,9 +2,57 @@ SELECT drop_function('user_to_json');
...
+  changemaker_roles_json JSONB := NULL::JSONB;
+  funder_roles_json JSONB := NULL::JSONB;
+  data_provider_roles_json JSONB := NULL::JSONB;
 BEGIN
+  changemaker_roles_json := (
+    SELECT jsonb_object_agg(
+      changemaker_role_maps.changemaker_id, changemaker_role_maps.role_map
+    )
+    FROM (
+      SELECT changemaker_roles.changemaker_id, jsonb_object_agg(changemaker_roles.access_type, TRUE) AS role_map
```
bickelj, line 16: 5. What does `TRUE` do here? (I can investigate later but want to finish this review quickly).

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -132,6 +132,9 @@ erDiagram
 	Proposal }o--|| User : "is created by"
 	ProposalVersion }o--|| User : "is created by"
 	BulkUpload }o--|| User : "is created by"
+	User }o--o{ Changemaker : "is granted access within"
```
bickelj, line 135: 6. This almost says "`Changemaker` in `ChangemakerRole` is the object but it's still not clear.

bickelj (commented):
I was able to lint, build, and test this PR locally, great!

It is not clear to me if this change implements role-based access control or access control lists. In other words, I cannot tell if `Changemaker` in `ChangemakerRole` is the subject or the object. The word `Role` to me connotes subject. If `Changemaker` there is truly a subject, then I think the object is missing in the entity. If `Changemaker` there is the object, then I think the word `Role` is inappropriate. Perhaps `ChangeMakerAccessControl` or something like that would be more appropriate. 

I think at least one application of the access control and an associated test would help clarify.

src/types/index.ts
```diff
@@ -7,10 +7,13 @@ export * from './BulkUpload';
 export * from './Bundle';
 export * from './Changemaker';
 export * from './ChangemakerProposal';
+export * from './ChangemakerRole';
 export * from './CheckResult';
 export * from './DataProvider';
+export * from './DataProviderRole';
 export * from './express/AuthenticatedRequest';
 export * from './Funder';
+export * from './FunderRole';
```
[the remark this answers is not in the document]
slifty: I agree and for what it's worth I did explore this and implemented both ways locally to try it out -- the alternative unfortunately felt similarly awkward but also in a way that was limiting for future functionality.

**First the awkward;** A single table would mean either:

(1) Using polymorphic association: No foreign key relationships, but we can have a natural primary key
(2) Using null-able columns (what we do for `sources`): Foreign key relationships, but no natural primary keys

If we want both foreign keys and natural primary keys, having the three tables to represent the three types of role the only solution I came up with :-/

**Second, the the functionality / data consistency;** By having three tables for the three types of role relationship we can eventually change the `access_type` for a given org entity to be more granular and specific to the activities related to that organization.

For instance, we expect to have `edit_opportunities` and `edit_proposals` as separate access types in the context of a funder.  I was planning on building that in in a future iteration rather than here, but it doesn't feel like a stretch to consider now.

---

Not sure if that's at all compelling, but I think I at least reflected the exploration and the tradeoffs I found.

src/types/ChangemakerRole.ts
slifty: I'll describe the intention here and we can work from there to land on where to go:

The intention is that a record in this table would say "`userKeycloakUserUd` has permission to `accessType` data related to `ChangemakerId`.

For instance: `{ Alice } has permission to { edit proposal data } related to { Foo Inc. }`

src/database/migrations/0040-create-changemaker_roles.sql
slifty: That's a really good idea

src/database/initialization/user_to_json.sql
slifty: The result of this is an object where keys are access types, and values are booleans (well, `true` specifically).

So:

`{ edit: true }`

A client can then say something like `if (user.roles.funders.fordFoundation.edit) then ...`

That said, this might not be the right object structure.  Maybe an array of access types would be better.

src/database/migrations/0042-create-data_provider_roles.sql
jasonaowen, line 1: Why have these as three separate migrations? There's only a single commit, I think we want them all together - if one of the three were to somehow fail, we'd be left in an uncomfortable place.

src/database/migrations/0042-create-data_provider_roles.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE data_provider_roles (
+  user_keycloak_user_id UUID NOT NULL,
+  data_provider_short_code short_code_t NOT NULL,
+  access_type access_type_t NOT NULL,
+  created_by UUID NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  PRIMARY KEY (user_keycloak_user_id, data_provider_short_code, access_type),
+  FOREIGN KEY (created_by) REFERENCES users(keycloak_user_id) ON DELETE CASCADE,
```
jasonaowen, line 8: I find it more readable / intuitive to have single-column foreign keys defined as column constraints rather than table constraints:

```suggestion
  created_by UUID NOT NULL REFERENCES users(keycloak_user_id) ON DELETE CASCADE,
  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
  PRIMARY KEY (user_keycloak_user_id, data_provider_short_code, access_type),
```

Having the FK part of the column definition makes it more obvious to me what the column is, rather than having to look back up at the column once I see the FK table constraint.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: Regarding the name clarity: what do you think of `UserChangemakerRole`

Meaning "the user's role within a changemaker"

(and then the same for the other two)

src/database/migrations/0042-create-data_provider_roles.sql
slifty: Sounds good to me

src/database/initialization/user_to_json.sql
slifty: On reflection I think an array is more honest; the fact that `false` values don't get entries in the object makes the object misleading.  An array naturally communicates that it is only communicating what is present.

Updating this (and `TRUE` will no longer be there)

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: I've made this change, I think it is more clear (unresolve if not!)

src/types/ChangemakerRole.ts
slifty: The names AND attribute order are now all updated to hopefully make more intuitive sense.

slifty:
@bickelj you asked to show an example of using this in code but I think that's premature since there's no way to grant access via API in this PR (I can add it all to one PR but the PR will get much bigger!)

That said, one example of an application of this would be that a user with edit access to a foundation would be allowed to create proposal versions for any proposal associated with that foundation's opportunities, or add new proposals to an opportunity.

slifty:
@jasonaowen we talked about taking out `view` but I am confident that I'll be prioritizing the implementation of the view permissions in the next two weeks.

src/database/initialization/user_to_json.sql
bickelj: Oh, nice. I truly was asking and didn't mean to say "do a bunch more" but I'm glad that the question raised a better alternative!

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
bickelj: Between yesterday's review and today I thought perhaps `ChangemakerAccess` or now maybe `UserChangemakerAccess`, `UserChangemakerPermission`, `UserPermissionOnChangemaker` would be better, assuming we're talking about access controls on objects associated with that Changemaker. I think that's what we're talking about but when you say "within a changemaker" above, it makes me worry that we're substituting these ACLs for group membership at the same time. I would say "the user's access permissions on a changemaker's objects." In other words, these are truly User permissions and not Changemaker permissions. The User is the subject and the Changemaker is the object, right? I will re-review and see if that is consistent.

bickelj:
> @bickelj you asked to show an example of using this in code but I think that's premature since there's no way to grant access via API in this PR (I can add it all to one PR but the PR will get much bigger!)
> 
> That said, one example of an application of this would be that a user with edit access to a foundation would be allowed to create proposal versions for any proposal associated with that foundation's opportunities, or add new proposals to an opportunity.

This helps. I thought of another clarifying question. Does a row in the `[Yada]Changemaker[Yada]` table imply membership in that Changemaker or is it exclusively access to that Changemaker's stuff? For example, can I as an OTS team member have rows for my user in `[Yada]Changemaker[Yada]` and that not be misinterpreted as my membership of that Changemaker?

src/types/UserChangemakerRole.ts
```diff
@@ -0,0 +1,39 @@
+import { ajv } from '../ajv';
+import { AccessType } from './AccessType';
+import type { JSONSchemaType } from 'ajv';
+import type { Writable } from './Writable';
+import type { KeycloakUserId } from './KeycloakUserId';
+
+interface UserChangemakerRole {
+	readonly userKeycloakUserId: KeycloakUserId;
+	readonly accessType: AccessType;
+	readonly changemakerId: number;
```
bickelj, line 10: Nice. I take this order to be the subject-permission-object scheme I was looking for. I don't think of this as a role but I suppose it could be thought of as a role. Usually I think of role as the hat one wears in an organization, such as "I do software engineering: I have a role of software engineer" and "I also administer some systems: I also have a role of system administrator." In other words, roles are associated with people, not with the objects in the access control. After the roles are established there would be a separate modeling of  permissions on objects granted to certain roles.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -132,6 +132,9 @@ erDiagram
...
+	User }o--o{ Funder : "is granted access within"
```
bickelj, line 136: Replacing "within" with "to" or "on" I think clears it up.

bickelj:
Maybe I actually want the group membership thing to be modeled somewhere. That would be one more level of indirection. Right now it looks like User to Changemaker relationships where the relation indicates access to that Changemaker's objects. But nowhere can I declare "this User is a member of that Changemaker" and then bind that Changemaker to have access to objects. What I just mentioned gives up simpler modeling in the software to gain simpler management for customers in the long run. If Groups or Keycloak Roles are granted permissions to objects instead of Users, adding a User gives that User all the permission needed straightaway.

src/types/index.ts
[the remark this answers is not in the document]
bickelj: I think I'm OK with the three tables as long as they are access/permissions tables (associated with objects) rather than roles tables (implying subjects).

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: I like `UserPermissionOnChangemaker` -- and then in stead of `access_type` it maybe should be `permission`

Instead of `roles` we then have `permissions`

Thanks @bickelj for the refinements!

src/types/UserChangemakerRole.ts
slifty: Very true -- I think the word `role` made sense originally when we were thinking in terms of `administrator, editor, viewer` but when we switched to actions (aka permissions) the word should have changed.  I agree with your naming suggestion in an earlier comment.

slifty:
> If Groups or Keycloak Roles are granted permissions to objects instead of Users, adding a User to a Group or Role gives that User all the permission needed straightaway.

Agreed!

What do you think of sticking with this simpler model for this PR, but exploring the potential of moving to a more advanced (and Keycloak) driven one as a future iteration?

bickelj:
Yes, use this as a first increment. I think it can be adapted pretty easily to fit whatever we do in Keycloak.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: Futher improvement from our call right now: `PermissionOnChangemaker` is even better as it becomes more flexible down the line (we may have this table grant permissions to groups instead of / as well as individual users)

slifty: We landed on `UserChangemakerPermissions` for now; we can adjust this name later if / when the scope expands beyond individual users.

src/openapi.json
```diff
@@ -966,13 +966,142 @@
 					}
 				]
 			},
+			"AccessType": {
```
bickelj, line 969: I see this might need an update.

bickelj (approved):
OpenAPI spec needs an update to use "Permission" language but otherwise skimming over this it looks great!

Issue #1274, opened 2024-10-22
Create a CHANGEMAKER_DATA.md file in `/docs`


slifty:
This should document the intent of the current logic (the scenarios) and should also note some of the known "edge cases" / potential limitations of our current logic.

bickelj:
I should have read the OP here, but is it essential to include the scenarios or the edge cases? The doc as I wrote it briefly describes how to use the `/changemakers` endpoint and includes details of how the heuristic currently works.

bickelj:
Aside or GitHub gripe: why don't Project status transitions appear directly in the feed? Maybe in another 10 years GitHub will reach feature parity with Redmine.

bickelj:
I named it `/docs/CHANGEMAKER_DATA.md` because it is specifically about changemakers.

PR #1300, opened 2024-10-31
Add documentation for "gold" Changemaker data

bickelj:
Without this commit, one would be left to read the source code to try to make sense of the values coming out of `GET /changemakers`. While source code may be well-commented, it does not capture the high-level intent that drove people to write it. This commit adds documentation with some details about the `/changemakers` endpoint and special attention to the method it uses to retrieve field values.

Issue #1274


bickelj:
Failed lint on a markdown file, wow. Thanks, `prettier`! :facepalm:

hminsky2002 (approved):
To the extent that this explained our golddata system to me just reviewing it, cool! Approved, though I'd imagine @slifty and @jasonaowen should probably look through this first

docs/CHANGEMAKER_DATA.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Storing Changemaker data
+
+To add a Changemaker to the PDC, use `POST /changemakers`. Note, however, that
+this endpoint only registers a Changemaker by tax ID and name. It is not the
+endpoint to add other Changemaker attributes such as location or contacts. One
+writes Changemaker data to PDC via `POST /proposalVersions` and its associated
+endpoints. In other words, proposal data contains much data about Changemakers,
+therefore proposal data is an important source of Changemaker data within the
+PDC. As of this writing proposals are the sole source of Changemaker data.
```
slifty, line 14: Right now this section is awkward because we actually just don't support the use case that it seems to be aimed at: editing / managing changemaker data in the context of that changemaker.

I would probably adjust this to just say "changemaker data cannot be directly written to the PDC, though that functionality is on the roadmap.  Instead, changemaker data is aggregated from any proposals that have been submitted by the changemaker"

PR #1301, opened 2024-10-31
Start a glossary

bickelj:


bickelj:
I failed to mention this in the commit, but I structured it in a way that might be usable by a tool such as [glossarify](https://www.npmjs.com/package/glossarify-md) in the future.

slifty (approved):
So sorry for the delay of this review -- looks like a great start, thanks for kicking it off!

Issue #1138, opened 2024-08-06
Implement gold (aka best) organization data retrieval


bickelj:
Following #1087 which has had sufficient thought, do it, implement it, iterate on it.

slifty:
This is actually done via https://github.com/PhilanthropyDataCommons/service/pull/1129

There is another iteration, but we'll open a new issue for that.

PR #1314, opened 2024-11-07
Update base fields to have org scope

bickelj:
Without this change the "gold" data feature does not return gold data. This is because the base fields need 'organization' scope. This changes the relevant base fields in one fell swoop.

Issue #1138


bickelj:
@slifty As I pushed this I thought maybe you will want this change to be done via a GET followed by a PUT in the API.

slifty:
To your point above @bickelj I don't think that a migration is the right way to adjust seed data or production data.

I think we should update the seed sql in this PR instead and then for any instances just run some scripted curls to make these changes via the existing API instance (any script to do that will be simple and IMHO doesn't need to be included in the repository, but could be part of the PR description instead)

bickelj:
@slifty I wouldn't call it simple, but it seems to do the trick:

```bash
API_DOMAIN='api-test.philanthropydatacommons.org'; TOKEN='yada.yada.yada'; curl -4 https://${API_DOMAIN}/baseFields \
  | jq -c '.[] | select(.shortCode | startswith("organization_")) | select(.scope == "proposal") | {id: .id, label: .label, description: .description, scope: "organization", dataType: .dataType, shortCode: .shortCode}' \
  | tr -d '{' \
  | sed 's/"id":\([0-9]*\),\(.*\)/--next -X '\''PUT'\'' -H '\''content-type: application\/json'\'' -H '\''Authorization: Bearer '${TOKEN}''\'' -d '\''{\2'\'' https:\/\/'${API_DOMAIN}'\/\1 /g' \
  | xargs -0 | tr '\n' ' '
```

Except that a real token is too long: `xargs: argument line too long`. So take the output, 
then copy a chunk, prepend curl command, paste. Get new tokens as needed.

Then after pasting and running all that, run this to verify, this should come up empty because it filters on starting with "organization" and scope is "proposal" (all should be "organization" now):
```bash
API_DOMAIN='api-test.philanthropydatacommons.org'; curl -4 https://${API_DOMAIN}/baseFields \
  | jq -c '.[] | select(.shortCode | startswith("organization_")) | select(.scope == "proposal") | {id: .id, label: .label, description: .description, scope: "organization", dataType: .dataType, shortCode: .shortCode}' \
  | tr -d '{' \
```

So that was an interesting exercise but I think running a SQL command would have been the better way.

bickelj:
I just ran the proposed seed on a fresh database from main at 2af1e4b05425896e2b857cd652ab241576aeba69 and it succeeded.

slifty (approved):
w00t!

Also yes I stand corrected on the "simple" ;)

(this update to the seed looks good, and thanks for sharing the way to manually update on live data)

PR #1324, opened 2024-11-13
Add `userChangemakerPermission` endpoints

slifty:
This PR adds a set of endpoints for the creation and deletion of changemaker permissions for users.

Similar endpoints need to be added for the dataProvider and funder permissions, but I figure starting here to confirm the pattern makes sense.

This PR also fixes a bug related to the User type as it relates to permissions.

Related to Issue #150

### DIFF
--- a/src/__tests__/userChangemakerPermissions.int.test.ts
+++ b/src/__tests__/userChangemakerPermissions.int.test.ts
@@ -1,0 +1,156 @@
+import request from 'supertest';
+import { app } from '../app';
+import {
+	createChangemaker,
+	createOrUpdateUserChangemakerPermission,
+	loadSystemUser,
+} from '../database';
+import { expectTimestamp, loadTestUser } from '../test/utils';
+import {
+	mockJwt as authHeader,
+	mockJwtWithAdminRole as authHeaderWithAdminRole,
+} from '../test/mockJwt';
+import { keycloakUserIdToString, Permission } from '../types';
+
+describe('/users/changemakers/:changemakerId/permissions/:permission', () => {
+	describe('PUT /', () => {
+		it('returns 401 if the request lacks authentication', async () => {
+			const user = await loadTestUser();
+			const changemaker = await createChangemaker({
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+			});
+			await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/${changemaker.id}/permissions/${Permission.MANAGE}`,
+				)
+				.send({})
+				.expect(401);
+		});
+
+		it('returns 401 if the authenticated user lacks permission', async () => {
+			const user = await loadTestUser();
+			const changemaker = await createChangemaker({
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+			});
+			await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/${changemaker.id}/permissions/${Permission.MANAGE}`,
+				)
+				.set(authHeader)
+				.send({})
+				.expect(401);
+		});
+
+		it('returns 400 if the userId is not a valid keycloak user ID', async () => {
+			await request(app)
+				.put(`/users/notaguid/changemakers/1/permissions/${Permission.MANAGE}`)
+				.set(authHeaderWithAdminRole)
+				.send({})
+				.expect(400);
+		});
+
+		it('returns 400 if the changemaker ID is not a valid ID', async () => {
+			const user = await loadTestUser();
+			await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/notanId/permissions/${Permission.MANAGE}`,
+				)
+				.set(authHeaderWithAdminRole)
+				.send({})
+				.expect(400);
+		});
+
+		it('returns 400 if the permission is not a valid permission', async () => {
+			const user = await loadTestUser();
+			await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/1/permissions/notAPermission`,
+				)
+				.set(authHeaderWithAdminRole)
+				.send({})
+				.expect(400);
+		});
+
+		it('creates and returns the new user changemaker permission when user has administrative credentials', async () => {
+			const user = await loadTestUser();
+			const changemaker = await createChangemaker({
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+			});
+
+			const response = await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/${changemaker.id}/permissions/${Permission.EDIT}`,
+				)
+				.set(authHeaderWithAdminRole)
+				.send({})
+				.expect(201);
+			expect(response.body).toEqual({
+				changemakerId: changemaker.id,
+				createdAt: expectTimestamp,
+				createdBy: user.keycloakUserId,
+				permission: Permission.EDIT,
+				userKeycloakUserId: user.keycloakUserId,
+			});
+		});
+
+		it('creates and returns the new user changemaker permission when user has permission to manage the changemaker', async () => {
+			const user = await loadTestUser();
+			const changemaker = await createChangemaker({
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+			});
+			await createOrUpdateUserChangemakerPermission({
+				userKeycloakUserId: user.keycloakUserId,
+				changemakerId: changemaker.id,
+				permission: Permission.MANAGE,
+				createdBy: user.keycloakUserId,
+			});
+			const response = await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/${changemaker.id}/permissions/${Permission.EDIT}`,
+				)
+				.set(authHeader)
+				.send({})
+				.expect(201);
+			expect(response.body).toEqual({
+				changemakerId: changemaker.id,
+				createdAt: expectTimestamp,
+				createdBy: user.keycloakUserId,
+				permission: Permission.EDIT,
+				userKeycloakUserId: user.keycloakUserId,
+			});
+		});
+
+		it('does not update `createdBy`, but returns the user changemaker permission when user has permission to manage the changemaker', async () => {
+			const user = await loadTestUser();
+			const systemUser = await loadSystemUser();
+			const changemaker = await createChangemaker({
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+			});
+			await createOrUpdateUserChangemakerPermission({
+				userKeycloakUserId: user.keycloakUserId,
+				changemakerId: changemaker.id,
+				permission: Permission.MANAGE,
+				createdBy: systemUser.keycloakUserId,
+			});
+			const response = await request(app)
+				.put(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/${changemaker.id}/permissions/${Permission.MANAGE}`,
+				)
+				.set(authHeader)
+				.send({})
+				.expect(201);
+			expect(response.body).toEqual({
+				changemakerId: changemaker.id,
+				createdAt: expectTimestamp,
+				createdBy: systemUser.keycloakUserId,
+				permission: Permission.MANAGE,
+				userKeycloakUserId: user.keycloakUserId,
+			});
+		});
+	});
+});
--- a/src/handlers/userChangemakerPermissionsHandlers.ts
+++ b/src/handlers/userChangemakerPermissionsHandlers.ts
@@ -1,0 +1,92 @@
+import { createOrUpdateUserChangemakerPermission } from '../database';
+import {
+	isAuthContext,
+	isId,
+	isKeycloakUserId,
+	isPermission,
+	isTinyPgErrorWithQueryContext,
+	isWritableUserChangemakerPermission,
+} from '../types';
+import {
+	DatabaseError,
+	FailedMiddlewareError,
+	InputValidationError,
+} from '../errors';
+import type { Request, Response, NextFunction } from 'express';
+
+const putUserChangemakerPermission = (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+): void => {
+	if (!isAuthContext(req)) {
+		next(new FailedMiddlewareError('Unexpected lack of auth context.'));
+		return;
+	}
+
+	const { userKeycloakUserId, changemakerId, permission } = req.params;
+	const createdBy = req.user.keycloakUserId;
+
+	if (!isKeycloakUserId(userKeycloakUserId)) {
+		next(
+			new InputValidationError(
+				'Invalid userKeycloakUserId parameter.',
+				isKeycloakUserId.errors ?? [],
+			),
+		);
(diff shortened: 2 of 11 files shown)


slifty:
I see there are a lot of untested lines; I'll add tests for those now.

bickelj:
@slifty I tried this locally but could not seem to add a permission without having a permission (circular dependency). How do we bootstrap permissions? Should an admin be able to add a permission?

slifty:
@bickelj an admin is able to add permissions.

src/openapi.json
```diff
@@ -2621,6 +2659,107 @@
...
+						"description": "The id of a changemaker.",
+						"in": "path",
+						"required": true,
+						"schema": {
+							"type": "integer"
+						}
+					},
+					{
+						"name": "permission",
+						"description": "The permission to be granted.",
```
bickelj, line 2748: To be deleted, here.

bickelj (approved):
I skimmed the code and didn't see anything objectionable. I was able to build, lint, and run this successfully locally. I verified that I could not add a permission for a non-existent keycloak ID or a non-existent changemaker ID. And that I could not add a permission as a non-admin. Once I became an admin I could add a permission and delete a permission.

So we probably will want to do some more deep testing and review once permissions are enabled, or perhaps I should review this in more depth another time, but I think this is good enough to merge at the moment. @slifty If you'd like a more thorough review I can probably look at this tomorrow.

src/middleware/requireChangemakerPermission.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { Request } from 'express';
+
+const requireChangemakerPermission =
+	(permission: Permission) =>
+	(req: Request, res: Response, next: NextFunction) => {
+		if (!isAuthContext(req)) {
+			next(new UnauthorizedError('The request lacks an AuthContext.'));
+			return;
+		}
+		if (req.role?.isAdministrator === true) {
```
bickelj, line 14: I'm not sure if this should precede or follow the validation of changemaker ID. It seems right now that check is skipped if you're an admin. However if we check it before checking for admin, we'd be revealing valid and invalid changemaker IDs. Then again, we'd only be revealing those to already authenticated users who would presumably have access to those IDs anyway. Hmm.

src/handlers/userChangemakerPermissionsHandlers.ts
```diff
@@ -0,0 +1,151 @@
...
+		next(error);
+	});
+};
+
+const putUserChangemakerPermission = (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+): void => {
+	if (!isAuthContext(req)) {
```
bickelj, line 80: Is this a repeat of what can be done using `requireAuthentication` in the routes declared below in `usersRouter.ts`? Why not use `requireAuthentication` there?

slifty: This is a necessary type guard that allows us to treat `req` as an AuthContext

src/handlers/userChangemakerPermissionsHandlers.ts
```diff
@@ -0,0 +1,151 @@
...
+	InputValidationError,
+} from '../errors';
+import type { Request, Response, NextFunction } from 'express';
+
+const deleteUserChangemakerPermission = (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+): void => {
+	if (!isAuthContext(req)) {
```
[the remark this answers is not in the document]
slifty: In this case you are right: we do NOT need this type guard because `req` is not accessed as an AuthContext in the handler.

src/openapi.json
```diff
@@ -2621,6 +2659,61 @@
 					}
 				}
 			}
+		},
+		"/users/{userKeycloakUserId}/changemakers/{changemakerId}/permissions/{permission}": {
```
jasonaowen, line 2663: This is currently added in the first commit (correct user type); does it belong in the second (add put endpoint)?

src/middleware/requireChangemakerPermission.ts
slifty: There is no secret being validated here -- just "is this a GUID" which is part of the API spec.

jasonaowen: The admin check should come first, for exactly the reason you said! What is the advantage to checking the changemaker ID first?

src/openapi.json
slifty: Good catch thanks

src/database/queries/userChangemakerPermissions/deleteOne.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM user_changemaker_permissions
+  WHERE user_keycloak_user_id = :userKeycloakUserId
```
jasonaowen, line 2: This line should not be indented:

```suggestion
WHERE user_keycloak_user_id = :userKeycloakUserId
```

I know I need to dust off the SQL linter PR! :sweat_smile:

src/database/queries/userChangemakerPermissions/deleteOne.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM user_changemaker_permissions
```
jasonaowen, line 1: I believe this is the first time we're deleting anything from the database! Which brings to mind the question of audit logs. I can talk myself into thinking that's a separate, deferrable concern, but I want that to be a conscious decision.

The alternative, I think, would be for the database to have `not_before` and `not_after` columns, and "deleting" a permission would update the `not_after` from `NULL` to `now()`. That would give us history which we could surface someday, without building a whole separate audit log layer.

What do you think?

src/__tests__/userChangemakerPermissions.int.test.ts
```diff
@@ -0,0 +1,292 @@
...
+import { expectTimestamp, loadTestUser } from '../test/utils';
+import {
+	mockJwt as authHeader,
+	mockJwtWithAdminRole as authHeaderWithAdminRole,
+} from '../test/mockJwt';
+import { keycloakUserIdToString, Permission } from '../types';
+import { NotFoundError } from '../errors';
+
+describe('/users/changemakers/:changemakerId/permissions/:permission', () => {
+	describe('PUT /', () => {
```
jasonaowen, line 17: Can we have a test for adding an already existing / duplicate permission?

src/database/queries/userChangemakerPermissions/deleteOne.sql
slifty: Would it be enough to just have `not_after` at this point?  I'm not sure we have a use case for `not_before` yet.

I'm open to it, though obviously it results in more complex queries / the possibility that we'll fail to add an appropriate WHERE clause some day!

jasonaowen: Sure, until we have that use case it'd just be identical to `created_at` anyway, and when that time comes the migration could copy `created_at` into the new column.

bickelj: This sounds good to me. To be clear, the suggestion is to keep the HTTP delete endpoint as an HTTP delete, however, avoid scrubbing the database and use the above technique?

src/middleware/requireChangemakerPermission.ts
bickelj: Right. My thought was that the check on changemaker ID here is skipped. But that is OK because it appears to be re-checked later in the handler ~L85 of `putUserChangemakerPermission`. This seems appropriate because the check at L19 below is exclusively for the sake of looking up permissions ~L26.

And as for the secret, right too, it's a format/type check, not checking whether the value exists. Moving along...

src/handlers/userChangemakerPermissionsHandlers.ts
bickelj: Almost all of our routes have a `requireAuthentication` middleware but this one does not. I think that's what's giving me pause.

I take "auth" in your comment to mean both authentication and authorization. We already have a middleware for requiring authentication, but we're not using it, perhaps because we are already required to use it here in the authorization middleware? And that perhaps due to TypeScript not knowing the details of Express middleware such that we can't declare this middleware as having Request be of type Request-with-AuthContext even when we already went through a `requireAuthentication` middleware?

Regardless, I think the type check is OK here for the reason you stated (needing to access an actually present `user` object on `req`), but maybe we should re-use `requireAuthentication` on the route because it's weird to not have our ubiquitous authentication middleware not checking for authentication first. In general, I think validating things exactly once is best, but when it comes to authentication and authorization, more checks might be OK, especially given the limitations of the typechecker in an express middleware context. If we could have the type checker recognize that this code can only be called from middleware that already checked this type that might be cool but it's clear enough and low cost enough to check before use here.

Aside: I realize you are calling this a non-business-logic type check, but type checks serve the purpose of serving the business too, if only indirectly, and all our coding is intended to serve the business. So I have a quibble with the phrase "business logic" in general. This is an aside because it doesn't change your point or mine either way.

I might could be convinced that the two-in-one authorization-middleware-implicitly-checks-authentication is also OK, I'm still noodling over it.

src/middleware/requireChangemakerPermission.ts
```diff
@@ -0,0 +1,38 @@
+import { Response, NextFunction } from 'express';
+import { InputValidationError, UnauthorizedError } from '../errors';
+import { isAuthContext, isId } from '../types';
+import type { Permission } from '../types';
+import type { Request } from 'express';
+
+const requireChangemakerPermission =
```
bickelj, line 7: This is an authorization middleware that checks whether the currently authenticated user has the given permission on the changemaker in the URL parameters. It is distinct from the handler that creates or deletes such permissions that this authorization handler checks, but happens to be used by the route leading to those handlers. It is also distinct from our authentication-checking handler but authentication is also checked here.

I'm trying to think if this could use a better name but perhaps not. It is in the `middleware` directory and not the `handlers` directory. I think it would be worth adding a comment on the function, though, basically my first paragraph, to remind our future selves.

slifty: I'm not sure I understand, but to clarify the intention: this is express middleware (both in form and function) so `middleware` is the appropriate location.

The handler requires that the calling path have a `changemakerId` parameter, as well as an auth setting.  This is self documented IMHO because the function is simple and those checks are made.  It can be used on any path that has a `changemakerId` URL parameter.

src/database/queries/userChangemakerPermissions/deleteOne.sql
slifty: @bickelj yes that's correct

src/handlers/userChangemakerPermissionsHandlers.ts
slifty: > perhaps because we are already required to use it in the authorization middleware requireChangemakerPermission

You may have missed the saga, but yes, and we had to refactor for this reason: https://github.com/PhilanthropyDataCommons/service/issues/1210

> In general, I think validating things exactly once is best

This isn't there for validation, it is a type guard / it is for type safety.  There is no other way to achieve type safety (we could cast it, blindly, but that would not be type safe which is why we took on #1210)

> if we could have the type checker recognize that this code can only be called from middleware that already checked this type that might be cool

We cannot :(

> it's weird to not have our ubiquitous authentication middleware checking for authentication first

`requireAuthentication` is not ubiquitous: it is currently only used for endpoints that ONLY require authentication but no additional authorization -- that's the purpose of that middleware, to simply require authentication.  We also have `requireAdministratorRole` which is used for endpoints that require that additional permission, and now this is a third rule set.  Based on the names, would you not expect those pieces of middleware to imply a requirement of authorization?  If not, then I could see it as confusing.  If so, then I think it would be redundant.

Side note: Even if we do include that extra auth middleware it would not change the definition of `requireChangemakerPermission` because that will still need the type guards.

The intention of these three distinct tiers of auth requirements is to define what the rules for accessing the route is going to be.  Alternatively, we could move to a single auth middleware that has some kind of settings parameter to specify expectations, but I think at this point it is simpler to just have distinct middleware until we have a wider range of auth rules.

bickelj: > > perhaps because we are already required to use it in the authorization middleware requireChangemakerPermission
> 
> You may have missed the saga, but yes, and we had to refactor for this reason: #1210

Right, right. That makes sense.

> > In general, I think validating things exactly once is best
> 
> This isn't there for validation, it is a type guard / it is for type safety. There is no other way to achieve type safety (we could cast it, blindly, but that would not be type safe which is why we took on #1210)

Type guards, safety, and checking are a kind of validation. We don't guard types just because the compiler tells us to do so but because we want some assurance that a thing is the expected thing. I recognize that type guards/checking/etc are a limited form of validation and we want to do further validation on the contents but I don't think there's a hard distinction here. I also understand that there is a limitation within the context of typescript and express middleware with regard to types. It's OK with me to re-check types several times given the limitations.

> > if we could have the type checker recognize that this code can only be called from middleware that already checked this type that might be cool
> 
> We cannot :(

Gotcha. It is what it is.

> > it's weird to not have our ubiquitous authentication middleware checking for authentication first
> 
> `requireAuthentication` is not ubiquitous: it is currently only used for endpoints that ONLY require authentication but no additional authorization -- that's the purpose of that middleware, to simply require authentication for routes that need nothing more than authentication. We also have `requireAdministratorRole` which is used for endpoints that require that additional permission, and now this is a third rule set. Based on the names, would you not expect those pieces of middleware to imply a requirement of authorization? If not, then I could see it as confusing. If so, then I think it would be redundant.
> 
> This isn't directly relevant, but I want to note that even if we do include that extra auth middleware it would not change the definition of `requireChangemakerPermission` because that will still need the type guards.
> 
> The intention of these three distinct tiers of auth requirements is to define what the rules for accessing the route is going to be. Alternatively, we could move to a single auth middleware that has some kind of settings parameter to specify expectations, but I think at this point it is simpler to just have distinct middleware until we have a wider range of auth rules.

I'm OK with it, on reflection. My first review was pretty rushed. During my second review, I think it crystalized.

I think it's overstating it to say the handlers and middleware here are self-documenting. I think it becomes clearer after a long hard look, but I also think given the circular reference it may be worth a comment.

src/middleware/requireChangemakerPermission.ts
bickelj: I meant to leave this comment here instead of in the other thread:

I think it's overstating it to say the handlers and middleware here are self-documenting. I think it becomes clearer after a long hard look, but I also think given the circular reference it may be worth a comment.

bickelj: But technically, yeah, you're right, it's a simple function, just read it carefully three or four times and you're good to go : )

bickelj:
FWIW my approval still stands after all that haggling.

src/middleware/requireChangemakerPermission.ts
slifty: @bickelj I hear you -- given that point I think that it would be appropriate to add a docblock to all three "auth" middlewares to make it explicit what their intention is.

src/database/queries/userChangemakerPermissions/deleteOne.sql
slifty: I'm going to move delete into a separate PR, so marking this as resolved for now.

src/__tests__/userChangemakerPermissions.int.test.ts
slifty: Great catch, and for the record adding this test exposed a bug 🎉

PR #1333, opened 2024-11-23
Split up lint jobs

jasonaowen:
Run each linter in a separate GitHub Actions job, so that what exactly is wrong is more obvious at a glance, and to more easily add additional linters (like #321).

We will need to reconfigure our merge requirements, which are currently expecting the single job `lint`.

PR #1342, opened 2024-11-26
Split test action into two jobs

jasonaowen:
The GitHub Action for running our tests was doing two things: checking that the tests passed, and then generating and uploading a test coverage report.

Unfortunately, our tests are somewhat slow, so running them twice in the same job takes a few minutes. Split them up so they can be run in parallel.

See also https://github.com/PhilanthropyDataCommons/service/pull/1333#issuecomment-2501752011 .

As with that PR, we'll need to update the branch protection rules.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 jobs:
-  target_node:
+  tests:
     runs-on: ubuntu-latest
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


jasonaowen:
The split test checks took 1m 49s for the tests, and 2m 1s for the coverage, down from 3m 0s for the combined check on #3591.

slifty (approved):
Thanks for this -- I wonder if the upstream bug that forced us to run twice ever got fixed...

jasonaowen:
> I wonder if the upstream bug that forced us to run twice ever got fixed...

I don't think I remember that; tell me more?

PR #1349, opened 2024-12-02
Add DELETE endpoint for UserChangemakerPermission

slifty:
This PR adds the ability to `DELETE` various permissions.

Related to #1351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Manage user changemaker permissions using `GET`, `PUT`, and `DELETE` on `/user/{keycloakUserId}/changemakers/{changemakerId}/permissions/{permission}`.
+
 ### Changed
 
 - Upgraded to use OpenAPI Specification 3.1.
--- a/src/__tests__/userChangemakerPermissions.int.test.ts
+++ b/src/__tests__/userChangemakerPermissions.int.test.ts
@@ -4,6 +4,8 @@ import {
 	createChangemaker,
 	createOrUpdateUserChangemakerPermission,
 	loadSystemUser,
+	loadUserChangemakerPermission,
+	removeUserChangemakerPermission,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
 import {
@@ -11,6 +13,7 @@ import {
 	mockJwtWithAdminRole as authHeaderWithAdminRole,
 } from '../test/mockJwt';
 import { keycloakUserIdToString, Permission } from '../types';
+import { NotFoundError } from '../errors';
 
 describe('/users/changemakers/:changemakerId/permissions/:permission', () => {
 	describe('PUT /', () => {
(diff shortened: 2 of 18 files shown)


src/__tests__/userChangemakerPermissions.int.test.ts
```diff
@@ -153,4 +155,206 @@ describe('/users/changemakers/:changemakerId/permissions/:permission', () => {
...
+			await request(app)
+				.delete(
+					`/users/${keycloakUserIdToString(user.keycloakUserId)}/changemakers/${changemaker.id}/permissions/${Permission.MANAGE}`,
+				)
+				.set(authHeaderWithAdminRole)
+				.send()
+				.expect(404);
+		});
+
+		it('returns 404 if the permission had previously been deleted', async () => {
```
slifty, line 236: I wanted to highlight this for consideration.  The internet disagrees on this -- should it be 404 the second time, or 204.  What does idempotent mean in this context?

It's 404 if it never existed of course.

bickelj (changes requested):
I was able to `lint`, `build`, `test:ci`, `migrate`, and `start` this locally. I got the expected responses when attempting to create permissions for non-existent entities. I was able to add a permission as expected and then delete it as expected. For example, an added and then deleted permission:

```
pdc=> select * from user_changemaker_permissions ;
        user_keycloak_user_id         | permission | changemaker_id |              created_by              |          created_at           |           not_after           
--------------------------------------+------------+----------------+--------------------------------------+-------------------------------+-------------------------------
 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | edit       |              5 | 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | 2024-12-03 11:20:21.078377-06 | 2024-12-03 11:21:15.407213-06
(1 row)
```

However, after re-deleting that same permission, the `not_after` was updated in the same row, like this:

```
pdc=> select * from user_changemaker_permissions ;
        user_keycloak_user_id         | permission | changemaker_id |              created_by              |          created_at           |           not_after           
--------------------------------------+------------+----------------+--------------------------------------+-------------------------------+-------------------------------
 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | edit       |              5 | 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | 2024-12-03 11:20:21.078377-06 | 2024-12-03 11:27:24.499753-06
(1 row)
```

And then after re-adding the same permission, the `not_after` was cleared:

```
pdc=> select * from user_changemaker_permissions ;
        user_keycloak_user_id         | permission | changemaker_id |              created_by              |          created_at           | not_after 
--------------------------------------+------------+----------------+--------------------------------------+-------------------------------+-----------
 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | edit       |              5 | 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | 2024-12-03 11:20:21.078377-06 | 
(1 row)
```

I think the case of a repeated `delete` should compare `now()` to the existing `not_after` and if `now()` is later, do not update the `not_after` because that would be redundant and erase the actual expiration of the permission.

I'm less sure of what I expect for the last case (adding a previously deleted permission) but in the spirit of retaining what actually happened, as complex as it is, I think I want new non-overlapping-in-time rows.

bickelj:
Perhaps it is too much to try to track the entire permissions history directly in the permissions table. Instead, we can allow simple `insert` and `delete` and [enable postgres `mod` logs](https://github.com/PhilanthropyDataCommons/deploy/issues/119). We can look at the database logs if a question arises as to the history of permissions changes. Or if we wish to keep a history in-band in the database itself, we could use a history table off to the side: an audit log table.

src/__tests__/userChangemakerPermissions.int.test.ts
slifty: @bickelj I wanted to highlight the above comment, since you started to explore some of the questions in it below.  Specifically this question of: `If 204, should it update the not_after to the most recent call, or should not_after not be touched if the permission had already expired?`

slifty:
> Perhaps it is too much to try to track the entire permissions history directly in the permissions table.

I agree that if the goal is simply supporting an audit that the above solution would make a simple design much more complex and prone to having errors in future queries.

I think it makes more sense to have an audit log system (which we had an issue fore: #135) than to turn the permissions tables into audit logs themselves.  Whether / when an audit log system becomes an engineering priority I think should be driven by use case + client + @kfogel, if that makes sense.

bickelj:
Given what was discussed in the meeting, I think keep the `not_after` column until we have a better substitute because it's something rather than nothing.

slifty:
@bickelj I just created #1355 to capture the audit log discussion.

src/__tests__/userChangemakerPermissions.int.test.ts
slifty: @jasonaowen @bickelj I do think the above question *is* something we should answer, even without an audit log.

If someone deletes a permission that has already been deleted we should essentially ignore the second request, correct?  Should it return 404, or 204?

bickelj: @slifty I think it should not change `not_after` if it already expired and should return 204 or 404. I'm not certain about the 204 or 404 question, tbh.

src/database/queries/userDataProviderPermissions/insertOrUpdateOne.sql
```diff
@@ -2,13 +2,15 @@ INSERT INTO user_data_provider_permissions (
...
   :userKeycloakUserId,
   :permission::permission_t,
   :dataProviderShortCode,
-  :createdBy
+  :createdBy,
+  null
 )
-ON CONFLICT (user_keycloak_user_id, permission, data_provider_short_code)
-DO NOTHING
+ON CONFLICT (user_keycloak_user_id, permission, data_provider_short_code) DO UPDATE
```
bickelj, line 14: It would seem that this reduces the chance of interleaving queries on this row, but I wasn't confident that it guaranteed no interleaving (e.g. an insert followed by update followed by the insert's do update on conflict), though this part of the postgres documentation suggests exactly one insert or one update will happen:
> ON CONFLICT DO UPDATE guarantees an atomic INSERT or UPDATE outcome; provided there is no independent error, one of those two outcomes is guaranteed, even under high concurrency. This is also known as UPSERT — “UPDATE or INSERT”.

I suppose if there is a concurrent insert/update and delete, it's anybody's guess as to who should win the race anyway.

src/database/queries/userChangemakerPermissions/deleteOne.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE user_changemaker_permissions
+  SET not_after = now()
+  WHERE user_keycloak_user_id = :userKeycloakUserId
+  AND permission = :permission::permission_t
+  AND changemaker_id = :changemakerId;
```
bickelj, line 5: Something like `AND ( not_after >= now() OR not_after IS NULL )`

src/database/initialization/user_to_json.sql
```diff
@@ -16,6 +16,9 @@ BEGIN
       SELECT user_changemaker_permissions.changemaker_id, jsonb_agg(user_changemaker_permissions.permission) AS permissions
       FROM user_changemaker_permissions
       WHERE user_changemaker_permissions.user_keycloak_user_id = "user".keycloak_user_id
+        AND (user_changemaker_permissions.not_after IS NULL
+          OR user_changemaker_permissions.not_after > now()
+        )
```
bickelj, line 21: Slight style inconsistency.

bickelj (approved):
After our discussion today and reviewing in more detail, this looks good to me. I think 404 is an appropriate response for `delete` when a resource has already been "deleted." A future audit log will remove the need for the `not_after` column.

src/__tests__/userChangemakerPermissions.int.test.ts
jasonaowen: Yeah, I agree that it should return 404. My reading of this is that "idempotent" means that repeating the same request should not cause issues, and if the server returns 404 on the second request, the state is the same after one or two `DELETE`s: the permission is gone.

src/types/UserChangemakerPermission.ts
```diff
@@ -10,6 +10,7 @@ interface UserChangemakerPermission {
 	readonly changemakerId: number;
 	readonly createdBy: KeycloakUserId;
 	readonly createdAt: string;
+	readonly notAfter?: string;
```
jasonaowen, line 13: We want `notAfter` to always be present, don't we? Just sometimes unset? I know you went to a lot of trouble wrangling types for `Writable`!

Or, maybe we don't need to read `notAfter` outside of the database at all yet, and we can just omit it from the type definition and the SQL query.

src/database/queries/userChangemakerPermissions/selectByPrimaryKey.sql
```diff
@@ -0,0 +1,5 @@
+SELECT user_changemaker_permission_to_json(user_changemaker_permissions.*) AS "object"
+FROM user_changemaker_permissions
+WHERE user_keycloak_user_id = :userKeycloakUserId
+  AND changemaker_id = :changemakerId
+  AND permission = :permission
```
jasonaowen, line 5: Do we want a `AND (not_after > now() OR not_after IS NULL`? Or do we also want to expose "deleted" rows? Are we using this outside of tests yet?

src/openapi.json
```diff
@@ -1163,6 +1163,12 @@
 						"type": "string",
 						"format": "date-time",
 						"readOnly": true
+					},
+					"notAfter": {
+						"type": "string",
+						"format": "date-time",
+						"readOnly": true,
+						"nullable": true
```
jasonaowen, line 1171: Do we want to expose this in the API? I feel like there's an argument to be made that this is an implementation detail, and it's okay to not return "deleted" permissions.

jasonaowen (changes requested):
It's not clear to me that either application code or the API needs to be aware of how we "delete" permissions. It could be handled entirely within the database (via queries with appropriate `WHERE` clauses).

On the other hand, if we do want to expose the idea of expired permissions, let's do so consistently: the `notAfter` property should always be present in returned JSON, just set to `null`.

(I don't feel very strongly about these, so if you passionately disagree, feel free to dismiss this review!)

src/__tests__/userChangemakerPermissions.int.test.ts
slifty: Thanks y'all!  I'll update to that approach.

src/types/UserChangemakerPermission.ts
slifty: Side question: TS guidelines say "use undefined, not null" -- https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

What do you think about all that?  This wouldn't change your request of having it be required (just possibly undefined) e.g. `notAfter: string | undefined`

as for the other points I'll get back to ya, I did circle around some of those and all three were implemented at one point but that doesn't mean I disagree with them.

src/openapi.json
slifty: It seemed to me that `notAfter` could also eventually allow for people to set expiring permissions (though we are not opting to support it) and in that lens it could be included as... foreshadowing!

I'm fine not exposing it.  I'm not sure how we would test it without writing a direct SQL query if we aren't exposing it, though maybe that's a-ok to not test (and instead just verify that no value gets returned when you try to select it)!

Leaving it as is is basically partially implementing expiring permission support; making it more obscure is also fine and means we don't have a partial implementation of something; the code wasn't that hard to write and can be easily re-written later.

bickelj: I lean toward leaving it out: simpler, committing to less, yagni, etc.

src/types/UserChangemakerPermission.ts
jasonaowen: I mean, the top of that page is pretty explicit about its target audience! :stuck_out_tongue: 

I think the semantics of `null` more closely align with what we're intending to convey, and is analogous to the SQL `NULL` that's backing it: it's not that we have this variable that hasn't been initialized yet; we know the value of `notAfter`, and it is explicitly missing.

slifty: Oh right, so it is!  I have to re-remember that every year or two

src/database/initialization/user_to_json.sql
slifty: Fixed this by adding an `is_expired()` function!

src/database/queries/userChangemakerPermissions/selectByPrimaryKey.sql
slifty: Per the other conversation this is now updated so `not_after` is purely an internal concept to the DB; it's not exposed to node at all and any expired record is treated as if it doesn't exist for all queries.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Manage user changemaker permissions using `GET`, `PUT`, and `DELETE` on `/user/{keycloakUserId}/changemakers/{changemakerId}/permissions/{permission}`.
```
slifty, line 12: This includes other verbs because the changelog was not properly updated when those were added.

PR #1359, opened 2024-12-04
Move SQLFluff `requirements.txt` to subdirectory

### LINKED ISSUE #1356: Deployment failure 2024-12-03
This issue is to investigate a notification about a DO deployment failure last night, December 3.

jasonaowen:
Having `requirements.txt` in the root of the project causes DigitalOcean App Platform to detect the app as needing the Python buildpack. Adding that buildpack seems to throw away the Node output(?), and so the migrate step fails to run:

> Error: Cannot find module '/workspace/dist/scripts/migrate.js'

Even if that did not cause a build failure, we do not want this dev/CI tool to be present in the deployed image.

Move the `requirements.txt` file into a new subdirectory, and configure dependabot to look at that new subdirectory weekly.

Resolves #1356 Deployment failure 2024-12-03


jasonaowen:
I was able to deploy this on api-test.pdco by editing the app spec to point to this branch rather than `main`, and it successfully deployed: https://cloud.digitalocean.com/apps/21057031-b8f6-4acd-a82a-912e2607e7e7/deployments/b463cb15-1ade-4b94-a48d-2dd525cfe8e3

I'll switch it back now so that it'll work when this gets merged.

.github/workflows/lint.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           python-version: '3.13'
       - name: Install SQLFluff
-        run: 'pip install -r requirements.txt'
+        run: 'pip install -r sqlfluff/requirements.txt'
```
slifty, line 35: My initial thought was "this directory name seems too narrow" but until we have another python requirement, this is truly the only thing it exists for and we can always change it another day.

slifty (approved):

PR #1236, opened 2024-10-03
Deploy on DigitalOcean App Platform

jasonaowen:
> [DigitalOcean App Platform](https://docs.digitalocean.com/products/app-platform/) is a Platform-as-a-Service (PaaS) offering that allows developers to publish code directly to DigitalOcean servers without worrying about the underlying infrastructure.

The majority of the work in migrating is in configuring the application on the DigitalOcean side. To do so, some improved documentation and logging is helpful, as is capturing the configuration into an [app template](https://docs.digitalocean.com/products/app-platform/how-to/add-deploy-do-button/).

Issue https://github.com/PhilanthropyDataCommons/deploy/issues/118 Migrate to DigitalOcean Apps

PR #1364, opened 2024-12-05
Update to Node 22

### LINKED ISSUE #1363: Upgrade to Node 22
Our node version policy says we will use the active version of node.  That version is 22 as of October!

slifty:
This PR bumps everything Node-Version-Related to version 22.

Resolves #1363


.github/workflows/lint.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.17.0
+          node-version: 22
```
slifty, line 15: Since we've been talking about pinned version granularity: even if we do decide to keep the pin for the test CI, I don't think there is much value in pinning a patch version for the lint CI

.github/workflows/test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v4
         with:
-          node-version: 20.17.0
+          node-version: 22
```
bickelj, line 30: Whether we know the production version in advance or the test version in advance, we should deploy what we test and test what we deploy.

Right now, production shows this:
```
$ node --version
v20.15.1
```

We pretty much don't know what version will come with the DO buildpack until we do a test deployment. If we accept this spirit of embracing uncertainty around which version we're running, then yeah, there is no point in trying to make it match, and 22 should be declared here. It is still too loosey-goosey for my taste.

bickelj (approved):
I don't know what version of node we'll end up on or if this deployment will fail when we merge this PR. The worst case is a rollback to a previously working version will happen. So I guess it's worth a shot, though, just to see what happens.

.github/workflows/lint.yml
jasonaowen: I agree! In which case, can we point this at the `.node-version` file?

```suggestion
          node-version-file: .node-version
```

jasonaowen (approved):
This settles the discussion in #1360 nicely - thanks, @slifty!

I agree that pinning an exact version of Node.js makes less sense with DigitalOcean Apps, and from what I've seen, using just the major version is the common practice.

jasonaowen:
Oh, it looks like we need to update the version section of the README - and, indeed, I should have done so back in #1236. Would you like to fix that in this PR, @slifty? Or should I follow up with another?

.github/workflows/test.yml
slifty: @bickelj appreciated all around, and I want to make sure you know I do see the value in what you're saying; I think it's worth leaning into that for now given the overall tradeoffs (and, as you say, the DO side of things too).

slifty:
I'll fix the readme now and also on the office hours @jasonaowen pointed out that the docker version should be updated too!  I'll add that in and then merge.

.github/workflows/lint.yml
bickelj: While I disagree with using the DO buildpack and prefer to know what we're deploying in advance of deploying it so that we can test it, given the DO buildpack situation and the preferences of others I think referencing the .node-version file is nice.

slifty:
@bickelj understood and agreed -- I'm going to open up an issue so we can explore that more robustly, separate from the version upgrade.

PR #1368, opened 2024-12-05
Add GET proposalVersionById route

### LINKED ISSUE #1366: Add GET ProposalVersion endpoint
We should add a handler to the `proposalVersionsHandlers.ts` endpoint that can retrieve a proposalVersion by ID. This is needed for https://github.com/PhilanthropyDataCommons/front-end/issues/915, but we agree we should have one for general use.

hminsky2002:
Closes #1366 

I'll note I did not update the changelog because I wasn't sure if this justified a change (as it wouldn't break anything), but can gladly update if requested!

### DIFF
--- a/src/__tests__/proposalVersions.int.test.ts
+++ b/src/__tests__/proposalVersions.int.test.ts
@@ -6,6 +6,7 @@ import {
 	createBaseField,
 	createOpportunity,
 	createProposal,
+	createProposalVersion,
 	loadSystemSource,
 	loadTableMetrics,
 } from '../database';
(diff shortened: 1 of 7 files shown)


src/__tests__/proposalVersions.int.test.ts
```diff
@@ -34,6 +35,67 @@ const createTestBaseFields = async () => {
...
+			const testUser = await loadTestUser();
+			await createProposal({
+				externalId: 'proposal-1',
+				opportunityId: 1,
+				createdBy: testUser.keycloakUserId,
+			});
+			await createApplicationForm({
+				opportunityId: 1,
+			});
+			const newProposalVersion = await createProposalVersion({
```
bickelj, line 55: While this follows some of the existing conventions, and therefore I don't think my objection is reason to hold up this PR, I prefer to get the database-generated IDs explicitly rather than making the assumption that ID 1 was inserted. Something for future PRs, perhaps.

src/openapi.json
```diff
@@ -1823,6 +1823,51 @@
 				}
 			}
 		},
+		"/proposalVersions/{proposalVersionId}": {
+			"get": {
+				"operationId": "getProposalVersionById",
+				"summary": "Gets a specific proposalVersion.",
```
bickelj, line 1829: Here I would say "proposal version." instead of "proposalVersion."

src/openapi.json
```diff
@@ -1823,6 +1823,51 @@
...
+				"tags": ["Proposals"],
+				"security": [
+					{
+						"auth": []
+					}
+				],
+				"parameters": [
+					{
+						"name": "proposalVersionId",
+						"description": "The PDC-generated ID of a proposalVersion.",
```
bickelj, line 1839: Here I would say "proposal version." instead of "proposalVersion."

bickelj (approved):
I was able to `build`, `run`, and try this code locally. It seemed to work fine. I needed a fresh authentication token to access the endpoint and when I asked for a proposal version by ID I got it back. It looked like the right stuff that I got back. Nothing in the code appears objectionable, at least nothing new objectionable.

My vote is to rebase, push, and merge when the checks pass.

src/__tests__/proposalVersions.int.test.ts
slifty: The same can be said for the other items -- it's all being loaded right above so there's no cost to just `const  opportunity = createOpportunity...` and then accessing `opportunity.id` in the appropriate places.

Also the prefix `new` here isn't contextually relevant so I'd just call it `proposalVersion` or `testProposalVersion`.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -212,6 +214,30 @@ const postProposalVersion = (
...
 };
 
+const getProposalVersion = (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+): void => {
+	const { proposalVersionId } = req.params;
+	if (!isId(proposalVersionId)) {
+		next(new InputValidationError('Invalid request body.', isId.errors ?? []));
```
slifty, line 224: This is a copy pasted error (as in, the error exists in another endpoint so I totally feel for you) -- proposalVersionId is a query parameter, not part of the request body, so this error should be updated to say `invalid query parameter`

slifty (approved):
Small text change request but looks good!

I am approving since I don't need to re-review; I do agree with @bickelj's comments too.

src/handlers/proposalVersionsHandlers.ts
bickelj: Nice catch!

PR #1373, opened 2024-12-10
Add Keycloak organization UUID to data providers, funders, and changemakers

bickelj:
Given that an organization might exist in Keycloak prior to the introduction of the data provider row here or vice versa, we intend to support setting the Keycloak organization ID optionally. This change allows the software to defer linking a PDC Keycloak organization to a PDC service data provider until both exist.


bickelj:
@slifty Obviously I need tests, but wanted to get your opinion on this sooner than later.

src/types/DataProvider.ts
```diff
@@ -1,15 +1,20 @@
...
 
 interface DataProvider {
 	readonly shortCode: ShortCode;
 	name: string;
+	keycloakOrganizationId?: KeycloakId | null;
 	readonly createdAt: string;
 }
 
-type WritableDataProvider = Writable<DataProvider>;
+// Using `Writable` here instead of explicit `Pick` messes up the optional field.
```
slifty, line 13: I wrote a patch for this in another PR, just removed it and made it its own thing so we can merge it and benefit from it in this case / not depend on the other place. https://github.com/PhilanthropyDataCommons/service/pull/1378

Let's merge that and then you won't need to remove use of `Writable` here.

slifty (commented):
Full review incoming (I have a question that will be covered in like 2m of voice chat if we can swing it, otherwise I'll just ask it here and we can type it all out!)

bickelj:
Before I forget, the justification for storing the UUID at all (rather than doing runtime linking based on names) is that organization A person B may wish to grant permission to organization C, therefore the organization C data is not present in person B's JWT. We could have a design where the PDC service queries the PDC Keycloak instance directly for lists of organizations, however, we'd prefer to keep it simple where the JWT has what we need, and allow grants of permission only to actually-used/actually-logged-in users and organizations. It is also convenient to see "organization C has Keycloak UUID [yada] while organization E is still NULL" (meaning no one from organization E has used PDC yet. The assumption I had was that person D (part of organization C) might log in at some point without any action on permissions at all, but at that moment the user's JWT can be used to link up organization C to a UUID. Later, when person B wants to grant permissions to organization C, now the linkage is available, and the organization C's UUID can be set in the permissions table.

src/database/queries/dataProviders/insertOrUpdateOne.sql
jasonaowen, line 1: This change appears to have been accidentally included in the first commit, the rename commit, rather than with the migration that adds the new column. Would you mind moving it to the second commit, please?

src/types/DataProvider.ts
```diff
@@ -1,15 +1,20 @@
 import { ajv } from '../ajv';
+import { KeycloakId, keycloakIdSchema } from './KeycloakUserId';
 import type { JSONSchemaType } from 'ajv';
-import type { Writable } from './Writable';
 import type { ShortCode } from './ShortCode';
 
 interface DataProvider {
 	readonly shortCode: ShortCode;
 	name: string;
+	keycloakOrganizationId?: KeycloakId | null;
```
jasonaowen, line 9: As discussed [elsewhere](https://github.com/PhilanthropyDataCommons/service/pull/1378#discussion_r1881251198), why make this key optional? It definitely should be nullable! But when should it be missing?

src/handlers/dataProvidersHandlers.ts
```diff
@@ -83,6 +84,20 @@ const putDataProvider = (
 		);
 		return;
 	}
+	const { keycloakOrganizationId } = req.params;
```
jasonaowen, line 87: Should this be added to the OpenAPI spec?

src/handlers/dataProvidersHandlers.ts
```diff
@@ -83,6 +84,20 @@ const putDataProvider = (
...
+		keycloakOrganizationId !== null &&
+		!isKeycloakId(keycloakOrganizationId)
+	) {
+		next(
+			new InputValidationError(
+				'Invalid Keycloak organization id.',
+				isKeycloakId.errors ?? [],
+			),
+		);
+		return;
```
jasonaowen, line 99: Mind adding a few tests around requests with undefined / null / non-uuid `keycloakOrganizationId`?

jasonaowen (changes requested):
Thanks @bickelj! This makes sense, although I have not tested it.

src/handlers/dataProvidersHandlers.ts
bickelj: Of course, yes!

src/handlers/dataProvidersHandlers.ts
bickelj: I shall.

src/types/DataProvider.ts
bickelj: One advantage to making `keycloakOrganizationId` optional is it truly is optional when creating a data provider. The way I  implemented "optional" on the relational database side of things is a nullable column on the existing data providers entity. In the API, both the API of the TypeScript code and the HTTP API, we require one less thing from callers that create a data provider. This also means the existing tests become regression tests and new tests that supply the optional parameter can test the optional parameter.

I suppose we could instead add a table to the database that relates PDC Keycloak IDs to PDC service data providers. That would be a more faithful database representation of "optional." The reason I did not do that is the anticipation of dealing with three more tables and the associated joins that result. Perhaps this is the wrong reason to pre-emptively denormalize with a nullable column. If we add the separate tables and make both the PDC Keycloak ID and PDC service data provider/funder/changemaker short code/ID non-null, would we make the `keycloakOrganizationId` optional but not null? Would that satisfy?

Or, I suppose, we could force API callers to include the `keycloakOrganizationId` every time, even if it is null. Would you expect this in the case of using a separate table for the optional attribute?

bickelj:
I added some more detail to the commit messages based on conversation with @slifty where some assumptions were not made explicit and I'd rather have those in the git history than squirreled away on GH. I will continue to follow up on the other feedback. This is also rebased on the Writable type changes.

src/types/DataProvider.ts
slifty: I don't think Jason is asking about database implementation, but rather API design.

Basically: is it more clear to have an attribute that is sometimes there and sometimes isn't, as opposed to a value that is always there and can sometimes be null.

jasonaowen: @bickelj yes, as @slifty says, I agree with your database design! My question is: when should we accept a `DataProvider` object that does not have a `keycloakOrganizationId` instead of an (logically equivalent) object that has `keycloakOrganizationId` set to `null`:

```ts
let missing: DataProvider = {
  shortCode: s,
  name: "missing KC org id",
  createdAt: "now",
};

let nulled: DataProvider = {
  shortCode: s,
  name: "null KC org id",
  createdAt: "now",
  keycloakOrganizationId: null,
};
```

I believe the only place we're constructing a `DataProvider` in TypeScript code (as opposed to SQL) is the `putDataProvider` handler call to `createOrUpdateDataProvider`, and in that instance there is always a (possibly null) `keycloakOrganizationId` key.

src/handlers/dataProvidersHandlers.ts
```diff
@@ -98,6 +113,7 @@ const putDataProvider = (
 		const dataProvider = await createOrUpdateDataProvider({
 			shortCode,
 			name,
+			keycloakOrganizationId,
```
jasonaowen, line 116: This is explicitly setting the nullable key, as opposed to omitting the key entirely when it is `null`.

jasonaowen (commented):
On making `DataProvider` accept `null` for `keycloakOrganizationId` in API calls

src/types/DataProvider.ts
```diff
@@ -20,8 +22,13 @@ const writableDataProviderSchema: JSONSchemaType<WritableDataProvider> = {
...
 			type: 'string',
 		},
+		keycloakOrganizationId: {
+			...keycloakIdSchema,
+			// I wish we could do this:
+			// nullable: true,
+		},
 	},
-	required: ['name'],
+	required: ['name', 'keycloakOrganizationId'],
```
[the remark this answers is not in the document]
bickelj: So almost, but not quite, back to where I started.

src/handlers/dataProvidersHandlers.ts
bickelj: I think this note was intended as a hint regarding an earlier note about undefined vs null. I don't think anything else is needed here, right?

src/types/DataProvider.ts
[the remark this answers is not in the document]
bickelj: Seems to work, thank you, @jasonaowen.

bickelj:
The test failure I think has to do with the call to [jsonb_strip_nulls](https://github.com/PhilanthropyDataCommons/service/blob/00fe59c308edf2b403673da02732916749dbaf55/src/database/initialization/source_to_json.sql#L25). I pushed because I will need to think about how to resolve this and wouldn't mind the extra eyes on that issue. I think the call is needed to help pick one of funder, data provider, changemaker, but then it happens to also eliminate this mandatory-but-null-allowed key.

src/types/DataProvider.ts
bickelj: @jasonaowen I wasn't sure whether you were offering a preference or pointing out an inconsistency in what I pushed, thus the digression about the database implementation. Regardless, what I pushed did not actually work. Pushing served its purpose of getting good feedback such that I didn't need to re-implement this three times across three entities. I agree that it is simpler (and better) to consistently include the key every time everywhere and let it be `null` and I've taken that approach in recent pushes.

src/handlers/dataProvidersHandlers.ts
bickelj: After looking again, I think we have ample tests for Data Provider, but we needed to strictify some of them. For example, with `toMatchObject` didn't care whether the key was there with null or not, but `toEqual` or `toStrictEqual` gave the expected failures if the key was not present.

src/types/DataProvider.ts
```diff
@@ -1,11 +1,13 @@
...
+import { KeycloakId, keycloakIdSchema } from './KeycloakUserId';
+import { Writable } from './Writable';
 import type { JSONSchemaType } from 'ajv';
-import type { Writable } from './Writable';
 import type { ShortCode } from './ShortCode';
 
 interface DataProvider {
 	readonly shortCode: ShortCode;
 	name: string;
+	keycloakOrganizationId: KeycloakId | null;
```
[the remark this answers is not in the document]
jasonaowen: Looks similar! But I'm not sure if it's the same or not.

src/database/migrations/0045-alter-changemakers-add-keycloak_organization_id.sql
jasonaowen, line 1: I think we want this and the previous migration to be a single file - if one succeeds and the next fails, we'd be in an inconsistent state. In general, I've advocated for no more than one migration per commit.

src/database/initialization/source_to_json.sql
```diff
@@ -22,16 +23,23 @@ BEGIN
...
-    'funder', funder_json,
-    'changemakerId', source.changemaker_id,
-    'changemaker', changemaker_json,
-    'createdAt', source.created_at
-  ));
+  -- The reason to go through the following hassle is we want null organization IDs returned.
+  -- At the same time, we want to omit non-existent changemakers, funders, and data providers.
+  SELECT jsonb_object_agg(name, value)
+  INTO source_json
+  FROM (VALUES
```
jasonaowen, line 30: This is a clever solution, @bickelj!

It took me a minute to understand, because it's depending on changes elsewhere: the other `_to_json.sql` files have new nullable `keycloakOrganizationId` entries, which were getting stripped out by `jsonb_strip_nulls`. Filtering out the top level without filtering out the nested `NULL`s is a good idea.

I spent a few minutes trying to figure out an alternative solution to see how it compared; I tried merging two JSONB objects with `||`, or leveraging `coalesce()`, and neither worked without `jsonb_set_lax`, which is not nearly as elegant as this!

src/database/initialization/source_to_json.sql
```diff
@@ -22,16 +23,23 @@ BEGIN
...
+    ('id', to_jsonb(source.id)),
+    ('label', to_jsonb(source.label)),
+    ('dataProviderShortCode', to_jsonb(source.data_provider_short_code)),
+    ('dataProvider', data_provider_json),
+    ('funderShortCode', to_jsonb(source.funder_short_code)),
+    ('funder', funder_json),
+    ('changemakerId', to_jsonb(source.changemaker_id)),
+    ('changemaker', changemaker_json),
+    ('createdAt', to_jsonb(source.created_at))
```
jasonaowen, line 39: Are these calls to `to_jsonb` needed to avoid mixing types?

```psql
pdc=> select jsonb_object_agg((values ('id', 3), ('createdAt', now()))); 
ERROR:  VALUES types integer and timestamp with time zone cannot be matched
LINE 1: ... jsonb_object_agg((values ('id', 3), ('createdAt', now())));
```

src/openapi.json
```diff
@@ -252,6 +257,11 @@
 						"type": "string",
 						"example": "Example Inc."
 					},
+					"keycloakOrganizationId": {
+						"description": "If available, the PDC Keycloak Organization ID for this changemaker.",
+						"type": "string",
+						"format": "uuid"
+					},
```
jasonaowen, line 264: Does this need `nullable: true`?

Since we have a few copies of this now, is it worth extracting a type to use as a `$ref`?

src/types/Funder.ts
```diff
@@ -1,11 +1,16 @@
...
-import type { Writable } from './Writable';
 import type { ShortCode } from './ShortCode';
 
 interface Funder {
 	readonly shortCode: ShortCode;
 	name: string;
+	// We do not really want "undefined" here, only null. See
+	// https://github.com/ajv-validator/ajv/issues/2283 and/or
+	// https://github.com/ajv-validator/ajv/issues/2163.
+	keycloakOrganizationId: KeycloakId | null | undefined;
```
jasonaowen, line 13: Good job working through the type issues, @bickelj!

jasonaowen (approved):
This looks good to me - excellent work, @bickelj!

I have not run it or tested it, but I believe I understand it.

src/database/initialization/source_to_json.sql
bickelj: @jasonaowen I wish I could take credit for it, but it was adapted from some SO post. I probably should include a [link to it](https://stackoverflow.com/a/58310353) there in the comments.

src/database/initialization/source_to_json.sql
bickelj: @jasonaowen I guess. Again, thanks to some SO post here, it gave an error without and stopped giving an error with the call, so I kept the calls to `to_jsonb`.

src/database/migrations/0045-alter-changemakers-add-keycloak_organization_id.sql
bickelj: I suppose all three should be in one commit, then. Or should the one commit rather be split into two? I think the latter makes most sense, three commits, one per entity, even though it could be conceptually thought of as one change.

src/database/initialization/source_to_json.sql
bickelj: I added the link.

src/database/migrations/0045-alter-changemakers-add-keycloak_organization_id.sql
bickelj: I made a commit for each entity.

src/openapi.json
bickelj: Done, nice suggestion!

PR #1374, opened 2024-12-10
Test against current Node.js weekly

### LINKED ISSUE #1367: Testing against "Current" node
In https://github.com/PhilanthropyDataCommons/service/pull/1360 @jasonaowen and @bickelj raised some really good points around the value of keeping an eye on the "upcoming" version of Node so that if something big is looming we have a chance to address it proactively before it switches to active and we need to address it reactively.

One idea that I threw out there was instead of doing that test in the context of a PR we might consider a weekly scheduled run of the tests against `Current`.  This way if something changes in `Current` in a way that will impact us (or we add a feature that won't work in `Current`) we know it in a timely manner, but also we decouple that warning from our own active development.

All this is to say...

Let's update our policy to reference the wisdom of paying attention to `Current`, and also add that weekly test against current (which is 23 right now).

(Unless of course folks do feel that PR is the right place to do it, in which case we should still make that improvement but it would be in that action instead)

jasonaowen:
Add a scheduled task to run against the [current](https://github.com/actions/setup-node/blob/39370e3970a6d050c480ffad4ff0ed4d3fdee5af/README.md#supported-version-syntax) version of Node.js, so that we can be aware of upcoming breaking changes.

Resolves #1367 Testing against "Current" node

### DIFF
--- a/.github/workflows/test-current-node.yml
+++ b/.github/workflows/test-current-node.yml
@@ -1,0 +1,53 @@
+name: Test on current Node.js
+
+on:
+  schedule:
+    - cron: '00 00 * * 1'
+  workflow_dispatch:
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: current
+          check-latest: true
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run build
+        run: npm run build
+
+      - name: Run all tests (without coverage)
+        run: npm run test:ci -- --coverage=false
+        env:
+          PGHOST: localhost
+          PGPORT: 5432
+          PGUSER: postgres
+          PGPASSWORD: postgres
+          AUTH_SERVER_ISSUER: https://totally-fake-server-name/realms/pdc
+          OPENAPI_DOCS_AUTH_CLIENT_ID: pdc-fake-client-id
+          S3_ACCESS_KEY_ID: fake-access-key-id
+          S3_ACCESS_SECRET: fake-access-key-secret
+          S3_BUCKET: fake-s3-bucket
+          S3_ENDPOINT: https://fake-s3-endpoint
+          S3_PATH_STYLE: false
+          S3_REGION: fake-s3-region


.github/workflows/current.yml
```diff
@@ -0,0 +1,53 @@
+name: Test on current Node.js
+
+on:
+  schedule:
+    - cron: '00 00 * * Mon'
```
jasonaowen, line 5: This is (intended to be) run at 00:00 UTC on Mondays; that's 4 am Pacific / 7 am Eastern. Happy to choose another time if that doesn't work for folks!

.github/workflows/current.yml
```diff
@@ -0,0 +1,53 @@
...
+  workflow_dispatch:
```
jasonaowen, line 6: This should let us manually run it whenever we want, but I don't see a way to start a new workflow from a branch; I think it has to be on `main` already.

.github/workflows/current.yml
jasonaowen: Although it turns out using weekday names is a Vixie-ism that isn't part of the POSIX standard that GitHub uses; fixed to be `1` instead of `Mon`.

.github/workflows/current.yml
```diff
@@ -0,0 +1,53 @@
...
+
+on:
+  schedule:
+    - cron: '00 00 * * 1'
+  workflow_dispatch:
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    services:
```
bickelj, line 11: This isn't specific to this PR, but I notice a lot of boilerplate in our workflow `.yml` files. Is there a such thing as includes or some other way to de-duplicate?

.github/workflows/current.yml
```diff
@@ -0,0 +1,53 @@
+name: Test on current Node.js
```
slifty, line 1: Could we put this job in the existing `test.yml` file?  If not, could we pick a more meaningful name for this file?  (e.g. `test_against_current_node` or `scheduled_test`)

slifty (changes requested):
Looks good -- main concern is the name of the file is a bit unintuitive to me but happy to have pushback.

I agree with @bickelj that it would be nice to find a way to avoid redundant boilerplate, but also agree we could open an issue for that and tackle it in future if it would require much effort to address.

.github/workflows/current.yml
jasonaowen: I believe to have a different [trigger](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#on), the jobs must be in a different file. It'd definitely be possible to run it on PRs from the existing file!

I'll rename it.

.github/workflows/current.yml
jasonaowen: It looks like their current options to [avoid duplication](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication) are intended for a much larger scale. At a glance, setting up composite actions might help - bundling together checkout, setup-node, and `npm ci` - but it's not obvious to me that the added complexity is a better trade-of than the duplication.

slifty (approved):
Thanks for the new name (and thanks for remembering to use `-` instead of `_` as word separators)

PR #1378, opened 2024-12-11
Improve Writable utility type

slifty:
This PR fixes a bug in our `Writable` utility type which cause it to lose `optional` fields.

You can test it out by adding an optional field to a type and running it through `Writable` and seeing that the new type is now properly formed.

### DIFF
--- a/src/types/Writable.ts
+++ b/src/types/Writable.ts
@@ -15,6 +15,21 @@ type WritableKeys<T> = {
 	[P in keyof T]: IfEquals<{ [Q in P]: T[P] }, { -readonly [Q in P]: T[P] }, P>;
 }[keyof T];
 
+type OptionalKeys<T> = {
+	// map each attribute to either itself or never, depending on if it is optional
+	[P in keyof T]: {} extends Pick<T, P>
+		? P /* If an empty Object type extends a type that
+		       only contains attribute P then we know attribute P is optional,
+		       since {} extends { a?: string } but not { a: string } */
+		: never; // otherwise don't keep it
+}[keyof T]; /* Pull out the mapped attributes, which are now
+               exclusively either "never" or the attributes with optional types) */
+
+// By definition if a key is not optional then it is required.
+// This means we can get the list of required keys by getting
+// all keys and then removing any optional keys
+type RequiredKeys<T> = Exclude<keyof T, OptionalKeys<T>>;
+
 type WritablePrimitive =
 	| undefined
 	| null
@@ -22,6 +37,19 @@ type WritablePrimitive =
 	| string
 	| number
 	| Function;
+
+type WritableObject<T> = {
+	// Capture required keys
+	[K in keyof T & RequiredKeys<T> & WritableKeys<T>]: Writable<T[K]>;
+} & {
+	// Capture optional keys
+	[K in keyof T & OptionalKeys<T> & WritableKeys<T>]?: Writable<T[K]>;
+} extends infer O
+	? {
+			// Flatten the two types (required and optional) into a single combined type
+			[K in keyof O]: O[K];
+		}
+	: never;
 
 export type Writable<T> = T extends WritablePrimitive
 	? T
@@ -32,5 +60,5 @@ export type Writable<T> = T extends WritablePrimitive
 			: T extends Set<infer X>
 				? Set<Writable<X>>
 				: T extends Object
-					? { [K in WritableKeys<T>]: Writable<T[K]> }
+					? WritableObject<T>
 					: T;
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,7 @@
 				"eslint-config-prettier": "^9.1.0",
 				"eslint-plugin-import": "^2.31.0",
 				"eslint-plugin-sort-exports": "^0.9.1",
+				"expect-type": "^1.1.0",
 				"jest": "^28.1.1",
 				"jsonwebtoken": "^9.0.2",
 				"mock-jwks": "^1.0.10",
@@ -5416,6 +5417,16 @@
 			},
 			"engines": {
 				"node": "^12.13.0 || ^14.15.0 || ^16.10.0 || >=17.0.0"
+			}
+		},
+		"node_modules/expect-type": {
+			"version": "1.1.0",
+			"resolved": "https://registry.npmjs.org/expect-type/-/expect-type-1.1.0.tgz",
+			"integrity": "sha512-bFi65yM+xZgk+u/KRIpekdSYkTB5W1pEf0Lt8Q8Msh7b+eQ7LXVtIB1Bkm4fvclDEL1b2CZkMhv2mOeF8tMdkA==",
+			"dev": true,
+			"license": "Apache-2.0",
+			"engines": {
+				"node": ">=12.0.0"
 			}
 		},
 		"node_modules/express": {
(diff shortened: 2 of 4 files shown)


src/types/Writable.ts
```diff
@@ -15,6 +15,12 @@ type WritableKeys<T> = {
 	[P in keyof T]: IfEquals<{ [Q in P]: T[P] }, { -readonly [Q in P]: T[P] }, P>;
 }[keyof T];
 
+type OptionalKeys<T> = {
+	[P in keyof T]: {} extends Pick<T, P> ? P : never;
```
jasonaowen, line 19: This creates a [union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types).

A key `P` (why the name `P`? `K` or `Key` seems clearer here) is optional if `{} extends Pick<T, P>` - that is, if an empty object is a valid assignment for [`Pick<T, P>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys), then the picked key must not be required.

So, `OptionalKeys<T>` is the union of all the optional keys.

src/types/Writable.ts
```diff
@@ -15,6 +15,12 @@ type WritableKeys<T> = {
...
+}[keyof T];
```
jasonaowen, line 20: What does `[keyof T]` do here? If I delete it, the TypeScript compiler and ESLint are both still happy. Is that because we're not using some edge case this addresses?

src/types/Writable.ts
```diff
@@ -15,6 +15,12 @@ type WritableKeys<T> = {
...
+
+type RequiredKeys<T> = Exclude<keyof T, OptionalKeys<T>>;
```
jasonaowen, line 22: Likewise, this is a union type of all the remaining keys, which by definition are not optional.

src/types/Writable.ts
```diff
@@ -32,5 +38,20 @@ export type Writable<T> = T extends WritablePrimitive
...
+					? {
+							// Capture required keys
+							[K in keyof T & RequiredKeys<T> & WritableKeys<T>]: Writable<
+								T[K]
+							>;
+						} & {
+							// Capture optional keys
+							[K in keyof T & OptionalKeys<T> & WritableKeys<T>]?: Writable<
+								T[K]
+							>;
```
jasonaowen, line 50: The crucial difference here being the `?` after the `[K in keyof T...]`, leaving the optional keys optional, and the required keys required.

src/types/Writable.ts
```diff
@@ -32,5 +38,20 @@ export type Writable<T> = T extends WritablePrimitive
...
+						} extends infer O
+						? {
+								// Flatten the two types (required and optional) into a single combined type
+								[K in keyof O]: O[K];
+							}
+						: never
```
jasonaowen, line 56: Label the [intersection type](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types) as `O`, so that we can do an identity pass on it.

Is this necessary? I was able to delete this clause and it still compiles / builds / passes ESLint:

```suggestion
						}
```

src/types/Writable.ts
jasonaowen, line 1: I see a use case listed in https://github.com/PhilanthropyDataCommons/service/pull/1373#discussion_r1880701441, but I'll raise the same question here as I did on https://github.com/PhilanthropyDataCommons/service/pull/1349#discussion_r1868571402: do we _want_ to enable optional writable keys? I find it is almost always more clear to have a key with an explicit `null` value than it is to have a key sometimes be missing.

jasonaowen (changes requested):
This is some dark magic, but let me make sure I understand it...

jasonaowen:
It occurs to me that _some_  kind of automated testing would be useful for types this complex, but I don't know what that would look like.

A quick search turns up [How to test your types](https://www.totaltypescript.com/how-to-test-your-types), which looks promising!

src/types/Writable.ts
slifty: I agree that we won't have the use case soon, but I also think it's worth improving the utility type to fix this bug since this is now the third time it's come up (even if we ultimately don't go in that direction).  It's just more feature complete at the TS level.

src/types/Writable.ts
slifty: Given how magic and complex all this is, would it be worth adding some of this narrative in the form of comments?

src/types/Writable.ts
slifty: They're happy, but it changes what `OptionalKeys` actually accomplishes

My understanding is that the `{}` block here is creating a mapped object where the attributes are keys of T, and values are either P or never.  Then the `[keyof T]` is going through each key and for keeping anything that isn't `never`.

With the `[keyof T]`:

image


Without `[keyof T]`:
image

So, without keeps the non-optional keys in there (but calls it `never`) <-- the presence of the key in the object means it gets kept later on.

slifty:
Added tests for types 🎉; Will see if I can hop on the office hours call to talk through the other questions and get a sense of whether you think comments would be useful to add before merging.

src/types/Writable.ts
jasonaowen: Aha! Excellent, thank you very much for the explanation and demonstration!

src/types/Writable.ts
jasonaowen: Aha, I see that now too. Thank you for walking through this with me!

src/types/Writable.ts
jasonaowen: @slifty and I talked today, and worked out a refactoring that makes things clearer, and he plans to add more comments, so I'm marking this as resolved.

.eslintrc.json
```diff
@@ -41,6 +41,14 @@
 					"src/**/*.test.*.{ts,js}"
 				]
 			}
+		],
+		"jest/expect-expect": [
```
jasonaowen, line 45: Hm, I'm not able to resolve this, and neither is CI:

> error  Definition for rule 'jest/expect-expect' was not found

Is there a corresponding dev package that needs to be installed?

src/types/Writable.ts
slifty: I've added comments... do they make it clearer, who knows.

.eslintrc.json
slifty: Huh -- I added it per the library documentation but taking it out results in no issues or complaints.

Removing that change entirely.

src/types/Writable.ts
```diff
@@ -15,6 +15,18 @@ type WritableKeys<T> = {
 	[P in keyof T]: IfEquals<{ [Q in P]: T[P] }, { -readonly [Q in P]: T[P] }, P>;
 }[keyof T];
 
+type OptionalKeys<T> = {
+	// map each attribute to either itself or never, depending on if it is optional
+	[P in keyof T]: {} extends Pick<T, P>
+		? P // If an empty Object type extends a type that only contains attribute P then we know attribute P is optional, since {} extends { a?: string } but not { a: string }
+		: never; // otherwise don't keep it
+}[keyof T]; // Pull out the mapped attributes (which are now exclusively either "never" or the attributes with optional types)
```
jasonaowen, line 23: These are great! Mind wrapping them?

jasonaowen (approved):
This looks good - thank you for talking through this with me, @slifty!

src/types/Writable.ts
slifty: Will do -- though prettier ruins everything and puts half the comment after the `:` grumble grumble

slifty: Ok fixed it using `/* */`

This is why I never use comments 😅

Issue #1375, opened 2024-12-10
Validate GitHub Actions workflow files


jasonaowen:
The YAML files in `.github/workflows` need to be correctly shaped for workflows to run, but it's not obvious to me when or how they checked by GitHub Actions itself. Finding incorrect syntax - and, hopefully, suggesting corrected syntax - is something that should at least be checked on pull requests, and ideally be easily checkable locally.

I found a few options so far, but none of them are perfect:

- [action-validator](https://github.com/mpalmer/action-validator) has an NPM package, but the CLI only does one file at a time, and does not handle globs; we'd have to write some kind of wrapper script.
- action-validator gets its schema from [schemastore](https://github.com/SchemaStore/schemastore) (https://json.schemastore.org/github-workflow.json) which could be used with [ajv-cli](https://ajv.js.org/guide/environments.html#command-line-interface). ajv-cli does accept globs, but we'd either have to download the schema locally, vendor it, or add schemastore as a git submodule.
- [actionlint](https://github.com/rhysd/actionlint) seems to be the most thorough, can be run locally, and has its own [GitHub Action](https://github.com/marketplace/actions/actionlint) to make adding it to CI easy, but it's written in Go, so that'd be yet another language dependency; folks would either need to download the binary or have Go installed to compile it locally.

We could also just not do this, since we don't update our actions all that often, but I found it helpful to have some validation while working on PR #1374.

Thoughts, @bickelj @hminsky2002 @slifty?

slifty:
@jasonaowen I'll take a closer look at this shortly but in the meantime: when you open a PR I believe GitHub runs the actions defined in the branch, so for most of our actions would that be enough?

I agree that for scheduled actions that doesn't quite do the trick.

jasonaowen:
Now that I look more closely, actioncheck actually does [a lot](https://github.com/rhysd/actionlint/blob/v1.7.4/docs/checks.md) more than just validating the YAML, including shellcheck for `run:` steps and hardcoded credentials, and is what I'd advocate.

bickelj:
@jasonaowen I say we try one, no objection to the one you mentioned, actionlint. If it helps, keep using it. If it causes pain, find an alternative or stop using it. A [while ago I found act](https://github.com/PhilanthropyDataCommons/deploy/pull/107#issuecomment-2023430265) to be helpful when developing GH actions.

slifty:
This is resolved by #1379 right?

jasonaowen:
[...]

PR #1379, opened 2024-12-12
Lint GitHub Action workflows with actionlint

jasonaowen:
Use the [actionlint action](https://github.com/marketplace/actions/actionlint) to run [actionlint](https://github.com/rhysd/actionlint) on pull requests.

Issue #1375 Validate GitHub Actions workflow files

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,6 +6,15 @@ on:
   pull_request:
 
 jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .node-version
+      - name: Lint GitHub Action workflows
+        uses: raven-actions/actionlint@v2
   eslint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


bickelj (approved):
I started down the path of using `act` to run this locally, but that proved troublesome (I need a GH token to do it locally? what?), so I looked at https://github.com/PhilanthropyDataCommons/service/actions/runs/12288589684/job/34292643249?pr=1379#step:4:421 and the code here instead. It looks good to me.

.github/workflows/lint.yml
```diff
@@ -6,6 +6,15 @@ on:
   pull_request:
 
 jobs:
+  actionlint:
```
slifty, line 9: I'm realizing now that it's too bad that there's no way to have a linter that runs locally (I don't like things that require a PR to validate / can't be perfected and checked by local tooling without having to generate repo noise), but this is still an improvement and it doesn't seem like there are comparable local tools out there!

slifty (approved):

.github/workflows/lint.yml
jasonaowen: Oh, I meant to document this in the README - you can run it locally! Go to https://github.com/rhysd/actionlint and either compile it locally or grab a release binary. I'll document that before merging.

PR #1380, opened 2024-12-12
Use lowercase for SQL function and type names

jasonaowen:
As discussed in https://github.com/PhilanthropyDataCommons/service/pull/1259#discussion_r1878726661 and https://github.com/PhilanthropyDataCommons/service/pull/1259#discussion_r1878727869, set the SQLFluff capitalization policy for function names and type names to be lowercase.

### DIFF
--- a/.sqlfluff
+++ b/.sqlfluff
@@ -10,6 +10,9 @@ exclude_rules =
 indent_unit = space
 tab_space_size = 2
 
+[sqlfluff:rules:capitalisation.functions]
+extended_capitalisation_policy = lower
+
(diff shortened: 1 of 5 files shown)


jasonaowen:
I merged #1259 into this branch locally and ran SQLFluff, and I don't think there's anything controversial in any of its suggestions! Maybe we'll find more things to argue about as we go, but I think this puts us in a great place.

slifty (approved):
Linting really shouldn't make me as happy as it does.

Thank you for this!

bickelj (approved):
LGTM!

PR #1381, opened 2024-12-12
Ignore engine spec when testing with latest Node

jasonaowen:
We are defining the version of Node we support with the [`engines` directive](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#engines) in `package.json` - as we should be. However, when testing against the latest version of Node, we need to override that. Turn off [`engine-strict`](https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict) via the command line, so that Node will install our dependencies.

Follow-up to #1374

### DIFF
--- a/.github/workflows/test-current-node.yml
+++ b/.github/workflows/test-current-node.yml
@@ -31,7 +31,7 @@ jobs:
           check-latest: true
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --no-engine-strict
 
       - name: Run build
         run: npm run build


slifty (approved):

jasonaowen:
- [Initial failing run](https://github.com/PhilanthropyDataCommons/service/actions/runs/12305767562)
- [Successful run against this branch, pre-merge](https://github.com/PhilanthropyDataCommons/service/actions/runs/12305946222)
- [Successful run against `main`, post-merge](https://github.com/PhilanthropyDataCommons/service/actions/runs/12306057412)

.github/workflows/test-current-node.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           check-latest: true
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --no-engine-strict
```
bickelj, line 34: Elegant solution!

PR #1384, opened 2024-12-16
Use built-in type definitions for uuid

jasonaowen:
[Version 11 of the uuid library](https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md#1100-2024-10-26) included a migration to TypeScript, so the externally-provided `@types/uuid` is no longer needed. Uninstall it.

We upgraded to version 11 in PR https://github.com/PhilanthropyDataCommons/service/pull/1285.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,6 @@
 				"@types/pg": "^8.11.10",
 				"@types/supertest": "^6.0.2",
 				"@types/swagger-ui-express": "^4.1.7",
-				"@types/uuid": "^10.0.0",
 				"@typescript-eslint/eslint-plugin": "^7.18.0",
 				"@typescript-eslint/parser": "^7.18.0",
 				"copyfiles": "^2.4.1",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -10320,6 +10313,7 @@
 				"https://github.com/sponsors/broofa",
 				"https://github.com/sponsors/ctavan"
 			],
+			"license": "MIT",
```
bickelj, line 10316: Strange that this line was added.

bickelj (approved):
It is always nice to see unneeded code removed. I did not try it locally but the tests, IIRC, include some use of Uuid, so since all that passed I think this is going to be OK.

package-lock.json
jasonaowen: I noticed that too; I think I updated `npm`, but I'm not sure what version from, and I can't find anything obvious in the [npm changelog](https://github.com/npm/cli/blob/release/v10/CHANGELOG.md). The [upstream `package.json` license](https://github.com/uuidjs/uuid/blame/v11.0.3/package.json#L21)  hasn't changed in almost a decade.

PR #1392, opened 2024-12-21
Enable SQLFluff indenting

jasonaowen:
When introducing SQLFluff in PR #321, I intentionally left the [`layout.indent` rule](https://docs.sqlfluff.com/en/stable/reference/rules.html#rule-layout.indent) disabled. There were already so many non-whitespace changes involved that I thought reindenting the world would be too much.

Enable and configure the `layout.indent` rule, and apply it to our SQL code.

I configured it to use tabs, like the rest of our codebase. Beyond the type of indentation, there are several options available for [configuring layout](https://docs.sqlfluff.com/en/stable/configuration/layout.html); I chose the ones that made the most sense to me.

For the ease of reviewability, I split this into four commits: manually reformatting to avoid automatic reformatting issues, the automatic reformatting that includes multiline changes, the whitespace-only automatic reformatting, and the configuration changes. Note that **if you rebase this PR** you will need to **update the `.git-blame-ignore-revs` file** with the new commit hash.

I apologize for dropping this and then leaving; I meant to get it done earlier, but time got away from me. Please merge when/if you feel ready.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,5 @@
 # Invoking prettier for the first time
 b16d4388c0457cd6fe02fc09947712d6f3afed7c
+
+# Re-indenting SQL with SQLFluff
+b555be039a94a139f31a1f0d1a8ccf9fcaa89c88
--- a/.sqlfluff
+++ b/.sqlfluff
@@ -2,13 +2,14 @@
 dialect = postgres
 templater = placeholder
 exclude_rules =
-  layout.indent,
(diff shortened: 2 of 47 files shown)


bickelj (approved):
No objection. I think I was on another project when we switched from spaces to tabs. You say	tomato, I say   tomato.

slifty (approved):
Thank you for yet another improvement <3

slifty:
There were merge conflicts so I had to rebase (sorry for the stolen valor)

Issue #1291, opened 2024-10-29
Keycloak groups to assist with PDC object permissions


slifty:
In #1093 @bickelj asked some questions about using Keycloak for managing permissions and didn't really get a direct answer (sorry Jesse, there were enough moving parts / trying to scope out the core use cases)!

Right now I'm just about finished with the MVP of our permissions based on  which will involve a roles table, but it would be reasonable to take a moment to reflect on what a Keycloak-based granular permissions system would look like and what the benefits / negatives would be.

bickelj:
I just looked at the latest Keycloak docs. Lo and behold, version 26 has something we might like: [Organizations](https://www.keycloak.org/docs/26.0.4/server_admin/index.html#_managing_organizations).

In my comment in #1093 I am glad I explicitly linked to the particular minor version of Keycloak I was referring to because that version (one of the versions starting with 25) did not have Organizations.

Off the cuff, a design comes to mind:
* Embrace such Organizations in Keycloak (of course necessitating a Keycloak upgrade).
* Let an Organization in Keycloak be one of PDC-land Funder, Changemaker, or DataProvider.
* Add one (enum-ish) [attribute to each Keycloak Organization](https://www.keycloak.org/docs/26.0.4/server_admin/index.html#_managing_attributes_) denoting "this Organization is a Funder" vs a Changemaker or DataProvider.
* Ensure [the Keycloak-supplied JWT includes Organization attributes](https://www.keycloak.org/docs/26.0.4/server_admin/index.html#mapping-organization-claims_server_administration_guide), especially `id` which should be a `UUID`.
* An integration with the currently proposed #1284 could be pretty seamless: let the proposed permissions-on-Organizations table to allow either a UUID that is a Keycloak ID or an Organization ID. Or we could come up with something else that depends less on our tables and more on Keycloak if we find that possible.

bickelj:
We have Keycloak 26 in production now, so we can use either Keycloak groups (any Keycloak version) or Keycloak organizations (Keycloak >= 26).

The commits related to #1250/#1284 allow individual users to be granted permission on a given organization's (changemaker's, funder's, or data provider's) stuff. In other words, the `changemaker`, `funder`, and `data_provider` refer not to the user (subject) or the user's associations, but a grant of permission to the user for the given organization's objects. Something is still missing, namely, grouping users together into changemakers, funders, and data providers.

jmergy:
[...]

PR #1395, opened 2024-12-24
Let changemaker Keycloak organization ID be set

bickelj:
Without this commit, there would be no way for the PDC service code to set a changemaker's Keycloak organization UUID after initial creation. With this commit a function is available to set it. While the capability is not yet exposed to the HTTP API, the first use of the function is expected to be within the service, perhaps in middleware, and not yet exposed to users. Another reason to defer exposure in the HTTP API is to figure out whether we want to use `PUT` or `PATCH` for this purpose. Unlike the funder and data provider entities, changemaker has a synthetic key of `id` due to a compound natural key of `taxId` plus `name`.

Issue #1291

### DIFF
--- a/src/__tests__/changemakers.int.test.ts
+++ b/src/__tests__/changemakers.int.test.ts
@@ -17,6 +17,7 @@ import {
 	loadSystemUser,
 	loadTableMetrics,
 	createOrUpdateDataProvider,
+	updateChangemaker,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
 import { mockJwt as authHeader } from '../test/mockJwt';
@@ -31,8 +32,10 @@ import {
 	Opportunity,
 	PostgresErrorCode,
 	Source,
+	stringToKeycloakId,
 	User,
 } from '../types';
+import { NotFoundError } from '../errors';
 
 const insertTestChangemakers = async () => {
 	await createChangemaker(db, null, {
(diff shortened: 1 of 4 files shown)


src/database/queries/changemakers/updateKeycloakOrganizationId.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE changemakers
```
slifty, line 1: Would you be OK with sticking with our current (albiet little used) pattern of universal update queries with optional parameters?   You can see and example in `bulkUploadTasks/updateById.sql`.

This would just mean renaming the query to `updateById` -- I don't think you even need to use the COALESCE stuff that the bulkUploadTask query uses since there's only one field that can be updated here.

(The thinking is that we don't want to have to make an entire new query for every possible combination of fields that get updated for a given entity type.)

slifty (changes requested):
Consistency (and flexibility, I suppose) request

Also just to clarify: we allow it to be set already, right?  This allows it to be `updated` yes?  May be worth tweaking the commit message if so just to prevent confusion for future archivists.

src/database/queries/changemakers/updateKeycloakOrganizationId.sql
bickelj: My reasoning at the time was that there is only one field to be updated as of this commit, therefore no need to generalize prematurely, we could generalize when another field needs an update. Then again, I wonder if we want to allow tax ID or name to change for an existing changemaker. I also probably did not want to open that potential can of worms. But in short, yes, I can go ahead and genericize in advance.

bickelj:
> Also just to clarify: we allow it to be set already, right? This allows it to be `updated` yes? May be worth tweaking the commit message if so just to prevent confusion for future archivists.

Yes and no. We allow it to be set on initial creation, but the most common case will be automated initial creation prior to existence in Keycloak. We need to be able to set it after initial creation. We had decided that we would not require the PDC Keycloak Organization to exist prior to the PDC service changemaker existing. And that is OK, as long as we can set it later. The use case here is for a middleware to graze on JWTs and update changemakers as links are found. This change isn't needed for Data Providers and Funders because they already have a `PUT` endpoint and corresponding functions.

PR #1396, opened 2024-12-30
Add `PUT` and `DELETE` endpoints for UserFunderPermissions

slifty:
This PR adds two new endpoints that make it possible to create and delete UserFunderPermissions in the PDC.

It uses the same patterns that were established for UserChangemakerPermissions

Related to #1250

### DIFF
--- a/src/__tests__/userFunderPermissions.int.test.ts
+++ b/src/__tests__/userFunderPermissions.int.test.ts
@@ -4,6 +4,8 @@ import {
 	createOrUpdateFunder,
 	createOrUpdateUserFunderPermission,
 	loadSystemUser,
+	loadUserFunderPermission,
+	removeUserFunderPermission,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
 import {
@@ -11,6 +13,7 @@ import {
 	mockJwtWithAdminRole as authHeaderWithAdminRole,
 } from '../test/mockJwt';
 import { keycloakIdToString, Permission } from '../types';
+import { NotFoundError } from '../errors';
 
 describe('/users/funders/:funderShortcode/permissions/:permission', () => {
 	describe('PUT /', () => {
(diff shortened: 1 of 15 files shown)


slifty:
The swagger spec + changelog need to be updated before this can be merged.

src/openapi.json
```diff
@@ -2828,6 +2857,107 @@
 					}
 				}
 			}
+		},
+		"/users/{userKeycloakUserId}/fundera/{funderShortCode}/permissions/{permission}": {
```
slifty, line 2861: `fundera` => `funders`

src/__tests__/userFunderPermissions.int.test.ts
```diff
@@ -118,7 +121,7 @@ describe('/users/funders/:funderShortcode/permissions/:permission', () => {
 
 			const response = await request(app)
 				.put(
-					`/users/${keycloakIdToString(user.keycloakUserId)}/fundera/${funder.shortCode}/permissions/${Permission.EDIT}`,
```
hminsky2002: Methinks this should be in the previous commit

src/__tests__/userFunderPermissions.int.test.ts
```diff
@@ -144,7 +147,7 @@ describe('/users/funders/:funderShortcode/permissions/:permission', () => {
 				userKeycloakUserId: user.keycloakUserId,
 				funderShortCode: funder.shortCode,
 				permission: Permission.MANAGE,
-				createdBy: user.keycloakUserId,
+				createdBy: systemUser.keycloakUserId,
```
hminsky2002, line 150: this also should be in the previous commit, methinks

hminsky2002 (changes requested):
This looks good to me -- I tested locally and had all runs passing. I noticed two lines in the final commit that looked like they should be part of the previous commit and marked them as such!

hminsky2002:
@slifty a question more than a comment -- as it stands since we use an upsert for a funder permission, is it true that a user and funder can only have one userFunderPermission entity between the two of them? If so, does that mean that the different permission levels (Manage and Edit) are meant to be structured hierarchically? I.e ever permission is a subset of a larger permission. That seems reasonable enough, but wanted to clarify if we're going to be adding more permissions further down the line.

slifty:
> @slifty a question more than a comment -- as it stands since we use an upsert for a funder permission, is it true that a user and funder can only have one userFunderPermission entity between the two of them? If so, does that mean that the different permission levels (Manage and Edit) are meant to be structured hierarchically? I.e ever permission is a subset of a larger permission. That seems reasonable enough, but wanted to clarify if we're going to be adding more permissions further down the line.

The primary key for UserFunderPermission is actually *three* things (the specific permission being the third thing).  So it's the case that there can only be one row per combindation of user + funder + permission (e.g. you can't grant edit twice, but you CAN grant edit and view).

slifty:
Good catches -- ready for re-review

hminsky2002 (approved):
Splendid, thanks @slifty

bickelj (commented):
I'm a bit late. But I tried this locally after merge to main and it seemed to work as expected, both for bad/non-existent IDs and good/existing ones.

PR #1400, opened 2025-01-02
Streamline lint / format scripts

### LINKED ISSUE #1399: Integrate sqlfluff steps in our scripts
When we merged #321 we added some CLI commands to support it but also ended up fragmenting the commands so devs can no longer locally fixup or lint their project in a single command.

Let's add the sqlfluff to our universal `lint` and re-introduce a singular `format` command.

This does mean that the python dependency becomes required to do any linting or formatting, but I think that decision was made already when we made the (valid!) decision to use sqlfluff in the first place.

slifty:
This PR cleans up our scripts so devs can just run `npm run format` and `npm run lint` to format and lint all types of file.

Resolves #1399


hminsky2002 (approved):
Necessary addition, thanks @slifty.

bickelj (commented):
This is better, thank you!

PR #1403, opened 2025-01-02
Add `PUT` and `DELETE` endpoints for UserDataProviderPermissions

### LINKED ISSUE #1250: Support granting of permissions for users in the context of organizational entities

### LINKED ISSUE #1351: Support removal of user permissions
With #1250 we allow the creation of permission relationships within organizational entities.  We also need to support removing those permissions.

This will be done using a `not_after` field.  Eventually this means we might allow the creation of temporary permissions, though for now there isn't a use case so that functionality won't yet be exposed via API.

slifty:
This PR adds the final set of permission endpoints for our three organizational entity types.

Resolves #1250 
Resolves #1351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
 ### Added
 
 - Manage user-funder permissions using `PUT`, and `DELETE` on `/user/{keycloakUserId}/funders/{funderShortCode}/permissions/{permission}`.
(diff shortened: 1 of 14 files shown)


hminsky2002 (approved):
i see no pattern deviations from #1396, looks good and tests are passing locally. approved!

bickelj:
Likewise late to this one but I tried adding and deleting (and viewing the results in /user) these permissions and it seemed to work as expected.

Issue #1407, opened 2025-01-03
Improve db utilities to consistently accept an optional authentication state


slifty:
As part of #1406 we will need queries to only return data that the caller is allowed to access.  Some of this might happen at the route definition but at least for collection queries the database will need to be able to filter as part of the query itself.

Part of this issue should be to reflect on whether the query should ALWAYS accept auth contexts and implement auth logic at the query level in all cases, or whether we should rely on middleware logic whenever possible.

slifty:
I might be missing something, but I'm 99% sure that we now have moved all DB operations to generators!

PR #1413, opened 2025-01-07
Add authentication to bundle loaders

slifty:
This PR does two things:

(1) It DRYs up our bundle loaders by moving the core logic to a generator function and updating our bundle loader definitions to use the generator

(2) It updates the generator to universally accept an auth context as the first parameter, which then passes the appropriate auth information to the bundle queries.

This doesn't actually *use* the authentication in our queries but it is a necessary precursor for that use.

Related to #1407


bickelj:
On first skim, this looks pretty nifty and helpful.

src/database/operations/applicationForms/loadApplicationFormBundle.ts
```diff
@@ -1,21 +1,7 @@
...
-		},
-		'application_forms',
-	);
-	const entries = bundle.entries.map((entry) => entry.object);
-	return {
-		...bundle,
-		entries,
-	};
-};
+export const loadApplicationFormBundle = generateLoadBundleOperation<
```
hminsky2002, line 4: awesome, cheers to following patterns!

src/database/initialization/changemaker_to_json.sql
```diff
@@ -2,7 +2,7 @@ SELECT drop_function('changemaker_to_json');
 
 CREATE FUNCTION changemaker_to_json(
 	changemaker changemakers,
-	keycloakUserId uuid DEFAULT NULL
```
hminsky2002, line 5: so to clarify, we are renaming this field, but not the field actually attached to the user Schema which is still named 'keycloakUserId', to differentiate it in context? I.e even though it's the same value sourced from the User, we want to refer to it from the changemaker context?

src/__tests__/baseFields.int.test.ts
```diff
@@ -589,6 +589,7 @@ describe('/baseFields', () => {
 			const after = await loadTableMetrics('base_field_localizations');
 			const baseFieldLocalizations =
 				await loadBaseFieldLocalizationsBundleByBaseFieldId(
+					undefined,
```
hminsky2002, line 592: Just wondering if undefined is semantically the value we want to use when we want to ignore the first parameter -- I suppose so, since we're saying the request is undefined. that seems acceptable

hminsky2002 (approved):
@slifty awesome! A masterclass in DRY'ing. Approving because I've tested locally with all tests passing, and believe that everything is handled correctly. 

My sense here is that this commit is justified since we already adhere so strictly to a pattern of bundle loaders, as seen in how uniformly each refactor to a generator function was. I have been probing my head to think of any sort of entity or use case in the future that could have a different enough functionality that using this generator would be limiting, but I think the task of loading a paginated list of entities is broad enough that that shouldn't be a problem. Something to consider moving forward, considering there are many patterns we adhere to strictly that might be able to be DRY'd up. 

 For mine own sake I left a few clarifying comments, and am gonna add a questions here:

Despite the changing of a field in one of our database initializations, that field naming is handled internally by our api, and thus no updates are needed to our open api spec?

src/database/initialization/changemaker_to_json.sql
slifty: Essentially yes -- basically this parameter is *intended* to signal the authentication context of the querying user, it's not simply a generic instance of a keycloak user ID (you can see this intent in the way it's used in the query).

Giving it the more specific name makes its purpose more explicit and leaves less room for confusion.

It's also a prelude to the fact that we're soon going to be starting to pass auth contexts around to our various functions in a much more formal / predictable way, but that's not part of this PR.

src/__tests__/baseFields.int.test.ts
slifty: Not only that but it's the only value we *can* use (per the type definition of `AuthContext | undefined`

But yes semantically it works too because what were saying is "the auth context is not defined, you should go forward behaving however you would want to behave without an auth context"

slifty:
@hminsky2002

> Despite the changing of a field in one of our database initializations, that field naming is handled internally by our api, and thus no updates are needed to our open api spec?

Correct, this is all internal.

> I have been probing my head to think of any sort of entity or use case in the future that could have a different enough functionality that using this generator would be limiting

So far we have not, and of course in that case one could simply not use the generator.

> Something to consider moving forward, considering there are many patterns we adhere to strictly that might be able to be DRY'd up

Stay tuned for future PRs (e.g. #1415 and the one I'm about to make for create / update!)

PR #1415, opened 2025-01-07
Add AuthContext to load query operations

slifty:
This PR continues the work of DRYing out our DB operators by using a generator for load operations; it also adds in the auth context for load operations via that generator.

Related to #1407

### DIFF
--- a/src/database/operations/generators/generateLoadItemOperation.ts
+++ b/src/database/operations/generators/generateLoadItemOperation.ts
@@ -1,6 +1,10 @@
 import { db } from '../../db';
 import { NotFoundError } from '../../../errors';
-import type { JsonResultSet } from '../../../types';
+import {
+	getIsAdministratorFromAuthContext,
+	getKeycloakUserIdFromAuthContext,
+} from '../../../types';
+import type { AuthContext, JsonResultSet } from '../../../types';
 
 /**
  * Generates an item loader function for a specific query and table.
@@ -14,18 +18,26 @@ import type { JsonResultSet } from '../../../types';
  *
  * @returns {Function} A function that takes query parameters, limit, and offset, and returns a promise that resolves to a bundle of entries and total count.
  */
- const generateLoadItemOperation = <T, P extends [...args: unknown[]]>(
-	queryName: string,
-	entityType: string,
-	parameterNames: { [K in keyof P]: string },
-) => {
-	return async (...args: [...P]): Promise<T> => {
+const generateLoadItemOperation =
+	<T, P extends [...args: unknown[]]>(
+		queryName: string,
+		entityType: string,
+		parameterNames: { [K in keyof P]: string },
+	) =>
+	async (authContext: AuthContext | null, ...args: [...P]): Promise<T> => {
+		const authContextKeycloakUserId =
+			getKeycloakUserIdFromAuthContext(authContext);
+		const authContextIsAdministrator =
+			getIsAdministratorFromAuthContext(authContext);
 		const queryParameters = parameterNames.reduce(
 			(acc, parameterName, index) => ({
 				...acc,
 				[parameterName]: args[index],
 			}),
-			{},
+			{
+				authContextKeycloakUserId,
+				authContextIsAdministrator,
+			},
 		);
 
 		const result = await db.sql<JsonResultSet<T>>(queryName, queryParameters);
@@ -38,6 +50,5 @@ import type { JsonResultSet } from '../../../types';
 		}
 		return object;
 	};
-};
 
 export { generateLoadItemOperation };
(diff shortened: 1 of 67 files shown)


src/database/operations/generators/generateItemLoader.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param {Object} parameterNames - An object mapping parameter indices to their names within the query.
+ *
+ * @returns {Function} A function that takes query parameters, limit, and offset, and returns a promise that resolves to a bundle of entries and total count.
+ */
+
+import { db } from '../../db';
+import { NotFoundError } from '../../../errors';
+import type { JsonResultSet } from '../../../types';
+
+const generateItemLoader = <T, P extends [...args: unknown[]]>(
```
slifty, line 18: I think that as we add verbs (create, update, etc) it will be cleaner for this to be called `generateLoadItemOperation` rather than `generateItemLoader`.

e.g. `generateCreateOrUpdateItemOperation` is a cleaner name than something like `generateItemCreateOrUpdater`

A little more verbose, but hash-tag-worth imho

src/database/operations/generators/generateLoadItemOperation.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Generates an item loader function for a specific query and table.
+ *
+ * @template T - The type of the item being loaded.
+ * @template P - The type of the parameters for the query. Use labeled tuples so the generated function has a pretty type definition.
+ *
+ * @param {string} queryName - The name of the query to execute.
+ * @param {string} tableName - The name of the table to load metrics from.
```
slifty, line 8: This is copypasta (the parameter is `entityType`)

src/database/operations/generators/generateLoadItemOperation.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @template T - The type of the item being loaded.
+ * @template P - The type of the parameters for the query. Use labeled tuples so the generated function has a pretty type definition.
+ *
+ * @param {string} queryName - The name of the query to execute.
+ * @param {string} entityType - The name of entity type being loaded, for use in error messages.
+ * @param {Object} parameterNames - An object mapping parameter indices to their names within the query.
+ *
+ * @returns {Function} A function that takes query parameters, limit, and offset, and returns a promise that resolves to a bundle of entries and total count.
+ */
```
bickelj, line 12: This documentation is excellent and I want to see it when I poke at the calls to this function in other parts of the code. But codium doesn't show it unless it precedes the function, a few lines down.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -45,7 +45,7 @@ const assertApplicationFormExistsForProposal = async (
 
 	let proposal: Proposal;
 	try {
-		proposal = await loadProposal(proposalId);
+		proposal = await loadProposal(undefined, proposalId);
```
bickelj, line 48: Passing `undefined` here and elsewhere feels wrong, somehow. Passing `null` would feel wonky too but at least we typically think of `null` as a potential value. Shouldn't `undefined` be reserved for cases when things are not defined?

When I tried to pass nothing by deleting `undefined` I of course got a typescript error because it expects an arg. When I tried to pass `null` it doesn't like that either because the generator does not generate `| null` as part of the signature. So I don't know if there's a quick fix here.

Is it possible to make the auth context an optional argument?

bickelj (commented):
I was able to `lint`, `build`, and `start`, and calls to get individual entities succeeded, great!

I also see a "+363 -479" lines summary which means that even with the comments and import statements this is a net reduction in code, also great!

I am not sure about the use of `undefined` in this situation. See in-line comment.

src/handlers/proposalVersionsHandlers.ts
slifty: Regarting optional: in this case I think auth context is so critical to think about that we should force devs to choose to explicitly not pass one.

That said, I'm completely fine with `null` instead of `undefined`; it sounds like that would address the root concern, right?

This would apply to the other generator too which is no problem at all.

bickelj: I think so. Is there any pitfall you thought of that is avoided by passing `undefined` instead of `null`? If it only feels better to me to pass `null` and there is some reason to pass `undefined` I can live with it.

slifty: No I think you're right!  I think this is what `null` is for, and it doesn't seem like `undefined` gives any particular benefits (we can still do things like `??` with null just like with undefined)

bickelj (approved):

Issue #1410, opened 2025-01-06
Deployment failures 2025-01-03 and 2025-01-05


bickelj:
This ticket is to investigate the deployment failures on Jan 3 and Jan 5.

bickelj:
First one:
```
[2025-01-03 21:28:04] │ -----> Installing dependencies
[2025-01-03 21:28:04] │        Installing node modules
[2025-01-03 21:28:12] │        
[2025-01-03 21:28:12] │        > postinstall
[2025-01-03 21:28:12] │        > [ ! -d "venv" ] && python3 -m venv venv; ./venv/bin/pip install -r sqlfluff/requirements.txt
[2025-01-03 21:28:12] │        
[2025-01-03 21:28:12] │        The virtual environment was not created successfully because ensurepip is not
[2025-01-03 21:28:12] │        available.  On Debian/Ubuntu systems, you need to install the python3-venv
[2025-01-03 21:28:12] │        package using the following command.
[2025-01-03 21:28:12] │        
[2025-01-03 21:28:12] │            apt install python3.10-venv
[2025-01-03 21:28:12] │        
[2025-01-03 21:28:12] │        You may need to use sudo with that command.  After installing the python3-venv
[2025-01-03 21:28:12] │        package, recreate your virtual environment.
[2025-01-03 21:28:12] │        
[2025-01-03 21:28:12] │        Failing command: /workspace/venv/bin/python3
[2025-01-03 21:28:12] │        
[2025-01-03 21:28:12] │        sh: 1: ./venv/bin/pip: not found
[2025-01-03 21:28:12] │        npm error code 127
[2025-01-03 21:28:12] │        npm error path /workspace
[2025-01-03 21:28:12] │        npm error command failed
[2025-01-03 21:28:12] │        npm error command sh -c [ ! -d "venv" ] && python3 -m venv venv; ./venv/bin/pip install -r sqlfluff/requirements.txt
[2025-01-03 21:28:12] │        npm error A complete log of this run can be found in: /tmp/npmcache.sUAS8/_logs/2025-01-03T21_28_04_967Z-debug-0.log
```

Second one:
```
[2025-01-06 02:53:42] │ -----> Installing dependencies
[2025-01-06 02:53:42] │        Installing node modules
[2025-01-06 02:53:50] │        
[2025-01-06 02:53:50] │        > postinstall
[...]

PR #1417, opened 2025-01-07
Try adding `python3-pip` to `Aptfile`

bickelj:
The previous attempt did not resolve a Digital Ocean deployment failure, `pip` is still not found. It appears the package from `Aptfile` was in fact installed so explicitly install `pip` too.

Issue #1410

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -1,1 +1,2 @@
 python3-venv
+python3-pip


slifty (approved):
WE'RE DOIN' IT LIVE!

bickelj:
I suppose we could do an off-to-the-side example app in Digital Ocean, but given that we have a broken pipeline, I think it is OK to try it here.

PR #1419, opened 2025-01-08
Add python3 to build

slifty:
We're trying to get python to work as expected in the build process in Digital Ocean Apps.  I'm not sure it's actually been installed, so fixing that here.

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -1,2 +1,3 @@
+python3
 python3-venv
 python3-pip


slifty:
Since this is just related to CI and there's no way to verify it is the right solution without merging to main, I'm going to just merge it in without a review.

If this doesn't work then I expect the next PR to simply be removing this file and re-enstating our docker builds, allowing us to move off of DigitalOcean's build patform entirely.

bickelj:
The other two packages would include `python3` transitively, then again I think I see what you mean here, that it's not `python3-minimal`? In any case, it appears to have not solved : (

PR #1420, opened 2025-01-08
Resurrect the docker image build/push to GHCR

bickelj:
Prior to this commit, the Digital Ocean deployment pipeline would pull the new source code, re-run lint, re-run tests, build a docker image, and then deploy. The problem with this approach is that it includes too much, does too much, and is entirely outside our CI/CD steps at the canonical source forge (GH). The problem is revealed when running the `lint` step at Digital Ocean: no Python environment. But then an attempt to add Python to the environment both fails and does not make much sense. There doesn't need to be a Python environment, nor any development environment for that matter, present in the production deployment image. Only the already-linted, already-compiled, and already-tested code should go into production. Instead of trying to remove steps from the DO pipeline, this commit restores a docker image build and push when a merge to main occurs. Note that a controversial token is not used in this step and the inter-repository trigger that previously used such a token not restored.

The approach here is to build an artifact out of components that have already been linted, compiled, and tested, and do not include any development tools along the way. See the `npm prune --omit=dev` command. Note that this does not perfectly meet the objective but further improvements can be done in later commits, for example, making this step contingent on previous workflow steps and declaring `uses` the exact same deployment image in the lint and test workflows.

On the DO App Platform side, there is an option to run a docker image rather than compiling and building one from source. Other benefits come from this approach. The version of node used by DO was not clear until looking deep into logs or running a shell on the production runtime. The version of node was not able to controlled, either. With the approach of running an existing docker image, the full runtime, including node version, is known and controlled. A further benefit is when an issue occurs in production that cannot be easily reproduced with source code, the exact image running in production can be run on a local development environment by pulling from the image repository.

With this commit, the ability to use the option of running a GHCR image on the DO App Platform is present because the image is publicly available.

Issue #1410
https://github.com/PhilanthropyDataCommons/deploy/issues/118 PR #1236

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,45 @@
+name: Build and push docker image
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: philanthropydatacommons/service
+
+on:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  build-and-push-docker-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - run: echo VERSION="$(git log -1 --date=unix --pretty=format:'%cd' | date --utc '+%Y%m%d')"-"$(git log -1 --pretty=format:'%h')" >> "$GITHUB_ENV"
+      - name: Build with node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .node-version
+      - run: npm ci
+      - run: npm run build
+      - run: npm prune --omit=dev
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          platforms: linux/amd64
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.VERSION }}, ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache,mode=max
--- a/Aptfile
+++ b/Aptfile
@@ -1,3 +1,0 @@
-python3
-python3-venv
-python3-pip


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM bitnami/node:22-debian-12
+FROM bitnami/node:22.13.0-debian-12-r0
```
slifty, line 1: Pinning in this level of specificity means we do not benefit from security updates automatically.

I'm not sure the benefit -- the built image that we test against *will* be locked in on a given version at the time of deployment, so locking in simply prevents us from those updates, correct?  If an updated version is used which does not pass tests it would simply not be deployed, right?

Let's keep this as written / not create the dev chore of updating patch and minor versions of debian or node.

slifty (changes requested):
Thank you for this!

There are some action lint flags to be addressed; also have a comment about sticking with just major-version pinning (I think we had landed on major being good enough?)

The other thing is: could you remove the Aptfile in this PR since the PR obviates the need for that file.

bickelj:
Nice catch by actionlint there.

Dockerfile
bickelj: While I still don't agree with the fuzzy versioning, I can abide it here for the reason that "the built image will be locked on an [unknown] version at the time of deployment" and therefore the ability to reproduce production issues using the production image is available. Since that capability is not present today, it's still a step in the right direction from the present to be fuzzy here.

The purpose of locking a version here is not only for reproducibility from production backward, but also from source code forward. Not only should we have the ability to get and run what's in production but I should be able to build what's in production from source twice on different days and expect to get the same results. In this limited case I am OK with changing it back. That should not be misinterpreted as me thinking it's a good idea, though.

bickelj: On a more technical note, I am not sure whether we'll get "automatic" updates by using a fuzzy version here, that remains to be seen. I'm OK with trying it out, though.

.github/workflows/build.yml
```diff
@@ -0,0 +1,45 @@
...
+      contents: read
+      packages: write
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - run: echo VERSION="$(git log -1 --date=unix --pretty=format:'%cd' | date --utc '+%Y%m%d')"-"$(git log -1 --pretty=format:'%h')" >> "$GITHUB_ENV"
+      - name: Build with node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22.13.0
```
slifty, line 25: Not trying to belabor the patch vs major point, but this build step can be on `22` rather than specifying an older version, right? (I know it isn't older now, but it will be soon)

slifty (approved):
Thank you!

Left a question about patch stuff but it's not blocking / I defer to whether you think it's consistent to shift it to Major only vs patch, but if you think having patch here is more important than the other places that's fine by me.

bickelj:
@slifty Good catches. I removed the Aptfile, linting is now passing, and I begrudgingly restored fuzzy node runtime versioning.

.github/workflows/build.yml
bickelj: Also good catch, I can make it match the other flows here, again despite my better judgment.

bickelj:
Something I'd like to have in future (no need to spend the time here and now on it) is to run the docker image build steps on every pull request but only push the image on the merge to main. It might also be nice to run some simple tests on the image after build. Along similar lines mentioned in the commit, make this build step contingent on passage of lint, test, etc., also use the same docker base image that we use to deploy in all these github actions.

Issue #1147, opened 2024-08-14
Improve database utilities to require a transaction / db object


slifty:
Right now all of our database operation functions use the global `db` export instead of taking a db context as a parameter.  This means that the caller is unable to define, for instance, a transaction context when invoking various database transactions.

Let's change the function signatures for all of these operations to require the db context as a parameter.

PR #1421, opened 2025-01-08
Add AuthContext to create / update operations

slifty:
This PR continues the process of updating our db operation utilities to use a generator, and then improving that generator to accept an authentication context.

It also has the generator accept an optional db which means all create / update operations will now support transactions instead of just a few of them.

Related to #1147

### DIFF
--- a/src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
+++ b/src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
@@ -1,0 +1,41 @@
+import { db as defaultDb } from '../../db';
+import type { JsonResultSet } from '../../../types';
+
+// This may seem silly but it is necessary to get all keys of all
+// possible types in the event the type is a Union (e.g. A | B | C)
+type KeysOfUnion<T> = T extends T ? keyof T : never;
+
+/**
+ * Generates a function that will invoke a specific query for data insertion
+ *
+ * @template T - The type of the item being created.
+ * @template I - The type of object to be passed for data insertion.
+ *
+ * @param {string} queryName - The name of the query to execute.
+ * @param {Object} savedAttributes - A list of the attribute names which should be converted to query parameters.
+ *
+ * @returns {Function} A function that takes query parameters, limit, and offset, and returns a promise that resolves to a bundle of entries and total count.
+ */
+ const generateCreateOrUpdateItemOperation =
+	<T, P extends Record<string, unknown>>(
+		queryName: string,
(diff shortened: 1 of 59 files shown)


src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { db as defaultDb } from '../../db';
+import {
+	getIsAdministratorFromAuthContext,
+	getKeycloakUserIdFromAuthContext,
+} from '../../../types';
+import type { AuthContext, JsonResultSet } from '../../../types';
+
+// This may seem silly but simply doing `keyof T` will not account for the possibility of `T` being a union type.
+type KeysOfUnion<T> = T extends T ? keyof T : never;
```
bickelj, line 21: I don't understand but I am not sure I want to either.

src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Generates a function that will invoke a specific query for data insertion
+ *
+ * @template T - The type of the item being created.
+ * @template I - The type of object to be passed for data insertion.
+ *
+ * @param {string} queryName - The name of the query to execute.
+ * @param {Object} savedAttributes - A list of the attribute names which should be converted to query parameters.
+ *
+ * @returns {Function} A function that takes query parameters, limit, and offset, and returns a promise that resolves to a bundle of entries and total count.
+ */
```
bickelj, line 11: This comment should be moved to just above the function it describes so that IDEs can make use of it.

bickelj (approved):
A heroic refactoring, thank you!

It is a little jarring every time I see the `null` as the first argument, but as @slifty mentioned earlier, that's the point. Developers need to be aware of whether these functions need the auth context or not. 'Tis important.

I was able to `build`, `lint`, `migrate`, and `start` without issue. I successfully created an opportunity and funder, each of which should call into the new/modified code.

The only change I request is that the doc block be next to the function it describes.

hminsky2002 (approved):
A heroic refactoring indeed, couldn't find any mistakes in here. Excellent work @slifty

src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
slifty: I tried to update the comment to be more useful (but I agree it's better to just close your eyes sometimes when it comes to TypeScript)

PR #1422, opened 2025-01-09
Trigger a deployment when workflows succeed

bickelj:
The new `deploy` workflow should trigger a deployment of the newly built (in the `build` workflow) image to Digital Ocean App Platform when several previous tasks such as `test` and `build` succeeded.

The `build` task is assumed to have updated the `latest` image tag and therefore the deployment command called by `curl` here should deploy the image just created in most cases. There may still be a chance of interleaving workflow executions causing the unintended image to be deployed. Until a more robust solution is found, this may have to do.

The situation as of writing is that no new code is delivered to production so this commit is a step toward getting an automated deployment working in a test environment and we expect to improve it to deploy to production.

Issue #1410

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,11 @@
+# GitHub Workflow Notes for the Philanthropy Data Commons service
+
+## Deploy
+
+The deploy task requires an API key, secret `DIGITAL_OCEAN_TOKEN`, and an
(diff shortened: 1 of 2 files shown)


slifty (approved):
Makes sense to me, and it looks good (standard caveat of "I'm not sure how to test CI beyond just reviewing the code itself)

bickelj:
Trying it, fingers crossed.

PR #1423, opened 2025-01-09
Improve the deploy script

bickelj:
The previous script would attempt to run several times or not at all. This commit introduces a different barrier action that might work. It also appears that the `workflow_run` might expect names of workflows rather than filenames of workflows, this uses names.

Issue #1410


bickelj:
Trying this sans review because it is currently broken, no harm in trying a new one.

PR #1424, opened 2025-01-09
Lengthen deploy barrier timeout

bickelj:
Given that the tests usually take around six minutes, set a timeout of
several (four) times that, 24 minutes. This should help the barrier
action work (more) properly.

Correct the deploy script's variable substitution

The correct form is like `${{ secrets.VARIABLE }}`.


bickelj:
This is very close to working, of course, any post-merge review is welcome. The situation now is that no code is making it to prod, so getting the deployment pipeline up again is important.

PR #1426, opened 2025-01-09
Improve deploy script

bickelj:
Clean up deploy script
 
Set the environment more simply and correctly, pass secrets through the `env` as documented, explicitly mark the URL in `curl` call.

Correct a parameter to deploy barrier action

Issue #1410


bickelj:
The situation is still this: trying to unblock the deployment pipeline, getting this deploy script to work.

PR #1438, opened 2025-01-13
Deploy to Production Digital Ocean environment

bickelj:
To limit the chances of interleaving workflow runs deploying an
unintended version of PDC, condition the build/push step on successful
completion of lint, test, and sdk tasks.

Use names for steps for friendlier presentation in GH UIs.

Use fuzzy versioning for actions to match our current practice.

Improve deployment workflow documentation

Include production deployment in deploy script

Deploy to production when the call to deploy to test succeeds. Without
this commit, the deployment actions only target the test environment.

While this deployment pipeline does not (yet) validate that the test
environment is working after deployment to test, it at least validates
that the command to deploy succeeded via `needs: deploy-to-test-env`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,7 +9,7 @@ jobs:
   deploy-to-test-env:
     runs-on: ubuntu-latest
     environment: Test
-    name: deploy-to-test
+    name: Deploy to Test environment
     env:
       TOKEN: ${{ secrets.DIGITAL_OCEAN_TOKEN }}
       URL: ${{ vars.DEPLOYMENT_URL }}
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -9,12 +9,13 @@ jobs:
...
     runs-on: ubuntu-latest
     environment: Test
-    name: deploy-to-test
+    name: Deploy to Test environment
     env:
       TOKEN: ${{ secrets.DIGITAL_OCEAN_TOKEN }}
       URL: ${{ vars.DEPLOYMENT_URL }}
     steps:
-      - uses: kachick/wait-other-jobs@v3.6.0
+      - name: Wait for lint, test, build, and other workflows to succeed
```
slifty, line 17: I believe GitHub Actions have built in ways to have workflows trigger other workflows, so you don't need to use a 3rd party tool to accomplish this (and also I think this will give some UI sugar to boot).

Check out https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_runbranchesbranches-ignore

If it does what I am claiming (I could be wrong!) then I think this concept should probably also be applied to `build`.

slifty (changes requested):
I *think* you can avoid using a 3rd party tool and can use `workflow_run` instead, so requesting that change (if I'm wrong I apologize in advance!)

.github/workflows/deploy.yml
bickelj: No, unfortunately `workflow_run` will trigger the workflow N times for N prior jobs, and those will trigger regardless of success of the previous job. This is why there are a plethora of 3rd party tools.

bickelj: Perhaps a better tack would be to unify the jobs in a single workflow such that `needs` could be used, I think that works as expected out of the box unlike `workflow_run`.

PR #1439, opened 2025-01-14
Fix build script JSON

bickelj:
No trailing commas allowed.

Issue #1410


bickelj:
This is a downside to the `kachick/wait-other-jobs@v3` action: there is a JSON object as input and that input is not validated by actionlint.

However, the overall action flow worked properly. When the `build.yml` workflow here failed, the deployment to test workflow failed in turn as it should have. When the deployment to test workflow failed, the deployment to production was skipped as it should be.

PR #1444, opened 2025-01-15
Update Digital Ocean deploy template

bickelj:
This reflects the recent change in the deployment style. The structure matches what is used as of this commit.

Issue #1410

### DIFF
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -1,74 +1,85 @@
-spec:
-  databases:
-    - name: pdc-service-db
-  envs:
-    - key: AUTH_SERVER_ISSUER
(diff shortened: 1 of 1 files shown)


bickelj:
That's a bit of a conundrum. Lint fails on an already-existing, already-deployed, already-working format of a file. Hmm.

bickelj:
I used `npx prettier .do --write` with the actual test and prod yaml files present in that directory and re-uploaded the test file to the test environment. It re-deployed successfully and still shows (when viewing in DO interfaces) the undesired formatting version. So there seems to be no harm in keeping the better-formatted one here in our repo as DO will accept it.

slifty (approved):
Seems reasonable to me!

PR #1448, opened 2025-01-17
Remove seed file

### LINKED ISSUE #1441: Remove Seed file
With the merging of #1259, we no longer need a seedfile in the repository. Populating the database with basefields is now handled by the basefields copy worker

hminsky2002:
Closes #1441 , and rids us finally of the seed file! Woohoo


slifty (approved):
Looks good!  Hooray!

PR #1452, opened 2025-01-21
Create a jest group for dependabot

slifty:
This PR makes it so Dependabot will combine ts-jest and jest dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,9 @@ updates:
       eslint:
         patterns:
           - '*eslint*'
+      jest:
+        patterns:
+          - '*jest*'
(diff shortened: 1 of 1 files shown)


hminsky2002 (approved):
Neato! Are we still thinking about migrating to vitest in the future? I did that for the front-end repo and I believe for many of the standard jest functions it's almost exactly the same

slifty:
@hminsky2002 it wasn't on my immediate radar.  I imagine it won't be quite as straightforward as frontend since we're doing jest specific things here (mocks, teardown, etc).  (that said, make an issue for it and let's discuss over there rather than in this PR!)

PR #1459, opened 2025-01-23
Use generator for update operations

slifty:
This continues the process of using generators for our db operations.  This in turn results in use of `db` and `authContext` parameters.

Related to #1407

### DIFF
--- a/src/__tests__/changemakers.int.test.ts
+++ b/src/__tests__/changemakers.int.test.ts
@@ -17,7 +17,6 @@ import {
 	loadSystemUser,
 	loadTableMetrics,
 	createOrUpdateDataProvider,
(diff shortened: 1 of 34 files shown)


hminsky2002 (approved):
All changes here seem to have been made correctly, I looked for any remaining methods that may have missed a refactor and found none. Approving, great work @slifty

bickelj (approved):
Having skimmed the code and seeing another approval already, this looks fine to me.

I wonder if the new base fields error handling area should have new tests given that it's new code.

slifty:
Ack I'm gonna hit merge since this is approved -- I don't disagree with there being value in tests, but I also know this is needed and I'm sick so don't want to delay.

PR #1468, opened 2025-01-29
Allow updates to changemakers by ID

bickelj:
The purpose of this change is to allow a keycloakOrganizationId to be added to a changemaker after a changemaker has been created. Often a changemaker will be created by bulk upload, for example, prior to the Keycloak organization to be associated with the changemaker.

The exposure of the new endpoint `PATCH /changemakers` allows an administrator, after creating a Keycloak organization in Keycloak, to associate the relevant changemaker in the PDC service.

Issue #1291 Keycloak groups to assist with PDC object permissions

### DIFF
--- a/src/__tests__/changemakers.int.test.ts
+++ b/src/__tests__/changemakers.int.test.ts
@@ -19,7 +19,10 @@ import {
 	createOrUpdateDataProvider,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
-import { mockJwt as authHeader } from '../test/mockJwt';
+import {
+	mockJwt as authHeader,
+	mockJwtWithAdminRole as adminUserAuthHeader,
+} from '../test/mockJwt';
 import {
 	BaseField,
 	BaseFieldDataType,
@@ -31,6 +34,7 @@ import {
 	Opportunity,
 	PostgresErrorCode,
 	Source,
+	stringToKeycloakId,
 	User,
 } from '../types';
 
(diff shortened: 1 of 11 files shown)


src/__tests__/changemakers.int.test.ts
```diff
@@ -849,4 +853,83 @@ describe('/changemakers', () => {
 			});
 		});
 	});
+
+	describe('PATCH /', () => {
```
hminsky2002, line 857: We should probably add a test for when the passed changemaker id does not refer to an existing changemaker

src/openapi.json
```diff
@@ -1819,6 +1836,59 @@
...
+					"required": true,
+					"content": {
+						"application/json": {
+							"schema": {
+								"$ref": "#/components/schemas/PartialChangemaker"
+							}
+						}
+					}
+				},
+				"responses": {
```
hminsky2002, line 1870: similarly, if the referenced changemaker doesn't exist we should expect an error (would that be a 422?)

src/__tests__/changemakers.int.test.ts
hminsky2002, line 857: We should probably add a test to handle the case where the referenced changemaker does not exist in the database

hminsky2002 (changes requested):
This seems good! Left a note about adding handling for the case when the referenced changemaker does not exist. I think a patch as opposed to a put makes sense since we don't want to create a new changemaker in the case it doesn't exist yet

src/database/queries/changemakers/updateById.sql
```diff
@@ -1,4 +1,9 @@
 UPDATE changemakers
-SET keycloak_organization_id = :keycloakOrganizationId
+SET
+	tax_id = coalesce(:taxId, tax_id),
+	name = coalesce(:name, name),
+	keycloak_organization_id = coalesce(
```
bickelj, line 5: An unfortunate side effect of the intersection of `coalesce` and the rest of our code is that we cannot nullify an already-set keycloak organization ID here. If we literally pass `null` as the desired ID, `coalesce` leaves the existing UUID. We can't distinguish "I'm setting this as `null`" from "I'm not setting this at all."

bickelj: I think this is something to be worked out later, though, as we have a workaround where it is a limited use case where an organization gets removed and if needed we could set the keycloak_organization_id to an all-zeroes UUID.

Created #1472 to track.

src/openapi.json
bickelj: Fixed, and other adjustments made in the codebase to make this possible. Thank you!

hminsky2002 (approved):
Looks good to me, approving!

PR #1473, opened 2025-01-31
UserChangemakerPermission required fields in doc

bickelj:
Set the required fields in OpenAPI spec for UserChangemakerPermission.

Issue #1291 Keycloak groups to assist with PDC object permissions

### DIFF
--- a/src/openapi.json
+++ b/src/openapi.json
@@ -1272,7 +1272,13 @@
 						"readOnly": true
 					}
 				},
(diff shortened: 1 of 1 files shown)


slifty (approved):
Wooo!  I kept seeing those required fields and thinking "... that doesn't look right" so I'm very glad this is finally fixed up.  Thank you

bickelj:
da96818 was only a rebase on main.

PR #1474, opened 2025-01-31
Add documentation on PDC data permissions

bickelj:
Issue #1291

### DIFF
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -1,0 +1,157 @@
+# Permissions
+
+The Philanthropy Data Commons (PDC) uses [Keycloak] to manage users and groups.
+Within the PDC service (not PDC Keycloak), permissions can be granted on data to
+particular users and groups by reference to their respective Keycloak UUIDs.
+
+Keycloak implements open standards for authentication and authorization such as
+[OpenID Connect] which in turn uses a variety of other open standards. Much of
+the terminology in this document comes from such standards or Keycloak and is
+beyond the scope of this document. The focus of this document is to provide an
+overview of data permissions within the PDC and how to grant or revoke such
+permissions.
+
+[Keycloak]: https://www.keycloak.org/documentation
+[OpenID Connect]: https://openid.net/developers/how-connect-works/
+
+## Example Workflow
+
+This is a high level overview of a workflow whereby a changemaker gains the
+ability to grant permissions on its data to other parties. Background details
+and how to perform most of these steps follow in this document.
+
+1. Funder bulk uploads changemakers and corresponding changemaker proposals.
+2. One of the above changemakers requests access to the PDC.
+3. PDC admin adds Keycloak organization for changemaker.
+4. PDC admin adds Keycloak users (if needed) for changemaker.
+5. PDC admin sets Keycloak organization membership (if needed) for changemaker.
+6. PDC admin links the new Keycloak organization to PDC service using its UUID.
+7. PDC admin grants `manage` permission to changemaker using PDC service.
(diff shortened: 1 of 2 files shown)


docs/PERMISSIONS.md
```diff
@@ -0,0 +1,139 @@
...
+by logging into the account management interface e.g. URL `realms/pdc/account`.
+
+Administrators of the PDC Keycloak realm can see organizations, respective
+organization UUIDs, and organization members by logging into the Keycloak admin
+interface, visiting the PDC realm, selecting "Organizations", and selecting an
+organization. The organization UUID appears in the URL.
+
+[Keycloak organizations]: https://www.keycloak.org/docs/26.1.0/server_admin/index.html#_managing_organizations
+
+## How to Link Keycloak Organizations to PDC Service Entities
```
slifty, line 99: What happens when this association is formed / what does actually mean?  It seems like with `userGroups` we don't actually need to know the direct associations any longer (since permissions will be granted directly).

If there isn't a direct use case I'm wondering if we should remove this part of the documentation (and even the functionality?).

slifty (approved):
This looks really really good thank you for documenting it so clearly.

I had one question about the semantic linking between keycloak organization and funder / changemaker / data provider in the PDC.  If there is no use case for that right now we might consider removing it from the documentation entirely to make it clearer that the way you assign permissions to members of a keycloak organization is via PDC userGroups

docs/PERMISSIONS.md
bickelj: Use cases that come to mind for linking funders/changemakers/dataproviders to their respective Keycloak Organizations.

1. User Interfaces where the users do not know anything about Keycloak but know that they would like to grant Funder F view access to their resources. By having the link present in PDC Service, this is straightforward. Funder F is Keycloak Organization O.
2. Validation that permissions are granted only to known Keycloak Organization UUIDs, much in the same way that there is validation of Keycloak User UUIDs to only be those that have been seen. One reason it's not present in #1477 is we have three entities to search rather than one so you can't have a foreign key that searches three tables, it will have to be an application-level validation.
3. Implicit permissions for a changemaker. We may want to have some basic level of permission, whatever that may be, automatically granted to an organization (or a group within that organization). For example, rather than the PDC admin explicitly granting `view` access on changemaker C data to nebulous/unknown flat list of Keycloak organizations among which is the actual changemaker, the PDC admin associates Changemaker C with Keycloak Organization O, and members of Changemaker C automatically have some kind of access to their own proposal data.

I can see the view that with our current design there is no link strictly necessary because we're granting access to the UUID, not the funder/changemaker/dataprovider. Technically, I guess we could make it work, and it might be more explicit in that you'd have to have a PDC Admin grant your group access like any other. We could have an unvalidated, opaque UUID that is presumably a Keycloak UUID but the PDC service would not know anything about it. Users would then have to ask a PDC admin for the current list of Keycloak Organizations (or members of that group) in order to have the power to grant another group access. Heck, they might even have to ask a PDC admin for their own Keycloak Organization UUID, because not all of them would have access to the Keycloak Admin interface.

What benefits would we get by *not* associating them and would those outweigh these?

bickelj: Thanks and yes. I think these are all in the realm of potentials. Something that occurred to me is that some of this might apply to users as well. Although for users all the data (e.g. first and last name) is exclusively in Keycloak whereas the relationship of funder to org can be exclusively in the database. Let us chat.

bickelj: I added a contextual paragraph justifying setting the link within the PDC service rather than not setting and rather than having the link set in Keycloak.

PR #1477, opened 2025-02-03
Create `UserGroupPermissions` entities

### LINKED ISSUE #1469: Create organizationPermission entities
Related to #1291 

We've decided that for the time being, we're going to approach organization permissions similarly to how we already handle userPermissions. Per our discussion in #1291, we've settled on adding entities and handlers for organization permissions that copy the pattern for userPermissions, (e.g a `organization_changemaker_permissions` table to match `user_changemaker_permissions`). Separate entities will avoid having to distinguish between user and organization keycloak ids at entity creation. Instead we will handle organization permissions through an `/organizations` route. Unlike user permissions, organization permissions will not reference the users table, and permission handling will be handled by simply checking all permissions for a matching organization keycloak id.

hminsky2002:
Closes #1469 

This PR brings in userGroup Permissions entities, which follow the pattern for existing userPermissions entities. These are meant to allow organizations to provide users with permissions for managing other entities in the pdc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.16.7 2025-02-05
+
 ### Added
 
 - Manage user-funder permissions using `PUT`, and `DELETE` on `/user/{keycloakUserId}/funders/{funderShortCode}/permissions/{permission}`.
 - Manage user-data-provider permissions using `PUT`, and `DELETE` on `/user/{keycloakUserId}/funders/{funderShortCode}/permissions/{permission}`.
-
-### Added
-
+- Manage user-group-changemaker permissions using `PUT`, and `DELETE` on `/userGroups/{keycloakOrganizationId}/changemaker/{changemakerId}/permissions/{permission}`.
+- Manage user-group-data-provider permissions using `PUT`, and `DELETE` on `/userGroups/{keycloakOrganizationId}/dataProvider/{dataProviderShortCode}/permissions/{permission}`
+- Manage user-group-funder permissions using `PUT`, and `DELETE` on `/userGroups/{keycloakOrganizationId}/funders/{funderShortCode}/permissions/{permission}`
 - Manage user-funder permissions using `PUT`, and `DELETE` on `/user/{keycloakUserId}/funders/{funderShortCode}/permissions/{permission}`.
 
 ## 0.16.6 2024-12-23
--- a/src/__tests__/userGroupChangemakerPermissions.int.test.ts
+++ b/src/__tests__/userGroupChangemakerPermissions.int.test.ts
@@ -1,0 +1,316 @@
+import request from 'supertest';
+import { app } from '../app';
+import {
+	db,
+	createChangemaker,
+	createOrUpdateUserGroupChangemakerPermission,
+	loadUserGroupChangemakerPermission,
+	removeUserGroupChangemakerPermission,
+} from '../database';
+import { expectTimestamp, loadTestUser } from '../test/utils';
+import {
+	mockJwt as authHeader,
+	mockJwtWithAdminRole as authHeaderWithAdminRole,
+} from '../test/mockJwt';
+import { KeycloakId, keycloakIdToString, Permission } from '../types';
+import { NotFoundError } from '../errors';
+
+const mockKeycloakOrganizationId = '123e4567-e89b-12d3-a456-426614174000';
+
+describe('/userGroups/changemakers/:changemakerId/permissions/:permission', () => {
+	describe('PUT /', () => {
+		it('returns 401 if the request lacks authentication', async () => {
+			const organizationAsChangemaker = await createChangemaker(db, null, {
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+				keycloakOrganizationId: mockKeycloakOrganizationId,
+			});
+			const changemaker = await createChangemaker(db, null, {
+				taxId: '11-1111112',
+				name: 'Example Inc.',
+				keycloakOrganizationId: null,
+			});
+			await request(app)
+				.put(
+					`/userGroups/${keycloakIdToString(organizationAsChangemaker.keycloakOrganizationId as KeycloakId)}/changemakers/${changemaker.id}/permissions/${Permission.MANAGE}`,
+				)
+				.send({})
+				.expect(401);
+		});
+
+		it('returns 401 if the authenticated user lacks permission', async () => {
+			const organizationAsChangemaker = await createChangemaker(db, null, {
+				taxId: '11-1111111',
+				name: 'Example Inc.',
+				keycloakOrganizationId: mockKeycloakOrganizationId,
+			});
+
+			const changemaker = await createChangemaker(db, null, {
+				taxId: '11-1111112',
+				name: 'Example Inc.',
+				keycloakOrganizationId: null,
+			});
+			await request(app)
+				.put(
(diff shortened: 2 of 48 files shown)


slifty:
@hminsky2002 there's a typo in a test you should fix up when you get a minute (I'll still review the rest since that's small!)

slifty (changes requested):
Meta comment before I get into the meat of the review: these commits are organized so that each sub-step of the *entire* set of permission features is added over four parts.

I'd suggest inverting that, so that each commit provides a specific complete -- and completely documented -- feature.

This would mean one commit for userGroupChangemakerPermission, one for userGroupFunderPermission, and one for userGroupDataProviderPermission.

This means the context of, for instance, the swagger specification is paired with the code and tests for that endpoint.  It also means that there's a commit where DB operations have been created even though they aren't used anywhere, which means it's dead code (even if that's only for a short period of time).

src/database/queries/userGroupChangemakerPermissions/deleteOne.sql
```diff
@@ -0,0 +1,7 @@
+UPDATE user_group_changemaker_permissions
+SET not_after = now()
+WHERE
+	keycloak_organization_id = :keycloakOrganizationId
+	AND permission = :permission::permission_t
```
slifty, line 5: I realize this is copy pasta but why do we cast to permission_t here (and elsewhere) and not everywhere.

I'd say lets not cast unless we have to.

slifty (changes requested):
Looking good -- a few little changes (plus the larger commit organization change request from earlier)

bickelj:
> Meta comment before I get into the meat of the review: these commits are organized so that each sub-step of the _entire_ set of permission features is added over four parts.
> 
> This means the context of, for instance, the swagger specification is paired with the code and tests for that endpoint. It also means that there's a commit where DB operations have been created even though they aren't used anywhere, which means it's dead code (even if that's only for a short period of time).
> 
> I'd suggest inverting that, so that each commit provides a specific complete -- and completely documented -- feature.
> 
> This would mean one commit for userGroupChangemakerPermission, one for userGroupFunderPermission, and one for userGroupDataProviderPermission.

Another solution, given the circumstances, is to squash it all.

src/database/migrations/0047-create-user-group-permissions-tables.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE user_group_changemaker_permissions (
+	keycloak_organization_id uuid NOT NULL,
+	permission permission_t NOT NULL,
```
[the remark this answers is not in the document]
bickelj: Maybe it's the English-speaker in me: I still like the subject verb object order. On the other hand, Latin and Yoda good are.

src/database/operations/userGroupChangemakerPermissions/createOrUpdateUserGroupChangemakerPermission.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '../../../types';
+
+const createOrUpdateUserGroupChangemakerPermission =
+	generateCreateOrUpdateItemOperation<
+		UserGroupChangemakerPermission,
+		InternallyWritableUserGroupChangemakerPermission,
+		[]
+	>(
+		'userGroupChangemakerPermissions.insertOrUpdateOne',
+		['keycloakOrganizationId', 'permission', 'changemakerId', 'createdBy'],
```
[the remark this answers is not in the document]
slifty: https://github.com/PhilanthropyDataCommons/service/issues/1479

src/types/UserGroupChangemakerPermission.ts
```diff
@@ -0,0 +1,41 @@
+import { ajv } from '../ajv';
+import { Permission } from './Permission';
+import type { JSONSchemaType } from 'ajv';
+import type { Writable } from './Writable';
+import type { KeycloakId } from './KeycloakId';
+
+interface UserGroupChangemakerPermission {
+	readonly keycloakOrganizationId: KeycloakId;
+	readonly permission: Permission;
```
bickelj, line 9: While order is technically not significant here, for consistency this should be in the Latin/Cherokee/Yoda order of subject-object-verb.

src/openapi.json
```diff
@@ -3344,6 +3443,309 @@
...
+						"in": "path",
+						"required": true,
+						"schema": {
+							"type": "string",
+							"format": "uuid"
+						}
+					},
+					{
+						"name": "dataProviderShortCode",
+						"description": "The shortCode of a funder.",
```
bickelj, line 3672: This should be "data provider" (it was also incorrect in the original/other entities).

src/openapi.json
```diff
@@ -3344,6 +3443,309 @@
...
```
bickelj, line 3725: Likewise here, should be "data provider."

Maybe add the suggested language from https://github.com/PhilanthropyDataCommons/service/pull/1478/files as well if you think that's useful.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Manage user-funder permissions using `PUT`, and `DELETE` on `/user/{keycloakUserId}/funders/{funderShortCode}/permissions/{permission}`.
 - Manage user-data-provider permissions using `PUT`, and `DELETE` on `/user/{keycloakUserId}/funders/{funderShortCode}/permissions/{permission}`.
+- Manage user-group-changemaker permissions using `PUT`, and `DELETE` on `/userGroups/{keycloakOrganizationId}/changemaker/{changemakerId}/permissions/{permission}`.
+- Manage user-group-data-provider permissions using `PUT`, and `DELETE` on `/userGroups/{keycloakOrganizationId}/dataProvider/{dataProviderShortCode}/permissions/{permission}`
+- Manage user-group-funder permissions using `PUT`, and `DELETE` on `/userGroups/{keycloakOrganizationId}/funders/{funderShortCode}/permissions/{permission}`
```
bickelj, line 16: You reminded me that I forgot to update the changelog for some other entities, whoops! Thank you!

My following comment is not particular to this pull request, but more for the team in general. I don't think it makes sense to put stuff under "unreleased" at this stage of the PR, because once it gets merged, it will be released. So I suggest incrementing the patch or more likely minor version (how do we decide, @slifty?) in the OpenAPI spec and guessing at a date here, maybe today. If it turns out you are going to smash "merge" tomorrow, then update the date before doing so.

bickelj (changes requested):
@hminsky2002 This is excellent. I was able to `lint`, `build`, `migrate`, and `start` without issue. When I tried out the endpoints locally via the OpenAPI pages, they all appeared to return the expected responses and cross-checking the database contents all looked correct too. I noted a couple of minor changes above. I confess I did not read through the code in detail, but skimming shows that it appears to match the original patterns. Thank you again for your help on this! If you would like me to push any of these suggested changes let me know.

bickelj (approved):

bickelj:
@hminsky2002 Thank you once again. We can resolve the changelog separately, I think.

bickelj:
@hminsky2002 The changelog and version update looks great to me!

src/__tests__/userGroupChangemakerPermissions.int.test.ts
```diff
@@ -0,0 +1,316 @@
...
+			const changemaker = await createChangemaker(db, null, {
+				taxId: '11-1111112',
+				name: 'Example Inc.',
+				keycloakOrganizationId: null,
+			});
+			await request(app)
+				.put(
+					`/userGroups/${keycloakIdToString(organizationAsChangemaker.keycloakOrganizationId as KeycloakId)}/changemakers/${changemaker.id}/permissions/${Permission.MANAGE}`,
+				)
+				.send({})
```
bickelj, line 37: I apparently missed this when reviewing before. I don't think we should be required to send an empty body. @slifty what do you think?

Issue #1485, opened 2025-02-06
Support fiscal sponsor relationships between changemakers in our entity model


bickelj:
The goal is to formally model who fiscally sponsors whom. For the purposes of the PDC backend, we will constrain this relationship to be between changemakers only. This does not imply, however, that a real-life organization cannot wear multiple hats in the PDC. In other words, it is conceivable that a real life organization "Change the World LLC" could be primarily a funder but on occasion a grant-seeking organization as well, in which case PDC would model this with one row in the funders table and one row in the changemakers table.

Because a grant-seeker may have multiple fiscal sponsors and a fiscal sponsor may sponsor multiple grant-seekers, we will model this appropriately, with a many-to-many relationship between changemakers in a separate table.

bickelj:
Given that the PDC aims to be changemaker-centric, the best API modeling approach is to treat sponsorship as an attribute of the sponsee, the grant recipients, which we typically call changemakers. This means that permissions for adding and removing sponsor data are tied to the sponsee rather than sponsor in the initial implementation. I think that best fits the approach of empowering even the smallest changemaker, and I would guess that in a sponsor-sponsee relationship the sponsee is typically the smaller of the two changemakers.

slifty:
@bickelj is this resolved now?

bickelj:
@slifty Yes.

PR #1490, opened 2025-02-07
Add ability to link fiscal sponsors to changemaker

bickelj:
A fiscal sponsor in PDC terms is another changemaker. To add a fiscal sponsor to a changemaker, call the new endpoint `PUT /changemakers/{changemakerId}/fiscalSponsors/{fiscalSponsorChangemakerId}` To remove the relationship, call the corresponding `DELETE` endpoint.

The `GET /changemakers/{changemakerId}` endpoint now returns a list of fiscal sponsors. The list is composed of shallow changemakers to avoid a technical issue of potential infinite recursion. So the caller will not see the `fields` or `fiscalSponsors` within the `fiscalSponsors` returned by `GET /changemakers/{changemakerId}`. To get the detailed data on a changemaker, simply call the same `GET` with the sponsor ID.

There is some ambiguity about whether this explicitly modeled fiscal sponsorship relationship in PDC is intended to be only current relationships or both current and past relationships. Assume for now that it is both current and past, reserving `DELETE` for mistaken data entry.

Circular sponsorship graphs are supported but not direct self-sponsorship cycles.

Issue #1485 Support fiscal sponsor relationships between changemakers


src/database/initialization/changemaker_to_json.sql
```diff
@@ -7,7 +7,19 @@ CREATE FUNCTION changemaker_to_json(
...
+  fiscal_sponsors_json JSONB;
 BEGIN
+  -- Terminate changemaker recursion by passing null here. This also means no deep fields on sponsors.
+  SELECT jsonb_agg(changemaker_to_json(fiscal_sponsors.*, NULL))
+  INTO fiscal_sponsors_json
+  FROM changemakers AS fiscal_sponsors
+  INNER JOIN fiscal_sponsorships fs
+    ON fs.fiscal_sponsor_changemaker_id = fiscal_sponsors.id
+    AND fs.fiscal_sponsee_changemaker_id = changemaker.id
+    AND (fs.not_after IS NULL or fs.not_after > now())
```
slifty, line 19: Use the SQL helper function we have `is_expired(fs.not_after)`

so `NOT is_expired...`

src/database/operations/fiscalSponsorships/createOrUpdateFiscalSponsorship.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * A type designed to deal with the backend DB implementation and our generator
+ * functions that probably does not need to be used elsewhere in the code. It
+ * has two Ids that will be validated as Ids before ever getting to this point.
+ * The type is only here and not exported because it is used twice here. If it is
+ * found to be useful elsewhere, hoist it out to our general `types` directory
+ * and export it from there following the usual pattern.
+ */
+interface FiscalSponsorship {
+	fiscalSponseeChangemakerId: Id;
```
slifty, line 14: You probably considered this already but I could see `sponsee` being an unnecessary descriptor since we have `fiscalSponsor` as a descriptor of the other field.

In either case you also might consider removing `fiscal` in both places since the context of *fiscal* sponsorship is already established by the table name.

(these are both "thought" level feedback, please don't consider this a change request if you don't agree)

src/handlers/changemakersHandlers.ts
```diff
@@ -162,9 +165,92 @@ const patchChangemaker = (
...
+	res: Response,
+	next: NextFunction,
+): void => {
+	if (!isAuthContext(req)) {
+		next(new FailedMiddlewareError('Unexpected lack of auth context.'));
+		return;
+	}
+	const { changemakerId, fiscalSponsorChangemakerId } = req.params;
+	if (!isId(changemakerId)) {
+		next(
```
slifty, line 179: This is a testable case so probably worth covering it.

src/handlers/changemakersHandlers.ts
```diff
@@ -162,9 +165,92 @@ const patchChangemaker = (
...
+			new InputValidationError(
+				'Invalid changemakerId parameter.',
+				isId.errors ?? [],
+			),
+		);
+		return;
+	}
+	if (!isId(fiscalSponsorChangemakerId)) {
```
slifty, line 187: Likewise this is a testable case

src/handlers/changemakersHandlers.ts
```diff
@@ -162,9 +165,92 @@ const patchChangemaker = (
...
+};
+
+const deleteChangemakerFiscalSponsor = (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+): void => {
+	const { changemakerId, fiscalSponsorChangemakerId } = req.params;
+	if (!isId(changemakerId)) {
+		next(
```
slifty, line 219: Let's add a test for this

src/handlers/changemakersHandlers.ts
```diff
@@ -162,9 +165,92 @@ const patchChangemaker = (
...
+			new InputValidationError(
+				'Invalid changemakerId parameter.',
+				isId.errors ?? [],
+			),
+		);
+		return;
+	}
+	if (!isId(fiscalSponsorChangemakerId)) {
+		next(
```
slifty, line 228: And a test for this as well!

src/types/Changemaker.ts
```diff
@@ -12,6 +12,10 @@ interface Changemaker {
 	// https://github.com/ajv-validator/ajv/issues/2283 and/or
 	// https://github.com/ajv-validator/ajv/issues/2163.
 	keycloakOrganizationId: KeycloakId | null | undefined;
+	// For permissions purposes, treat sponsors as attributes of grant-seekers.
```
slifty, line 15: I'm not sure we should document this here because I think it risks becoming stale -- if the behavior of how this entity is constructed changes I don't think that we would think to update this comment, since the implementation of this logic is deeper down stream / in the SQL.

On the other end: I don't think that the end user would benefit from this documentation / it should probably go in the swagger specification if we do feel it is worth noting.

--- 

More importantly, this comment actually signals that the maybe the code itself isn't being clear / honest, and we're relying on a comment to mitigate that debt.

What do you think of, instead of having this comment, we make the documentation explicit in the type definition?

The way I think this would work is you would have a `ShallowChangemaker` interface defined above, and then you would define `Changemaker` to extend `ShallowChangemaker` and include `fiscalSponsors: ShallowChangemaker[]` plus whatever other "deep" attributes a Changemaker has that a shallow one lacks.

This should also then be reflected in the swagger spec (which will then be more complete!).

src/database/initialization/changemaker_to_json.sql
```diff
@@ -7,7 +7,19 @@ CREATE FUNCTION changemaker_to_json(
 RETURNS jsonb AS $$
 DECLARE
   proposal_field_values_json JSONB;
+  fiscal_sponsors_json JSONB;
 BEGIN
+  -- Terminate changemaker recursion by passing null here. This also means no deep fields on sponsors.
+  SELECT jsonb_agg(changemaker_to_json(fiscal_sponsors.*, NULL))
```
slifty, line 13: I alluded to this on call but yeah I do think this is pretty unintuitive -- why is it that a null auth context prevents recursion?  It feels like it's just a circumstance of some unrelated business logic decision.  Explicit would be safer and clearer!

One approach that feels right for now to me would be to have an additional optional boolean of "shallow" that defaults to false, this also would align with the more-transparent pattern of defining a `ShallowChangemaker` type in our swagger / type definitions, and we ensure that this function adheres to that specification if the bool is true.

slifty (changes requested):
This looks great; a few suggestions!

src/database/initialization/changemaker_to_json.sql
bickelj: After seeing your PR a moment ago, I thought you might mention it here, you're right!

src/types/Changemaker.ts
bickelj: Oh, "permissions purposes" here doesn't have to do with deep or shallow retrieval. It is trying to say "instead of having a separate API endpoint of `/fiscalSponsorships` or instead of showing `fiscalSponsees`, the model is changemaker-centric." Maybe I should say that. The OpenAPI spec doesn't need to justify the modeling decision. And this comment needs clarification, agree.

src/database/initialization/changemaker_to_json.sql
bickelj: I suppose you're right. Sigh.

src/database/operations/fiscalSponsorships/createOrUpdateFiscalSponsorship.ts
bickelj: I started there but changed it to include `fiscal` to be crystal clear in all contexts. Especially since I didn't expose a `FiscalSponsorship` type and instead pass the IDs as args, in that context it is nice to have the parameter names match the database columns and so forth. It's verbose but I still think it's better to keep it.

src/handlers/changemakersHandlers.ts
```diff
@@ -26,6 +28,7 @@ import {
 	extractProposalParameters,
 } from '../queryParameters';
 import type { Request, Response, NextFunction } from 'express';
+import {} from '../database/operations/fiscalSponsorships/createOrUpdateFiscalSponsorship';
```
bickelj, line 31: Hrm? What if I remove this?

src/types/Changemaker.ts
slifty: This comment was meant in reference to the whole block -- basically, what I mean is that we shouldn't have to explain that fiscalSponsors is not actually an array of `Changemaker[]` but in fact is an array of `ShallowChangemaker[]`

What I'm getting at is that our type would be more accurate if we created two types of entity.

Making the type more precise would, for instance, mean that a developer would never try to say `changemaker.fiscalSponsors[0].fiscalsponsors` because that entity would simply not exist (and the type would reflect it).

bickelj: Right, agree, that's in line with your below comment. I still need to clarify my comment here, or move it, because it didn't have anything to do with all that.

slifty: Ahhh got it (sorry!)

src/handlers/changemakersHandlers.ts
bickelj: Nothing bad happens.

src/database/initialization/changemaker_to_json.sql
bickelj: The more I consider this though, the more I think this is shoe-horning or rushing a more important design decision debate into this PR that is kind of unrelated. It's not unrelated but the pattern is already here of bringing non-Changemaker fields into an actual Changemaker. In other words, I think of those shallow attributes as defining a Changemaker, while bringing along ProposalFieldValues and so forth as enriching or extending a DeepChangemaker. So I might rather rename Changemaker to DeepChangemaker or ChangemakerDeep rather than call what is actually a Changemaker a ShallowChangemaker. Perhaps we can come to quick consensus but again we have gone back and forth over it in the past.

Also, the stuff returned is in fact a legal Changemaker, no one should ever ask for `changemaker.fiscalSponsors[0].fiscalsponsors` because you should iterate the list which could always be empty. There is no guarantee of values in `fields` even if you ask for deep fields.

I can see the counter argument that "fiscal sponsors are changemakers so it is more urgent." That is fair too.

src/database/initialization/changemaker_to_json.sql
bickelj: Nice catch, thanks, fixed.

src/types/Changemaker.ts
bickelj: Oh, and now I see the two lines below that comment, which can be removed, my mistake!

bickelj: I removed the comment and moved that first comment into paragraph form in the commit message. Perhaps that is the most appropriate place for counterfactuals. I have a note in the issue about it already but I think in-band is better.

src/database/initialization/changemaker_to_json.sql
bickelj: @slifty  On reflection and having reviewed https://github.com/PhilanthropyDataCommons/service/issues/174, I think in our current mode of "enrich some `GET` endpoints with useful depth and breadth by default" we could add a `shallow` parameter and have it be `false` by default. That would mean "shallow" is the special snowflake, we keep `Changemaker` as the enriched/broad/deep/high-level type, and as-needed create `ChangemakerShallow` or `ShallowDepthChangemaker` or `EssentialChangemaker` or whatever as requested. I think I can live with this pattern and it does not conflict with the comments in the previous issue regarding menu orders. I'll start doing it now.

bickelj: Oh, and for now, we only need to have the `shallow` parameter used at the lower levels, such as the database functions, rather than at the higher level of the API.

bickelj: I start with ShallowChangemaker now, as @slifty suggested and then build Changemaker from that, both in our types and also in our openAPI types. We still have to manually construct the verification objects in tests. That's OK. It should be refactored now. Not as well tested manually but it's there.

hminsky2002 (approved):
It looks like all changes posed by @slifty have been addressed, and I can confirm a successful `build`, `lint`, and `run` on local machine, so adding my approval. Nice work @bickelj!

bickelj:
@hminsky2002 Thanks for the review! I made two minor changes after your review:
1. Updated the changelog to mention the addition of `ShallowChangemaker` and addition of `fiscalSponors` to `Changemaker` (kind of an important change for existing consumers).
2. Added a test for "delete fiscalsponsorship on non-existing row returns 404".

I think these are minor enough to not need re-review, so assuming tests pass, I'll merge. Of course if you notice anything now or later, everything is always up for review and correction.

Issue #700, opened 2024-01-05
Explore a better pattern for async express handlers


slifty:
(This issue is created from [a PR conversation](https://github.com/PhilanthropyDataCommons/service/pull/679#discussion_r1442444855))

Express 4.x does not support asynchronous handlers, which means if our handlers rely on promises we have been using `then` and `catch` directly.  This is OK if only one promise has to be run in sequence, but it can be limiting if we have more than one asynchronous operation that has to occur as part of a handler!

This is why `async` / `await` syntax exists.  The issue thread talks about some solutions, which might involve an async wrapper function or might involve upgrading to express 5.x which is currently in beta but supports asynchronous handlers.

slifty:
Express 5 seems very close to release, so we may be able to just migrate to address this!

PR #1491, opened 2025-02-10
Store ephemeral user group associations

### LINKED ISSUE #1483: Create user-group associations "cache" middleware
A user's "group" associations are provided by JWT.  We don't want to have to pass a long list of user group IDs to every query so instead we want middleware which will compare the JWT list with a database-level mirror that is intended to reflect that list.  The middleware would be responsible for making sure the database mirror is sync'd with the jwt.

slifty:
This PR expands our auth middleware to store a user's group associations as part of the process of establishing the authentication context.

It does not actually use those associations when populating a user's roles (This will be in a future PR).

Resolves #1483


src/middleware/addUserContext.ts
```diff
@@ -44,13 +43,41 @@ const addUserContext = (
 		return;
 	}
 
-	selectOrCreateUser(stringToKeycloakId(keycloakUserId))
-		.then((user) => {
-			(req as AuthenticatedRequest).user = user;
-			next();
+	Promise.all([
```
slifty, line 46: This entire function is nasty right now because we don't have `async` middleware and cannot use `await` as a result.

I started to refactor to use `async` but determined that it was a complex enough proposition to warrant its own issue and PR.  This is related to #700 which I think might be worth taking on sooner than later.

slifty:
I'm adding a test to ensure that the rows are properly inserted by the middlware

hminsky2002:
> I'm adding a test to ensure that the rows are properly inserted by the middlware

One of my comments was going to be adding some amount of tests so I will wait for your updates to give final review!

src/types/express/AuthenticatedRequest.ts
```diff
@@ -20,6 +21,12 @@ interface ObjectWithAuthWithRealmAccessRoles {
 	};
 }
 
+interface ObjectWithAuthWithOrganizations {
+	auth: {
+		organizations: Record<string, { id: KeycloakId }>;
```
bickelj, line 26: I was thinking this should be a list, but you're right, Keycloak does it as an object with organization short names as keys in an object.

src/database/operations/ephemeralUserGroupAssociations/createEphemeralUserGroupAssociation.ts
```diff
@@ -0,0 +1,17 @@
...
+	EphemeralUserGroupAssociation,
+	InternallyWritableEphemeralUserGroupAssociation,
+} from '../../../types';
+
+const createEphemeralUserGroupAssociation = generateCreateOrUpdateItemOperation<
+	EphemeralUserGroupAssociation,
+	InternallyWritableEphemeralUserGroupAssociation,
+	[]
+>(
+	'ephemeralUserGroupAssociation.insertOne',
```
bickelj, line 12: Likely should be `ephemeralUserGroupAssociations.insertOne` (plural).

src/database/queries/ephemeralUserGroupAssociations/insertOne.sql
```diff
@@ -0,0 +1,15 @@
+INSERT INTO ephemeral_user_group_associations (
+	user_keycloak_user_id,
+	user_group_keycloak_organization_id
+) VALUES (
+	:userKeycloakUserId,
+	:userGroupKeycloakOrganizationId
+)
+ON CONFLICT (
+	user_keycloak_user_id, permission, user_group_keycloak_organization_id
```
bickelj, line 9: There is no `permission` column available here.

src/database/queries/ephemeralUserGroupAssociations/insertOne.sql
```diff
@@ -0,0 +1,15 @@
...
+) DO UPDATE
+SET user_keycloak_user_id = user_keycloak_user_id
```
bickelj, line 11: Ambiguous column error here.

bickelj (changes requested):
Overall, as awkward as it is, I still strongly support leaving users and groups up to PDC Keycloak while managing PDC service object permissions within the PDC service.

The ephemeral table approach is a good starting point. I wonder, though, if we are going from JSON to SQL rows back to JSON again unnecessarily, and if we can do something more JSONy with postgres in the future. For example, if we ephemerally store the whole validated JWT (maybe sans signature), then we have a single thing to pass to postgres queries instead of multiple things. This idea only occurred to me while reviewing here and I haven't tried writing anything that would make use of such a JWT in postgres. So it may not slay the complexity dragon but merely move it into the queries that make use of the JWT.

In any case, there are some plain old errors or typos that need to be fixed before this can be merged. See inline.

bickelj:
I am kind of surprised that so many tests pass since not a single authenticated endpoint works on the Swagger UI. I guess that means we don't integrate our mandatory middleware into those tests. Maybe we are missing a couple of tests that capture that, though I don't know how difficult that would be with supertest and jest.

slifty:
> I am kind of surprised that so many tests pass since not a single authenticated endpoint works on the Swagger UI. I guess that means we don't integrate our mandatory middleware into those tests. That cannot be true because we have to use auth headers to do anything in the software. ~Maybe we are missing a couple of tests that capture that, though I don't know how difficult that would be with supertest and jest.~

It's because I didn't update our tests to actually include an ephemeral organization association (and as you pointed out most of the code is broken) WHOOPS!

src/middleware/addUserContext.ts
```diff
@@ -44,13 +43,41 @@ const addUserContext = (
...
+		createOrUpdateUser(db, null, {
+			keycloakUserId: stringToKeycloakId(keycloakUserId),
+		}),
+		removeEphemeralUserGroupAssociationsByUserKeycloakUserId(
```
bickelj, line 50: I don't think we should remove here. The reason is other queries might be in flight and silently fail due to groups not being present. That raises the question, then, of when/where/how to remove stale groups. Hmm.

bickelj: Maybe an expiration would solve it. Then the remove operation can safely run in a blanket manner (i.e. remove all rows where notAfter is before now). This assumes we can guess at how many seconds would suffice for the in-flight requests to finish. Perhaps six or sixty would work well enough.

slifty: OK I have an idea: NEVER REMOVE -- but instead add a `not_after` column which is auto-populated / updated on insert to give the row a "life" of 1 hour (or whatever we think is reasonable).

bickelj: That sounds good to me. Shrinking the life shrinks the window for a query to finish successfully although maybe it is the window to start the query. Extending the life extends the amount of time it takes to revoke access by removing somebody from an organization. Our JWTs are around a half hour right now, probably should be back at five minutes, so IDK, I guess I'd recommend something more like one to thirty minutes.

Oh, or how about this: expire it when the JWT expires (with some minimum, such as two minutes)?

slifty: JWT expiration is a good idea -- we might need to think carefully about how ephemeral permissions + bulk upload work, since those jobs are run via a queue and in theory might not be started so quickly, but I think we're OK for now.

src/middleware/addUserContext.ts
```diff
@@ -44,13 +46,42 @@ const addUserContext = (
...
+						userKeycloakUserId: keycloakUserId,
+						userGroupKeycloakOrganizationId: keycloakOrganizationId,
+						notAfter: jwtExpiration
+							? new Date(
+									Math.max(
+										jwtExpiration,
+										MINIMUM_EPHEMERAL_USER_GROUP_ASSOCIATION_TIMEOUT_SECONDS,
+									) * 1000,
+								).toISOString()
+							: null,
```
bickelj, line 65: Should it still expire here with the minimum?

src/database/operations/users/__tests__/loadUserByKeycloakUserId.int.test.ts
```diff
@@ -0,0 +1,286 @@
...
+			createdBy: systemUser.keycloakUserId,
+		});
+
+		const populatedUser = await loadUserByKeycloakUserId(
+			db,
+			null,
+			user.keycloakUserId,
+		);
+
+		expect(populatedUser).toEqual({
```
bickelj, line 164: Woot!

bickelj (approved):
It looks like all my concerns were addressed. Note I did not try this locally but reviewed some of the changes and noticed the very good tests added. LGTM if you are comfortable with it.

src/middleware/addUserContext.ts
slifty: Good question -- if a JWT never expires, should we be taking that at face value?

bickelj: Hmm, maybe I only pushed the problem back a step with the JWT thing or JWT is really unrelated. I was looking for an objective (as opposed to an arbitrary) expiration time that would guarantee this request to succeed but not much more than that. We're still left wanting that educated guess. I would still have it expire to something by default. If that ends up being `MINIMUM_EPHEMERAL_USER_GROUP_ASSOCIATION_TIMEOUT_SECONDS` maybe we're better off ignoring the JWT and focusing on the request. I might think about it more clearly in the morning too.

slifty: Yep; agreed -- I also think think I won't allow NULL at the DB level since the table name makes it clear the intent is these ARE ephemeral, and so everything really needs to have an expiration.  I'll make that tweak.

I think JWT is still a good idea just from a "it is the responsibility of the system that defines the auth context to define the expiration of that auth context" -- and then the only reason we're overriding that for the "minimum" is that the JWT is intended as needing to be valid at the *start* of a call, whereas this table needs to be valid until the end of the call, so it makes sense we'd have that (modest) minimum.

I do think that it's reasonable to defensively add a maximum as well, though -- and it could truly be something like an hour -- which would cover the "never expires" case.

slifty: I decided not to do a maxmim for now just in that name of "separation of responsibility" -- but everything else is now implemented!  Going to go over it again and then smoosh merge.

PR #1499, opened 2025-02-13
Associate funders with opportunities

### LINKED ISSUE #1498: Add a `funder` attribute to opportunities
So far opportunities in the PDC have been visible to any authenticated user; as we implement #1406 we want to link funders and opportunities. This was always in the plans, and now we have a use case!

This will also involve updating the BulkUpload to require a funderId in addition to a source, since bulk uploads create funders.

We need to decide if we want opportunities to be able to have `NULL` field.  If we don't then we need to determine how to migrate since there are existing opportunities (do we create a system funder like we have for other similar cases).

slifty:
This PR associates funders with opportunities and bulk uploads.

This was always part of the plan, but we need to implement it now in order to support permissions on opportunity and proposal creation.

Resolves #1498

### DIFF
--- a/src/__tests__/funders.int.test.ts
+++ b/src/__tests__/funders.int.test.ts
@@ -5,6 +5,7 @@ import {
 	createOrUpdateFunder,
 	loadFunder,
 	loadTableMetrics,
+	loadSystemFunder,
 } from '../database';
 import { expectTimestamp } from '../test/utils';
 import {
@@ -20,14 +21,8 @@ describe('/funders', () => {
 			await agent.get('/funders').expect(401);
 		});
 
-		it('returns an empty bundle when no data is present', async () => {
-			await agent.get('/funders').set(authHeader).expect(200, {
-				entries: [],
-				total: 0,
-			});
-		});
-
 		it('returns all funders present in the database', async () => {
+			const systemFunder = await loadSystemFunder(db, null);
 			await createOrUpdateFunder(db, null, {
 				shortCode: 'theFundFund',
 				name: 'The Fund Fund',
@@ -54,8 +49,9 @@ describe('/funders', () => {
 						name: 'The Fund Fund',
 						keycloakOrganizationId: null,
 					},
+					systemFunder,
 				],
-				total: 2,
+				total: 3,
 			});
 		});
 	});
--- a/src/database/migrations/0052-add-funder-opportunity-associations.sql
+++ b/src/database/migrations/0052-add-funder-opportunity-associations.sql
@@ -1,0 +1,28 @@
+INSERT INTO funders (short_code, name) VALUES (
+	'pdc', 'System Placeholder Funder'
+);
+
+CREATE OR REPLACE FUNCTION system_funder_short_code()
+RETURNS varchar AS $$
+BEGIN
+    RETURN 'pdc';
+END;
+$$ LANGUAGE plpgsql;
+
+ALTER TABLE opportunities
+ADD COLUMN funder_short_code short_code_t NOT NULL REFERENCES funders (
+	short_code
+) DEFAULT 'pdc';
+
+ALTER TABLE opportunities
+ALTER COLUMN funder_short_code DROP DEFAULT;
+
+ALTER TABLE bulk_upload_tasks
+ADD COLUMN funder_short_code short_code_t NOT NULL REFERENCES funders (
+	short_code
+) DEFAULT 'pdc';
+ALTER TABLE bulk_upload_tasks
+ALTER COLUMN funder_short_code DROP DEFAULT;
+
+COMMENT ON COLUMN bulk_upload_tasks.funder_short_code IS
+'The funder to associate with opportunities created by this bulk upload task.';
--- a/src/database/queries/bulkUploadTasks/insertOne.sql
+++ b/src/database/queries/bulkUploadTasks/insertOne.sql
@@ -1,5 +1,6 @@
 INSERT INTO bulk_upload_tasks (
 	source_id,
+	funder_short_code,
 	file_name,
 	source_key,
 	status,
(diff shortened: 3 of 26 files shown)


bickelj (commented):
I understand why a funder needs to be associated with an opportunity, but I'm not sure I understand why a funder must be associated with a bulk upload. I see the note "Since bulk uploads create opportunities we needed to add a funder attribute to bulk uploads as well" so that makes some sense. It is not necessarily saying "I am a funder doing this bulk upload" but rather saying "these data came from this funder"? If I'm a data provider, I mark the source as me but the funder as the funder?

src/__tests__/funders.int.test.ts
```diff
@@ -20,14 +21,8 @@ describe('/funders', () => {
 			await agent.get('/funders').expect(401);
 		});
 
-		it('returns an empty bundle when no data is present', async () => {
-			await agent.get('/funders').set(authHeader).expect(200, {
-				entries: [],
-				total: 0,
-			});
-		});
-
```
bickelj, line 29: Did this need to be removed?

src/database/migrations/0052-add-funder-opportunity-associations.sql
```diff
@@ -0,0 +1,28 @@
+INSERT INTO funders (short_code, name) VALUES (
+	'pdc', 'The Philanthropy Data Commons'
```
bickelj, line 2: This could be 'unknown' or 'dummy' or something like that.

bickelj (approved):
I have yet to run it or try it but I see no red flags when looking over the code.

src/__tests__/funders.int.test.ts
slifty: Yes indeed -- there is no longer a case of there being no data now that we have a system funder.

src/database/migrations/0052-add-funder-opportunity-associations.sql
slifty: Fine with `Unknown` -- this will be applied to real data though, so whatever we use we should assume will be rendered to real users.

src/database/queries/funders/selectSystemFunder.sql
```diff
@@ -0,0 +1,3 @@
+SELECT funder_to_json(funders.*) AS object
+FROM funders
+WHERE short_code = system_funder_short_code();
```
hminsky2002, line 3: Where is this function defined? Or is this not a function but rather some sweet sql syntax

src/database/queries/bulkUploadTasks/insertOne.sql
```diff
@@ -1,12 +1,14 @@
 INSERT INTO bulk_upload_tasks (
```
hminsky2002, line 1: I don't know if it's necessary per-se, but since we have a bulkUploadTask 'updateById', do we want to add an option to update the funder_short_code for it as well? If not, is it because  we're only concerned with what the value was at creation?

hminsky2002 (approved):
Looks great! Left two comments more for my own understanding than anything else, but approving

src/database/queries/funders/selectSystemFunder.sql
slifty: It's in the new migration file

src/database/queries/bulkUploadTasks/insertOne.sql
slifty: Good call -- that should definitely be included as something that can be updated!

src/database/migrations/0052-add-funder-opportunity-associations.sql
slifty: @bickelj FWIW we *essentially* are doing that -- nobody is going to have the ability to write to this funder (except a superuser or our tests) since nobody will have write access to this funder.

I just mean that the data in there today will get this so it'll show up in API responses for that data!

slifty: I'm leaning towards keeping it as is: PDC isn't going to be funding things, but that's more of a feature than a bug -- we don't want this funder to generally be used on real data going forward, it's almost a migration + test artifact.

Labeling it `Unknown` makes it more likely, I think, that someone in future might try to use it to convey the funder is unknown (but in reality we will never want an unknown funder, which is why we aren't allowing `null` values).

Labeling it `Dummy` has the downside of production data having a `Dummy` label associated with it.  That said, maybe that's OK and it will just put the burden on correcting the data instead.

Labeling it `The Philanthropy Data Commons` is clear in the future -- people would only use that funder if something is literally funded by PDC which will never happen -- though it does mean that legacy data is being "funded" by PDC which is not really true.  That legacy data is synthetic right now, however, so this is possibly OK?

No perfect solution unfortunately 😭

src/database/queries/bulkUploadTasks/insertOne.sql
slifty: I'm seeing we don't allow updates to sourceId right now either -- I'm thinking maybe I'll just leave it as is for now, until we have a use case for allowing updates to funder for some reason.

hminsky2002: @slifty that seems reasonable, no need to add the functionality if we explicitly do not want people changing the value right now

src/database/migrations/0052-add-funder-opportunity-associations.sql
bickelj: How about "Deprecated" or "Legacy Data -- Do Not Use"?

slifty: I think "PDC System" or something like that could strike a reasonable balance?  Main thing is that while this is for legacy data right now, it isn't always (e.g. when it is used in tests, or even possibly in future if we have a use case of a bulk upload run by PDC / not linked to a funder for some reason)

All that said, let's ask @kfogel today since it'll be a quick decision and we can then smoosh merge!

slifty: OK we locked in on "PDC System Placeholder Funder"

Issue #1406, opened 2025-01-03
Utilize permissions


slifty:
With #1250 we have the concept of user permissions; let's start to use them!

Part of this issue will be determining what endpoints need permissions to be added, so I won't try to list that out in the issue description.

slifty:
In starting on this I realized that we're going to want a way to convey organizational association without having to pass queries an unbounded list of org IDs that came from the JWT.  The reason this is not desired is that, well, it's unbounded / unindexed / etc.

I spoke with @bickelj and we came up with the idea of creating a `user_organization_relationship_cache` table (name TBD) which would be ensured to be populated correctly *in a given API call* via middleware.  This would work by having our auth middleware compare the user organization state with the jwt, and if there is a disparity it would update that state.

From that point our queries would be able to JOIN on that table when looking up / loading / sharing user permissions.

slifty:
Part of this issue will be to delineate the implemented permissions / rules associated with various grants.

We're thinking this should go in a markdown file (and may as well be added as part of the same commits around the implementation lines)

slifty:
I'm gonna go through our entities and lay out what I think the rules are:


* **ApplicationForm**:
  * `GET`
    * ??
  * `POST`
    * `edit` permission on the `funder` defined by the `application form's opportunity`.

* **BulkUploadTask**:
  * `GET`
    * `read` permission on the `funder` or `data provider` defined by the `bulk upload task's source`.
  * `POST`
    * `edit` permission on the `funder` or `data provider` defined by the `bulk upload task's source`.

* **BaseFieldsCopyTask**:
  * `GET`
    * superuser only.
  * `POST`
    * superuser only.

* **ChangemakerProposal**
  * `GET`
    * `read` permission on the `organization` defined by the `proposal's source`.
    * `read` permission on the `funder` associated with the `proposal's opportunity`.
    * `read` permission on the `changemaker` associated with the `changemaker proposal`.
  * `POST`
    * `edit` permission on the `organization` defined by the `proposal's source`.

* **Changemaker**
  * `GET`
    * ??
  * `POST`
    * ??

* **Presigned Post**
  * `POST`
    * ??

* **DataProvider**
  * `GET`
    * ??
  * `PUT`
    * superuser only.

* **Funder**
  * `GET`
    * ??
  * `PUT`
    * superuser only.

* **Opportunity**
  * `GET`
    * ??
  * `PUT`
    * `edit` permission on the `funder` defined by the `opportunity`

[...]

PR #1504, opened 2025-02-18
Add permissions to opportunities

slifty:
This PR adds a permission check to the `POST /opportunities` endpoint so that a user cannot create an opportunity without edit permission on the related funder.

Related to #1406

### DIFF
--- a/src/__tests__/opportunities.int.test.ts
+++ b/src/__tests__/opportunities.int.test.ts
@@ -5,9 +5,12 @@ import {
 	createOpportunity,
 	loadTableMetrics,
 	loadSystemFunder,
+	createOrUpdateUserFunderPermission,
+	loadSystemUser,
 } from '../database';
-import { expectTimestamp } from '../test/utils';
+import { expectTimestamp, loadTestUser } from '../test/utils';
 import { mockJwt as authHeader } from '../test/mockJwt';
+import { Permission } from '../types';
 
 describe('/opportunities', () => {
 	describe('GET /', () => {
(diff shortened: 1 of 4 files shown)


slifty:
The untested lines fall into the "we have decided not to bend over backwards to test" category

src/__tests__/opportunities.int.test.ts
```diff
@@ -128,8 +131,16 @@ describe('/opportunities', () => {
 			await request(app).post('/opportunities').expect(401);
 		});
 
-		it('creates and returns exactly one opportunity', async () => {
+		it('creates and returns exactly one opportunity when edit funder permission is set', async () => {
```
hminsky2002, line 134: Should we have tests for all types of permissions here (Manage,Edit,View) or is edit the only relevant permission for opportunities?

src/handlers/opportunitiesHandlers.ts
```diff
@@ -71,6 +83,16 @@ const postOpportunity = (
 		);
 		return;
 	}
+	if (
```
hminsky2002, line 86: Ah, i suppose we're only checking if it has edit permissions in this handler

hminsky2002 (approved):
Looks great! Indeed quite minimal. Left a comment about testing other permission types but since we're only adding a check for edit in one handler then there's nothing else to test!

src/__tests__/opportunities.int.test.ts
slifty: Edit is the only one that would allow you to do it (and we do already have a test to ensure that it fails if you don't have edit).

We could add tests to confirm edit is the ONLY permission that allows it, which would be more complete / make sure we didn't accidentally over-power other permissions.

PR #1506, opened 2025-02-19
Document fiscal sponsorship modeling

bickelj:
Issue #1485 Support fiscal sponsor relationships between changemakers

### DIFF
--- a/docs/CHANGEMAKER_DATA.md
+++ b/docs/CHANGEMAKER_DATA.md
@@ -13,6 +13,14 @@ endpoints. Changemaker data cannot currently be directly written to the PDC,
 though that functionality is on the roadmap. Instead, changemaker data is
 aggregated from any proposals that have been submitted by the changemaker.
 
+## Registering fiscal sponsorship
+
+To register one changemaker as a fiscal sponsor of another changemaker
+explicitly (in addition to the proposal data reflecting fiscal sponsors), use
+`PUT /changemakers/:changemakerId/fiscalSponsors/:fiscalSponsorChangemakerId`
+where the `:changemakerId` is the sponsee. A changemaker may fiscally sponsor
+many other changemakers and vice versa.
+
 ## Viewing changemaker data
 
 To see changemaker data in the PDC, use `GET /changemakers`. This endpoint
@@ -24,6 +32,10 @@ for the given changemaker, PDC returns exactly one prioritized value. Each
 returned value may come from a separate data source, such as a proposal to a
 funder, a data platform provider (DataProvider in PDC), or the changemakers
 themselves. As of this writing values come solely from proposals.
+
+Rich field values ("gold" data) are returned in changemaker `fields`.
+
+Registered fiscal sponsors are returned in changemaker `fiscalSponsors`.
 
 ### Data prioritization or conflict resolution ("gold" data)
 
--- a/docs/ENTITY_RELATIONSHIP_DIAGRAM.md
+++ b/docs/ENTITY_RELATIONSHIP_DIAGRAM.md
@@ -139,6 +139,7 @@ erDiagram
 	User }o--o{ Changemaker : "is granted permissions for"
 	User }o--o{ Funder : "is granted permissions for"
 	User }o--o{ DataProvider : "is granted permissions for"
+	Changemaker }o--o{ Changemaker : "may fiscally sponsor"
 ```
 
 ## Narrative


docs/CHANGEMAKER_DATA.md
```diff
@@ -13,6 +13,13 @@ endpoints. Changemaker data cannot currently be directly written to the PDC,
 though that functionality is on the roadmap. Instead, changemaker data is
 aggregated from any proposals that have been submitted by the changemaker.
 
+## Registering fiscal sponsors
+
+To register one changemaker as a fiscal sponsor of another changemaker
+explicitly (in addition to the proposal data reflecting fiscal sponsors), use
+`PUT /changemakers/:changemakerId/fiscalSponsors/:fiscalSponsorChangemakerId`
```
kfogel, line 20: I see that `fiscalSponsors` is right there in the API route.  But we're not treating fiscal sponsorship specially, as a type of relationship, right?  For example, if we were to define "sharing staff" as a type of relationship, then would there be an endpoint `/changemakers/:changemakerId/sharesStaff/:sharedStaffChangemakerId`?  I.e., for every arbitrary relationship we define, part of making that definition is choosing an endpoint name?

docs/CHANGEMAKER_DATA.md
```diff
@@ -25,6 +32,10 @@ returned value may come from a separate data source, such as a proposal to a
 funder, a data platform provider (DataProvider in PDC), or the changemakers
 themselves. As of this writing values come solely from proposals.
 
+Rich field values ("gold" data) are returned in changemaker `fields`.
```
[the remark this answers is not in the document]
bickelj: The purpose of having two separate sentences here is to describe the two separate fields returned in the changemaker response body.

Both the parenthetical and the preceding paragraph provide the context for this note. Should I put it into the above paragraph? Rich is a better description than "gold" but I wanted to link the two here in case you've only ever heard of "gold" data. Should I eliminate either of those words? Ditch "rich" and keep "gold"?

docs/CHANGEMAKER_DATA.md
```diff
@@ -13,6 +13,13 @@ endpoints. Changemaker data cannot currently be directly written to the PDC,
 though that functionality is on the roadmap. Instead, changemaker data is
 aggregated from any proposals that have been submitted by the changemaker.
 
+## Registering fiscal sponsors
```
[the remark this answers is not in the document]
bickelj: Fixed and added.

docs/CHANGEMAKER_DATA.md
bickelj: The task in #1485, as I understood it, was to explicitly model fiscal sponsorship in our entity model. If there is another task in the future to "explicitly model sharing staff in our entity model" then yes we would need to add endpoints to allow for that relationship to be modeled.

Did we miss the mark?

slifty: I would not say we missed the mark based on the current use case and scope of work @bickelj -- if that happens in future we can implement support for generic entity relationships, but for now we have one very specific kind of relationship to model.

So to @kfogel's question: not necessarily, no!  I don't think this feature was intended to create a new pattern for future features so much as it was intended to meet the needs around fiscal sponsorship as understood today.

bickelj: Ah, I see what you are suggesting @slifty. Yes, right now we have a case of "one special relationship", but if we shift to a case of "many special relationships" that is an opportune moment to make a more generic endpoint, such as `/changemakers/:changemakerId/:otherEntityName/:otherEntityId/:specialRelationshipEnumValue` or what have you.

kfogel: @bickelj @slifty If we did that future genericization, we would have a compatibility path by which we could transition the old `fiscalSponsors` way to the new `:specialRelationshipEnumValue` way?  I mean, I think the answer to that is "obviously yes", but I wanted to raise it here explicitly.  I understand that "premature generalization is the root of all evil", but I also want to make sure we're not locking the door against future evil... or something like that...

slifty: Well the main issue is... this PR is a documentation PR for a feature that was already built / approved / merged / deployed.

Could we open up a new issue to capture the rest of this discussion, or Karl should we roll back that feature and re-implement?

Issue #940, opened 2024-04-22
Reflect on our `403` vs `404` response codes


slifty:
There was some discussion in [a PR](https://github.com/PhilanthropyDataCommons/service/pull/934) recently about whether we should always return the same status code for entities regardless of whether they don't exist at all, or just "don't exist" based on the permissions of a given user (aka the user is not authorized to access the resource).

This is something we should reach consensus on and implement consistently one way or another.

reefdog:
Copying over my minor contribution:

> 403 is semantically sensible (and spec-correct), but the design of showing 404 for private resources so as not to confirm entity presence is established best practice now, AFAIK. (Case in point, hit this URL while logged in: https://github.com/reefdog/PrivateTest)

Our use case is mildly different in that we currently use numeric IDs rather than, say, org-name-based-slugs. But each leaks data:

- a 403 for `/organizations/200` + a 404 for `/organizations/201` tells people we probably have 200 organizations. (Of course, so does our `/organizations` endpoint's `total` attribute, but at least we're doing that on purpose.)
- a 404 for `/organizations/red-cross` means we don't have it, but a 403 means we do.

In both these cases, can I imagine a horror-story outcome? No. But the way I think of it: if you are only granted access to chapters 1, 2,  and 3 of a book, you should also not have access to the table of contents that reveals chapter titles for the rest of the chapters. (Unless we explicitly decide we _want_ you to. Then, great. Populate `/organizations` with the same limited data they could get by brute-forcing the endpoint.)

slifty:
I agree that a more conservative approach is more secure and also fairly standard in my experience (I'm used to this meaning universal 404 rather than universal 403, but I think it COULD go either way?)

Knowing an ID is a real ID feels super mundane, but the thing about exposing mundane data is that I think it possibly *could* create situations where the presence of a new entity combined with some other kind of third party data might result in some kind of strange opsec edge case.  I'd rather just not have to think about those edge cases!

bickelj:
[...]

PR #1508, opened 2025-02-20
Add permission check for BulkUploadTask creation

slifty:
This PR adds permission checks to the `POST /tasks/bulkUploads` endpoint.

This also corrects a bug where administrators were not automatically considered as having all permissions for funder authorization checks.

Related to #1406

### DIFF
--- a/src/authorization.ts
+++ b/src/authorization.ts
@@ -5,7 +5,8 @@ const authContextHasFunderPermission = (
 	funderShortCode: ShortCode,
 	permission: Permission,
 ): boolean =>
-	auth.user.permissions.funder[funderShortCode] !== undefined &&
-	auth.user.permissions.funder[funderShortCode].includes(permission);
+	auth.role.isAdministrator ||
+	(auth.user.permissions.funder[funderShortCode] !== undefined &&
+		auth.user.permissions.funder[funderShortCode].includes(permission));
 
 export { authContextHasFunderPermission };
--- a/src/handlers/bulkUploadTasksHandlers.ts
+++ b/src/handlers/bulkUploadTasksHandlers.ts
@@ -6,6 +6,7 @@ import {
 	loadBulkUploadTaskBundle,
 } from '../database';
 import {
+	Permission,
 	TaskStatus,
 	isAuthContext,
 	isTinyPgErrorWithQueryContext,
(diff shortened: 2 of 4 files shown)


src/authorization.ts
```diff
@@ -5,7 +5,8 @@ const authContextHasFunderPermission = (
 	funderShortCode: ShortCode,
 	permission: Permission,
 ): boolean =>
-	auth.user.permissions.funder[funderShortCode] !== undefined &&
-	auth.user.permissions.funder[funderShortCode].includes(permission);
+	auth.role.isAdministrator ||
```
slifty, line 8: I believe this is the right place for this check (rather than having an explicit check every time a funder permission is tested).

If we ever have a case where we don't want the administrator flag to grant permission for some reason then I think that is a significant enough departure from the intention of `superuser` that a refactor would be appropriate at that point.

src/handlers/bulkUploadTasksHandlers.ts
```diff
@@ -48,6 +51,11 @@ const postBulkUploadTask = (
 	const { sourceId, funderShortCode, fileName, sourceKey } = req.body;
 	const createdBy = req.user.keycloakUserId;
 
+	if (!authContextHasFunderPermission(req, funderShortCode, Permission.EDIT)) {
+		next(new UnprocessableEntityError('The specified funder was not found'));
```
hminsky2002, line 55: Would this error message be more accurate as 'specified funder permission was not found'?

hminsky2002 (approved):
One small note that doesn't effect approval, and a note agreeing on giving administrator roles access to any given entity. Looks great @slifty !

src/handlers/bulkUploadTasksHandlers.ts
slifty: *technically* yes -- but actually we don't want to expose that information to the unauthorized user.  If the user can't see a funder they shouldn't be given hints that the funder exists.

I'll add in a conditional to say "if the user can see it then say they don't have permission to create, but otherwise say the funder doesn't exist"

All that said, I think in general our "errors" are inconsistent and leak information, and when we take on #940 this should be part of that in either case.

slifty: I'm just gonna modify the wording to be more mysterious...

`You do not have write permissions on a funder with the specified short code`

hminsky2002: Splendid

PR #1511, opened 2025-02-21
Add permissions to `POST /proposals`

slifty:
This PR adds permission checks to the `POST /proposals` endpoint so that only users with write permissions on a proposal's opportunity's funder is able to create new proposals in response to the opportunity.

Related to #1406

### DIFF
--- a/src/handlers/proposalsHandlers.ts
+++ b/src/handlers/proposalsHandlers.ts
@@ -106,21 +106,20 @@ const postProposal = (
 	const { externalId, opportunityId } = req.body;
 	const createdBy = req.user.keycloakUserId;
 
-	createProposal(db, null, {
-		opportunityId,
-		externalId,
-		createdBy,
-	})
-		.then((proposal) => {
-			res.status(201).contentType('application/json').send(proposal);
-		})
-		.catch((error: unknown) => {
-			if (isTinyPgErrorWithQueryContext(error)) {
-				next(new DatabaseError('Error creating proposal.', error));
-				return;
-			}
-			next(error);
+	(async () => {
+		const proposal = await createProposal(db, null, {
+			opportunityId,
+			externalId,
+			createdBy,
 		});
+		res.status(201).contentType('application/json').send(proposal);
+	})().catch((error: unknown) => {
+		if (isTinyPgErrorWithQueryContext(error)) {
+			next(new DatabaseError('Error creating proposal.', error));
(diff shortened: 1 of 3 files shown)


src/handlers/proposalsHandlers.ts
```diff
@@ -106,21 +106,20 @@ const postProposal = (
...
-		})
-		.catch((error: unknown) => {
-			if (isTinyPgErrorWithQueryContext(error)) {
-				next(new DatabaseError('Error creating proposal.', error));
-				return;
-			}
-			next(error);
+	(async () => {
+		const proposal = await createProposal(db, null, {
+			opportunityId,
```
hminsky2002, line 128: mm this is good -- so we're doing this because we'll eventually add other async functions in this block that we don't want to keep chaining?

src/handlers/proposalsHandlers.ts
```diff
@@ -107,6 +112,18 @@ const postProposal = (
 	const createdBy = req.user.keycloakUserId;
 
 	(async () => {
+		const opportunity = await loadOpportunity(db, null, opportunityId);
+		if (
+			!authContextHasFunderPermission(
+				req,
```
hminsky2002, line 118: these would be those other async functions, got it

src/handlers/proposalsHandlers.ts
```diff
@@ -118,6 +135,14 @@ const postProposal = (
 			next(new DatabaseError('Error creating proposal.', error));
 			return;
 		}
+		if (error instanceof NotFoundError) {
```
hminsky2002, line 138: Could i get a refresher on the UnprocessableEntityError? I know we just brought it in XD. We use it to more accurately represent a not-found error in the db to the api user?

hminsky2002 (approved):
Looks good, tested locally. Left some clarifying comments for myself

src/handlers/proposalsHandlers.ts
slifty: Yes (this is what I was trying to explain in the commit message)

src/handlers/proposalsHandlers.ts
slifty: UnprocessableEntity is the 422 HTTP status code -- it's saying that the format of the request was correct, but for some reason (e.g. the content of the request) the request cannot be processed.

PR #1513, opened 2025-02-21
Add permissions to proposal changemaker post

slifty:
This PR adds permission checks to the `POST /changemakerProposals` endpoint.  Prior to this there were no checks and any authenticated user could associate any proposal with any changemaker.

Now only users with edit permissions on the funder associated with the proposal's opportunity can make those associations.

Related to #1406


hminsky2002 (approved):
Looks great to me!

PR #1518, opened 2025-02-27
Add permissions to POST /proposalVersions

slifty:
This PR adds a check that limits creation of new proposal versions to those with write access to the funders who sponsored the opportunity.

In future we may add additional permissions (e.g. changemakers associated with the proposal).

Related to #1406

### DIFF
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -188,6 +188,7 @@ funder, data provider, or other changemaker entirely.
 - Allows users to create sources for the funder.
 - Allows users to create new proposals for the funder's opportunities.
 - Allows users to create new changemaker relationships with proposals for the funder's opportunities.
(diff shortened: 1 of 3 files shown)


hminsky2002 (approved):
Looks good, didn't see any pattern breaks from previous PRs. I know i've already asked this but could I get a refresh on what the intended use of the other permission types besides EDIT are? I think it's fairly self-evident why we are using EDIT permissions for a POST request, but am wondering when and in what context we'll be validating other permission types

slifty:
@hminsky2002 VIEW isn't used yet but would be used for read checks (that's what I'm working on now)

MANAGE is already used -- it lets you manage permissions for that entity.

slifty:
Also I see failing tests... sorry!  Fixing that now.

PR #1520, opened 2025-02-27
Add permissions to GET /applicationForms

slifty:
This PR adds our first permission checks for viewing data.  Specifically we are now limiting access to loading application forms so that only users with `view` access to a given funder can view that funder's opportunities' application forms.

Right now I'm implementing edit and view as completely distinct permissions, which creates some potentially un-intuitive situations (e.g. getting edit permission alone is not enough to create things, you need both edit and view because you need to be able to view the related items).

We can add additional permission cases in a future dev cycle; starting more restrictive and deciding as a group to become more permissive seems like a safer decision.

Related to #1406

### DIFF
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -193,3 +193,7 @@ funder, data provider, or other changemaker entirely.
 #### Manage
 
 - Allows users to modify the permissions for the funder for other users and user groups.
+
+#### View
+
+- Allows users to view the application forms associated with the funder's opportunities.
--- a/src/handlers/applicationFormsHandlers.ts
+++ b/src/handlers/applicationFormsHandlers.ts
@@ -51,7 +51,7 @@ const getApplicationForm = (
 	res: Response,
 	next: NextFunction,
 ): void => {
-	const { id: applicationFormId } = req.params;
+	const { applicationFormId } = req.params;
 	if (!isId(applicationFormId)) {
 		next(new InputValidationError('Invalid request.', isId.errors ?? []));
 		return;
--- a/src/handlers/applicationFormsHandlers.ts
+++ b/src/handlers/applicationFormsHandlers.ts
@@ -51,12 +51,17 @@ const getApplicationForm = (
 	res: Response,
 	next: NextFunction,
 ): void => {
+	if (!isAuthContext(req)) {
+		next(new FailedMiddlewareError('Unexpected lack of auth context.'));
+		return;
+	}
+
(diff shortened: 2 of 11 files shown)


src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -395,7 +395,7 @@ describe('processBulkUploadTask', () => {
 			entries: [applicationForm],
 		} = await loadApplicationFormBundle(db, null, NO_LIMIT, NO_OFFSET);
 		if (applicationForm === undefined) {
-			fail('The application form was not created');
```
hminsky2002, line 398: fail?? I didn't even know there was a fail function. is this a jest function?

docs/PERMISSIONS.md
```diff
@@ -193,3 +193,7 @@ funder, data provider, or other changemaker entirely.
 #### Manage
 
 - Allows users to modify the permissions for the funder for other users and user groups.
+
+#### View
```
hminsky2002, line 197: The commit could be a bit more descriptive, also are you actually fixing a typo anywhere? 
It seems like you're doing three small fixes in this one commit, which is fine if they don't fit anywhere else, but maybe mention that in the commit message

src/handlers/applicationFormsHandlers.ts
```diff
@@ -51,12 +51,17 @@ const getApplicationForm = (
...
+		return;
+	}
+
 	const { applicationFormId } = req.params;
 	if (!isId(applicationFormId)) {
 		next(new InputValidationError('Invalid request.', isId.errors ?? []));
 		return;
 	}
-	loadApplicationForm(db, null, applicationFormId)
+	loadApplicationForm(db, req, applicationFormId)
```
hminsky2002, line 68: nice, love to see these getting activated!

hminsky2002 (approved):
Looks great, really moving here.

src/tasks/__tests__/processBulkUploadTask.int.test.ts
slifty: Nope!  It was just... never caught?!  I don't even know -- it's wild to me this has been here this whole time 😂

docs/PERMISSIONS.md
slifty: Oooo I bet I accidentally included something in here that wasn't intended.  I'll review (it was meant to just fix a typo)

PR #1527, opened 2025-02-28
Add permission check GET `/proposal` endpoints

slifty:
This PR adds permission checks to the GET `/proposals` endpoints.

Related to #1406


hminsky2002 (approved):
Again i don't see any pattern breaks, approving

PR #1529, opened 2025-02-28
Add permissions to `GET` `/opportunities` endpoints

slifty:
This PR limits access to opportunities to those who have view access on the related funder.

Related to #1406

### DIFF
--- a/src/__tests__/opportunities.int.test.ts
+++ b/src/__tests__/opportunities.int.test.ts
@@ -9,7 +9,10 @@ import {
 	loadSystemUser,
 } from '../database';
 import { expectTimestamp, loadTestUser } from '../test/utils';
-import { mockJwt as authHeader } from '../test/mockJwt';
+import {
+	mockJwt as authHeader,
+	mockJwtWithAdminRole as authHeaderWithAdminRole,
+} from '../test/mockJwt';
 import { Permission } from '../types';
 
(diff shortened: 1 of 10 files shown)


src/__tests__/opportunities.int.test.ts
```diff
@@ -61,7 +64,7 @@ describe('/opportunities', () => {
 		});
 	});
 
-	describe('GET /:id', () => {
+	describe('GET /:opportunityId', () => {
```
hminsky2002, line 104: maybe i missed this but shouldn't there be a change to the opportunities router for the id parameter within this pr for it to still function? i see in the opportunities router this change has already been implemented, but then shouldn't testing have been failing?

hminsky2002 (approved):
Code looks correct and CI/CD passing but I left a comment regarding the parameter change from 'id' to 'opportunityId', maybe i'm too lost in the sauce of the code here but wanted to bring it up!

src/__tests__/opportunities.int.test.ts
slifty: This is just a test name / it's basically just a typo correction

hminsky2002: @slifty ah right, it's the test name, not the actual request. cool!

PR #1534, opened 2025-03-06
Add unique constraint on source label and fkeys

### LINKED ISSUE #1455: Add unique constraint on name and foreign key for sources

hminsky2002:
This PR adds unique constraints on a sources label and it's potential foreign keys to other entities (organization, dataProvider, funder). Notably, it does allow for two sources to exist with the same label so long as they are related to a distinct entity.

Closes #1455

### DIFF
--- a/src/__tests__/sources.int.test.ts
+++ b/src/__tests__/sources.int.test.ts
@@ -18,7 +18,7 @@ import {
 	mockJwt as authHeader,
 	mockJwtWithAdminRole as adminUserAuthHeader,
 } from '../test/mockJwt';
-import { Permission } from '../types';
+import { Permission, PostgresErrorCode } from '../types';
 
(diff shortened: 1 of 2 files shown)


bickelj (approved):
The constraints look and work great. Thanks for this!

I was able to `lint`, `build`, `migrate`, `start`, and `test` this locally. When looking at the database table in `psql` it looks good. When trying to add a second source with the same label to the same funder or changemaker or data provider, it appropriately fails.

The only thing I wonder, and this is more general and not specific to this pull request, is whether we should be intentionally surfacing postgres errors in 4xx codes to callers. There may be unexpected circumstances where it is better to surface a postgres error than to hide it, but maybe we should wrap some of these in a more tidy way someday. I don't think it really matters for now, just a thought.

PR #1540, opened 2025-03-19
Upgrade to Express v5

### LINKED ISSUE #1493: Upgrade to Express 5
https://expressjs.com/en/support/

Time to move to the latest and greatest, unless it turns out it isn't of course.  This should pave the way for being able to use asynchronous middleware and route handlers, which in turn would address #700

slifty:
This PR upgrades our version of express to v5 and fixes some resulting type complaints from eslint and TypeScript.

Resolves #1493

### DIFF
--- a/src/auth/jwtOptions.ts
+++ b/src/auth/jwtOptions.ts
@@ -18,7 +18,7 @@ export const jwtOptions: Params = {
 		jwksUri: `${AUTH_SERVER_ISSUER}/protocol/openid-connect/certs`,
 		rateLimit: true,
 		jwksRequestsPerMinute: 100,
-	}) as GetVerificationKey,
+	}) as unknown as GetVerificationKey,
 	issuer,
 	algorithms: ['RS256'],
 	credentialsRequired: false,
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -55,6 +55,12 @@
 					}
 				]
 			}
+		},
+		{
+			"files": ["**/*.test.ts"],
+			"rules": {
+				"@typescript-eslint/no-misused-promises": "off"
+			}
 		}
 	],
 	"env": {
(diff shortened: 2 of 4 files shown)


hminsky2002:
@slifty cool! before I review, i did take a look at the [express v5 migration guide](https://expressjs.com/en/guide/migrating-5.html) and it seems that there's a fair amount of updated functionality. Just wanted to check -- have you also gone through this document and reviewed the major updates/ is this PR just to get us over the line to be in express 5, and from here on there might be more v5 functionality we can use if we so choose?

slifty:
@hminsky2002 the intention is to do the bump in this PR and actually *use* the improvements later.

Our test coverage / CI passing suggests to me that we didn't have breaking changes but you're right that this is a major enough improvement that we should manually confirm things too, just in case there's a situation where the tests pass incorrectly.

slifty:
Just to confirm: looks like the endpoints are working.  I tried both an unauthenticated endpoint and authenticated endpoint.

hminsky2002:
@slifty excellent, well on that note the changes seem perfectly reasonable to me. approving!

hminsky2002 (approved):

src/auth/jwtOptions.ts
```diff
@@ -18,7 +18,7 @@ export const jwtOptions: Params = {
 		jwksUri: `${AUTH_SERVER_ISSUER}/protocol/openid-connect/certs`,
 		rateLimit: true,
 		jwksRequestsPerMinute: 100,
-	}) as GetVerificationKey,
+	}) as unknown as GetVerificationKey,
```
bickelj, line 21: I was puzzled by this but I'm also puzzled by my earlier comment but I see that this has been overcome by a later version of mockJWKS, hurray!

Issue #1547, opened 2025-03-21
Remove unnecessary assert operations


slifty:
There are a series of `assert` operations in the code base which may no longer be necessary now that (1) we are increasingly relying on our queries to understand authorization context and (2) our delete operations are being updated to use a generator that consistently will throw NotFoundErrors if no record was deleted.

For example, `assertUserGroupChangemakerPermissionExists` is used in one place (when trying to remove a `UserGroupChangemakerPermission`, in order to first see if the permission exists).  The query itself is written in a way that will not delete the record (and will throw a NotFoundError) if the record doesn't exist or had already been deleted.

Let's go through our assert operations and see if they are needed / if a quick refactor would make it possible to remove them.

This is somewhat related to #1407 since these operations haven't been migrated to use auth yet, and if we can remove them then we won't need to migrate them.

slifty:
I'm actually going to mark this as complete -- we no longer have assert operations from a `db/operations` PoV.  There are a few bespoke `assertFoo` functions that are defined in specific handlers.  It may make sense to remove those, but in the short term I think the intention of this particular issue is now complete.

PR #1551, opened 2025-03-21
Remove unnecessary assert operations

slifty:
This PR removes a slew of assert operations which are not actually necessary now that our delete queries will throw a NotFoundError if an item did not exist to be delted.

Related to #1547


bickelj (approved):
This looks great to me, a wonderful subtraction. I skimmed the code here and see that the tests pass here.

I still see an `assertBaseFieldExists` but that should not hold up this PR.

slifty:
+1 to the callout that there are a handful of artisanal asserts inside our handlers still in here (I'm hoping to go around and check on those as well at some point).

PR #1557, opened 2025-03-25
Ensure transaction is complete before responding

### LINKED ISSUE #1550: `POST /proposalVersions` provides a response before its transaction is committed
```
Summary of all failing tests
FAIL src/__tests__/proposalVersions.int.test.ts (5.907 s)
  ● /proposalVersions › POST / › creates exactly the number of provided field values

    expect(received).toEqual(expected) // deep equality

    Expected: 2
    Received: 0

      395 | 				],
      396 | 			});
    > 397 | 			expect(after.count).toEqual(2);
          | 			                    ^
      398 | 		});
      399 |
      400 | 		it('returns 400 bad request when no proposal id is provided', async () => {

      at Object.<anonymous> (src/__tests__/proposalVersions.int.test.ts:397:24)


Test Suites: 1 failed, 44 passed, 45 total
Tests:       1 failed, 470 passed, 471 total
Snapshots:   0 total
Time:        176.215 s
```

This has now happened a few times and we should look into what's going on.

What's particularly strange to me is that the body response is not failing (which signals that the values *are* being created...).

slifty:
This PR fixes a bug where a handler was resolving inside of a transaction; this could have meant double-responses in the event of a failed transaction, but more immediately it was resulting in intermittent test failures as the test would sometimes load data from the pre-transaction data state.

Resolves #1550

### DIFF
--- a/src/handlers/proposalVersionsHandlers.ts
+++ b/src/handlers/proposalVersionsHandlers.ts
@@ -163,7 +163,7 @@ const postProposalVersion = async (req: Request, res: Response) => {
 			),
 			assertSourceExists(sourceId),
 		]);
-		await db.transaction(async (transactionDb) => {
+		const finalProposalVersion = await db.transaction(async (transactionDb) => {
 			const proposalVersion = await createProposalVersion(transactionDb, req, {
 				proposalId,
 				applicationFormId,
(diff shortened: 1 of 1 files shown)


bickelj (approved):
Wow, nice catch. I mean, I thought this was going to be a test change but it was actually our tests revealing a race condition. I didn't run it locally, but looking at the code here this makes sense: all the statements are awaited and the transaction must finish before returning data. I checked locally to see if any other use of transactions is in the codebase but I don't find any others.

Thank you for this!

PR #1563, opened 2025-03-27
Use auth context agent for `createdBy` population

### LINKED ISSUE #1479: Use auth context for `createdBy` population
We have a `createdBy` field in a few places.  Now that we have auth context provided for creation operations / queries consistently we actually should use that instead of accepting a `createdBy` query parameter.

slifty:
This PR updates all of our `createdBy` attributes to be populated using the user ID of the agent's auth context.

The only "wonky" aspect of this change is task runners need to generate a fake auth context based on the user ID associated with the task; in future we might consider somehow passing a true auth context to the task runner, but for now this seemed like a reasonable (and reasonably contained) approach.

Resolves #1479

### DIFF
--- a/src/tasks/processBulkUploadTask.ts
+++ b/src/tasks/processBulkUploadTask.ts
@@ -23,6 +23,7 @@ import {
 	loadChangemakerByTaxId,
 	updateBulkUploadTask,
 	loadSystemUser,
+	loadUserByKeycloakUserId,
 } from '../database/operations';
 import { TaskStatus, isProcessBulkUploadJobPayload } from '../types';
 import { fieldValueIsValid } from '../fieldValidation';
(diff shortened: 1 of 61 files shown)


hminsky2002 (approved):
If PRs were measured solely by numbers of files changed, I'd rate this 61. Looks good to me, all tests passing locally and it looks like you've covered every operation. Great work @slifty!

src/tasks/processBulkUploadTask.ts
```diff
@@ -325,19 +326,31 @@ export const processBulkUploadTask = async (
 		});
 		csvReadStream.pipe(parser);
 		let recordNumber = 0;
+
+		// This is a monkey patch to create an "auth context" for the sole purpose
+		// of populating `createdBy`.  This is shallow, and if we ever update our create
+		// queries to require a full auth context, this will not be sufficient.
+		const userAgentCreateAuthContext = {
```
hminsky2002, line 333: Out of curiosity, what is the approach you're thinking of for upgrading this to a full auth context? Would that be one that would be created as part of a migration specifically for the task runner?

slifty: I thought of a few possible approaches

(1) add a JWT attribute to tasks and have the taskrunner re-parse-and-turn-them-into-auth-contexts.
(2) Add an auth context attribute to tasks as json.
(3) Start storing auth contexts in a table, and link it to tasks. 
(4) ???

That said, part of why I decided to avoid shaving that yak is they all had tradeoffs and would have added some complexity.  I do think it is something to consider eventually though, and should make an issue to at least capture the thought.

hminsky2002: Seems like a solid list. Would the auth context be passed from the user in 1-2? I feel like the task inheriting the authorization from the user that ran the task makes the most sense, since a user without the proper authorization shouldn't have been able to run the task in the first place?

slifty: The task would be acting on behalf of the user who created the task (so yes!)

PR #1564, opened 2025-03-28
299 validate currency fields

### LINKED ISSUE #299: Validate and normalize currency
We have several base fields relating to money: `organization_operating_budget`, `proposal_amount_requested`, `proposal_budget`, `proposal_project_budget`, and so on.

This shares some of the challenges as #282, in that we are trying to represent multiple disparate concepts into a single field, but this has the advantage of having less variation in those concepts: we always want to have a number, and a unit of currency. For example, "7 USD", or "1000000 CAD". We should use the currency unit codes defined by [ISO_4217](https://en.wikipedia.org/wiki/ISO_4217).

This also shares some similarities with #104, in that we are encoding a pair of data. As with names and addresses, I'd advocate for not splitting this information up, but rather encoding it into a single piece of data. For now, as our schema uses strings for everything, I'd suggest strings in the form `{amount} {code}`, where `amount` is the number without any [digit grouping separators](https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping) (commas) and using a period (`.`) for the decimal separator, and `code` is the 3-letter ISO 4217 currency code.

We may have some data that allows for a range of values, eg a operating budget of "one to two million dollars (US)", currently encoded as $1M-2M. If so, we'll need to figure out what to do about such data; perhaps that a field that has two values, each a currency, representing the low and high values of that range?

hminsky2002:
This PR adds a currency validation to our existing base field type
validation, introucing the ['validator' package](https://github.com/validatorjs/validator.js) which validates
ISO_4217 Currency Codes. 

It it is an opinionated validation per @jasonaowen 's specifications in Issue #299, where a valid currency is in the form '{amount} {code}', where amount is a number without any digit grouping separators, uses a period as the decimal separator (with a required two decimal values), and code is a 3-letter ISO 4217 currency code. 

I understand adding a new package is never a trivial task, happy to rethink the approach here, but from a preliminary investigation validator seems like it's at least as well-maintained as 'phone' (also, 'validator' offers mobile phone validation! I removed phone from the project and replaced it with validator's mobile phone validation and am passing all unit tests).

Closes #299


src/fieldValidation.ts
```diff
@@ -24,7 +24,24 @@ const isString = ajv.compile({
 	type: 'string',
 });
 
-const isPhoneNumberString = (value: string) => phone(value).isValid;
+const isPhoneNumberString = (value: string) => validator.isMobilePhone(value);
```
slifty, line 27: Sanity check: `isMobilePhone` is just what they call the phone validation in general?  Might actually be worth a quick comment to that effect so future readers know that this is expected to validate non-mobile numbers as well.

slifty (changes requested):
One test name tweak

src/fieldValidation.ts
hminsky2002: @slifty yes! i wondered the same thing, but it seems to be the only phone related function in the library. I'll add a comment to clarify.

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -82,4 +82,45 @@ describe('field value validation against BaseFieldDataType', () => {
...
 	});
+	test('validate a valid currency as CURRENCY', () => {
+		expect(
+			fieldValueIsValid('1000000.00 CAD', BaseFieldDataType.CURRENCY),
+		).toBe(true);
+		expect(fieldValueIsValid('7.00 USD', BaseFieldDataType.CURRENCY)).toBe(
+			true,
+		);
+	});
+	test('validate a valid currency as CURRENCY', () => {
```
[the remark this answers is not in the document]
hminsky2002: good catch...not...not! which is to say, good catch

slifty (approved):
Yay!

PR #1565, opened 2025-03-28
Add delete query, operation and endpoint for sources

### LINKED ISSUE #1411: Add DELETE endpoints for `/sources`
As we start to load more and more data now in 2025, we'll need some ability to delete bad records imported or created. It has not been a issue in the past, but as we have more and more coming it, we probably need the capability here through the API for admin.

hminsky2002:
This PR adds a delete operation for sources, which can be used to delete a source IF it is not linked to any other data in the PDC. In the future we may want to broach deleting linked data, but currently this is intended for the use case where a source is created by mistake.

Closes #1411


src/__tests__/sources.int.test.ts
```diff
@@ -550,4 +555,81 @@ describe('/sources', () => {
...
+
+			await agent
+				.delete(`/sources/${localSource.id}`)
+				.type('application/json')
+				.set(adminUserAuthHeader)
+				.expect(422);
+
+			const after = await loadTableMetrics('sources');
+
+			expect(before.count).toEqual(2);
```
slifty, line 631: Let's start moving our before / after checks a bit!

We don't actually need to validate that before is 2 -- rather, we just want to run:

`expect(after.count).toEqual(before.count)`

Does that sound reasonable?

src/__tests__/sources.int.test.ts
```diff
@@ -550,4 +555,81 @@ describe('/sources', () => {
...
+				.set(adminUserAuthHeader)
+				.expect(200);
+
+			const after = await loadTableMetrics('sources');
+
+			expect(before.count).toEqual(2);
+			expect(after.count).toEqual(1);
+		});
+
+		it('throws an error when it tries to delete a source that is associated with a proposal', async () => {
```
slifty, line 591: Let's instead say `Returns 422 Unprocessable Content`

This is more descriptive and more accurate (it doesn't throw an error!)

src/__tests__/sources.int.test.ts
```diff
@@ -550,4 +555,81 @@ describe('/sources', () => {
...
+				createdBy: testUser.keycloakUserId,
+			});
+			const applicationForm = await createApplicationForm(db, null, {
+				opportunityId: 1,
+			});
+			await createProposalVersion(db, null, {
+				proposalId: proposal.id,
+				applicationFormId: applicationForm.id,
+				sourceId: localSource.id,
+				createdBy: testUser.keycloakUserId,
```
slifty, line 619: I'm about to hit merge in a way that makes you sad and for this I'm sorry

slifty (changes requested):
I left a comment out of this review which also is included whoops!

Requested change to the test name, *suggested* change to the assertions.

Also, I'm about to merge the `createdBy` pr which will unfortunately impact these tests, happy to hop on a call to explain what you have to do differently if you have any questions.

src/__tests__/sources.int.test.ts
hminsky2002: not just reasonable, that's slick! makes a lot more sense since we care about the difference in values being zero, not the difference equaling an arbitrary value.

hminsky2002:
@slifty should be good to go, also realized i forgot to add swagger spec so tossed that in as well

slifty (approved):

Issue #1568, opened 2025-03-28
Enable `no-empty-servers` lint rule


slifty:
We don't have the `servers` section defined in our swagger spec; the recommended redocly ruleset would have [no-empty-servers](https://redocly.com/docs/cli/rules/oas/no-empty-servers) enabled.

This has enough potential implications on our deployment that I would like to investigate it separately from #1177.

slifty:
We looked into this and determined that we do NOT want this rule enabled -- since we're using swaggerUI it appears that the "server" is handled in a smart way by that frontend.

PR #1569, opened 2025-03-28
Use `redocly` for swagger linting

### LINKED ISSUE #1177: Switch our openAPI linter to `@redocly/cli`

slifty:
This PR switches to a *much better linter* holy moly!

It also enables the recommended strict rules (minus one, which we have #1568 to cover eventually).

There are still more improvements to be made to our spec, including more rules we might want to consider enabling, but this is a solid baseline and a meaningful set of changes.

Resolves #1177


slifty:
Humans please take note that I asked a robot to review *as a direct threat*

bickelj (approved):
This is uuuge, in a good way, of course!

When checking the license(s), I found a confusing (non-existent) license here: https://www.npmjs.com/package/@redocly/openapi-docs?activeTab=code, however, in other places on GH I see MIT (Expat) license. All good, I suppose.

When looking over the dependencies in `package-lock.json`, I saw some telemetry stuff. So I checked if it phoned home, and it appears redocly phones home on `npm run lint:openapi` (as does `npm`, apparently):
redocly_phones_home_2025-03-31_12-23-46

I found [the docs to disable](https://redocly.com/docs/realm/setup/reference/telemetry#opt-out) which is via environment variable, confusingly. Perhaps we can add that to our `.env.example`, but it seems like everybody has to individually make sure to set this variable to avoid telemetry. Regardless, if our previous solution made web service calls to a central service, `redocly` is a big step in the right direction.

I was able to run it locally, change a rule and see its effect, and I compared our rules with the recommended rules to see that we are mostly more strict than theirs.

I cannot understand why people think it's an error to not put the direct link to an instance of a service that implements an API in the API docs. I understand why they want to do it, for those services that are single-instance, but I don't think it's an error to strictly separate API from service. Thus the name of our repository.

bickelj:
Oh, I also see that our CI file may need to be updated to account for the new name `openapi` rather than `swagger`, or perhaps we can keep the old name.

slifty:
> Oh, I also see that our CI file may need to be updated to account for the new name openapi rather than swagger, or perhaps we can keep the old name.

I did already update our CI as part of this PR I believe, unless I'm missing an instance somewhere (If you're talking about the check -- that will be a change in the github UX once this is merged)

I do want to move to the less vendor-specific name (my understanding is that swagger is just one tool around the openapi spec).

Issue #1178, opened 2024-09-10
Break our openAPI json file into reusable pieces rather than a single monolith


slifty:

slifty:
We can use this repo as inspiration: https://github.com/PermanentOrg/stela/tree/main/packages/api/docs/future

PR #1588, opened 2025-04-07
Separate parameters in OpenAPI specification

slifty:
This PR continues our effort to break the OpenAPI documentation in to more granular components, specifically moving on the `parameters` section into individual files.

Related to #1178


hminsky2002 (approved):
Make it bigger! More commits! Just kidding, this is very digestible and looks good, didn't notice any typos and seems like all parameters were covered

PR #1590, opened 2025-04-07
Separate OpenAPI responses

slifty:
This PR separates our lonely `NotFound` response into its own file.  This may seem silly but it eventually we'll need this when we move routes to their own files too.

Related to #1178


hminsky2002 (approved):
Swell!

Issue #1559, opened 2025-03-25
Implement audit logging (database row level, 1 of 3)


bickelj:
Do the things thought of in #1355

bickelj:
It is expected that audit logs can be presented in an administrative interface. It is also expected that some reads would be included, not only writes. This from a meeting 2025-04-01 (no joke).

bickelj:
<details>
<summary>Trying audit-trigger locally</summary>

When adding [this trigger-audit SQL](https://github.com/iloveitaly/audit-trigger/blob/a1046331d8073b8444d90bda9b4700809b77800d/audit.sql) directly to a migration file (does not yet pass lint) we end up with this table and these triggers:

```
pdc=> \d audit.logged_actions
                                                 Table "audit.logged_actions"
      Column       |           Type           | Collation | Nullable |                        Default                         
-------------------+--------------------------+-----------+----------+--------------------------------------------------------
 event_id          | bigint                   |           | not null | nextval('audit.logged_actions_event_id_seq'::regclass)
 schema_name       | text                     |           | not null | 
 table_name        | text                     |           | not null | 
 relid             | oid                      |           | not null | 
 session_user_name | text                     |           |          | 
 action_tstamp_tx  | timestamp with time zone |           | not null | 
 action_tstamp_stm | timestamp with time zone |           | not null | 
 action_tstamp_clk | timestamp with time zone |           | not null | 
 transaction_id    | bigint                   |           |          | 
 application_name  | text                     |           |          | 
 client_addr       | inet                     |           |          | 
 client_port       | integer                  |           |          | 
 client_query      | text                     |           |          | 
 action            | text                     |           | not null | 
 row_data          | jsonb                    |           |          | 
 changed_fields    | jsonb                    |           |          | 
 statement_only    | boolean                  |           | not null | 
 row_id            | bigint                   |           |          | 
Indexes:
    "logged_actions_pkey" PRIMARY KEY, btree (event_id)
    "logged_actions_action_idx" btree (action)
    "logged_actions_action_tstamp_tx_stm_idx" btree (action_tstamp_stm)
    "logged_actions_relid_idx" btree (relid)
Check constraints:
    "logged_actions_action_check" CHECK (action = ANY (ARRAY['I'::text, 'D'::text, 'U'::text, 'T'::text]))
pdc=> \df audit.*
                                                                           List of functions
 Schema |       Name       | Result data type |                                               Argument data types                                               | Type 
[...]

Issue #1266, opened 2024-10-21
Graphile-worker runs `UPDATE` statements very frequently


bickelj:
From [deploy#112](https://github.com/PhilanthropyDataCommons/deploy/issues/112#issuecomment-2422624674): can we go back to the pg-as-a-queue library and set some less aggressive polling on it?

As of #630, there are multiple `UPDATE` statements per second running on the postgres database. While postgres handles this and the queue works, it has the unintended consequences of spamming the logs whenever `log_statement = 'mod'` and higher-than-expected resource utilization when the PDC is truly idle.

Perhaps there is a way to tune the library to be less aggressive.

bickelj:
Screenshot of DB CPU usage and load average at DO before this change, showing around 15% to 20% CPU usage and 0.2 to 2 load average:

<img width="1472" height="917" alt="Image" src="https://github.com/user-attachments/assets/392acd03-19a2-42bc-be50-fc831a294e70" />

bickelj:
No major change visible in the CPU usage or process queue (load) yet:

<img width="1481" height="924" alt="Image" src="https://github.com/user-attachments/assets/0697c980-8e18-42c2-8d19-c3b6e6acc967" />

bickelj:
Update, minimal difference in the CPU and load graphs, but certainly there would be fewer annoying `UPDATE` statements in logs if we were able to change `log_statement`:

<img width="1488" height="916" alt="Image" src="https://github.com/user-attachments/assets/4ed3f5ad-6f37-41b8-9b60-33ff309f2301" />

PR #1607, opened 2025-04-14
Use triggers on DB tables to log data changes

bickelj:
Database triggers provide a low-level audit trail that is available regardless of whether a data change was made by the application or by a user manually logged into the database, for example, via `psql`.

This change adds triggers to existing tables that log data changes to a new table called `audit_logs`.

The Postgres wiki has an existing example script called `audit-trigger` that has been extended or modified by a variety of folks. This audit trigger script formed the basis of this change, specifically the version shown in the second line of the migration file named `0055-create-audit-log.sql`. Several adaptations have been made, namely:
* Included the license directly in the file.
* Reformatted to satisfy SQLfluff.
* Removed the separate schema as it does not serve PDC well. Instead,
* renamed the tables and triggers to include `audit` for clarity.
* Added calls to the `audit_table` for existing tables in migration.
* Removed extraneous wrapper functions.
* Removed the assumption of an integer PK named `id`.
* Log calls to `audit_table` and `deaudit_table` as action 'A', 'S'.
* Added the session/connection/pid.

To begin auditing a table (a new one, for example), call the `audit_table` function with the table name. To stop auditing a table, call the `deaudit_table` with the table name. To see which tables are currently being audited, `select * from audited_tables`.

Issue #1559 Implement audit logging

(This is the first of several planned layers of audit logging)

### DIFF
--- a/src/database/migrations/0055-create-audit-log.sql
+++ b/src/database/migrations/0055-create-audit-log.sql
@@ -1,0 +1,458 @@
+-- Adapted from
+-- https://github.com/iloveitaly/audit-trigger/blob/a1046331d8073b844/audit.sql
+-- under the following license.
+--
+-- PostgreSQL Audit Trigger Example
+-- Copyright (c) 2013, PostgreSQL Global Development Group
+--
+-- Permission to use, copy, modify, and distribute this software and its
+-- documentation for any purpose, without fee, and without a written agreement
+-- is hereby granted, provided that the above copyright notice and this
+-- paragraph and the following two paragraphs appear in all copies.
+--
+-- IN NO EVENT SHALL THE UNIVERSITY OF CALIFORNIA BE LIABLE TO ANY PARTY FOR
+-- DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES, INCLUDING
+-- LOST PROFITS, ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS
+-- DOCUMENTATION, EVEN IF THE UNIVERSITY OF CALIFORNIA HAS BEEN ADVISED OF THE
+-- POSSIBILITY OF SUCH DAMAGE.
+--
+-- THE UNIVERSITY OF CALIFORNIA SPECIFICALLY DISCLAIMS ANY WARRANTIES,
+-- INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+-- AND FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+-- ON AN "AS IS" BASIS, AND THE UNIVERSITY OF CALIFORNIA HAS NO OBLIGATIONS TO
+-- PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+-- An audit history is important on most tables. Provide an audit trigger that
+-- logs to a dedicated audit table for the major relations.
+--
+-- This file should be generic and not depend on application roles or
+-- structures as it's being listed here:
+--
+--	https://wiki.postgresql.org/wiki/Audit_trigger_91plus
+--
+-- This trigger was originally based on
+--   http://wiki.postgresql.org/wiki/Audit_trigger
+-- but has been completely rewritten.
+--
+-- Should really be converted into a relocatable EXTENSION, with control and
+-- upgrade files.
+--
+-- Audited data. Lots of information is available, it's just a matter of how
+-- much you really want to record. See:
+--
+--   http://www.postgresql.org/docs/9.1/static/functions-info.html
+--
+-- Remember, every column you add takes up more audit table space and slows
+-- audit inserts.
+--
+-- Every index you add has a big impact too, so avoid adding indexes to the
+-- audit table unless you REALLY need them. The json GIN/GIST indexes are
+-- particularly expensive.
+--
+-- It is sometimes worth copying the audit table, or a coarse subset of it that
+-- you're interested in, into a temporary table where you CREATE any useful
+-- indexes and do your analysis.
+--
+CREATE TABLE audit_logs (
+	event_id bigserial PRIMARY KEY,
+	schema_name text NOT NULL,
+	table_name text NOT NULL,
+	relid oid NOT NULL,
+	session_user_name text,
+	action_tstamp_tx timestamp with time zone NOT NULL,
+	action_tstamp_stm timestamp with time zone NOT NULL,
+	action_tstamp_clk timestamp with time zone NOT NULL,
+	transaction_id bigint,
+	application_name text,
+	client_addr inet,
+	client_port integer,
+	pid integer,
+	client_query text,
+	action text NOT NULL CHECK (action IN ('I', 'D', 'U', 'T', 'A', 'S')),
+	row_data jsonb,
+	changed_fields jsonb,
+	statement_only boolean NOT NULL
+);
+REVOKE ALL ON audit_logs
+FROM public;
+COMMENT ON TABLE audit_logs IS
+'History of auditable actions on audited tables, from '
+'audit_if_modified_func()';
+COMMENT ON COLUMN audit_logs.event_id IS
+'Unique identifier for each auditable event';
+COMMENT ON COLUMN audit_logs.schema_name IS
+'Database schema audited table for this event is in';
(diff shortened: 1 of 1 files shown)


bickelj:
@slifty and @hminsky2002 I kept the formatting as SQLfluff intended as a nudge against SQLfluff. How could it allow such terrible formatting?

src/database/migrations/0055-create-audit-log.sql
```diff
@@ -0,0 +1,439 @@
...
+ORDER BY
+	schema,
+	auditedtable;
+COMMENT ON VIEW audited_tables IS $body$ View showing all tables with auditing
+set up.Ordered by schema,
+	then table.$body$;
+
+-- Enable audit logs on PDC tables
+SELECT
+	audit_table('application_form_fields') AS aff,
```
slifty, line 415: do we need the `AS foo`s here (the output of this select isn't used, right?)

src/database/migrations/0055-create-audit-log.sql
```diff
@@ -0,0 +1,439 @@
...
+	application_name text,
+	client_addr inet,
+	client_port integer,
+	client_query text,
+	action text NOT NULL CHECK (action IN ('I', 'D', 'U', 'T', 'A', 'S')),
+	row_data jsonb,
+	changed_fields jsonb,
+	statement_only boolean NOT NULL
+);
+REVOKE ALL ON audit_logs
```
slifty, line 74: I know this is all boilerplate but in the spirit of knowing what's up -- do you know what this clause is accomplishing (the `REVOKE ... FROM public`).  Does this narrow which users are able to interact with audit_logs beyond what we have already configured?

If we eventually want to make these logs queryable via API will this mean our db user doesn't have access?

src/database/migrations/0055-create-audit-log.sql
```diff
@@ -0,0 +1,439 @@
...
+--
+-- Audited data. Lots of information is available, it's just a matter of how
+-- much you really want to record. See:
+--
+--   http://www.postgresql.org/docs/9.1/static/functions-info.html
+--
+-- Remember, every column you add takes up more audit table space and slows
+-- audit inserts.
+--
+-- Every index you add has a big impact too, so avoid adding indexes to the
```
slifty, line 47: This makes me 👀 -- specifically, if we can't add an index I'm suspecting that we're not going to end up wanting to expose it via API.  I am *NOT* suggesting any change in this PR just flagging it for future conversation (we don't even know if we want to expose things via API as opposed to dumping logs to files that get munged and parsed by a system optimized around log analysis such as Grafana Loki)

slifty (approved):
Nice stuff.

I had some questions for my own knowledge and one lint-level comment that doesn't matter ("do we need these alias definitions?")

Other than that, w00t!

I ran this locally, tried inserting a base field, and the log entry was written.

src/database/migrations/0055-create-audit-log.sql
bickelj: It is worth taking a closer look, good call.

Before looking,  my rudimentary understanding is that this means only the owner and superuser will be able to use/see the table directly. At the moment, I think we use the same database user at runtime as migration-time, so there should be no issue querying it at runtime. Looking...

src/database/migrations/0055-create-audit-log.sql
bickelj: First, there are several indices on the audit table out of the box. So I doubt we need to add some. It should be ready to query:
```
Indexes:
    "audit_logs_pkey" PRIMARY KEY, btree (event_id)
    "audit_logs_action_idx" btree (action)
    "audit_logs_action_tstamp_tx_stm_idx" btree (action_tstamp_stm)
    "audit_logs_relid_idx" btree (relid)
```

If our task is to build ELK or something like it, then we are going down an abortive path here with the DB triggers. The advantage of using database triggers for audit logs is we can continue building our APIs and UIs in the same manner as the rest of the application without adding dependencies. We can still change course later if it is not working as we wish.

I think of log aggregation and interfaces into them as tools for the producers of the software, not for the consumers. To extend my favorite metaphor, it's a display inside the kitchen, not something in the dining room.

Having written software that captured logs and forwarded them to a log aggregation/display/query tool, I suspect the trigger approach will be more reliable. I expect triggers to capture more than 99.999% of changes to data (when active). In the cases where triggers do not capture them, it would be in a situation where the software is simply unavailable.

If our goal is to be able to actually trace whence a datum came, we need the complete log, not the last couple of years or months of logs, such as what log aggregation tools usually expect. If we need to prune these data, I don't think it will be so simple as "delete the earliest year." It would be more like "keep the last N changes to currently surfaced data, however old or new."

You do highlight a danger, though, which I think this warning highlights: storing every dang interaction with the software means this will quickly become the biggest table. We will need to vigilantly watch it and suppress data that is obviously useless while keeping data that may potentially be useful.

src/database/migrations/0055-create-audit-log.sql
bickelj: I think this was a SQLfluff requirement. Let me check.

src/database/migrations/0055-create-audit-log.sql
```diff
@@ -0,0 +1,439 @@
...
+	action_tstamp_tx,
+	action_tstamp_stm,
+	action_tstamp_clk,
+	transaction_id,
+	application_name,
+	client_addr,
+	client_port,
+	client_query,
+	action,
+	statement_only)
```
bickelj, line 259: Given that some calls can have filters on them, it would be useful to include those args in the `row_data` so that we can see whether to expect certain changes or not.

src/database/migrations/0055-create-audit-log.sql
slifty: Ty ty this is great

Just to make sure what I was trying to get at is clear -- I really was NOT saying we don't need this or suggest this PR isn't the right solution; I was just realizing that exposing this data via API might not be the right approach down the line as opposed to having the logs munged and parsed by a system designed to translate Vast Amounts Of Logs Into Interfaces (e.g. https://grafana.com/oss/loki/)

src/database/migrations/0055-create-audit-log.sql
slifty: 🤦🏻 @ sqlfluff

bickelj: Yes, blame SQLfluff. I tried to remove one of them and got this:
```
service$ npm run lint:sqlfluff

> lint:sqlfluff
> ./venv/bin/sqlfluff lint src/database/

== [src/database/migrations/0055-create-audit-log.sql] FAIL                                  
L: 424 | P:   2 | AL03 | Column expression without alias. Use explicit `AS`                  
                       | clause. [aliasing.expression]
All Finished 📜 🎉!                                                                          
service$ echo $?
1
```

That's, what, strike 8 against SQLfluff?

slifty: I think just leaving as is (with aliases) is fine for now (alternatively you could disable the rule via a comment). I was just curious the reason!

bickelj: I tried several test runs (of sources.int.ts) in a row with and without the independent statements and I cannot find a substantial performance difference. Both take about 35 to 38 seconds on my machine. Seems long for both. Perhaps I haven't turned off synchronous commit or something.

src/database/migrations/0055-create-audit-log.sql
bickelj: For sure, agree. We don't have either interface at the moment, so we will likely want the log aggregation in future, and this level of detail may fall into the kitchen rather than being presented in a UI. At the moment the obstacle to logging `mod` statements to postgres logs is the #1266 issue. If we can somehow suppress those at the library or schema level, we could use logs and log aggregation. Otherwise I think this audit table will be indispensable.

src/database/migrations/0055-create-audit-log.sql
bickelj: So since the same user running migrations is the runtime user, I am confident that user should be able to query at runtime. I use the same user locally too, and it can access the table.

As for the practical effect of this `REVOKE` statement, in our case, it should be a no-op, meaning any existing users will not actually have access to new tables by default. I tried setting up a user like I had done in test and prod, and regardless of the presence of `REVOKE...` the user could not access the new `audit_logs` table.

Still, the `REVOKE` statement gives me warm fuzzies in the obscure case that we have some other non-admin user that for whatever reason has access to the `pdc` database.

slifty (approved):
w00t!  Changes look good to me.

PR #1608, opened 2025-04-15
Use natural key for basefields

### LINKED ISSUE #1141: Use natural key `short_code` as primary key for base fields
We should drop the surrogate `id` key from base fields, and use the natural `short_code` key everywhere.

# Background

Currently, the base fields table looks like `id integer primary key, short_code text unique not null, created_at...`. Other tables that refer to the base fields table do so via the integer `id`. The integer `id` is also exposed in several endpoints, including `GET /baseFields/`, `POST /applicationForms/` and `PUT /baseFields/:id`.

Using a surrogate key (aka artificial key or synthetic key) is a trade-off. In general, some of its benefits include that it improves the performance of `JOIN`s, and limits the size of database tables with foreign keys. However, surrogate keys are (by definition) opaque, and so they reduce legibility anywhere the surrogate ID is used: `id=132` has no meaning to a reader, but `short_code=proposal_budget` does.

# The Problem

In the case of base fields in the PDC, the surrogate ID is both opaque *and* unstable.

It is opaque in that using the surrogate ID makes it more difficult to read the database, as you need to do additional joins to get the short code. If the foreign key used the short code, it would become much clearer. Likewise, the few URLs we have that include the base field ID convey very little meaning in themselves - just that it will give some information about some base field - whereas a URL that included the short code would more clearly communicate its purpose.

It is unstable in that two instances of the PDC service with the "same" base field (meaning the same short-code and, presumably, label) can have different surrogate IDs, depending on the order the base fields were added.

As a result, clients need to work more-or-less exclusively with short codes. A client cares about specifying the `proposal_name`, not specifying `base field id 103`. That means that their workflow always has to include looking up the list of base fields and finding the relevant ID so that they can feed it back into the system elsewhere.

For example, the front-end needs to show [a preselected list of base fields](https://github.com/PhilanthropyDataCommons/front-end/blob/35cb42479ba077cd0580dc402245490030c7c717/src/utils/proposals.ts#L27-L33) as column headers. It must list those fields by short code, since it can't guarantee what ID the API will use. Then, it has to find the relevant base field from the id-sorted list returned by the API to find the label & description. This leads to some pretty convoluted code, such as the [`getValueOfBaseField` function](https://github.com/PhilanthropyDataCommons/front-end/blob/35cb42479ba077cd0580dc402245490030c7c717/src/utils/proposalFields.ts#L19-L34) or [whatever this thing I wrote does](https://github.com/PhilanthropyDataCommons/front-end/blob/main/src/map-proposals.ts).

As another example, the code for #756 will have the same concern: it cannot assume that the base fields in the target instance have the same IDs as the source instance, so it will need to act by short code.

The front-end, and probably every other client that works with the PDC, would be much simpler if it could just use short codes everywhere.

Further, the primary benefit of using surrogate keys - performance - isn't relevant at our scale, and is unlikely to be any time soon. So we have the drawbacks, but not the benefits.

# Mutability

## Theory

One of the arguments sometimes used in favor of surrogate keys is the concern of updating the data: what do you do if the data you need to update is (part of) the primary key?

Database-wise, that's not a problem; we can simply specify `ON UPDATE CASCADE` on our [foreign keys](https://www.postgresql.org/docs/16/ddl-constraints.html#DDL-CONSTRAINTS-FK), and when you update the short code, the database will update the short code in each reference.

## URLs

The other concern around mutability is the changing of URLs, and in particular `PUT /baseFields/:id` and (soon) `PUT /baseFields/:id/:lang`. Currently, those endpoints are only available users with administrator access, and we can just talk to each of those few people in the uncommon case of renaming a short code. Even with that mitigation aside, I believe that legible URLs are more valuable than forever URLs (at least in this case - I'll have to reflect if I believe that more broadly!).

## Practice

While I anticipate that we will occasionally need to change a short code, I suspect it will be a rare occurrence. Changing a base field that is in use - regardless of the key strategy - will change the meaning of every application and proposal that refers to it. Therefore, I expect such changes to be limited to fixing typos and so on; changes that don't affect the meaning of the field.

# Refactoring

The process is slightly tedious but not at all difficult; something like:

- create new FKs pointing to `base_fields (short_code)` alongside existings FK pointing to `base_fields (id)`
- populate the new column
- make the new column `NOT NULL`, if needed
- drop the old FK column
- set the `short_code` to be the new primary key
- drop the `id` column

# Broader implications

I am not suggesting that we use natural keys everywhere in the PDC. I'm not sure where else the opportunity even exists, and if and when we identify one, I'd want to take it on a case-by-case basis. I think it's a clear win with base fields, but that doesn't mean natural keys would necessarily be better elsewhere.

# Further reading
- [Primary Keyvil, reprised by Josh Berkus](https://www.databasesoup.com/2015/03/primary-keyvil-reprised.html)
- [Celko On SQL: Identifiers and the Properties of Relational Keys](https://www.informationweek.com/data-management/celko-on-sql-identifiers-and-the-properties-of-relational-keys)
- [Celko on SQL: Natural, Artificial and Surrogate Keys Explained](https://www.informationweek.com/it-sectors/celko-on-sql-natural-artificial-and-surrogate-keys-explained)
- [Choosing a Primary Key: Natural or Surrogate?](https://agiledata.org/essays/keys.html)
- [You'll regret using natural keys by Mark Seemann](https://blog.ploeh.dk/2024/06/03/youll-regret-using-natural-keys/)
- https://github.com/PhilanthropyDataCommons/service/issues/72

hminsky2002:
A monstrous refactor of the baseFields entity to use a natural key instead of the autogenerated numerical id. This is split into two commits:

1. which removes the 'PUT' endpoint and update related operations for the baseField entity, as we'd discussed that since we are Pkey'ing on a natural key, we no longer want to be able to allow the user to update the shortcode of a basefield (as we'd then lose reference to the entity)

2. which contains the database migration, query updates, and complete code refactor to handle the pkey change in baseField (and fkeys in applicationFormField and baseFieldLocalization)

Understandably, commit 2 is hard to digest, and I'm open to splitting it up, but I decided that it was the best way to ensure that the commits were not breaking.

Closes #1141


src/__tests__/baseFields.int.test.ts
```diff
@@ -319,183 +319,6 @@ describe('/baseFields', () => {
 		});
 	});
 
-	describe('PUT /:baseFieldId', () => {
```
slifty, line 322: We don't want to remove PUT!

There are a lot of things about base fields that are updatable that are NOT the short code (the scope, the description, the label, etc).  We just want the short code to not be an editable component.

I think you can remove this commit entirely, actually, and when shortCode is updated to be `readonly` it will naturally stop being editable by the PUT (you will need to make some changes to the generated operations + queries to remove `shortCode` as a parameter.)

src/database/migrations/0055-alter-base_fields-use-shortcode-as-primary-key.sql
```diff
@@ -0,0 +1,62 @@
...
+UPDATE base_field_localizations
+SET base_field_short_code = base_fields.short_code
+FROM base_fields
+WHERE base_field_localizations.base_field_id = base_fields.id;
+UPDATE application_form_fields
+SET base_field_short_code = base_fields.short_code
+FROM base_fields
+WHERE application_form_fields.base_field_id = base_fields.id;
+
+
```
slifty, line 17: Extra whitespace (another ding against sqlfluff!)

src/database/migrations/0055-alter-base_fields-use-shortcode-as-primary-key.sql
```diff
@@ -0,0 +1,62 @@
...
+ALTER TABLE application_form_fields
+ALTER COLUMN base_field_short_code SET NOT NULL;
+
+-- 4. Drop the existing foreign key constraint on base_field_id.
+ALTER TABLE base_field_localizations
+DROP CONSTRAINT base_field_localizations_base_field_id_fkey;
+ALTER TABLE application_form_fields
+DROP CONSTRAINT fk_base_field;
+
+-- 5. Drop the existing primary key constraint on base_field_localizations.
```
slifty, line 30: I may be wrong (plz fact check; maybe you already did), but I think all these constraints get dropped automatically when the column is dropped.  If that is the case then I think you can remove these steps and JUST drop the columns.

src/openapi/api.json
```diff
@@ -1550,9 +1550,9 @@
 				}
 			}
 		},
-		"/baseFields/{baseFieldId}/localizations": {
```
slifty, line 1553: I'm sadly about to make a merge conflict by merging the path refactor.

That said, we should also be updating the appropriate schema definitions in the api spec.

src/__tests__/baseFields.int.test.ts
hminsky2002: Doy! Yeah that's the right call, still clearly valuable to have update operations on non-shortcode fields.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.18.0 2025-02-13
```
slifty, line 10: Remember to also update the version in api.json too (otherwise sdk won't trigger a rebuild)

src/database/migrations/0055-alter-base_fields-use-shortcode-as-primary-key.sql
hminsky2002: Seems to be the case! Dropped and ditched

src/openapi/api.json
hminsky2002: I gotta say, the work you did in refactoring our api spec made refactoring the actual schema and path definitions a breeze. Woohoo!

src/database/migrations/0057-alter-base_fields-use-shortcode-as-primary-key.sql
```diff
@@ -0,0 +1,52 @@
...
+UPDATE base_field_localizations
+SET base_field_short_code = base_fields.short_code
+FROM base_fields
+WHERE base_field_localizations.base_field_id = base_fields.id;
+UPDATE application_form_fields
+SET base_field_short_code = base_fields.short_code
+FROM base_fields
+WHERE application_form_fields.base_field_id = base_fields.id;
+
+
```
slifty, line 17: Harry, this extra line of whitespace will haunt you until the end of your days because it is in a migration file.

src/database/operations/baseFields/updateBaseField.ts
```diff
@@ -1,21 +1,14 @@
 import { generateCreateOrUpdateItemOperation } from '../generators';
-import type { BaseField, WritableBaseField } from '../../../types';
+import type { BaseField, ShortCode, WritableBaseField } from '../../../types';
 
 const updateBaseField = generateCreateOrUpdateItemOperation<
```
slifty, line 4: This also shouldn't exist!  We can now just use `createOrUpdateBaseField` everywhere.  Life gets so simple.

src/database/queries/baseFields/updateByShortCode.sql
```diff
@@ -1,9 +1,8 @@
 UPDATE base_fields SET
```
slifty, line 1: This file name should now be `createOrUpdateOne` and add in the upsert logic (borrowing from other upserts, such as the `UserGroupFunderPermission` operations)

src/handlers/baseFieldsHandlers.ts
```diff
@@ -40,12 +45,15 @@ const postBaseField = async (req: Request, res: Response) => {
...
-	req: Request<{ baseFieldId: string }>,
+	req: Request<{ baseFieldShortCode: string }>,
 	res: Response,
 ) => {
-	const baseFieldId = Number.parseInt(req.params.baseFieldId, 10);
-	if (Number.isNaN(baseFieldId)) {
-		throw new InputValidationError('Invalid id parameter.', isId.errors ?? []);
+	const { baseFieldShortCode } = req.params;
+	if (!isShortCode(baseFieldShortCode)) {
+		throw new InputValidationError(
```
slifty, line 53: Codecov is right on this one -- this should have test coverage.

src/handlers/baseFieldsHandlers.ts
```diff
@@ -54,60 +62,58 @@ const putBaseField = async (
...
+const getBaseFieldLocalizationsByBaseFieldShortCode = async (
+	req: Request<{ baseFieldShortCode: ShortCode }>,
 	res: Response,
 ) => {
-	const baseFieldId = Number.parseInt(req.params.baseFieldId, 10);
-	if (!isId(baseFieldId)) {
-		throw new InputValidationError('Invalid id parameter.', isId.errors ?? []);
+	const { baseFieldShortCode } = req.params;
+	if (!isShortCode(baseFieldShortCode)) {
+		throw new InputValidationError(
```
slifty, line 88: We should have a test for this one too.

src/handlers/baseFieldsHandlers.ts
```diff
@@ -28,10 +33,10 @@ const getBaseFields = async (req: Request, res: Response) => {
 };
 
 const postBaseField = async (req: Request, res: Response) => {
```
slifty, line 35: There's no longer such a thing as posting a base field -- only put (the route and openAPI spec should be removed too)

slifty (changes requested):
Apologies for another slew of changes requested but this is coming along!

One meta note: let's make sure that the naming conventions for and parameter patterns in here are consistent with other natural-key-based operations.  Specifically take a keen eye towards:

1. the query file name
2. the operation file name + parameters
3. the writable type

I'm pretty sure I gave the right guidance in my comments, but it's worth doing an extra check.  One thing to consider is that generally an upsert operation is assumed to be based on primary key so that doesn't need to be specified in the name nor, I think, would it be a separate parameter outside of the upsert object (e.g. `createOrUpdateBaseField(db, auth, { objectInfoIncludingPrimaryKey })`).

---

Also some tests we should add, and removal of some operations / routes that are obviated by the Newly Universal Operation

src/handlers/baseFieldsHandlers.ts
```diff
@@ -54,60 +48,53 @@ const putBaseField = async (
...
+const getBaseFieldLocalizationsByBaseFieldShortCode = async (
+	req: Request<{ baseFieldShortCode: ShortCode }>,
 	res: Response,
 ) => {
-	const baseFieldId = Number.parseInt(req.params.baseFieldId, 10);
-	if (!isId(baseFieldId)) {
-		throw new InputValidationError('Invalid id parameter.', isId.errors ?? []);
+	const { baseFieldShortCode } = req.params;
+	if (!isShortCode(baseFieldShortCode)) {
+		throw new InputValidationError(
```
slifty, line 69: Whoops, we should add a test for this case

src/types/BaseField.ts
```diff
@@ -117,19 +113,62 @@ const writableBaseFieldSchema: JSONSchemaType<WritableBaseField> = {
...
 		'label',
 		'description',
-		'shortCode',
 		'dataType',
 		'scope',
 		'valueRelevanceHours',
 	],
 };
 
+const internallyWritableBaseFieldSchema: JSONSchemaType<InternallyWritableBaseField> =
```
slifty, line 122: I don't think this (or the related type guard) is actually used anywhere, right?  So we wouldn't need to make it.

slifty (changes requested):
Two little things (which I will take on since you're on vacation!)

slifty:
Rebased, added that one test, removed the dead code, and force pushed, this means unfortunately I GET THE CREDIT MWAHHAHAHAA (jk sorry harry; I did add us both as co-authors though)

slifty (approved):
This is lookin' good!  I mentioned this in commit but I rebased which meant needing to update them migration file name / sequence, and added the missing test but changed no functionality!

slifty:
(since Harry isn't here for a while I'll smoosh the merge button)

hminsky2002:
@slifty (i'm not here, but thank you muchly for finishing this up :) )

PR #1610, opened 2025-04-15
Add relevance duration

### LINKED ISSUE #1578: Add the concept of "stale time" to base fields
We want to allow base fields to have an attribute indicating how long it should take for a given value to become stale (e.g. org names maybe never go stale, whereas annual budgets go stale after a year)

slifty:
This PR adds a new `relevanceDuration` attribute to base fields.  When not-null this attribute indicates if values associated with a given base field become less relevant after a certain amount of time.

This PR doesn't actually utilize the new attribute (e.g. when returning gold data for an organization)

Resolves #1578


src/openapi/components/schemas/BaseField.json
```diff
@@ -34,6 +34,11 @@
 			"type": "string",
 			"enum": ["proposal", "changemaker"]
 		},
+		"relevanceDuration": {
```
slifty, line 37: Would it be clearer to call this `valueRelevanceDuration` (as written the name *could* be interpreted as signaling how long the base field itself was relevant, rather than value relevance)

src/__tests__/applicationForms.int.test.ts
```diff
@@ -29,13 +29,15 @@ const createTestBaseFields = async () => {
 		shortCode: 'organizationName',
 		dataType: BaseFieldDataType.STRING,
 		scope: BaseFieldScope.ORGANIZATION,
+		relevanceDuration: null,
```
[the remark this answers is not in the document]
hminsky2002: Having gone through this myself on #1141 , i whole whole WHOLE heartedly agree! Would be very happy to work on this

src/openapi/components/schemas/BaseField.json
hminsky2002: Now that you've brought it up, I am inclined to say yes. If only because we do allow for updating baseField values.

src/tasks/__tests__/copyBaseFields.int.test.ts
```diff
@@ -38,7 +39,7 @@ const createTestBaseFieldsCopyTask = async (
 	});
 };
 
-const mockFirstNameBaseField = {
+const mockFirstNameBaseField: BaseField = {
```
hminsky2002, line 42: Nice! Didn't even think about doing this, also a useful pattern to adopt for making more data in the future

hminsky2002 (approved):
Having gone through a basefield refactor myself very recently, I did not see anything that you missed coverage on. Going to 'approve' but I'll note that I agree that sINCE you thought of it, value_relevance_duration might be a more appropriate name for this field.

src/database/migrations/0055-alter-base_fields-add-relevance_duration.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE base_fields
+ADD COLUMN relevance_duration integer;
```
bickelj, line 2: This means the maximum duration we support is under 69 years. I'm trying to think if we can reasonably expect a duration of, say, a century.

How about the current president of an organization? We would not expect that to last more than about 100 years, right?

I wish we could use a more explicit duration type such as postgres' `interval` here and TypeScript's `TimeDuration` elsewhere. That might add some complexity, though.

What would it mean to have a negative duration here? Is it allowed?

src/openapi/components/schemas/BaseField.json
bickelj: I concur.

src/types/BaseField.ts
```diff
@@ -55,6 +56,10 @@ const baseFieldSchema: JSONSchemaType<BaseField> = {
 			type: 'string',
 			enum: Object.values(BaseFieldScope),
 		},
+		relevanceDuration: {
+			type: 'number',
```
bickelj, line 60: Regarding my wistful wish for duration types: ah, but JSON and OpenAPI and AJV. Sigh.

Perhaps to make it more explicit that the unit is seconds, can we include the unit in the name? E.g. `valueRelevanceSeconds`?

src/database/migrations/0055-alter-base_fields-add-relevance_duration.sql
slifty: This is a good point and I wonder if we should just make it days.  I'm trying to think of a use case that would be in terms of hours.  It seems more likely that something would be 100 years than need the granularity of seconds, hours, or minutes.

slifty: UPDATE: spoke with @bickelj and we landed on hours as a nice middle ground here!

src/database/migrations/0055-alter-base_fields-add-relevance_duration.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE base_fields
+ADD COLUMN value_relevance_hours integer;
```
bickelj, line 2: Nice, thank you!

src/database/migrations/0055-alter-base_fields-add-relevance_duration.sql
```diff
@@ -0,0 +1,5 @@
...
+
+COMMENT ON COLUMN base_fields.value_relevance_hours IS
+'Duration (in seconds) for which values of this base field are considered relevant.'; -- noqa: LT05
```
bickelj, line 5: Comment should say "in hours".

src/openapi/components/schemas/BaseField.json
```diff
@@ -34,6 +34,11 @@
 			"type": "string",
 			"enum": ["proposal", "changemaker"]
 		},
+		"valueRelevanceHours": {
+			"type": ["integer", "null"],
+			"example": 60,
+			"description": "Duration (in seconds) for which values of this base field are considered relevant."
```
bickelj, line 40: Update to 24 and "in hours" here too.

src/types/BaseField.ts
```diff
@@ -101,8 +107,19 @@ const writableBaseFieldSchema: JSONSchemaType<WritableBaseField> = {
 			type: 'string',
 			enum: Object.values(BaseFieldScope),
 		},
+		valueRelevanceHours: {
+			type: 'number',
+			nullable: true as false, // see https://github.com/ajv-validator/ajv/issues/2163
+		},
```
bickelj, line 113: Do we want to limit it to positive integers here?

bickelj (approved):
I see some comment updates and a migration filename update needed but otherwise, great, thank you for expanding to hours!

src/types/BaseField.ts
slifty: Yeah very good idea ty

slifty: I decided to allow 0 as well just because who am I to not think of a situation where you wanna store data and immediately consider it irrelevant.

slifty:
This is mysterious.

Tests are failing: https://github.com/PhilanthropyDataCommons/service/actions/runs/14518162626/job/40732196705?pr=1610

But the failure is due to a name conflict with a migration file that does not exist on this branch (yet; it will, and I'll have to fix the problem, but that's because it exists on main).

image

Is it possible that the state of the test runner is not actually clean?  Wild!

slifty:
Update: ok it's just that the checkout action defaults to checking out the `merge comimt` (which includes current `main` changes).  This is a reasonable default behavior and explains what's up.  Nothing wild going on!

bickelj:
@slifty Oh! What about the changelog?

Issue #1612, opened 2025-04-16
Implement audit logging (application query level, 2 of 3)


bickelj:
Issue #1559 ended up being associated with the first increment of work on audit logging using database triggers, the stuff originally thought of in #1355. Implementation of table triggers to capture changes was straightforward but it would have been awkward to integrate user context into every log there, so we decided to have multiple levels of audit logs rather than relying solely on database triggers. The advantage of triggers on tables is it captures data changes made outside the application. The goal in this ticket is to augment the lower-level audit logs with logs produced by the application. The plan is to modify functions that generate functions that call SQL to do `INSERT`s into a new mid-level audit log table.

bickelj:
Something that neither the database triggers nor the application queries will capture is UNsuccessful calls, such as those that have an expired token and are not retried, and so forth. I think that is something that level 3 (HTTP requests, #1613) can most easily capture.

bickelj:
In prod now.

bickelj:
Reopening to complete revisions in response to review on #1615

bickelj:
Really actually done now : )

PR #1615, opened 2025-04-16
Audit generated SQL function calls

bickelj:
Database triggers provide a low-level audit trail that is available
regardless of whether a data change was made by the application or by
a user manually logged into the database, for example, via `psql`. The
triggers can be referred to as "level 1" and were implemented in pull
request https://github.com/PhilanthropyDataCommons/service/pull/1607. The SQL function call logging implemented in this
commit can be referred to as "level 2" audit logs.

The low-level audit rows are in `audit_logs`. This change extends the
audit capabilities with a new table `db_operations_audit_logs` and
causes inserts to that table whenever various generated SQL function
calls succeed.

The design of the new table is related to the original such that a
view, also added here as `unified_audit_logs`, can provide friendlier
access to all the audit logs of all levels. The `unified_audit_logs`
view can in future (and now does) select relevant columns, filter out
extraneous rows, and expand short text codes into human-readable form.

Issue https://github.com/PhilanthropyDataCommons/service/issues/1612 Implement audit logging (application query level, 2 of 3)

### DIFF
--- a/src/__tests__/dbOperationAuditLogs.int.test.ts
+++ b/src/__tests__/dbOperationAuditLogs.int.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, test } from '@jest/globals';
+import { db } from '../database/db';
+import { createDbOperationAuditLog } from '../database/operations/dbOperationAuditLogs/createDbOperationAuditLog';
+import { loadUnifiedAuditLogBundle } from '../database/operations/unifiedAuditLogs';
+import {
+	expectTimestamp,
+	getTestAuthContext,
+	NO_LIMIT,
+	NO_OFFSET,
+} from '../test/utils';
+
+describe('db operation audit logs appear', () => {
+	test('direct call to createDbOperationAuditLog appears', async () => {
+		const authContext = await getTestAuthContext(false);
+		const administratorAuthContext = await getTestAuthContext(true);
+		const queryParameters = {
+			keyOne: 'Value 1 of Pretend query 5099',
+			keyTwo: 'Value 2 of Pretend query 5099',
+		};
+		await createDbOperationAuditLog(db, authContext, {
+			queryName: 'Pretend query 5099',
+			queryParameters,
+		});
+		const unifiedAuditLogsViewRows = await loadUnifiedAuditLogBundle(
+			db,
+			administratorAuthContext,
+			NO_LIMIT,
+			NO_OFFSET,
+		);
+		expect(unifiedAuditLogsViewRows.entries).toContainEqual({
+			statementTimestamp: expectTimestamp,
+			userKeycloakUserId: authContext.user.keycloakUserId,
+			userIsAdministrator: false,
+			pid: expect.any(Number),
+			auditLevel: 2,
+			operation: 'Called query Pretend query 5099',
+			details: queryParameters,
+		});
+	});
+});
--- a/src/database/operations/dbOperationAuditLogs/createDbOperationAuditLog.ts
+++ b/src/database/operations/dbOperationAuditLogs/createDbOperationAuditLog.ts
@@ -1,0 +1,18 @@
+import TinyPg from 'tinypg';
+import { AuthContext } from '../../../types';
+
+export const createDbOperationAuditLog = async (
+	db: TinyPg,
+	authContext: AuthContext | null,
+	createValues: {
+		queryName: string;
+		queryParameters: object;
+	},
+): Promise<void> => {
+	await db.sql('dbOperationAuditLogs.insertOne', {
+		authContextKeycloakUserId: authContext?.user?.keycloakUserId,
+		authContextIsAdministrator: authContext?.role?.isAdministrator,
+		queryName: createValues.queryName,
+		queryParameters: createValues.queryParameters,
+	});
+};
--- a/src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
+++ b/src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
@@ -1,3 +1,4 @@
+import { createDbOperationAuditLog } from '../dbOperationAuditLogs/createDbOperationAuditLog';
 import { NoDataReturnedError } from '../../../errors/NoDataReturnedError';
 import {
 	getIsAdministratorFromAuthContext,
@@ -67,6 +68,11 @@ const generateCreateOrUpdateItemOperation =
 			throw new NoDataReturnedError(
 				'The database did not return a query result.',
 			);
+		} else {
+			await createDbOperationAuditLog(db, authContext, {
+				queryName,
+				queryParameters,
+			});
 		}
 		return object;
 	};
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
 	},
 	"type": "commonjs",
 	"devDependencies": {
+		"@jest/globals": "^29.7.0",
 		"@redocly/cli": "^1.34.2",
 		"@tsconfig/node22": "^22.0.1",
 		"@types/cors": "^2.8.17",
--- a/src/database/initialization/unified_audit_log_to_json.sql
+++ b/src/database/initialization/unified_audit_log_to_json.sql
@@ -1,0 +1,18 @@
+SELECT drop_function('unified_audit_log_to_json');
+
+CREATE FUNCTION unified_audit_log_to_json(
+	unified_audit_log unified_audit_logs
+)
+RETURNS jsonb AS $$
+BEGIN
+	RETURN jsonb_build_object(
+		'statementTimestamp', unified_audit_log.action_tstamp_stm,
+		'userKeycloakUserId', unified_audit_log.user_keycloak_user_id,
+		'userIsAdministrator', unified_audit_log.user_is_administrator,
+		'pid', unified_audit_log.pid,
+		'auditLevel', unified_audit_log.level,
+		'operation', unified_audit_log.operation,
+		'details', unified_audit_log.details
+	);
+END;
+$$ LANGUAGE plpgsql;
--- a/src/database/migrations/0058-create-db-operation-audit-logs.sql
+++ b/src/database/migrations/0058-create-db-operation-audit-logs.sql
@@ -1,0 +1,107 @@
+CREATE TABLE db_operation_audit_logs (
+	action_tstamp_stm timestamp with time zone NOT NULL
+	DEFAULT statement_timestamp(),
+	user_keycloak_user_id uuid REFERENCES users (keycloak_user_id),
+	user_is_administrator boolean,
+	pid integer NOT NULL DEFAULT pg_backend_pid(),
+	query_name text NOT NULL,
+	query_parameters jsonb
+);
+CREATE INDEX db_operation_audit_logs_action_tstamp_stm_idx
+ON db_operation_audit_logs (action_tstamp_stm);
+CREATE INDEX db_operation_audit_logs_action_user_keycloak_user_id_idx
+ON db_operation_audit_logs (user_keycloak_user_id);
+COMMENT ON TABLE db_operation_audit_logs IS
+'Audit log of calls to database-query functions made in the PDC service.';
+COMMENT ON COLUMN db_operation_audit_logs.action_tstamp_stm IS
+'The time at which the statement causing the `INSERT` to this row began. Can '
+'be used for temporal correlation with rows in lower-level audit_logs table. '
+'The audit_logs row will precede this row temporally.';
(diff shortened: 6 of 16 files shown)


joao-krauchenco-mottu (approved):

src/database/operations/generators/generateCreateOrUpdateItemOperation.ts
```diff
@@ -67,6 +67,12 @@ const generateCreateOrUpdateItemOperation =
 			throw new NoDataReturnedError(
 				'The database did not return a query result.',
 			);
+		} else {
+			await db.sql('auditLogs.insertOne', {
```
slifty, line 71: I know we can't use a generator for this since that would get really infinite really fast BUT we should still have this as a db operation utility such as `writeOperationAuditLog` which has the same signature pattern as other operations (e.g. params are an auth context, database, additional parameters).

The other value of sending the entire auth context to the operation and extracting exactly what we need from it is that we actually probably should store more than just the user ID (we should store the isAdmin state too)

In addition to giving us more type safety, using an operation also allows us to potentially handle errors differently if we ever decide that, for instance, failure to save an audit log should not disrupt the user's action (or on the contrary: if it SHOULD roll back a transaction!)

src/database/queries/auditLogs/insertOne.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO audit_logs_2 (
+	user_keycloak_user_id,
```
slifty, line 2: Let's also store the administrator status of the operating agent.

slifty (changes requested):
A legibility suggestion (table name), DRY suggestion (using an operation utility) and one piece of data I think we should also log (admin status)

src/database/migrations/0056-create-audit-logs-2.sql
```diff
@@ -0,0 +1,99 @@
+CREATE TABLE audit_logs_2 (
```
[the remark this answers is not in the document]
bickelj: Would you be OK with `audit_logs_service_operations`? I was thinking it is nice to keep all audit-log related stuff having a prefix `audit_log` but I could be persuaded otherwise.

src/database/queries/auditLogs/insertOne.sql
bickelj: It is being stored in the `queryParameters` along with the user ID. The reason for storing the (duplicative) user ID separately is the anticipation that we will want to be able to say "show me X user's actions." Do you think we'll have a commonly called query of "show me all administrative actions respective of users"?

Example row from the view in the original PR:
` 2025-04-16 15:17:30.980584-05 | 8f6e6dd9-d4af-45db-af50-712f7e962cd7 | 14554 |     2 | Called query applicationForms.selectWithPagination  | {"limit": 10, "offset": 0, "authContextKeycloakUserId": "8f6e6dd9-d4af-45db-af50-712f7e962cd7", "authContextIsAdministrator": true}`

src/database/migrations/0056-create-audit-logs-2.sql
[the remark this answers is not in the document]
bickelj: I admit I am not fully satisfied with the naming here, but we are not surfacing it anywhere (yet), perhaps we can rename if a better name comes up when doing that. For now I went with your original suggestion.

src/database/queries/auditLogs/insertOne.sql
bickelj: Added the extra column `user_is_administrator` to the table.

bickelj:
Odd [deadlock detected during tests](https://github.com/PhilanthropyDataCommons/service/actions/runs/14741843908/job/41381437915?pr=1615#step:7:384).

src/__tests__/dbOperationAuditLogs.int.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, expect, test } from '@jest/globals';
```
slifty, line 1: Why are we adopting a different pattern here?  We don't import these in any of our other tests!

src/__tests__/dbOperationAuditLogs.int.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { loadUnifiedAuditLogBundle } from '../database/operations/unifiedAuditLogs';
+import {
+	expectTimestamp,
+	getTestAuthContext,
+	NO_LIMIT,
+	NO_OFFSET,
+} from '../test/utils';
+
+describe('db operation audit logs appear', () => {
+	test('direct call to createDbOperationAuditLog appears', async () => {
```
slifty, line 13: We use `it` instead of `test` in all our other tests FWIW -- is there a reason to start using `test` here?

src/database/operations/dbOperationAuditLogs/createDbOperationAuditLog.ts
```diff
@@ -0,0 +1,18 @@
+import TinyPg from 'tinypg';
+import { AuthContext } from '../../../types';
+
+export const createDbOperationAuditLog = async (
```
slifty, line 4: We should avoid acronyms in function names (`database` is more clear and explicit than `db`)

bickelj: The function name follows the database table name. [See above](https://github.com/PhilanthropyDataCommons/service/pull/1615#discussion_r2047663030), we can continue to iterate on names there.

src/__tests__/dbOperationAuditLogs.int.test.ts
bickelj: Let me see what errors occur if I remove these. For whatever reason, codium was not happy without this.

bickelj: Weird. No bad thing occurs when I remove it and the associated dependency. :shrug: Thanks!

src/database/migrations/0056-create-audit-logs-2.sql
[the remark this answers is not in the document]
bickelj: I think your original suggestion plus your below suggested expansion of `db` to `database` is somewhat satisfactory, `DatabaseOperationAuditLogs`, but a database operation is kind of vague or less distinguished from plain audit logs generated via triggers. After a conversation with `qwen2.5-coder:14b`  I think `ApplicationQueryAuditLogs` is the least worst. Can you live with `ApplicationQueryAuditLogs`?

<details>
<summary>Full disclosure of chat with `qwen2.5-coder:14b`</summary>

```

$ ollama run qwen2.5-coder:14b
>>> I am having trouble coming up with a name for a database table and its rows. We use plural for table name and singular for the name of an object derived (via our custom ORM framework) fr
... om it. For example, we have a database table called `user_funder_permissions` and our ORM has a class `UserFunderPermission` to represent a single one of those objects. We have two audi 
... t log tables and one audit log view to represent user actions within the database or application, respectively. These audit logs are not from the database log file, they are higher level
...  than that because we will create an API within our application to serve these objects to administrative users. That is why we have tables for them. The first table is simply `audit_logs
... ` and is populated via postgres database triggers on our tables watching for data changes. The second table is populated by calls to our mid-level (not HTTP request level) ORM functions 
... that may read from or write to the tables in our database. The view is a unified view of these two tables and is OK, it's called `unified_audit_logs`. Originally I named the mid-level ta
... ble populated by our application-level/ORM calls to be `audit_logs_2` but that was unsatisfactory to a reviewer and he suggested `db_operation_audit_logs` which I implemented. Now when t
... he same reviewer sees a function following our pattern of naming with `DbOperationAuditLog` he thinks we shouldn't have a contraction or acronym in these function names. In other words, 
... we need a new name for this mid-level audit log table that is not `audit_logs`, not `audit_logs_2`, not `db_operation_audit_logs`, and does not contain an acronym in it. Can you come up 
... with ten possible names for this second audit log table, keeping in mind it is logging calls to utility functions within our application-level custom ORM framework?
Certainly! Here are ten possible names for the mid-level audit log table that align with your requirements:

1. **ApplicationAuditLogs**
2. **AppLevelAuditLogs**
3. **ORMActivityLog**
4. **CustomORMAuditLogs**
5. **UtilityFunctionCallsLog**
6. **ProgrammaticAuditLogs**
7. **NonTriggerAuditLogs**
8. **MidLevelAuditLogs**
9. **HigherLevelAuditLogs**
10. **SystemGeneratedAuditLogs**

These names avoid using contractions, acronyms, or the term "2," while still clearly indicating that these logs are related to application-level operations performed via your custom 
ORM framework.

>>> Can you come up with ten more names that do not include the acronym "ORM"?
Certainly! Here are ten additional names for the mid-level audit log table, avoiding the acronym "ORM":

1. **ApplicationActivityLog**
2. **AppOperationsLog**
3. **CustomFrameworkAuditLogs**
4. **ProgrammaticOperationsLog**
5. **NonTriggerOperationsLog**
6. **MidLevelOperationsLog**
7. **HigherLevelOperationsLog**
8. **SystemGeneratedOperationsLog**
9. **UtilityFunctionCallsLog**
10. **CustomFunctionActivityLog**

These names maintain clarity and relevance while avoiding the use of "ORM" or any other acronyms, and they effectively convey the purpose of logging application-level operations.

>>> I think I like the word "activity" in there. These will occur when a database query occurs. Can you give me ten more names for this mid-level audit log table that include the word "Activ
... ity" and takes into account that these are SQL queries?
Of course! Here are ten additional names for the mid-level audit log table that include the word "Activity" and reflect the context of database queries:

1. **DatabaseQueryActivityLog**
2. **SQLActivityLog**
3. **DBQueryActivityLogs**
4. **DatabaseOperationActivityLog**
5. **QueryExecutionActivityLog**
6. **SQLExecutionActivityLog**
7. **DatabaseInteractionActivityLog**
8. **AppSQLActivityLog**
9. **CustomSQLActivityLog**
10. **ApplicationSQLActivityLog**

These names highlight the activity of executing SQL queries and provide a clear indication of their purpose within your application's database interactions.

>>> I also like the end of the name to be "AuditLogs" and perhaps either "Query" or "Activity" in there but I'm still not quite satisfied with any suggestions, can you come up with ten more 
...  new suggested names for this audit log table of sql queries called by our application?
Certainly! Here are ten more names that include the word "AuditLogs," either "Query" or "Activity," and reflect the context of SQL queries executed by your application:

1. **SQLQueryAuditLogs**
2. **DatabaseActivityAuditLogs**
3. **AppQueryExecutionAuditLogs**
4. **CustomQueryAuditLogs**
5. **ApplicationActivityAuditLogs**
6. **DBActivityAuditLogs**
7. **QueryActivityAuditLogs**
8. **CustomSQLQueryAuditLogs**
9. **ProgrammaticQueryAuditLogs**
10. **MidLevelSQLActivityAuditLogs**

These names aim to provide a clear and descriptive representation of the audit logs capturing SQL queries executed by your application.

>>> Suppose the reviewer simply suggested that we expand `DbOperationAuditLogs` to `DatabaseOperationAuditLogs`? That is great, except it is kind of ambiguous what a Database Operation is, i
... t seems overly broad. The same problem exists with `DatabaseActivityAuditLogs` because how is that different from the lower-level trigger audit logs? I like `AppQueryExecutionAuditLogs` 
... but it contains a contraction `App` but expansion to `ApplicationQueryExecutionAuditLogs` is probably too long. Maybe something like `ApplicationQueryAuditLogs`? Can you make ten more si
... milar to that? No contractions, no acronyms.
Certainly! Here are ten names that focus on SQL query execution and avoid using contractions or acronyms:

1. **ApplicationSQLQueryAuditLogs**
2. **AppLevelSQLQueryAuditLogs**
3. **CustomSQLQueryActivityAuditLogs**
4. **MidLevelSQLQueryAuditLogs**
5. **HigherLevelSQLQueryAuditLogs**
6. **ProgrammaticSQLQueryAuditLogs**
7. **NonTriggerSQLQueryAuditLogs**
8. **SystemGeneratedSQLQueryAuditLogs**
9. **UtilitySQLQueryAuditLogs**
10. **AppDrivenSQLQueryAuditLogs**

These names emphasize SQL query execution while maintaining clarity and avoiding ambiguity.

```

</details>

src/__tests__/dbOperationAuditLogs.int.test.ts
bickelj: I limited the import to `expect` in #1630 and made a comment explaining why it is there.

src/__tests__/dbOperationAuditLogs.int.test.ts
bickelj: Resolved in #1630 .

PR #1617, opened 2025-04-21
Remove fp-ts dependency

slifty:
This PR removes an unused dependency.  I apparently had added it when adding the newtypes, and suspect it was accidentally included / was something I looked into but didn't end up using.


hminsky2002 (approved):
Ok, but i still think Whoever added this dependency should be subject to a kangaroo court style show trial

Issue #1558, opened 2025-03-25
Implement the concept of "good-as-of" for field values


slifty:
This feature need came from discussions with the PDC exec team.

When data is added to the PDC the source of the data might have added context of how "fresh" the data is.  This is useful information for our gold-data algorithm / conflict resolution.

We should:

- [ ] Allow the "good-as-of" date to be assigned to a given field value (how should this work for bulk upload?)
- [ ] Use this date in the gold data / disambiguation logic.

Might be worth separating these two concepts into two or more issues once we decide to prioritize this ticket.

PR #1625, opened 2025-04-25
Add `goodAsOf` attribute

slifty:
This PR adds the concept of `goodAsOf` to proposal field values.  This concept isn't utilized for conflict resolution as part of this PR.

Issue #1558 Implement the concept of "good-as-of" for field values


src/database/db.ts
```diff
@@ -4,21 +4,26 @@ import { TinyPg } from 'tinypg';
...
 const initializationDirectory = path.join(__dirname, 'initialization');
+const logger = getLogger('db');
 
 const db = new TinyPg({
 	root_dir: [path.resolve(__dirname, 'queries')],
 });
 
+db.pool.on('connect', (client) => {
+	client.query("SET TIME ZONE 'UTC';").catch((err: unknown) => {
+		logger.error({ err }, 'Failed to set timezone');
```
slifty, line 15: I think this falls in the category if "I don't feel like writing a test that will simulate a database failure" but if anybody feels strongly to the contrary I can do it (it has been done in this code base!)

hminsky2002 (approved):
Looks good, everything is covered that would need to be! Well done @slifty! I guess normally i would want to see this paired with it's related-functionality in the PR, but since we've the time better to get this in now.

PR #1630, opened 2025-05-01
Audit logging: clean up service-level audit code

bickelj:
The previous merge had an unsightly contraction/initialism "db" in the names of tables and associated types. This commit renames the `db_operation_audit_logs` table and `DbOperationAuditLog` type to `service_query_audit_logs` and `ServiceQueryAuditLog` respectively.

This commit clarifies an unusual import that was added in a test. This commit uses imports in the usual manner.

Follows #1615 review of 54368a3dc713652b1cb94b930600f9bc04fb3418

Issue #1612: Implement audit logging (application query level, 2 of 3)


src/database/operations/serviceQueryAuditLogs/createServiceQueryAuditLog.ts
```diff
@@ -1,18 +1,20 @@
...
+const createServiceQueryAuditLog = async (
 	db: TinyPg,
 	authContext: AuthContext | null,
 	createValues: {
 		queryName: string;
 		queryParameters: object;
 	},
 ): Promise<void> => {
-	await db.sql('dbOperationAuditLogs.insertOne', {
+	await db.sql<object>('serviceQueryAuditLogs.insertOne', {
```
bickelj, line 12: `@typescript-eslint` was happier with this than the previous one after making the imports more consistent.

slifty (approved):
Thank you for these!  Sorry again for being so slow this week in review 😅

Excited to have this layer of logging!

src/database/migrations/0059-rename-db_operation_audit_logs.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE db_operation_audit_logs RENAME TO service_query_audit_logs;
```
bickelj, line 1: While looking at my local database in another context, this is enough to rename the table, but the indices are still not renamed.

bickelj:
I renamed some indices and a foreign key. That's the difference since last review.

Issue #1631, opened 2025-05-01
Deadlock detected


slifty:
CI failed for a dependabot test: https://github.com/PhilanthropyDataCommons/service/actions/runs/14767292535/job/41461149455?pr=1629

```
Summary of all failing tests
FAIL src/__tests__/proposalVersions.int.test.ts (8.152 s)
  ● /proposalVersions › POST / › Returns 409 Conflict if a provided application form field ID does not exist

    [raw_query]

    DROP SCHEMA test_1 CASCADE;

    error: deadlock detected

      at Parser.parseErrorMessage (node_modules/pg-protocol/src/parser.ts:369:69)
      at Parser.handlePacket (node_modules/pg-protocol/src/parser.ts:188:21)
      at Parser.parse (node_modules/pg-protocol/src/parser.ts:103:30)
      at Socket.<anonymous> (node_modules/pg-protocol/src/index.ts:7:48)


Test Suites: 1 failed, 45 passed, 46 total
Tests:       1 failed, 478 passed, 479 total
Snapshots:   0 total
Time:        206.119 s
Error: Process completed with exit code 1.
```

bickelj:
Two more instances on a recent merge to main, same test.

From [instance 1](https://github.com/PhilanthropyDataCommons/service/actions/runs/14798686941/job/41552028460#step:7:380):

```
FAIL src/__tests__/proposalVersions.int.test.ts (8.141 s)
  ● /proposalVersions › POST / › Returns 409 Conflict if a provided application form field ID does not exist

    [raw_query]

    DROP SCHEMA test_1 CASCADE;

    error: deadlock detected

      at Parser.parseErrorMessage (node_modules/pg-protocol/src/parser.ts:369:69)
      at Parser.handlePacket (node_modules/pg-protocol/src/parser.ts:188:21)
      at Parser.parse (node_modules/pg-protocol/src/parser.ts:103:30)
      at Socket.<anonymous> (node_modules/pg-protocol/src/index.ts:7:48)


Test Suites: 1 failed, 45 passed, 46 total
Tests:       1 failed, 473 passed, 474 total

```


From [instance 2](https://github.com/PhilanthropyDataCommons/service/actions/runs/14798686941/job/41552028454#step:7:672):

```
Summary of all failing tests
FAIL src/__tests__/proposalVersions.int.test.ts (8.441 s)
  ● /proposalVersions › POST / › Returns 409 Conflict if a provided application form field ID does not exist

[...]

PR #1633, opened 2025-05-05
Convert `Promise.all` calls to new `allNoLeaks`

bickelj:
This avoids deadlock during the test named 'Returns 409 Conflict if a
provided application form field ID does not exist'. The reason is that
`DROP SCHEMA` during teardown can be reached before another `Promise`
passed to `Promise.all` resolves when one `Promise` passed to
`Promise.all` rejects with an error. By using `Promise.allSettled` no
`Promise` will continue execution (be pending) following the call to
`await Promise.allSettled`. But `Promise.allSettled` has a cumbersome
API compared to `Promise.all`. Something like `Promise.all` that does
not reject until all `Promise` objects have either rejected or
resolved resolves (pun intended) the issue. It is OK to throw any
error associated with a rejected `Promise`, similar to the semantics
of `Promise.all`. Therefore a new function `allNoLeaks` is introduced
to wrap the tedious bits of `Promise.allSettled` while providing the
convenience of a `Promise.all`-like API but with no leaked `Promise`.

While the above case brought the issue to light, there are many other
calls to `Promise.all` where similar race conditions obtain. So the
new function is used in place of `Promise.all` everywhere else too.

Furthermore, the problem case of a call to `Promise.all` had only two
`Promise` objects sent and has been replaced with simple `await`s on
each function in sequence.

Issue https://github.com/PhilanthropyDataCommons/service/issues/1631: Deadlock detected

### DIFF
--- a/src/handlers/proposalVersionsHandlers.ts
+++ b/src/handlers/proposalVersionsHandlers.ts
@@ -23,6 +23,7 @@ import {
 } from '../errors';
 import { fieldValueIsValid } from '../fieldValidation';
 import { authContextHasFunderPermission } from '../authorization';
+import { allNoLeaks } from '../promises';
 import type { Request, Response } from 'express';
 import type {
 	Proposal,
@@ -119,7 +120,7 @@ const assertProposalFieldValuesMapToApplicationForm = async (
 			}
 		},
 	);
-	await Promise.all(applicationFormFieldQueries);
+	await allNoLeaks(applicationFormFieldQueries);
 };
 
 const postProposalVersion = async (req: Request, res: Response) => {
@@ -149,24 +150,22 @@ const postProposalVersion = async (req: Request, res: Response) => {
 				'You do not have write permissions on this proposal.',
 			);
 		}
-		await Promise.all([
-			assertApplicationFormExistsForProposal(
-				req,
-				applicationFormId,
-				proposalId,
-			),
-			assertProposalFieldValuesMapToApplicationForm(
-				applicationFormId,
-				fieldValues,
-			),
-		]);
+		await assertApplicationFormExistsForProposal(
+			req,
+			applicationFormId,
+			proposalId,
+		);
+		await assertProposalFieldValuesMapToApplicationForm(
+			applicationFormId,
+			fieldValues,
+		);
 		const finalProposalVersion = await db.transaction(async (transactionDb) => {
 			const proposalVersion = await createProposalVersion(transactionDb, req, {
 				proposalId,
 				applicationFormId,
 				sourceId,
 			});
-			const proposalFieldValues = await Promise.all(
+			const proposalFieldValues = await allNoLeaks(
 				fieldValues.map(async (fieldValue) => {
 					const { value, applicationFormFieldId } = fieldValue;
 					const applicationFormField = await loadApplicationFormField(
--- a/src/promises/allNoLeaks.ts
+++ b/src/promises/allNoLeaks.ts
@@ -1,0 +1,30 @@
+// Inspired by https://stackoverflow.com/questions/64928212/how-to-use-promise-allsettled-with-typescript#69451755
+const isRejected = (result: PromiseSettledResult<unknown>) =>
+	result.status === 'rejected';
+const isFulfilled = <T>(result: PromiseSettledResult<T>) =>
+	result.status === 'fulfilled';
+
+const getErrorFromReason = (reason: unknown): Error => {
+	if (reason instanceof Error) {
+		return reason;
+	}
+	return new Error(
+		`An unexpected error occurred while awaiting all tasks: ${JSON.stringify(reason)}`,
+	);
+};
+
+/**
+ * Returns results of given `Promise`s or throws an `Error` if any `Promise` rejects.
+ * `Promise.all` has a simple API but it leaks execution after one `Promise` rejects.
+ * `Promise.allSettled` does not leak execution but its API is awkward.
+ * Here we offer a simplified `Promise.all`-like API wrapping `Promise.allSettled` functionality.
+ * Use this function in place of `Promise.all` in almost all circumstances.
+ */
+export const allNoLeaks = async <T>(values: Promise<T>[]) => {
+	const settled = await Promise.allSettled(values);
+	const rejectedTask = settled.find(isRejected);
+	if (rejectedTask !== undefined) {
+		throw getErrorFromReason(rejectedTask.reason);
+	}
+	return settled.filter(isFulfilled)?.map((e) => e.value);
+};
--- a/src/promises/__tests__/allNoLeaks.unit.test.ts
+++ b/src/promises/__tests__/allNoLeaks.unit.test.ts
@@ -1,0 +1,69 @@
+import { fail } from 'assert';
+import { allNoLeaks } from '../allNoLeaks';
+
+describe('allNoLeaks', () => {
+	it('should throw an error after companion long-duration `Promise` completes', async () => {
+		let longTaskCompleted = false;
+		const longTask: Promise<void> = new Promise((resolve) => {
+			setTimeout(() => {
+				longTaskCompleted = true;
+				resolve();
+			}, 50);
+		});
+		const errorMessageToThrow = "errored, y'all!";
+		const errorTask = Promise.reject(new Error(errorMessageToThrow));
+		try {
+			// Substitute `Promise.all` here and the test fails, demonstrating the purpose of allNoLeaks.
+			await allNoLeaks([longTask, errorTask]);
+			fail('Expected an error to be thrown');
+		} catch (error) {
+			expect(longTaskCompleted).toBe(true);
+			expect(error).toEqual(new Error(errorMessageToThrow));
+		}
+	});
+
+	it('should throw an error with object reason', async () => {
+		const errorReasonToThrow = { weird: 'object' };
+		const errorTask = Promise.reject(errorReasonToThrow);
+		try {
+			await allNoLeaks([errorTask]);
+			fail('Expected an error to be thrown');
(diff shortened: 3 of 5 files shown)


bickelj:
@slifty This is only one way of refactoring to avoid deadlock and I think there may be better ways. I tried to de-uglify it by extracting the ugly into a function. In other words, this is the code that helped demonstrate `Promise.all` is a source of the deadlock problem, but I'd be open to other options.

bickelj:
My takeaway: genericize the already-extracted function but keep the rest specific to ProposalFieldValue. Takes an array of Promises, does the logic in one place. Replacement for `Promise.all` that includes the throw.

bickelj:
While iterating on this in order to pass all our lint rules, I cannot help but think there is no possible way we are the first to encounter this problem of `Promise.all`. Indeed we are not: https://www.npmjs.com/package/promise-all-reject-later, maybe there's another ready-made one out there too.

bickelj:
Before writing tests for our new seemingly-obvious wrapper around `Promise.all`, perhaps https://www.npmjs.com/package/p-all can be of service, specifically using [the `stopOnError=false` option](https://www.npmjs.com/package/p-all#stoponerror). Hat tip to `qwen2.5-coder:14b` for pointing me to this package.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -149,7 +170,7 @@ const postProposalVersion = async (req: Request, res: Response) => {
 				'You do not have write permissions on this proposal.',
 			);
 		}
-		await Promise.all([
+		const relationsExistSettled = await Promise.allSettled([
```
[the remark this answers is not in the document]
bickelj: Agree that this should be two `await ...theFunction...` statements. Fixed.

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -160,13 +181,17 @@ const postProposalVersion = async (req: Request, res: Response) => {
...
+			throw getErrorFromRejectedTask(formValidationRejectedTask);
+		}
 		const finalProposalVersion = await db.transaction(async (transactionDb) => {
 			const proposalVersion = await createProposalVersion(transactionDb, req, {
 				proposalId,
 				applicationFormId,
 				sourceId,
 			});
-			const proposalFieldValues = await Promise.all(
+			const proposalFieldValuesSettled = await Promise.allSettled(
```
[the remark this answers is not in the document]
bickelj: I did not do the single DB operation thing but did the `allSettledAndThrowOnRejection` as `allNoLeaks`.

bickelj:
@slifty I am curious what you think of this so far, but I also think there is one of two things yet to do before this should be merged:
1. Add a couple of tests for the new function `allNoLeaks` (or whatever we want to name it).
2. Find an already-existing package with a similar function. I cannot believe this is not out there in a maintained package.

src/promises/allNoLeaks.ts
```diff
@@ -0,0 +1,38 @@
...
+): input is PromiseRejectedResult => input.status === 'rejected';
+
+// From https://stackoverflow.com/questions/64928212/how-to-use-promise-allsettled-with-typescript#69451755
+const isFulfilled = <T>(
+	input: PromiseSettledResult<T>,
+): input is PromiseFulfilledResult<T> => input.status === 'fulfilled';
+
+// The contract of PromiseRejectedResult requires `reason` to be `any`. Thus lint is disabled here.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const getErrorFromReason = (reason: any): Error => {
```
slifty, line 13: We shouldn't use any -- that isn't safe -- but rather use `unknown` (which still means it could be anything, but if you try to access an attribute without a typeguard then typescript will scream at you)

Changing to `unknown` will let you remove the eslint override too!

src/promises/allNoLeaks.ts
```diff
@@ -0,0 +1,38 @@
+// From https://stackoverflow.com/questions/64928212/how-to-use-promise-allsettled-with-typescript#69451755
```
slifty, line 1: IANAL but are we're allowed to use code copied from stack overflow (I'm not sure that CC 4.0 SA is compatible with AGPL)?

Can we just write these ourselves?  They seem pretty trivial, and the code in that answer is kind of weird anyway (specifically, the type guard feature of this function doesn't belong / isn't doing anything)

I think we can just do:

```
const isRejected = (result: PromiseSettledResult) => result.status === 'rejected';
const isFulfilled = (result: PromiseSettledResult) => result.status === 'fulfilled';
```

and drop attribution since that's what we would have written all by ourselves!

src/promises/allNoLeaks.ts
```diff
@@ -0,0 +1,38 @@
...
+): input is PromiseFulfilledResult<T> => input.status === 'fulfilled';
+
+// The contract of PromiseRejectedResult requires `reason` to be `any`. Thus lint is disabled here.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const getErrorFromReason = (reason: any): Error => {
+	if (reason instanceof Error) {
+		return reason;
+	}
+	let explanation = 'An unexpected error occurred while awaiting all tasks.';
+	if (typeof reason === 'string') {
```
slifty, line 18: This would only happen if we (or a library we use) made a mistake and threw a string instead of an error, right?

Let's spare ourselves the drama and just not support exposing the content if someone does that particular non-standard behavior.

slifty (changes requested):
I like the name you landed on -- pithy and clear!

I don't think we should add a dependency for something as simple as this; I agree it should just be baked into promises already, but this isn't something that needs to be maintained so our 10-liner is good enough.

Added some requests which I think will let us be cleaner overall.

src/promises/allNoLeaks.ts
bickelj: Hmm, IIRC `unknown` might have come with its own trouble, but I'll try it again and see.

src/promises/allNoLeaks.ts
bickelj: Let me see about that improvement. I think/hope we can credit the SO answer for the approach without violating copyright.

src/promises/allNoLeaks.ts
bickelj: I almost had the simpler form but that would unfortunately swallow whatever error data we may actually have, which in the case that this actually surfaces would be incredibly useful and maddening if not present. An error can occur at any level (our code, our libraries, node, DB, OS, network, hardware) for many reasons. I think what I really want here is the (broad) to-string of `reason` rather than (narrow) case of it being a string.

bickelj:
The last two pushes were simply rebases.

src/promises/allNoLeaks.ts
```diff
@@ -0,0 +1,30 @@
+// Inspired by https://stackoverflow.com/questions/64928212/how-to-use-promise-allsettled-with-typescript#69451755
+const isRejected = (result: PromiseSettledResult<unknown>) =>
+	result.status === 'rejected';
+const isFulfilled = <T>(result: PromiseSettledResult<T>) =>
```
slifty, line 4: We don't need the <T> here

src/promises/allNoLeaks.ts
```diff
@@ -0,0 +1,30 @@
+// Inspired by https://stackoverflow.com/questions/64928212/how-to-use-promise-allsettled-with-typescript#69451755
+const isRejected = (result: PromiseSettledResult<unknown>) =>
```
slifty, line 2: We don't need `<unknown>` here

slifty (approved):
Just two unnecessary type generics in the helper functions that we should take out, but otherwise approved!

src/promises/allNoLeaks.ts
bickelj: OK, can you help with the resulting lint errors then?

src/promises/allNoLeaks.ts
bickelj: OK, can you help with the resulting lint errors then?

bickelj:
> Just two unnecessary type generics in the helper functions that we should take out, but otherwise approved!

I pushed your recommended change so you can see the resulting lint failures.

src/promises/allNoLeaks.ts
slifty: Yeah fancy those!  Clearly I'm wrong  regarding the `<unknown>` being assumed.

I am now curious though so I'm gonna spend just 5m learning

slifty:
Sorry -- disregard my incorrect understanding of the generics situation!  no changes requested whatsoever

bickelj:
Here's what TSC said:
```
Run npm run lint:tsc
  npm run lint:tsc
  shell: /usr/bin/bash -e {0}

> lint:tsc
> tsc --noEmit -p tsconfig.dev.json

Error: src/promises/allNoLeaks.ts(2,29): error TS2314: Generic type 'PromiseSettledResult' requires 1 type argument(s).
Error: src/promises/allNoLeaks.ts(4,30): error TS2314: Generic type 'PromiseSettledResult' requires 1 type argument(s).
Error: src/promises/allNoLeaks.ts(27,41): error TS2339: Property 'reason' does not exist on type 'PromiseSettledResult<Awaited<T>>'.
  Property 'reason' does not exist on type 'PromiseFulfilledResult<Awaited<T>>'.
Error: src/promises/allNoLeaks.ts(29,51): error TS2339: Property 'value' does not exist on type 'PromiseSettledResult<Awaited<T>>'.
  Property 'value' does not exist on type 'PromiseRejectedResult'.
Error: Process completed with exit code 2.
```

bickelj:
Rebased with the same commit as before removing the `<T>` and `<unknown>` because I think these are needed.

Issue #1621, opened 2025-04-22
Add the concept of "too sensitive to store data" to base fields


slifty:
This might take the form of a flag on base fields; it may also take the form of a new entity / set of api routes.

We'll make a decision about this as part of implementation, but the most important thing to realize is that whatever implementation happens, we MUST ensure that the rule is followed (i.e. now data is ever stored or retained related to these fields).

Note this may mean that audit logs have to be deleted which have sensitive data in them.

We should think through the fact that this would mean enabling this flag is considered disruptive.

slifty:
A question here for @kfogel and @jmergy would these base fields be published by the API (with the flag of "notStored" or whatever we land on)?

slifty:
I wanted to document some implementation possibilities here rather than barreling right in.

I'm thinking of not just what would be most intuitive from a data representation perspective, but also in terms of the act of registering a sensitive field and the correct REST API design.

On the one hand this is metadata associated with a base field; on another it is actually a form of deletion (so setting that metadata is an inherently destructive act).


## Storage
### Option A: Add an attribute of `BaseField` to reflect the state (e.g. `isSuperDuperSensitive` or `isDeactivated` or `isDeleted` or `deniedAfter`)

This would signal that the base field existed (along with any current base field metadata it had), but that it had been deactivated (possibly with an additional field to reflect a reason).  This would be simple, but it would mean there is a row in the table for a base field that we have determined to not be allowed which would mean we lose out-of-the-box DB-enforcement of disallowing relationships with disabled base fields.  We could look into things like triggers to add that protection, but the point is it opens up the door for logical errors.

The benefit of this approach, though, is that we don't *lose* anything about the disabled base field.  We get it's label, description, scope, localizations, etc for free without having to replicate these attributes in another data structure (and as attributes of a base field are added we don't have to repeat those additions in another entity.

This means that if we ever wanted to re-enable a base field for any reason that would become a trivial act.

Another major benefit is that it would explicitly disable the base field's short code forever (one could not accidentally re-insert the base field since the short code would already be in the system).

### Option B: Add a new entity type to reflect base fields that have been rejected (e.g. `DeactivatedBaseField`)

This entity would represent base fields that have been rejected for whatever reason.  This would be the "strongest" way to ensure there is no associated with a disabled base field (it is literally deleted, and so FK constraints would force cascading deletions).

The downsides / upsides of this are the inverse of Option A.

---

[...]

PR #1686, opened 2025-06-03
Add `sensitivityClassification`

slifty:
This PR adds the concept of a `sensitivityClassification` to `BaseFields`.  If a base field is marked as `FORBIDDEN` then values associated with that field are no longer exposed and new data for those fields cannot be accepted.

In an upcoming commit we will be adding mechanisms to purge data associated with forbidden fields as well so that it is not even stored.

Related to #1621


src/database/initialization/proposal_version_to_json.sql
```diff
@@ -3,32 +3,35 @@ SELECT drop_function('proposal_version_to_json');
...
-  INTO proposal_field_values_json
-  FROM proposal_field_values
-  WHERE proposal_field_values.proposal_version_id = proposal_version.id;
+	SELECT jsonb_agg(
+		proposal_field_value_to_json(proposal_field_values.*)
+		ORDER BY proposal_field_values.position, proposal_field_values.id DESC
+	)
+	INTO proposal_field_values_json
+	FROM proposal_field_values
+	INNER JOIN application_form_fields on proposal_field_values.application_form_field_id = application_form_fields.id
```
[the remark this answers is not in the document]
slifty: `proposal_field_value_to_json` simply packages a selected field value as json; we want to omit the field value from the points it is selected it for sure as at least one layer.

That said, you bring up a good idea: should we defensively omit the value attribute of a proposal field value to account for a potential bug / mistake in selection logic somewhere?  Maybe the answer to this is yes, but `value` is not a null field and we do make various commitments regarding the validity of a value according to the `isValid` attribute.  It makes me think that it would risk undermining data integrity overall.

I'll spend a moment checking to see if there's any way for `proposal_field_value_to_json` to throw some kind of error if it is invoked against a forbidden field value, though.  I think it's reasonable to say we would prefer a 500 error vs leaking forbidden data.

bickelj (commented):
Having only skimmed the code, I first went to try it locally.

1. I synced base fields from prod (hat tip to @hminsky2002 in #1259 which worked perfectly first try).
2. I created boilerplate funders, associated opportunities, sources, and proposals.
3. I created an application form with three fields, all at the time `restricted`.
4. I posted a proposalVersion with all three fields filled in, no problem.
5. In `psql` I updated one of those base fields to be `forbidden` then re-posted the same proposalVersion.

I successfully got an error, but I doubt this is the error we want. I think a 400 Bad Request (or some 4xx) would make more sense rather than 500 Internal Server Error.

Looking a little closer, this may be an issue with audit log code or something in the error throwing/handling between this PR and that code:
```json
{
  "level": 50,
  "time": 1749568326472,
  "pid": 49507,
  "hostname": "host",
  "source": "/home/user/code/ots/ext/PhilanthropyDataCommons/service/dist/middleware/errorHandler.js",
  "wrappedError": {
    "tinyPgError": {
      "name": "TinyPgError",
      "queryContext": {
        "id": "9ee0586f-064f-4d29-ae72-db516c94acbf",
        "sql": "INSERT INTO service_query_audit_logs (\n\tuser_keycloak_user_id,\n\tuser_is_administrator,\n\tquery_name,\n\tquery_parameters\n) VALUES (\n\t$1,\n\t$2,\n\t$3,\n\t$4\n);\n",
        "start": 1749568326469,
        "name": "serviceQueryAuditLogs_insertOne",
        "params": {
          "authContextKeycloakUserId": "8f6e6dd9-d4af-45db-af50-712f7e962cd7",
          "authContextIsAdministrator": true,
          "queryName": "proposalFieldValues.insertOne",
          "queryParameters": {
            "authContextKeycloakUserId": "8f6e6dd9-d4af-45db-af50-712f7e962cd7",
            "authContextIsAdministrator": true,
            "proposalVersionId": 3,
            "applicationFormFieldId": 4,
            "position": 50,
            "value": "Call me Ishmael",
            "isValid": true,
            "goodAsOf": "2025-06-10T15:10:13.673Z"
          }
        },
        "submit": 1749568326471,
        "wait_duration": 2,
        "active_duration": 0,
        "end": 1749568326471,
        "duration": 2,
        "error": {
          "length": 145,
          "name": "error",
          "severity": "ERROR",
          "code": "25P02",
          "file": "postgres.c",
          "line": "1498",
          "routine": "exec_parse_message"
        },
        "data": null
      }
    },
    "name": "DatabaseError"
  },
  "statusCode": 500
}
```

The postgres log says:

```
2025-06-10 10:12:06.470 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) ERROR:  Cannot insert proposal field value for forbidden application form field 6
2025-06-10 10:12:06.470 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) CONTEXT:  PL/pgSQL function prevent_inserting_forbidden_proposal_field_value() line 14 at RAISE
2025-06-10 10:12:06.470 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) STATEMENT:  INSERT INTO proposal_field_values (
		proposal_version_id,
		application_form_field_id,
		value,
		position,
		is_valid,
		good_as_of
	) VALUES (
		$1,
		$2,
		$3,
		$4,
		$5,
		$6
	)
	RETURNING proposal_field_value_to_json(proposal_field_values) AS object;
	
2025-06-10 10:12:06.471 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) ERROR:  current transaction is aborted, commands ignored until end of transaction block
2025-06-10 10:12:06.471 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) STATEMENT:  INSERT INTO service_query_audit_logs (
		user_keycloak_user_id,
		user_is_administrator,
		query_name,
		query_parameters
	) VALUES (
		$1,
		$2,
		$3,
		$4
	);
	
2025-06-10 10:12:06.471 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) ERROR:  current transaction is aborted, commands ignored until end of transaction block
2025-06-10 10:12:06.471 CDT [51454] 68484474.c8fe pdc@pdc 127.0.0.1(58736) STATEMENT:  INSERT INTO service_query_audit_logs (
		user_keycloak_user_id,
		user_is_administrator,
		query_name,
		query_parameters
	) VALUES (
		$1,
		$2,
		$3,
		$4
	);
	
```

The `GET /proposals` endpoint successfully returned the existing proposal and filtered out the verboten field.

After that, I added a Changemaker and linked all the proposals to that new Changemaker. The following endpoints successfully filtered out the forbidden field:
* `GET /changemakers`
* `GET /changemakerProposals`
* `GET /proposals`
* `GET /proposals/{proposalId}`
* `GET /proposalVersions`
* `GET /proposalVersions/{proposalVersionId}`

So far this is looking really really good.

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -20,9 +27,11 @@ WHERE
 			OR :search = ''
 		) THEN
 			TRUE
-		ELSE
-			proposal_field_values.value_search
+		ELSE (
```
bickelj, line 30: I know this is part of a series of `AND` clauses, but could it be out top as its own clause within the `WHERE`? Or even an `AND` on the `LEFT JOIN`? Doubt it makes any difference query-planner-wise but may be more readable.

src/handlers/applicationFormsHandlers.ts
```diff
@@ -77,24 +77,24 @@ const postApplicationForms = async (req: Request, res: Response) => {
...
-					...field,
-					applicationFormId: applicationForm.id,
-				}),
-			),
-		);
-		res
-			.status(201)
-			.contentType('application/json')
-			.send({
+		const finalApplicationForm = await db.transaction(async (transactionDb) => {
```
bickelj, line 80: Nice use of transaction here.

src/database/initialization/proposal_field_value_to_json.sql
```diff
@@ -5,23 +5,37 @@ CREATE FUNCTION proposal_field_value_to_json(
...
-  application_form_field_json JSONB;
+	is_forbidden BOOLEAN;
+	application_form_field_json JSONB;
 BEGIN
-  SELECT application_form_field_to_json(application_form_fields.*)
-  INTO application_form_field_json
-  FROM application_form_fields
-  WHERE application_form_fields.id = proposal_field_value.application_form_field_id;
+	SELECT EXISTS (
+		select 1
```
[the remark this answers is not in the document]
slifty: Oh good catch -- (fluff doesn't lint anything inside a function unfortunately :()

slifty:
I created https://github.com/PhilanthropyDataCommons/service/issues/1715 to capture the need to investigate audit log trigger impact on our error codes when they happen in the context of a rejected transaction.

src/database/queries/proposals/selectWithPagination.sql
slifty: Update: LEFT JOIN isn't what we want (we don't just want base_field to be null: we want to actively reject the *proposal field values* where base field is forbidden).  I am also wary that adding it outside of the case statement could lead to proposals being rejected when they have forbidden base fields in contexts outside of search, which would be incorrect.

I'm gonna just leave it as originally written, since truly we are just saying "if a search is being invoked, ignore records that matched against forbidden base fields"

Issue #1699, opened 2025-06-10
Move from airbnb eslint configuration


slifty:
Unfortunately our use of airbnb's eslint config has locked us into v8 of eslint.  This has become a critical point of tech debt for us.  They have [an issue to support v9](https://github.com/airbnb/javascript/issues/2961), but it has been open for over a year and hasn't had any activity for almost 7 months.

Time to move along!

A friend (@louh) recommended https://www.npmjs.com/package/eslint-config-love and it does look pretty good, so I'm gonna take a stab at setting that up.  This would then let us migrate to ESLint 9!

PR #1703, opened 2025-06-10
Upgrade to ESlint 9

### LINKED ISSUE #978: Adjust dependabot settings for ESLint 9

slifty:
This PR does two things:

1. It removes airbnb from our configuration
2. It upgrades our version of eslint to v9

The v8 lock was becoming a reasonably large problem so this is a great change to make!

Removing airbnb rules means we are less strict for the short term, but in a future PR I expect to add a new set of proposed rules.  I'm not doing that in this commit because I don't know what the scope of the new suggested rules will end up being and the configuration migration was significant enough on its own.

Resolves #978 
Related to #1699


bickelj (approved):
While I get a bunch of package warnings locally when doing this same `npm` uninstall/install I assume that this is due to `typescript-eslint` still being on 8 or other companion packages.

It is not great to be loose on lint for a time, but we still have the base eslint rules in place, code review, and a body of existing tests.

LGTM!

PR #1718, opened 2025-06-17
Add BaseField category

### LINKED ISSUE #1669: Add the concept of "BaseField Category"
This would be an attribute of base field -- TBD if it is an enum / schema-defined concept, or if `BaseFieldCategory` would be a table and BaseFields would reference that table.

Ideally this would be a schema attribute for various reasons (simplicity for synchronization, predictability).

For now we are planning to treat this as a schema / attribute instead of a new type.

slifty:
This PR replaces `BaseField.scope` with `BaseField.category`.  This change also expands the possible values for the field.

Resolves #1669


src/types/BaseField.ts
```diff
@@ -15,9 +15,18 @@ export enum BaseFieldDataType {
 	CURRENCY = 'currency',
 }
 
-export enum BaseFieldScope {
-	PROPOSAL = 'proposal',
+enum BaseFieldCategory {
+	PROJECT = 'project',
 	ORGANIZATION = 'organization',
+	NEEDS_ASSESSMENT = 'needs_assessment',
+	METHODOLOY = 'methodology',
```
bickelj, line 22: `METHODOLOGY`

bickelj (approved):
Well done, tidy, elegant, etc.!

For a minute I worried that there is a difference between a Proposal and a Project. Indeed there is. A Proposal is an actual application for funds, so it should be related more to Proposal Field Value than to Base Field. A Project could be actual or aspirational, so it should be related to Base Field. We might have a Project in mind without having applied for it yet. So it makes sense to me.

When syncing base fields from production, if I have a an old version locally vs production or vice versa, sync doesn't work. However, I do not think it's worth the effort to go and update the base field sync process to allow backward and forward compatibility.

src/types/BaseField.ts
slifty: METHODOOOOOLY!

PR #1719, opened 2025-06-17
Run `eslint --fix` on format

### LINKED ISSUE #1293: Add an eslint autofix step to our `format` step
We have a `format` script which is handy and runs prettier -- it should also run eslint autofix for any rules that prettier doesn't interact with and which could be fixed automatically.

slifty:
This PR adds a step to our `format` script to benefit from eslint's auto-fix functionality when possible.

Resolves #1293


bickelj (approved):
It gives no errors when running `npm run format` locally. LGTM.

PR #1721, opened 2025-06-19
Set up new lint rules (part 3)

slifty:
This PR enables another set of linting rules.

Many of these are trivial, but I want to call out two in particular:

1. No magic numbers :: this is generally a good rule, but we need to be careful to address it thoughtfully to prevent silly situations.  I disabled it for objects (the logic here is that objects have named attributes and the context of a number will often be semantically clear based on the object's structure).  This prevented situations where I'd have to turn: `{ page: { min: 0 } }` into `{ page: { min: PAGE_MINIMUM } }`.  It will still yell about things like accessing array indexes or `foo += 2`.  I don't love that, but we can revisit later.

2. No unsafe type assertions :: I see this as an important and explicitly good change even if it requires a little bit of boilerplate and an occasional `eslint-disable` in the context of mocking and `Newtypes`.  It forces us to stay honest with our typing and helps prevent situations where we override a statically analyzed type reality.

Related to #1699


src/handlers/presignedPostRequestsHandlers.ts
```diff
@@ -16,7 +18,7 @@ const generatePresignedPost = async (
 	createPresignedPost(s3Client, {
 		Bucket: S3_BUCKET,
 		Key: `unprocessed/${uuidv4()}`,
-		Expires: 3600, // 1 hour
+		Expires: SECONDS_PER_HOUR,
```
bickelj, line 21: In a place like this I don't think we can avoid specifying a number because it's not really a magic number. We're configuring a timeout.

While it is true that there are 3600 seconds per hour, using that value here is a bit confusing. I would rather see `ONE_HOUR_IN_SECONDS` or `1 * SECONDS_PER_HOUR` or `3600` with comment as before.

Better yet, we should use a date and time library such as [luxon](https://www.npmjs.com/package/luxon) and its companion [types](https://www.npmjs.com/package/@types/luxon). In this example we could declare the [duration](https://moment.github.io/luxon/#/tour?id=durations) here and convert it to what this library expects, pass the luxon obect if the library supports it, or declare a `ONE_HOUR` constant and then use that to convert to seconds here.

Thanks for your patience with this bikeshedding but I think every software engineer in the world has been bitten by dates, times, and durations!

bickelj (approved):
Thank you, @slifty, overall this is wonderful. Applying the rule exposed some awkward stuff.

I think the HTTP status code thing is a net improvement but I wish we didn't lose the HTTP status category in the process. I agree that the test refactoring related to this rule is desired in the long run but can be punted for now.

src/handlers/presignedPostRequestsHandlers.ts
bickelj: I created https://github.com/PhilanthropyDataCommons/service/issues/1722 for the date/time lib.

bickelj:
@slifty I just realized that I only reviewed one commit and then approved the whole PR. Going to the next ones.

src/middleware/requireChangemakerPermission.ts
```diff
@@ -24,7 +24,7 @@ const requireChangemakerPermission =
 			next(new UnauthorizedError('The request lacks an AuthContext.'));
 			return;
 		}
-		if (req.role?.isAdministrator === true) {
+		if (req.role?.isAdministrator) {
```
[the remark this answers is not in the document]
slifty: @bickelj From those docs it does seem that this rule will yell and scream if `strictNullChecks` is not enabled so we do have it set (via our tsconfigs)

src/handlers/presignedPostRequestsHandlers.ts
slifty: I'll rename this to `ONE_HOUR_IN_SECONDS` for now (and I see #1722 which will obviate this later)

bickelj: I assume you mean that the *rule* is obnoxious in this case because we are trying to express a magnitude of time numerically which is totally fine. I think we can disable it here.

slifty: Yeah sorry -- I wish this rule would allow us to say "1 day" (as just "1") * "seconds_per_day" or whatever.

src/types/Changemaker.ts
```diff
@@ -34,10 +31,13 @@ const writableChangemakerSchema: JSONSchemaType<WritableChangemaker> = {
...
 			...keycloakIdSchema,
-			nullable: true,
+			// This is a gross workaround for the fact that AJV does not support nullable types in TypeScript.
+			// See: https://github.com/ajv-validator/ajv/issues/2163
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+			nullable: true as false,
 		},
 	},
-	required: ['taxId', 'name'],
+	required: ['taxId', 'name', 'keycloakOrganizationId'],
```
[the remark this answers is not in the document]
slifty: Let's talk about this!

Our API design outside of this does not allow optional values -- we want our users to be explicit when something is null.

We see this in the following contexts:

- `BaseField.valueRelevanceHours`
- `ProposalFieldValue.goodAsOf`

We want our API clients to say "seriously, this changemaker doesn't have a keycloak association" (which makes it clear that they are expected to be the ones providing that association)

Basically: required parameters are explicit, optional parameters that translate to null values are implicit.  In general we want to prefer explicit.

---

Is there a use case (outside of convenience / less verbosity) in allowing the optional parameter in a POST?

src/__tests__/userGroupChangemakerPermissions.int.test.ts
```diff
@@ -78,7 +80,7 @@ describe('/userGroups/changemakers/:changemakerId/permissions/:permission', () =
 			});
 			await request(app)
 				.put(
-					`/userGroups/${keycloakIdToString(organizationAsChangemaker.keycloakOrganizationId as KeycloakId)}/changemakers/notanId/permissions/${Permission.MANAGE}`,
+					`/userGroups/${keycloakIdToString(organizationAsChangemaker.keycloakOrganizationId)}/changemakers/notanId/permissions/${Permission.MANAGE}`,
```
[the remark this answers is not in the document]
slifty: No -- we know it's either `KeycloakId` or `null` (since it's a Changemaker, and changemaker.keycloakOrganizationId is `KeycloakId | null).

`keycloakIdToString` accepts this type (`KeycloakId | null`) so all casting was doing was incorrectly removing the possibility that it could hold a `null` value

src/middleware/requireChangemakerPermission.ts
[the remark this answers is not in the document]
slifty: It's set in the base configuration we specify in our config

```
	"extends": "@tsconfig/node22/tsconfig.json",
```

Two things to note:

1. That extension *is* explicit (you have to know to dig into it, but it's not magic / it's semantically conveyed to anybody who wants to look).

2. This rule *will* error if it is run without `strictNullChecks` so all we are really relying on is the rule not being buggy.

Keep in mind that the base configuration sets a fair amount of things that our static analysis and overall build rely on.  If / when those things change in ways that violate our assumptions, we rely on our tooling to alert us.

---

Between those two things I do feel like we're super covered and safe as is, and don't need to override / replicate the config base to maintain that safety.  What do you think?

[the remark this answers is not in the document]
bickelj: Sounds good enough to me. I also just figured out the command that gives me the warm fuzzies:
```shell
$ npx tsc --showConfig
{
    "compilerOptions": {
        "lib": [
            "es2024",
            "esnext.array",
            "esnext.collection",
            "esnext.iterator"
        ],
        "module": "nodenext",
        "target": "es2022",
        "strict": true,
        "esModuleInterop": true,
        "skipLibCheck": true,
        "moduleResolution": "node16",
        "outDir": "./dist",
        "declaration": true,
        "resolveJsonModule": true,
        "noUncheckedIndexedAccess": true,
        "moduleDetection": "force",
        "allowSyntheticDefaultImports": true,
        "resolvePackageJsonExports": true,
        "resolvePackageJsonImports": true,
        "useDefineForClassFields": true,
        "noImplicitAny": true,
        "noImplicitThis": true,
        "strictNullChecks": true,
...
```

src/types/Changemaker.ts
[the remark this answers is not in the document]
bickelj: One discussion is what do we want to do across our API which I am not yet sure I'm ready to fight out here.

For the purposes of this PR, this change introduces a breaking API change in a lint PR.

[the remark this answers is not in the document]
slifty: (OH I should note something I see on re-read: I didn't mean to indicate that `convenience / less verbosity` alone isn't a fine reason to make an API design decision!  I just wanted to check if there are other things that might be on your mind that I should add to my own understanding)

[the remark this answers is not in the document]
slifty: OK!  cherry picked these out and I'm moving them to a new branch (will make a new issue)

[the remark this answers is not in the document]
slifty: Issue created: https://github.com/PhilanthropyDataCommons/service/issues/1725

src/openapi/components/schemas/DataProvider.json
```diff
@@ -10,13 +10,13 @@
...
+			"oneOf": [{ "$ref": "./keycloakOrganizationId.json" }, { "type": "null" }]
 		},
 		"createdAt": {
 			"type": "string",
 			"format": "date-time",
 			"readOnly": true
 		}
 	},
-	"required": ["id", "name", "createdAt"]
+	"required": ["id", "name", "keycloakOrganizationId", "createdAt"]
```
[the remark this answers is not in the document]
slifty: Moved all these changes to a new branch (for consideration over in the right place) -- really the discussion should be on the issue and I only have a branch to capture the change in case we decide to use it!

Issue #1725

src/__tests__/dataProviders.int.test.ts
```diff
@@ -130,7 +133,10 @@ describe('/dataProviders', () => {
 				.put('/dataProviders/Firework_-foo42')
 				.type('application/json')
 				.set(adminUserAuthHeader)
-				.send({ name: '🎆' })
+				.send({
+					name: '🎆',
+					keycloakOrganizationId: null,
```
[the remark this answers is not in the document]
slifty: Moved all these changes to a new branch (for consideration over in the right place) -- really the discussion should be on the issue and I only have a branch to capture the change in case we decide to use it!

Issue #1725

bickelj (approved):
Sorry for that misleading comment in the type!

src/__tests__/userGroupChangemakerPermissions.int.test.ts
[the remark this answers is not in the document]
slifty: Ahhh got ya -- just to explain in case you're curious:

Basically, this line isn't accessing `mockKeycloakOrganizationId` directly, but rather it's accessing `organizationAsChangemaker.keycloakOrganizationId` which is a value populated and returned by `createChangemaker`.  `createChangemaker`'s return type is `Changemaker` which can potentially have a null value for `keycloakOrganizationId`.  It's true that we know (because we know inside of the queries / database) that the value will be non-null based on the input, but all of that happens inside of a black box that TypeScript isn't able to track (even if it didn't happen at the db level, TS doesn't support conditional return types!)

If I replaced this line with:

```
`/userGroups/${keycloakIdToString(mockKeycloakOrganizationId)}/changemakers/notanId/permissions/${Permission.MANAGE}`,
```

then `null` wouldn't be a possibility for sure.  That said, in either case, we still don't need to cast `as KeycloakId` since TypeScript already knows the type!

src/constants/httpStatus.ts
```diff
@@ -0,0 +1,84 @@
...
+	SWITCHING_PROTOCOLS = 101,
+	PROCESSING = 102,
+	EARLY_HINTS = 103,
+}
+
+enum SUCCESSFUL {
+	OK = 200,
+	CREATED = 201,
+	ACCEPTED = 202,
+	NON_AUTHORITATIVE_INFORMATION = 203,
```
bickelj, line 12: Or is it `NON-AUTHORITATIVE_INFORMATION`? Probably fine as-is!

src/constants/httpStatus.ts
```diff
@@ -0,0 +1,84 @@
...
+	NOT_FOUND = 404,
+	METHOD_NOT_ALLOWED = 405,
+	NOT_ACCEPTABLE = 406,
+	PROXY_AUTHENTICATION_REQUIRED = 407,
+	REQUEST_TIMEOUT = 408,
+	CONFLICT = 409,
+	GONE = 410,
+	LENGTH_REQUIRED = 411,
+	PRECONDITION_FAILED = 412,
+	PAYLOAD_TOO_LARGE = 413,
```
bickelj, line 46: [IANA](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) calls this `CONTENT_TOO_LARGE` :shrug:

src/constants/httpStatus.ts
```diff
@@ -0,0 +1,84 @@
...
+	URI_TOO_LONG = 414,
+	UNSUPPORTED_MEDIA_TYPE = 415,
+	RANGE_NOT_SATISFIABLE = 416,
+	EXPECTATION_FAILED = 417,
+	IM_A_TEAPOT = 418,
```
bickelj, line 51: I think [IANA](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) is being no fun on this one, calling it `(Unused)`!?

src/constants/httpStatus.ts
```diff
@@ -0,0 +1,84 @@
...
+	MISDIRECTED_REQUEST = 421,
+	UNPROCESSABLE_ENTITY = 422,
```
bickelj, line 53: Oh, interesting. [IANA](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) would call this `UNPROCESSABLE_CONTENT`.

src/constants/httpStatus.ts
```diff
@@ -0,0 +1,84 @@
...
+	LOCKED = 423,
+	FAILED_DEPENDENCY = 424,
+	TOO_EARLY = 425,
+	UPGRADE_REQUIRED = 426,
+	PRECONDITION_REQUIRED = 428,
```
bickelj, line 58: [IANA](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) would call this `UPGRADE_REQUIRED`.

src/handlers/applicationFormsHandlers.ts
```diff
@@ -56,7 +60,10 @@ const getApplicationForm = async (
 	}
 
 	const applicationForm = await loadApplicationForm(db, req, applicationFormId);
-	res.status(200).contentType('application/json').send(applicationForm);
+	res
+		.status(HTTP_STATUS.SUCCESSFUL.OK)
```
bickelj, line 64: Beautifully verbose, taxonomical, and clear.

src/jobQueue.ts
```diff
@@ -10,6 +10,9 @@ import type {
 
 const logger = getLogger(__filename);
 
+const CONCURRENT_JOB_COUNT = 5;
+const POLL_INTERVAL_MS = 1000;
```
bickelj, line 14: Is this the source of the evil reported in #1266?

src/middleware/errorHandler.ts
```diff
@@ -17,17 +18,17 @@ const logger = getLogger(__filename);
...
+			return HTTP_STATUS.CLIENT_ERROR.BAD_REQUEST;
 		case PostgresErrorCode.CHECK_CONSTRAINT_VIOLATION.valueOf():
-			return 400;
+			return HTTP_STATUS.CLIENT_ERROR.BAD_REQUEST;
 		case PostgresErrorCode.INSUFFICIENT_RESOURCES.valueOf():
-			return 503;
+			return HTTP_STATUS.SERVER_ERROR.SERVICE_UNAVAILABLE;
 		default:
-			return 500;
+			return HTTP_STATUS.SERVER_ERROR.INTERNAL_SERVER_ERROR;
```
bickelj, line 31: :face_holding_back_tears: it's so pretty.

src/middleware/errorHandler.ts
```diff
@@ -132,7 +133,7 @@ export const errorHandler = (
 		? new DatabaseError('Error with a database operation.', err)
 		: err;
 	const statusCode = getHttpStatusCodeForError(wrappedError);
-	if (statusCode >= 500) {
+	if (Object.values(HTTP_STATUS.SERVER_ERROR).includes(statusCode)) {
```
bickelj, line 136: :smiling_face_with_tear: :sob: too beautiful!

src/test/mockJwt.ts
```diff
@@ -10,6 +11,8 @@ const getMockJwks = (jwksPath = '/protocol/openid-connect/certs'): JWKSMock =>
 
 const mockJwks = getMockJwks();
 
+const AN_ARBITRARILY_LONG_TIME = 1000000;
```
bickelj, line 14: Or is it arbitrarily short? :thinking:

bickelj (approved):
I don't know who the actual authority on the HTTP status codes is. An RFC pointed me to the IANA as the registry which in turn points back to RFCs. There were a handful of differences in wording but they are minor.

I double-approve with the new HttpStatus enum!

src/constants/httpStatus.ts
slifty: Wow.

I'm going to override / ignore IANA just for this specific instance.

wow.

src/constants/httpStatus.ts
slifty: Actually this one *is* `428 	Precondition Required` -- upgrade required is 426

What these things mean, who is really to know!

src/constants/httpStatus.ts
bickelj: Have that LLM write a GitHub action that as part of our build process updates that enum from the IANA text file...

PR #1735, opened 2025-06-24
Add Instructions field to ApplicationFormField

### LINKED ISSUE #1670: Add the concept of `instructions` to `ApplicationFormField`
Applications sometimes show additional instruction beyond just a label to people filling out their forms.

We want to be able to capture that kind of information alongside the form field definitions.

The thinking is this should be `instructions` (rather than `description`) to make it clear that this is information that was presented to applicants (as opposed to metadata intended for PDC users)

hminsky2002:
This commit adds the new 'instructions' field to the ApplicationFormField entity, which is intented to provide clarification and explicit instruction for a user to fill in a given form field. Furthermore, it also gives a record of the instructions a user was operating under at the time of filling out the field, which should help us interpret user input that varies widely from what we're expecting.

Closes #1670


src/database/migrations/0065-alter-application_form_fields-add-instructions.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE application_form_fields
+ADD COLUMN instructions text;
+
+COMMENT ON COLUMN application_form_fields.instructions IS
+'Instructions for the user to fill information into the application form field.'; -- noqa: LT05
```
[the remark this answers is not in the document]
hminsky2002: Ah neat, i did not realize you could automatically concatenate on separate lines! Also agreed on clarifying the comment, will touch that up

src/__tests__/applicationForms.int.test.ts
```diff
@@ -202,24 +202,28 @@ describe('/applicationForms', () => {
...
 				position: 1,
 				label: 'Anni Worki',
+				instructions: 'Please enter the number of years of work.',
 			});
 			await createApplicationFormField(db, null, {
 				applicationFormId: 3,
 				baseFieldShortCode: 'organizationName',
 				position: 2,
 				label: 'Org Nomen',
+				instructions: 'Please enter the name of the organization.',
```
slifty, line 212: This is moot since you did it already, but you should feel OK with just making this null.  It's more test boilerplate and null is just as valid a value as a string.

Since you've written the tests with values already no need to spend time removing them!

That said, make sure to create a test that explicitly checks to confirm that `null` is a settable value for creation.

src/database/migrations/0065-alter-application_form_fields-add-instructions.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE application_form_fields
+ADD COLUMN instructions text;
+
+COMMENT ON COLUMN application_form_fields.instructions IS
+'Instructions (supplementing or in addition to the label) for '
+'the user to fill information into the application form field.';
```
slifty, line 6: Missing a newline (do you have editorconfig properly enabled on your IDE? it should prevent this from happening / insert newlines on save)

src/openapi/components/schemas/ApplicationFormField.json
```diff
@@ -31,6 +31,10 @@
 			"type": "string",
 			"example": "Your First Name"
 		},
+		"instructions": {
```
slifty, line 34: Let's add a description to this field so it shows up in documentation for API users.

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -637,6 +642,8 @@ describe('processBulkUploadTask', () => {
 										baseFieldShortCode: 'organization_name',
 										createdAt: expectTimestamp(),
 										id: expectNumber(),
+										instructions: '',
+
```
slifty, line 646: huh -- why are there all these extra lines of whitespace in this file?

src/types/ApplicationFormField.ts
```diff
@@ -10,6 +10,7 @@ interface ApplicationFormField {
 	readonly baseField: BaseField;
 	position: number;
 	label: string;
```
slifty, line 12: The DB allows null already, but this line needs updating to support null too:

`instructions: string | null`

src/types/ApplicationFormField.ts
```diff
@@ -33,8 +34,11 @@ const writableApplicationFormFieldWithApplicationContextSchema: JSONSchemaType<W
 			label: {
 				type: 'string',
 			},
+			instructions: {
+				type: 'string',
```
slifty, line 38: I believe this will be `type: ['string', 'null']`

src/openapi/components/schemas/ApplicationFormField.json
```diff
@@ -31,6 +31,10 @@
...
+			"type": "string",
```
slifty, line 35: This will be `"type": ["string", "null"]`

slifty (changes requested):
+1 to the request to make this nullable; also some whitespace stuff.

src/types/ApplicationFormField.ts
hminsky2002: I think this suffers from the same issue as in valueRelevanceHours in basefields, where AJV does not support nullable types in TypeScript. I added that same disable/eslint comment to follow pattern!

slifty (approved):
Yay!

PR #1737, opened 2025-06-24
Add funder collaboratives

### LINKED ISSUE #1639: Create entity for Funder Collaboratives
Exchange has the notion of Funder Collaboratives, which are groups of funder organizations that are working together. 

Some additional functionality: 
- Notion of collaborative admins, who may or may not be a funder admin. 
- Implementation of membership changes. Currently, Exchange lets administrators for a collaborative send invitations to organizations.
- Collaboratives should be an entity type that could be used for controlling permissions.

hminsky2002:
This PR adds the concept of funder collaboratives to the pdc, as outlined in Issue #1639. The PR is split up into two major commits 

1. fdf7520c4fea759d692924bf3d859e0f20d405fc -> This commit adds in the `funderCollaborativeMembers` entity, which serves as the concrete relationship between a funder and a collaborative, extending the funders permissions onto the collaborative (but not vice versa) . Currently the only way to initialize a new funderCollaborative is for an admin to create one, as a 1-to-1 relationship between a collaborative funder and a non-collaborative funder. It also adds an `isCollaborative`attribute to the funders table, which we decided was the best way to be able to ensure that a collaborative funder can not be a collaborator on another collaborative funder, and that a non-collaborative funder cannot be collaborated on

2. f0b185cacecf18f54daefd36b5616fbae311078e -> This commit adds the `funderCollaborativeInvitations` entity, which supports the intended method of adding collaborators to a collaborative. A funder must have manage permissions on the collaborative in question in order to send an invitation to another funder. A funder that recieves an invitation has the option to either accept or reject an invitation. If they choose to accept, a funderCollaborativeMember entity is created between the invitee and the collaborative funder in question

Closes #1639

### DIFF
--- a/src/database/migrations/0067-add-funder-collaborative-members.sql
+++ b/src/database/migrations/0067-add-funder-collaborative-members.sql
@@ -1,0 +1,78 @@
+ALTER TABLE funders ADD COLUMN is_collaborative boolean NOT NULL DEFAULT false;
+CREATE TABLE funder_collaborative_members (
+	funder_collaborative_short_code short_code_t NOT NULL
+	REFERENCES funders (short_code) ON DELETE CASCADE,
+	member_funder_short_code short_code_t NOT NULL
+	REFERENCES funders (short_code) ON DELETE CASCADE,
+	-- If the user that created this happens to be deleted,
+	-- we still don't want to delete this.
+	created_by uuid REFERENCES users (keycloak_user_id)
+	ON DELETE SET NULL,
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+	not_after timestamp with time zone DEFAULT null,
+	PRIMARY KEY (funder_collaborative_short_code, member_funder_short_code),
+	CONSTRAINT no_self_membership CHECK (
+		funder_collaborative_short_code != member_funder_short_code
+	)
+
+);
+
+CREATE OR REPLACE FUNCTION prevent_inserting_non_collaborative_funder()
+RETURNS trigger AS $$
+DECLARE
+    forbidden BOOLEAN;
+BEGIN
+    SELECT EXISTS (
+			SELECT 1
+				FROM funders
+				WHERE funders.short_code = NEW.funder_collaborative_short_code
+					AND funders.is_collaborative = false
+		) INTO forbidden;
+
+    IF forbidden THEN
+        RAISE EXCEPTION 'Non-collaborative funder %', NEW.funder_collaborative_short_code
+            USING ERRCODE = '23514';
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION prevent_inserting_collaborative_member_funder()
+RETURNS trigger AS $$
+DECLARE
+    forbidden BOOLEAN;
+BEGIN
+    SELECT EXISTS (
+			SELECT 1
+				FROM funders
+				WHERE funders.short_code = NEW.member_funder_short_code
+					AND funders.is_collaborative = true
+		) INTO forbidden;
+
+    IF forbidden THEN
+        RAISE EXCEPTION 'Collaborative funder cannot be a member of another collaborative %', NEW.funder_collaborative_short_code
+            USING ERRCODE = '23514';
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE TRIGGER check_funder_is_non_collaborative
+BEFORE INSERT ON funder_collaborative_members
+FOR EACH ROW
+EXECUTE FUNCTION prevent_inserting_non_collaborative_funder();
+
+CREATE TRIGGER check_member_funder_is_collaborative
+BEFORE INSERT ON funder_collaborative_members
+FOR EACH ROW
+EXECUTE FUNCTION prevent_inserting_collaborative_member_funder();
+
+SELECT audit_table('funder_collaborative_members');
+
+COMMENT ON COLUMN funders.is_collaborative IS
+'To help the application prevent recursive collaboratives, a'
+'funder must be either a collaborative or not a collaborative.';
--- a/src/database/queries/funderCollaborativeMembers/selectByShortCode.sql
+++ b/src/database/queries/funderCollaborativeMembers/selectByShortCode.sql
@@ -1,0 +1,8 @@
+SELECT
+	funder_collaborative_member_to_json(funder_collaborative_members.*)
+		AS object
+FROM funder_collaborative_members
+WHERE
+	funder_collaborative_short_code = :funderCollaborativeShortCode
+	AND member_funder_short_code = :memberFunderShortCode
+	AND NOT is_expired(not_after);
--- a/src/handlers/fundersHandlers.ts
+++ b/src/handlers/fundersHandlers.ts
@@ -65,11 +65,12 @@ const putFunder = async (req: Request, res: Response): Promise<void> => {
 		);
 	}
 
-	const { name, keycloakOrganizationId } = body;
+	const { name, keycloakOrganizationId, isCollaborative } = body;
 	const funder = await createOrUpdateFunder(db, null, {
 		shortCode,
 		name,
 		keycloakOrganizationId,
+		isCollaborative,
 	});
 	res
 		.status(HTTP_STATUS.SUCCESSFUL.CREATED)
--- a/src/openapi/paths/funderCollaborativeMember.json
+++ b/src/openapi/paths/funderCollaborativeMember.json
@@ -1,0 +1,153 @@
+{
+	"get": {
+		"operationId": "getFunderCollaborativeMemberByShortCode",
+		"summary": "Gets a specific funder collaborative member.",
+		"tags": ["Funder Collaborative Members"],
+		"security": [
+			{
+				"auth": []
+			}
+		],
+		"parameters": [
+			{
+				"name": "funderCollaborativeShortCode",
+				"description": "The short code of a funder collaborative.",
+				"in": "path",
+				"required": true,
+				"schema": {
+					"type": "string"
+				}
+			},
+			{
+				"name": "memberFunderShortCode",
+				"description": "The short code of funder to be made a member of the collaborative.",
+				"in": "path",
+				"required": true,
+				"schema": {
+					"type": "string"
+				}
+			}
+		],
+		"responses": {
+			"200": {
+				"description": "The funder collaborative member.",
+				"content": {
+					"application/json": {
+						"schema": {
+							"$ref": "../components/schemas/FunderCollaborativeMember.json"
+						}
+					}
+				}
+			},
+			"404": {
+				"$ref": "../components/responses/NotFound.json"
+			}
+		}
+	},
+	"post": {
+		"operationId": "createOrUpdateFunderCollaborativeMember",
+		"summary": "Creates or updates a new funder collaborative member.",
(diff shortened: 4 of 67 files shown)


src/database/queries/funderCollaborativeMembers/selectByShortCode.sql
```diff
@@ -0,0 +1,3 @@
+SELECT funder_collaborative_member_tojson(funder_collaborative_members.*) AS object
```
bickelj, line 1: typo in `to_json`

src/database/migrations/0061-add-funder-collaboratives.sql
```diff
@@ -0,0 +1,22 @@
...
+	CONSTRAINT no_self_membership CHECK (
+		funder_collaborative_short_code != member_short_code
+	)
+);
+
+COMMENT ON COLUMN funders.is_collaborative IS
+'To help the application prevent recursive collaboratives, '
+'a funder must be either a collaborative or not a collaborative.';
+
+-- There is no way to check the `funders.is_collaborative` in the foreign table `funders` as part
```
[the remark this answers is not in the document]
bickelj: From https://www.postgresql.org/docs/17/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS:

> Note
>
> PostgreSQL does not support CHECK constraints that reference table data other than the new or updated row being checked. While a CHECK constraint that violates this rule may appear to work in simple tests, it cannot guarantee that the database will not reach a state in which the constraint condition is false (due to subsequent changes of the other row(s) involved). This would cause a database dump and restore to fail. The restore could fail even when the complete database state is consistent with the constraint, due to rows not being loaded in an order that will satisfy the constraint. If possible, use UNIQUE, EXCLUDE, or FOREIGN KEY constraints to express cross-row and cross-table restrictions.
>
> If what you desire is a one-time check against other rows at row insertion, rather than a continuously-maintained consistency guarantee, a custom [trigger](https://www.postgresql.org/docs/current/triggers.html) can be used to implement that. (This approach avoids the dump/restore problem because pg_dump does not reinstall triggers until after restoring data, so that the check will not be enforced during a dump/restore.)

src/database/migrations/0061-add-funder-collaboratives.sql
```diff
@@ -0,0 +1,22 @@
...
+  funder_collaborative_short_code short_code_t NOT NULL
+  REFERENCES funders (short_code) ON DELETE CASCADE,
+  member_short_code short_code_t NOT NULL
+  REFERENCES funders (short_code) ON DELETE CASCADE,
+  -- If the user that created this happens to be deleted, we still don't want to delete this.
+	created_by uuid REFERENCES users (keycloak_user_id)
+	ON DELETE SET NULL,
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+  PRIMARY KEY (funder_collaborative_short_code, member_short_code),
+	CONSTRAINT no_self_membership CHECK (
```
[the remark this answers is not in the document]
bickelj: Please let's do no recursion!

src/handlers/fundersHandlers.ts
```diff
@@ -79,8 +110,186 @@ const putFunder = async (req: Request, res: Response): Promise<void> => {
...
+			'Invalid request body.',
+			isWritableFunderCollaborativeInvitationForPatch.errors ?? [],
+		);
+	}
+
+	const { invitationStatus } = body;
+
+	const funderCollaborativeInvitation =
+		await updateFunderCollaborativeInvitation(db, req, {
+			funderShortCode: invitationShortCode,
```
hminsky2002, line 271: Noting here that I am very aware that this is ridiculous -- @slifty I know there's a better way than swapping variable names here, but I was struggling to think of what the best way to distinguish the `funderShortCode` from the `invitationShortCode`. In the POST route, we have `funders/{funderShortCode}/invitations/sent` with `invitationShortCode` included in the body, and i think that naming tracks because we're sending an invitation FROM {funderShortCode} TO {invitationShortCode}. However, in the PATCH route, the {funderShortCode} in question has now become the invitee. The right solution here might be reworking the PATHs slightly, but I didn't wanna make that change until you'd had a chance to review it

src/types/FunderCollaborativeInvitation.ts
```diff
@@ -0,0 +1,77 @@
...
+		Pick<
+			FunderCollaborativeInvitation,
+			'invitationStatus' | 'invitationShortCode'
+		>;
+
+type InternallyWritableFunderCollaborativeInvitation =
+	WritableFunderCollaborativeInvitation &
+		Pick<FunderCollaborativeInvitation, 'funderShortCode'>;
+
+type WritableFunderCollaborativeInvitationForPatch =
```
hminsky2002, line 32: @slifty also doing some typescript flailing here -- this was my 'it works' solution to having 'invitationShortCode' be a body parameter for POSTing but a path parameter for PATCHing. Very willing to change this

slifty (dismissed):
Lots-o-thoughts below; happy to pair on this.

src/handlers/fundersHandlers.ts
```diff
@@ -65,20 +77,327 @@ const putFunder = async (req: Request, res: Response): Promise<void> => {
...
 		keycloakOrganizationId,
+		isCollaborative,
+	});
+	res
+		.status(HTTP_STATUS.SUCCESSFUL.CREATED)
+		.contentType('application/json')
+		.send(funder);
+};
+
+const getFunderCollaborativeMember = async (
```
slifty, line 93: Let's put the `funderCollaborativeMember` handlers in their own handlers

src/handlers/fundersHandlers.ts
```diff
@@ -65,20 +77,327 @@ const putFunder = async (req: Request, res: Response): Promise<void> => {
...
+		throw new InputValidationError(
+			'Invalid request body.',
+			isFunderCollaborativeInvitationPatch.errors ?? [],
+		);
+	}
+
+	const { invitationStatus } = body;
+
+	const funderCollaborativeInvitation =
+		await createOrUpdateFunderCollaborativeInvitation(db, req, {
```
slifty, line 374: Let's make this a transaction (along with the creation of the membership) -- that way if one fails, the other fails too.

```
const transactionDb = await db.transaction();
...
await transactionDb.commit();
```

I believe we use transactions in proposal version POST so you can see an example there.

src/handlers/fundersHandlers.ts
```diff
@@ -65,20 +77,327 @@ const putFunder = async (req: Request, res: Response): Promise<void> => {
...
+	if (!isFunderCollaborativeInvitationPatch(body)) {
+		throw new InputValidationError(
+			'Invalid request body.',
+			isFunderCollaborativeInvitationPatch.errors ?? [],
+		);
+	}
+
+	const { invitationStatus } = body;
+
+	const funderCollaborativeInvitation =
```
slifty, line 373: We also talked about how this should be a new `updateFunderCollaborativeInvitation` method which will use the `updateFoo` pattern / take in a partial AND (importantly) fail if the entity does not already exist.

We are also changing `createOrUpdate` to just be `create...` and it will only succeed if the entity does not already exist.

src/handlers/fundersHandlers.ts
```diff
@@ -65,20 +77,327 @@ const putFunder = async (req: Request, res: Response): Promise<void> => {
...
+
+	if (!isFunderCollaborativeInvitationPatch(body)) {
+		throw new InputValidationError(
+			'Invalid request body.',
+			isFunderCollaborativeInvitationPatch.errors ?? [],
+		);
+	}
+
+	const { invitationStatus } = body;
+
```
slifty, line 372: There is a question of where to put any logic related to checking the "state" of an invitation and ensuring that the state change is actually allowed (e.g. we can't move from ACCEPTED to REJECTED).

This could be a trigger defined at the migration (maybe should!)
It could also be defined as a check in this handler.
It could also be defined as a check in the `update` query.

hminsky2002: ended up making it a migration trigger as this seemed like the safest option to guard against race conditions!

src/database/migrations/0067-add-funder-collaborative-members.sql
```diff
@@ -0,0 +1,77 @@
...
+	ON DELETE SET NULL,
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+	not_after timestamp with time zone DEFAULT null,
+	PRIMARY KEY (funder_collaborative_short_code, member_funder_short_code),
+	CONSTRAINT no_self_membership CHECK (
+		funder_collaborative_short_code != member_funder_short_code
+	)
+
+);
+
```
bickelj, line 19: We (really I) forgot a `SELECT audit_table('funder_collaborative_members');` here.

bickelj:
@hminsky2002 Since the first commit dwarfs my original commit, I think you should change the author to you.

bickelj (commented):
I was able to add funder collaboratives, add (non-collaborative) funders to those as members, and list those out. I was prevented from creating invalid combinations.

I half-expected to see `members` as a list in the usual `funder` object, but given that most of the time these funders are not collaboratives I can see why that would not be the case. It is kind of awkward to expose the membership relations themselves via a specific endpoint but it is probably the best alternative to the above.

I have a few comments scattered about here and there with stuff I found along the way.

Overall, it works! This is great! Thank you!

src/routers/fundersRouter.ts
```diff
@@ -18,4 +25,51 @@ fundersRouter.put(
...
+	funderCollaborativeMembersHandlers.getFunderCollaborativeMembers,
+);
+
+fundersRouter.get(
+	'/:funderShortCode/members/:memberFunderShortCode',
+	requireFunderPermission(Permission.MANAGE),
+	funderCollaborativeMembersHandlers.getFunderCollaborativeMember,
+);
+
+fundersRouter.post(
```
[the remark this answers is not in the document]
hminsky2002: I talked through this with @slifty and while it IS a createOrUpdate, it is only meant to be used as an update in the case that we are re-activating a funderCollaborativeMember (following the soft-delete not-after pattern). This is to say, as far as the user is concerned, we're making a new entity entirely.

src/openapi/paths/funderCollaborativeMember.json
```diff
@@ -0,0 +1,150 @@
...
+				"description": "The short code of a collaborative funder.",
+				"in": "path",
+				"required": true,
+				"schema": {
+					"type": "string"
+				}
+			},
+			{
+				"name": "memberFunderShortCode",
+				"description": "The short code of a member non-collaboraitve funder.",
```
bickelj, line 68: Typo here

src/openapi/paths/funderCollaborativeMember.json
```diff
@@ -0,0 +1,150 @@
...
+						}
+					}
+				}
+			},
+			"400": {
+				"$ref": "../components/responses/DatabaseError.json"
+			},
+			"401": {
+				"$ref": "../components/responses/Unauthorized.json"
+			}
```
bickelj, line 102: I also get a 422 here (currently undocumented) when adding a member that doesn't exist.

src/routers/fundersRouter.ts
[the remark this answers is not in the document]
bickelj: Along these lines, idempotent is the actual reason, not the particular method used. As an example, when I POST an existing entity, right now I get back a 201 created with an old/existing object rather than 200 OK. I think 200 OK might be better if no new resource was created.

bickelj (approved):
While I didn't test everything, I think this is in good enough shape now to merge, hurray!

PR #1745, opened 2025-07-01
Add documentation for Okta IdP integration

bickelj:


slifty (approved):
Looks good

My only concern honestly is that this feels like Keycloak documentation more than PDC documentation, and as a result I expect it will go stale.  Ideally we'd be able to just link to relevant Keycloak documentation 😰)

That said, we can deal with that if / when it goes stale instead

bickelj:
@slifty I share your concern about staleness. The answer, though, is that this is the result of a tested integration of a particular integration tech (SAML) with a particular IdP (Okta) and our particular configuration of Keycloak. So there is value in it even if it goes stale. I added some direct links to the Keycloak documentation in here. The reason I made them link to a particular version is that those links don't break over time and it clearly communicates the version against which these documents were written. What do you think?

PR #1752, opened 2025-07-08
Add `funder` filter to `GET /proposals`

### LINKED ISSUE #1747: Support `funder` filter on `/proposals`
The proposals endpoint accepts a `changemaker` filter, we also want to be able to filter by associated funder.

This is related to #1646 but this specific issue is more specific and straightforward so I wanted to pull it out.

What makes this different from search is this is a filter by funder ID as opposed to a faceted search such as `"funderName: foo"`

slifty:
This PR adds a `funder` filter to the `GET /proposals` endpoint.

Resolves #1747

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.20.1 2025-07-10
+
+### Added
+
+- `GET /proposals` now accepts a `funder` querystring filter which returns a subset of proposals based on the funder of the proposal's opportunity.
+
 ## 0.20.0 2025-06-17
 
 ### Changed
--- a/src/database/queries/proposals/selectWithPagination.sql
+++ b/src/database/queries/proposals/selectWithPagination.sql
@@ -1,5 +1,7 @@
 SELECT proposal_to_json(proposals.*) AS object
 FROM proposals
+	INNER JOIN opportunities
+		ON proposals.opportunity_id = opportunities.id
 	LEFT JOIN proposal_versions ON proposals.id = proposal_versions.proposal_id
 	LEFT JOIN
 		proposal_field_values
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.20.0 2025-07-08
+
+### Added
+
```
hminsky2002, line 11: should this not be 0.20.1 if we're adding something?

src/queryParameters/extractFunderParameters.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { JSONSchemaType } from 'ajv';
+import type { Request } from 'express';
+import type { ShortCode } from '../types';
+
+interface FunderParameters {
+	funderShortCode: ShortCode | undefined;
+}
+
+interface FunderParametersQuery {
+	funder: ShortCode | undefined;
```
hminsky2002, line 13: I feel like the name of this field should reflect that it's expecting a shortcode, like `queryFunderShortCode`? Is that too obtuse?

hminsky2002 (dismissed):
Looks good overall and cool that this took all of 25 minutes! Tests are comprehensive and things are running locally. I think you forgot to bump the version though (unless there is a pattern for not updating the version at all for tiny changes)

bickelj (approved):
It works. I am able to get proposals using a funder shortCode either as a `pdc-admin` or when my organization has been granted permission to view that funder, and I cannot see the proposals when neither of those is true.

At first I would have suggested we change the query parameter name to `funderShortCode` for consistency with the rest of the API but it appears we use the simplified (and frankly better) names elsewhere in query parameters. It is in URL parameters that we are more explicit.

It seems odd to call it `FunderParameters` when there is only one parameter. But I see that `PaginationParameters` has more than one and we have the potential to filter on more than one in future.

src/queryParameters/extractFunderParameters.ts
slifty: I hear you but will push back:

(1) we have the same pattern for `ChangemakerParameters` so this is mirroring that.
(2) The intention of the `changemaker` filter api design is this might eventually expand to be other types of identifier (e.g. a URI) as well.

This is a little different from our REST endpoints; it's a query parameter for filtering and so we can be more flexible as a result.

hminsky2002: Ok ok you and at @bickelj have convinced me -- i now think that in the context of a query parameter from the user perspective this makes sense!

src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -41,18 +43,18 @@ WHERE
...
-			FROM opportunities
-			WHERE
-				opportunities.id = proposals.opportunity_id
-				AND has_funder_permission(
-					:authContextKeycloakUserId,
-					:authContextIsAdministrator,
-					opportunities.funder_short_code,
-					'view'
-				)
+		has_funder_permission(
```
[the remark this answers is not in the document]
slifty: Agreed -- and it was probably unnecessarily less efficient as written before... whoops 😅

CHANGELOG.md
slifty: Bumped to 0.20.1 as mentioned!

Also I'm seeing the answer to the question I had on your PR -- the version in the spec hadn't been bumped at some point accidentally!

Issue #1773, opened 2025-07-14
Support OpportunityPermission grants


slifty:
We have the ability to give permissions on funders, changemakers, and data providers.  Time to add support for Opportunities!

This functionality is most immediately driven by #1677, but this feature will allow funders to have more granular control around specific opportunities.

This change should come with a new type of permission: `submit proposal` which determines which users are allowed to create new proposals against a given opportunity.

slifty:
For this stage I'm only implementing `MANAGE`, `EDIT`, `VIEW`, and `CREATE_PROPOSALS` since that is the immediate need.

Relatedly, I spoke with @bickelj and there may be some improvements to the permission system we want to make to avoid a ton of bespoke "verbs" that represent actions on sub-entities of a given entity (basically: instead of having a `CREATE_PROPOSALS` permission we may want to have a `{ entities: ['Opportunity', 'Proposal', ...]}` model attribute to a given permission which would be much more flexible.

PR #1777, opened 2025-07-14
Add `opportunity` permissions

slifty:
This PR creates a new `UserOpportunityPermission` and `UserGroupOpportunityPermission` grant type which will allow permissions to be granted against opportunities.

This is intended to covers use cases such as "Funder A would like to allow User B and Organization C to create new proposals within Opportunity 1" without providing permission across all of their opportunities.

This PR does not actually implement the resulting access, but it does create the permissions themselves.

Related to #1773


src/handlers/userGroupOpportunityPermissionsHandlers.ts
```diff
@@ -0,0 +1,111 @@
...
+		.send();
+};
+
+const putUserGroupOpportunityPermission = async (
+	req: Request,
+	res: Response,
+): Promise<void> => {
+	if (!isAuthContext(req)) {
+		throw new FailedMiddlewareError('Unexpected lack of auth context.');
+		return;
```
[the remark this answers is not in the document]
slifty: for anybody watching: the linter will catch this eventually (the rule is disabled right now just as part of a short term migration)

src/types/UserOpportunityPermission.ts
```diff
@@ -0,0 +1,41 @@
...
+		Pick<
+			UserOpportunityPermission,
+			'userKeycloakUserId' | 'opportunityPermission' | 'opportunityId'
+		>;
+
+const writableUserOpportunityPermissionSchema: JSONSchemaType<WritableUserOpportunityPermission> =
+	{
+		type: 'object',
+		properties: {},
+		required: [],
```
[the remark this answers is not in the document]
slifty: This one isn't correct, though speaks to #1770

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A new `PUT /userGroups/{keycloakOrganizationId}/opportunities/{opportunityId}/permissions/{opportunityPermission}` endpoint now exists
 - A new `DELETE /userGroups/{keycloakOrganizationId}/opportunities/{opportunityId}/permissions/{opportunityPermission}` endpoint now exists
 
+### Fixed
```
hminsky2002, line 17: I think this should probably be it's own PR, it being unrelated to any code changes made in the previous commit (though still about permissions). I could be persuaded though, why did you decide to group them?

hminsky2002 (changes requested):
Patterns all followed and code runs locally, I think this is the right shape for the entity here -- nice work @slifty !

src/database/migrations/0066-create-opportunity-permissions-tables.sql
```diff
@@ -0,0 +1,28 @@
...
+	) ON DELETE CASCADE,
+	opportunity_permission opportunity_permission_t NOT NULL,
+	opportunity_id int NOT NULL REFERENCES opportunities (id) ON DELETE CASCADE,
+	created_by uuid NOT NULL REFERENCES users (keycloak_user_id) ON DELETE CASCADE,
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+	not_after timestamp with time zone DEFAULT NULL,
+	PRIMARY KEY (user_keycloak_user_id, opportunity_id, opportunity_permission)
+);
+
+CREATE TABLE user_group_opportunity_permissions (
```
slifty, line 20: I just realized that we didn't run the audit log function on this.

I do think we should add a migration that sets up that function as a trigger on table create....

slifty: OK fixed!

src/database/migrations/0066-create-opportunity-permissions-tables.sql
```diff
@@ -0,0 +1,31 @@
...
+	keycloak_organization_id uuid NOT NULL,
+	opportunity_id int NOT NULL REFERENCES opportunities (id) ON DELETE CASCADE,
+	opportunity_permission opportunity_permission_t NOT NULL,
+	created_by uuid NOT NULL REFERENCES users (keycloak_user_id) ON DELETE CASCADE,
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+	not_after timestamp with time zone DEFAULT NULL,
+	PRIMARY KEY (keycloak_organization_id, opportunity_id, opportunity_permission)
+);
+
+SELECT audit_table('user_opportunity_permissions'
```
hminsky2002, line 30: Looks like this is causing the lint and testing ci/cd failure -- approving now and assuming this will be resolved to resolve ci/cd!

hminsky2002 (approved):

PR #1782, opened 2025-07-15
Remove `node-fetch`

### LINKED ISSUE #1781: Remove `node-fetch`
We are on Node 22 which has native support for fetch.

We were waiting for `nock` support before removing `node-fetch` but happy days that is now in place: https://github.com/nock/nock/issues/2397

Time to get rid of node fetch, which I believe will also remove this deprecation warning we see when running tests:

```
(node:19975) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

slifty:
This PR removes `node-fetch` as a dependency.  🎉 

Resolves #1781


bickelj (approved):
LGTM. It is always nice to see a net code decrease.

I am still befuddled by `fetch` versus `get`. I vaguely recall having the same reaction in the past and smarter people explained it to me. In any case, this change makes the code better than before and should be merged with haste.

slifty:
I mentioend this in Zulip but realize it should be here:

@Jesse Bickel I didn't remember the reason so looked it up but: https://github.com/nodejs/help/issues/4074

Basically fetch is promise based and implements an API that exists in more contexts (e.g. browsers) -- while that doesn't impact us directly, it does mean that if tooling is going to be built around one it will be built around fetch. There's no clear downside IMHO to using fetch but there are downsides to using http / https so that's the reason for using fetch.

PR #1790, opened 2025-07-17
Enable all remaining `love` rules

### LINKED ISSUE #1748: Remove the disabled `love` lint rules
We turned off airbnb rules (#1699)

We enabled a LOT of the new rules, but some remain.  This issue tracks the desire to re-enable those last few.

slifty:
This PR enables the remaining disabled `love` rules and then a few others that were disabled in the past for bonus points.

Resolves #1748

### DIFF
--- a/src/__tests__/applicationForms.int.test.ts
+++ b/src/__tests__/applicationForms.int.test.ts
@@ -808,21 +808,21 @@ describe('/applicationForms', () => {
 			await createTestBaseFields();
 			jest
 				.spyOn(db, 'sql')
-				.mockImplementationOnce(async () => ({
-					command: '',
-					row_count: 1,
-					rows: [
-						{
-							id: 1,
-							opportunityId: 1,
-							version: 1,
-							createdAt: new Date(),
-						},
-					],
-				}))
-				.mockImplementationOnce(async () => {
-					throw new Error('This is unexpected');
-				});
+				.mockReturnValueOnce(
+					Promise.resolve({
+						command: '',
+						row_count: 1,
+						rows: [
+							{
+								id: 1,
+								opportunityId: 1,
+								version: 1,
+								createdAt: new Date(),
+							},
+						],
+					}),
+				)
+				.mockReturnValueOnce(Promise.reject(new Error('This is unexpected')));
 			const result = await request(app)
 				.post('/applicationForms')
 				.type('application/json')
--- a/src/middleware/errorHandler.ts
+++ b/src/middleware/errorHandler.ts
@@ -56,45 +56,50 @@ const getNameForError = (error: unknown): string => {
 	return 'UnknownError';
 };
 
+const getHttpStatusCodeFromDatabaseError = (error: DatabaseError): number => {
+	const {
+		tinyPgError: {
+			queryContext: {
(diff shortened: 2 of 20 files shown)


src/middleware/errorHandler.ts
```diff
@@ -56,45 +56,50 @@ const getNameForError = (error: unknown): string => {
...
-		typeof error === 'string' &&
-		error.includes('exceeds maximum tokens per interval')
-	) {
+
+	if (isJwksRsaRateLimitError(error)) {
 		return HTTP_STATUS.SERVER_ERROR.SERVICE_UNAVAILABLE;
 	}
+
+	if (error instanceof Error) {
+		switch (error.constructor) {
```
bickelj, line 84: Weird. Switches used to be considered bad form not too recently. Maybe it varies by language. Here it is OK either way because of the return statements, including a return in the default.

src/promises/__tests__/allNoLeaks.unit.test.ts
```diff
@@ -1,14 +1,12 @@
...
 
 describe('allNoLeaks', () => {
 	it('should throw an error after companion long-duration `Promise` completes', async () => {
 		let longTaskCompleted = false;
-		const longTask = new Promise<void>((resolve) => {
-			setTimeout(() => {
-				longTaskCompleted = true;
-				resolve();
-			}, 50);
+		const longTask = setTimeout(50).then(() => {
```
bickelj, line 8: I wonder if this is for sure the same functionality. I can try it out to see. Well, yeah, the unit test I wrote here still works correctly, duh. :tada:

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -163,7 +164,12 @@ describe('processBulkUploadTask', () => {
 		);
 		const requests = await mockS3ResponsesForBulkUploadTaskProcessing(
 			bulkUploadTask,
-			`${__dirname}/fixtures/processBulkUploadTask/validCsvTemplate.csv`,
+			path.join(
```
bickelj, line 167: Fancy!

src/__tests__/applicationForms.int.test.ts
```diff
@@ -808,21 +808,21 @@ describe('/applicationForms', () => {
...
-							version: 1,
-							createdAt: new Date(),
-						},
-					],
-				}))
-				.mockImplementationOnce(async () => {
-					throw new Error('This is unexpected');
-				});
+				.mockReturnValueOnce(
+					Promise.resolve({
```
bickelj, line 812: Feels like a step backward but it's a narrow case, right?

bickelj (approved):
No red flags here, LGTM!

src/__tests__/applicationForms.int.test.ts
bickelj: Then groovy, hurray, yea `Promise`s! Finally those `async` pigs will pay for their crimes, eh, comrades?

bickelj: It's an immediately resolved promise, so yeah, I se how it makes sense.

PR #1795, opened 2025-07-18
Honor `create_proposals` opportunity permission

slifty:
This PR actually honors the `CREATE_PROPOSALS` and `VIEW` permission, allowing users with those permissions to create proposals.

Related to #1773

### DIFF
--- a/src/__tests__/proposals.int.test.ts
+++ b/src/__tests__/proposals.int.test.ts
@@ -19,6 +19,7 @@ import {
 	createOrUpdateUserFunderPermission,
 	createOrUpdateFunder,
 	createOrUpdateUserChangemakerPermission,
+	createOrUpdateUserOpportunityPermission,
 } from '../database';
 import { getAuthContext, loadTestUser } from '../test/utils';
 import {
@@ -37,6 +38,7 @@ import {
 	BaseFieldSensitivityClassification,
 	keycloakIdToString,
 	Permission,
+	OpportunityPermission,
 } from '../types';
 
 const createTestBaseFields = async () => {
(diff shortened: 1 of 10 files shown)


src/database/queries/opportunities/selectById.sql
```diff
@@ -2,9 +2,17 @@ SELECT opportunity_to_json(opportunities.*) AS object
...
-		opportunities.funder_short_code,
-		'view'
+	AND (
+		has_funder_permission(
+			:authContextKeycloakUserId,
+			:authContextIsAdministrator,
+			opportunities.funder_short_code,
+			'view'
+		)
+		OR has_opportunity_permission(
```
bickelj, line 12: If we view the opportunity as being "under" a funder, I think this should would be an `AND`. But maybe we want to keep them separate, you could view the opportunity without seeing funder info, thus `OR`. Hmm.

slifty: This is explicitly an or -- the whole point of these permissions is that users who are not members of the funder organization need to be able to make proposals.

src/__tests__/proposals.int.test.ts
```diff
@@ -1587,6 +1589,66 @@ describe('/proposals', () => {
...
+			const before = await loadTableMetrics('proposals');
+			await request(app)
+				.post('/proposals')
+				.type('application/json')
+				.set(authHeader)
+				.send({
+					externalId: 'proposal123',
+					opportunityId: opportunity.id,
+				})
+				.expect(422);
```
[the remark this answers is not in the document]
slifty: Actually I'm realizing that 422 may actually be right:  422 is saying "you tried to make a proposal for an opportunity that doesn't exist" -- we don't want to expose which opportunities exist if you don't have the right to view them.

Basically this aligns with the idea that 404 should be returned when loading an opportunity that you don't have access to.

Issue #1744, opened 2025-06-30
Create endpoint to determine requesting user's organization memberships


jayqi:
In Exchange, there will be functionality that is tied to the organization (funder or changemaker) that a user is a member of, e.g., structured interactions.

Exchange needs a way to determine what organizations PDC considers a user to be associated with (and maybe also permissions?) to be able to correctly manage functionality. 

Exchange will be making the request using the user's bearer token, so it should presumably only provide data about the requesting user.

slifty:
@bickelj the use case above I believe is related to contexts where a JWT doesn't exist; (I can explain more easily by voice, but it's related to some functionality that was shown at a meeting I was in with @jayqi last week).

I'm in the process of documenting some current state + possible paths forward.

jayqi:
Cross-posting my summary post from Zulip:

> - Exchange will need to keep track of user–KeycloakOrg and KeycloakOrg–Funder/Changemaker/DataProvider relationships, so that Exchange knows how to display or act on the right Funder/Changemaker/DataProvider for organization-centric features. 
>   - Some examples of these features would be:
>     - Structured interactions between two organizations
>     - A user writes proposal data to PDC on behalf of a changemaker ([#1677](https://github.com/PhilanthropyDataCommons/service/issues/1677))
>     - A user wants to view "my organization's proposals", i.e.,  all proposals associated with a funder/changemaker/data provider that the user is associated with
>   - How Exchange models this data is an Exchange implementation detail, but likely we'll do the flat organization model as has been suggested
> - Exchange should use Keycloak's `organizations[:].id` claims for a given user to determine whether they should be associated with a Keycloak organization. This is returned as part of the user's access token from Keycloak
>   - This will only be as up-to-date as the last time this user authenticated with Keycloak (either current session log in, or when the user's access token was last refreshed)
> - To determine the KeycloakOrg–Funder/Changemaker/DataProvider relationships, Exchange will use a still-to-be-implemented that is given a Keycloak organization ID and returns all `funderShortCode`/`changemakerId`/`dataProviderShortCode` values for the respective PDC entities that are associated with that Keycloak organization ID. 
>   - This will only tell Exchange about KeycloakOrg–Funder/Changemaker/DataProvider for which users with claims to those Keycloak orgs have logged into Exchange. I think this is fine for all of the features we have in mind.

PR #1804, opened 2025-07-22
Add an `/organizations/:keycloakId` endpoint

bickelj:
Without this endpoint, a caller would have to call all of the changemaker, data provider, and funder endpoints and sort through them to find the PDC ID for a Keycloak Organization UUID. With this endpoint, a caller can pass a Keycloak OrganizationUUID (from a JWT) and get exactly the changemaker, data provider, and funder associated.

Issue #1744 Add organizations endpoint


bickelj:
@slifty I'm having trouble getting the postgresql function to return 0 rows instead of 1 row, which is apparently what our generator function expects. I could modify the generator to check for a `null` result but I figure that is not a great solution.

slifty (approved):
This looks great!

bickelj:
[e87fe85](https://github.com/PhilanthropyDataCommons/service/commit/e87fe85d78cd184fda069d5f03988a8f13a894bc) to [79a35d7](https://github.com/PhilanthropyDataCommons/service/commit/79a35d7986caefa960bee973a24bb0cfe1fca248) was only a rebase on main.

Issue #1689, opened 2025-06-03
Support `file` types for base fields


slifty:
We want to support file-based data types.

We have some support for presigned posts.

slifty:
Documenting some related thoughts:

There are pros and cons to creating a base field type that could be considered to be either a file *or* text.  The pro is flexibility and native association between the two types of value, the con is meaningful additional complexity related to data access and validation (knowing a value is always of a given type allows you to develop explicit validation, rendering, processing, etc. logic).

At the field level value I do think we need to know if a field value is a free form string or a presigned-post-uploaded file URL *by the point of data entry*.  If the base field type is explicit ("this base field will always be associated with plain text") then that solves the problem.

If we create a hybrid base field type (e.g. `file-or-string`) or if we decide to treat some fields as "file is allowed instead of the specified base field type" then we would need our field values to have an `is_file` flag.

## A: `file` base field type

The "simplest" solution, I think, would be to have a `file` base field type which does not accept plain text.

Under this model these might be three distinct *BASE FIELDS*

1. Annual budget report (file)
2. Annual budget (number)
3. Annual budget narrative (string)

(2) could be treated as numeric (e.g. when we add faceted search we could define numeric ranges against this)
(1) and (3) could be accessed as text, but the mechanism behind the searches would of course be distinct since in the file case the file contents would need to be loaded or cached.

## B: `hybrid` base field type (either universal-to-any-type or specific-to-some-types)

Under this model, sometimes a base field's value is one type (e.g. `string`) but sometimes it is a URL that points to a file which would need to be processed in order to access the content.

This would mean that the expected type of a value is no longer captured in the value type defined in a given base field, but that a field value could override that type and specify it is a file.

bickelj:
@slifty FWIW I support (A) or some variation of (A), in other words having a separate base field of type "file" or type "pdf."

1. Not all fields should be allowed to be a file. The organization name or tax ID, for example, should not be files ever. I suppose this is still acknowledged in the (B) case and not every field would be set to allow files, so this is not a strong argument.
2. For the purposes of clarity and simplicity, separating the base fields will likely pay off in lower implementation complexity when implementing (the inevitable) search across these documents. We can more easily exclude file URLs from search results when they are explicitly marked as files and more easily include the contents of those files in search results when the two are strictly separated.

kfogel:
Agree with @bickelj FWIW.

slifty:
[...]

PR #1861, opened 2025-08-20
Add a `File` entity

slifty:
This PR adds a new `File` entity to the pdc.  Prior to this change users had no associations with files they uploaded to S3.

This new type replaces `PresignedPostRequest` in our API, and the endpoint is now `POST /files`.

The only other change is that `name` is now a required attribute.

Related to #1689


slifty:
Most of the lines codecov is complaining about are just the standard "we aren't bothering to test the middleware failure logic" (though I wonder if it would be possible to write a test that checks that for every routed endpoint that uses the auth middleware... 🤔 sorry that's not relevant here)

There *is* one line that it flags though, which is the check that the user requesting a presigned post is the user that created the referenced file.  Adding that test now!

src/constants/strings.ts
```diff
@@ -0,0 +1,2 @@
+export const UUID_PATTERN =
+	/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```
bickelj, line 2: If using UUIDs, should we use https://github.com/socketkit/ajv-uuid/tree/main, https://www.npmjs.com/package/ajv-formats, or something like those instead?

src/database/initialization/file_to_json.sql
```diff
@@ -0,0 +1,21 @@
+SELECT drop_function('file_to_json');
+
+CREATE FUNCTION file_to_json(file_record files)
```
bickelj, line 3: I prefer the name here, `file_record` or `fileRecord` rather than `file` or `File`. If I understand correctly, this commit allows a caller to `post` a record that registers intent to do two more steps, ultimately culminating in a file existing in an S3 bucket and linked to this record.

src/openapi/paths/files.json
```diff
@@ -0,0 +1,51 @@
+{
+	"post": {
+		"summary": "Create a new file record",
+		"description": "Creates a new file record that can be used to generate presigned upload URLs",
```
bickelj, line 4: Yes, this description is good, it is not creating a new file but a new file record. I would rename the other stuff to `FileRecord` too, to make sure we don't mistakenly think we have a file when we don't.

src/database/migrations/0069-create-files.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE files (
+	uuid uuid PRIMARY KEY DEFAULT gen_random_uuid(),
```
bickelj, line 2: Rather than identifying files by a random number, files self-identify by checksum. We could use `bytea` of some length (256 bits?) and use the `b2sum` of the file to identify it. The only trouble with my suggestion is that then it naturally de-duplicates these file records if the checksum is the primary key. So the key would have to be across multiple fields.

src/types/PresignedPostRequest.ts
```diff
@@ -1,10 +1,11 @@
 import { ajv } from '../ajv';
+import { uuidSchema } from './Uuid';
 import type { JSONSchemaType } from 'ajv';
 import type { PresignedPost } from '@aws-sdk/s3-presigned-post';
+import type { Uuid } from './Uuid';
 
 export interface PresignedPostRequest {
-	fileType: string;
-	fileSize: number;
+	fileUuid: Uuid;
```
bickelj, line 8: The removal of two fields and addition of one required field to the Pre-signed post endpoint is a breaking change. We should increment the major version of the software, perhaps to 1.0.0 if we haven't already released it.

src/openapi/paths/files.json
bickelj: Separately, this description should lead the caller to the next step. This is step one of three, so the user should be alerted to the next endpoint(s) needed to complete the process.

bickelj (dismissed):
I have not tried it locally yet. I'd rather give this feedback sooner (before trying it).

The pre-signed post file upload is already a tedious, error-prone process. This commit adds yet another step to it. So instead of two steps whereby a caller registers an intent to post followed by an actual post, this adds a third step, and also breaks the existing pre-signed post request step.

Ideally there would be a `POST /file` API endpoint where it literally means that: the caller posts actual file data. And that would be the end of it for callers, except perhaps to get the resulting ID from the response body. So that would be one step instead of three. Failing that, I think we could keep the cumbersome pre-signed post process at two steps, where the pre-signed post endpoint takes on the responsibilities of the proposed file endpoints.

If we cannot collapse the endpoints into one or two instead of three, then at least the endpoint should be named differently to highlight that it is not accepting a file.

Otherwise, looks great, tidy, and thank you! :smile:

bickelj:
Oh, and `CHANGELOG.md` needs an update.

src/database/initialization/file_to_json.sql
slifty: Hmm -- my only pause is that I've been on so many projects where we've re-organized away from `Record` prefix since it's generic (e.g. "Isn't everything a FooRecord?")

I don't disagree, but what does `Record` convey?  Why do we have `Users` instead of `UserRecords` for instance?

Either way we should be consistent for sure; will wait to resolve this thread before making a change in either direction.

src/types/PresignedPostRequest.ts
slifty: We aren't releasing major versions at this point, but yes this will be a version bump.

I wouldn't want to couple this particular change with a decision to exit the 0.x semver stage.

src/constants/strings.ts
hminsky2002: seconding this, since we're already using it for phone numbers and emails

src/database/initialization/file_to_json.sql
hminsky2002: @slifty yeah I think on reflection, I support not adding a `record` suffix to the entity type for redundancy concerns. I agree with @bickelj that "File" signals to me raw binary, but I'm swayed by the argument that it's a `File` as represented by the PDC.

bickelj: FWIW I don't like `FileRecord` either, I'd prefer the API does what it says and allow posting of a file. But that's not what it does, it allows registration of an intent to allow a second registration of intent and finally allowing a post of the file directly to S3.

> tell me your exact flows (register → upload → verify → make downloadable), and I’ll map them to endpoints and states with response examples.

I think ChatGPT assumed (so to speak) "which ultimately will have data uploaded to s3" means the endpoint, behind the scenes, would do the work of uploading the file to S3, not necessarily that the caller had to call `/files` and then the pre-signed post endpoint and then finally upload directly to S3. It clearly acknowledges the possibility of more complex workflows but its very first suggestion is "Clients shouldn’t have to know that bytes live in S3." In the current and proposed model, clients are forced to know that bytes live in S3 because they have to deal with pre-signed posts and the S3 API in order to upload.

Do we intend to have a `GET /file/{id}` that returns the actual bytes of the file via redirect or something like that? If so, I think `file` makes sense there. But if it's a handle, record, pointer, registration, or whatever else (not a file), I don't think it needs to be named "file."

src/database/migrations/0069-create-files.sql
bickelj: > Reflecting on all of the above: I'm thinking we should call this `storage_key` instead of `uuid` which makes it more explicit that this field is ultimately the key associated with the file in the storage service (or `s3_key` but that feels too implementation-specific).
> 
> I think with that name it's more intuitive as to why we would want that to be a generated UUID rather than a derived checksum

The S3 API is exposed to the client in this flow. So 'tis not a problem to make it explicit.

Regardless, I'd be fine with `storage_key`.

src/database/initialization/file_to_json.sql
slifty: Just to be clear: Presigned post flows are super common (file data being funneled through a server is an antipattern, to the point that the concept of presigned posts was engineered and implemented!)

Presigned post is not counter to "shouldn't have to know it's S3" -- posting to an endpoint that accepts the actual data as a separate step (i.e. a presigned post) is not exposing the use of S3 to process that presigned post.

I'm feeling increasingly strongly that `File` is correct...

src/database/migrations/0069-create-files.sql
slifty: I meant less safe in terms of "Anybody can predict a file's location on the internet based on its contents"

I do agree that storing a checksum and using that checksum is a good idea for a future augmentation of our presigned post flow; I just don't think it should be the identifier.

src/database/initialization/file_to_json.sql
slifty: We met and have reached a good next step ditch the idea of PresignedPostRequest, and instead when a user runs `POST /files` the resulting entity has a presignedPost.  This has lots of benefits and means we're not adding yet-another-step to file uploads.

(there was lots more discussion overall but this I think was the most critical actionable conclusion)

slifty:
Moving this back to draft since I'm in the middle of performing the refactor described above!

src/database/migrations/0069-create-files.sql
slifty: Created https://github.com/PhilanthropyDataCommons/service/issues/1871 (resolving this thread since it's now over there)

jest.config.ci.js
```diff
@@ -1,6 +1,31 @@
...
+		],
+	},
+};
+
 module.exports = {
 	...config,
-	setupFilesAfterEnv: ['<rootDir>/src/test/integrationSuiteSetup.ts'],
 	collectCoverage: true,
+	collectCoverageFrom: ['src/**/*.ts', '!src/**/*.test.*'],
+	projects: [
```
hminsky2002, line 18: @slifty and i went through this implementation together and pulled this out of [this jest documentation ](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig), which is unfortunately lacking in content. However, we determined that splitting up the CI testing into unit and integration using  jest proejcts was the best approach for the time being.

hminsky2002 (approved):
Approving this after pairing with dan on the PR for a little bit. I think the choice to make the file entity is the right one and will clarify the flow of file uploading moving forward.

src/types/File.ts
```diff
@@ -0,0 +1,40 @@
+import { ajv } from '../ajv';
+import type { Uuid } from './Uuid';
+import type { KeycloakId } from './KeycloakId';
+import type { JSONSchemaType } from 'ajv';
+import type { PresignedPost } from '@aws-sdk/s3-presigned-post';
+import type { Writable } from './Writable';
+
+interface File {
+	readonly id: number;
```
bickelj, line 9: Why not use `storageKey` here? I suppose the integer `id` provides insert order? But we have a `createdAt` that can be used.

src/openapi/paths/bulkUploadTasks.json
```diff
@@ -41,7 +41,7 @@
 	},
 	"post": {
 		"operationId": "addBulkUploadTask",
-		"summary": "Registers a bulk upload task. This lets PDC ingest data previously uploaded via a POST to the /presignedPostRequests endpoint.",
+		"summary": "Registers a bulk upload task. This lets PDC ingest a csv previously uploaded via the the Presigned Post flow started at the POST /files endpoint.",
```
bickelj, line 44: Double `the` and maybe should be `CSV`.

bickelj (commented):
I am going to see if I can use it locally.

src/database/migrations/0069-create-files.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE files (
+	id integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+	name text NOT NULL,
+	storage_key uuid UNIQUE DEFAULT gen_random_uuid(),
+	mime_type text NOT NULL,
+	size integer NOT NULL CHECK (size >= 0),
+	created_by uuid NOT NULL REFERENCES users (keycloak_user_id)
+	ON DELETE CASCADE,
+	created_at timestamp with time zone NOT NULL DEFAULT now()
+);
```
bickelj, line 10: Is the URL to see the file an implicit composition of the various S3 env vars, etc.? Should we add the expected URL to the file at the time of row insert in case of S3 env var changes?

bickelj:
I cannot seem to get a pre-signed post upload to work locally using `curl`. It is reminiscent of the time I tried before and eventually got it after 60+ attempts and asking @slifty who asked an LLM. If @hminsky2002 it works for you and/or bulk upload software, I have no objection to merging.

It would be nice to have the URL available in the `file` row. Perhaps that is a future thing for `GET` requests. Other minor stuff above.

src/database/migrations/0069-create-files.sql
slifty: The URL to see the file is a thing that will be generated separately, since there is no such thing as a universally-accessible-non-expiring-public-URL (that would be a security risk.

src/types/File.ts
bickelj: Suppose the (eventual) URL is the key. What bad thing happens then?

src/database/migrations/0069-create-files.sql
bickelj: Oh right, well, I assume no secret keys are in a URL, that would be bad ; ) ; )

If there is a secret key in a URL I suppose it would be in query parameters or after a pound sign or whatever, so I would not include those, my assumption is there is still a root URL onto which those keys get appended for temporary access.

src/types/File.ts
slifty: That's poor design -- the key should not be based on our back end storage implementations.

Primary keys should be identifiers, not attributes.

URI is an attribute of the file, not an identifier of the file.

src/database/migrations/0069-create-files.sql
slifty: There are secrets in the URL for sure (that's what makes it a presigned URL) but that's also the security model.  The idea is that the temporary presigned URL provides access for the user it was generated for and is, itself, the secret.

You can read more about it here -- https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html

It doesn't make much sense to me to store a URL that can't be used as opposed to structured attributes that signal where the object is stored + allows us to generate presigned urls using the S3 protocol.

src/openapi/paths/bulkUploadTasks.json
slifty: Actually I made this edit in an earlier iteration that touched on this endpoint (the current PR doesn't impact this at all).  Removing the change entirely.

src/database/migrations/0069-create-files.sql
bickelj: @slifty My double wink was to no avail. I recall having this discussion in the past regarding secrets in the URL. The wink was supposed to be a reference to that.

The link you shared has pre-signed URLs starting with a URL or perhaps URI, the Location or Identifier of a Resource (a document or file), and then includes as additional parameters the credentials to access that Resource.

If I'm the object owner, can I not make a specially formed request to GET that Resource without the parameters? In other words, all the stuff up to the parameters Locates or Identifies a Resource (file) that we are interested in.

PR #1869, opened 2025-08-26
Start versioning beyond version zero

bickelj:
Until recently, the PDC back-end software has largely been in an experimental or prototype state. At least two parties (besides the team developing the PDC back-end) call it. We also anticipate real data to be sent to the PDC soon. Even if the software remains experimental for a while, the API has matured over the past three-plus years since its first commit. It is time to call it version one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


bickelj:
@slifty It is a succinct and actionable discussion-starter, it's a PR : )

bickelj:
The purpose of this PR was to hit v1 at that time. Whenever we're ready we can re-open, we don't need to keep this hanging around.

PR #1879, opened 2025-08-29
Add docker compose for ci

bickelj:
Adds a `compose-ci.yml` file for automated testing of apps against the real/production PDC image alongside a development instance of Keycloak and S3 service.

### DIFF
--- a/compose-ci.yml
+++ b/compose-ci.yml
@@ -1,0 +1,281 @@
+# Exposes an instance of the PDC API with dependencies such as Keycloak
+#
+# Intended uses:
+# - Automated tests for PDC apps that call the PDC backend
+# - Local development of PDC apps that call the PDC backend
+#
+# What it does automatically:
+# 1. Creates docker volumes for databases and S3
+# 2. Creates databases and an S3 bucket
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 1.0.0 2025-09-05
```
slifty, line 10: Let's take this out -- this should be discussed in the form of an issue first.  Opened one here, but in the meantime let's rebase against main: https://github.com/PhilanthropyDataCommons/service/issues/1895

bickelj:
@slifty I intend to squash this once it is working on multiple persons' machines. Until then, treat it as one commit for review. The commit is merely a log of what actually happened.

bickelj:
@slifty Just demonstrated that this requires docker compose 2.30.0 and later due to use of `post_start` hooks. E.g. https://docs.docker.com/compose/how-tos/lifecycle/#services-lifecycle-hooks

bickelj:
Squashed and rebased.

PR #1884, opened 2025-09-02
Update API specification to 0.23.0

slifty:
This PR bumps the API version (which was already bumped in the CHANGELOG, but by mistake wasn't bumped in the actual spec)


hminsky2002 (approved):
Thank you for the speedy resolution on this @slifty

PR #1885, opened 2025-09-02
Add `file` type to `BaseFields`

slifty:
This PR adds support for a `file` type for our base fields.

This implementation declares a file value valid if it is a valid id string (e.g. a whole number greater than 0).

I had originally planned `isValid` to only be true if it *also* referenced a file that the proposal author had created themselves, but that type of check means we're caching permission state in the `isValid` attribute.  In reality, permission checks should be performed at the time of access and according to whatever the current business logic has been defined to entail.

This PR does not yet populate proposal field values with file metadata or accessible file URLs; this will be a larger lift!

Related to #1689

### DIFF
--- a/src/fieldValidation.ts
+++ b/src/fieldValidation.ts
@@ -1,5 +1,5 @@
 import validator from 'validator';
-import { BaseFieldDataType } from './types';
+import { BaseFieldDataType, isId } from './types';
 import { ajv } from './ajv';
 
 const isEmailString = ajv.compile({
@@ -55,6 +55,13 @@ const isCurrencyWithCodeString = (value: string): boolean => {
 	);
 };
 
+const isIdString = (value: string): boolean => {
+	if (value.trim() !== value) {
+		return false;
+	}
+	return isId(value);
+};
+
 export const fieldValueIsValid = (
 	fieldValue: string,
 	dataType: BaseFieldDataType,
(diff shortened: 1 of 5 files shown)


src/fieldValidation.ts
```diff
@@ -24,6 +24,11 @@ const isString = ajv.compile({
 	type: 'string',
 });
 
+const isIdString = ajv.compile({
```
slifty, line 27: I'm not going to change this in this PR, but I'm realizing that the `isFooString` vs `isFoo` is probably redundant.  In a future refactor we probably would just want to use `isId` from `types/Id.ts`

slifty: I changed my mind...  Using `isId` here instead.

src/fieldValidation.ts
```diff
@@ -1,6 +1,7 @@
 import validator from 'validator';
 import { BaseFieldDataType } from './types';
 import { ajv } from './ajv';
+import { isId } from './types';
```
bickelj, line 4: Address below lint thing.

slifty:
OK! Assuming CI actually passes...

So here's the tricky thing about this whole concept -- ultimately we want a field value of type `file` to return a file object (which has been decorated with the presigned downloadUrl, which is a whole concept).

This PR is a baby step towards that.  The reason I felt it was worth having as its own PR is that it enables form creation of file type and allows for data insertion (even if, to your point, data access is not yet actually where it needs to be).

---

What I imagine the future PR will do is:

1. expand the ProposalFieldValue type to have a `file?:` attribute which would be a `File` entity in the event that the field value actually maps to a valid PDC File.

2. Expand the `File` type to include a presigned access URL.

Those two changes would accomplish what we want, which is that when loading a proposal you get links to any files that it's field values reference without needing to make additional calls.

PR #1896, opened 2025-09-04
Update `BulkUploadTask` to reference `File` entities.

### LINKED ISSUE #1897: Migrate `BulkUploadTask` to reference `File` entities
The BulkUploadTask currently takes in information about S3 files but we now have a first class `File` type within the PDC.

We should update the bulk upload task to accept a file ID for the proposal data CSV.

We're soon going to also add additional files for a bulk upload (specifically, a proposalFileArchive or some other name in order to support #1858), but that work should be done as part of supporting the file field values rather than being part of this issue's work.

slifty:
This PR migrates our `BulkUploadTask` entity and corresponding processing script to use the new concept of a `File`.

It also:

1. Updates the task processor to remove the idea of bulk uploads files being moved around on S3 to represent their processing state.
2. Modifies our S3 utilities to potentially support multiple regions

Since this does not implement support for file type field values I decided to defer the addition of an additional file attribute (e.g. `attachmentsArchiveFile`) but that will be coming right around the corner!

Resolves #1897

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.24.0 2025-09-11
+
 ### Added
 
 - `BaseField` can now be of type `file`.
--- a/src/openapi/api.json
+++ b/src/openapi/api.json
@@ -3,7 +3,7 @@
 	"info": {
 		"title": "Philanthropy Data Commons API",
 		"description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for changemakers seeking grants.",
-		"version": "0.23.0",
+		"version": "0.24.0",
 		"license": {
 			"name": "GNU Affero General Public License v3.0 only",
 			"url": "https://spdx.org/licenses/AGPL-3.0-only.html"


src/s3.ts
```diff
@@ -21,22 +21,27 @@ const {
...
-export const s3Client = new S3({
-	forcePathStyle: S3_PATH_STYLE === 'true',
-	endpoint: S3_ENDPOINT,
-	region: S3_REGION,
-	credentials: {
-		accessKeyId: S3_ACCESS_KEY_ID,
-		secretAccessKey: S3_ACCESS_SECRET,
-	},
-});
+const s3Clients: Record<string, S3> = {};
```
bickelj, line 24: I see we have a client per region here, and it is populated below, more or less on demand.

It feels weird or wrong to me, somehow. Yet I see https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/s3-cross-region.html indicating that operating across regions is a problem with S3.

My understanding at the moment is that a bucket does exist in a particular "region" (datacenter) and therefore must be specified on bucket creation but it does not need to be specified when reading data.

The commit message indicates we may need a different region to read, but shouldn't we be able to read from any region?

In short, I wonder what problem this is solving. It seems like it does solve a problem but if so it's a strange problem of Amazon's making and a step backward from the old days where a URL was a URL and it could be anywhere in the world. I don't find myself adding a region parameter to my location bar for websites, for example.

Is `region` even a required parameter? It appears not, on a glance of the source:
```typescript
    /**
     * The AWS region to which this client will send requests
     */
    region?: string | __Provider<string>;
```

In other words, yes, a bucket is in a region, but we create our bucket outside the software, and the client here should work regardless of which region is specified. And a client when getting the actual file data will be routed appropriately by S3 outside the software here. So I'm scratching my head on the need for this additional complexity.

src/tasks/processBulkUploadTask.ts
```diff
@@ -4,7 +4,7 @@ import { finished } from 'node:stream/promises';
 import { parse } from 'csv-parse';
 import { requireEnv } from 'require-env-variable';
 import tmp from 'tmp-promise';
-import { getS3Client, S3_BULK_UPLOADS_KEY_PREFIX } from '../s3';
+import { getS3Client } from '../s3';
```
bickelj, line 6: :+1:

src/types/BulkUploadTask.ts
```diff
@@ -1,51 +1,44 @@
...
+import type { Id } from './Id';
 
 interface BulkUploadTask {
 	readonly id: number;
-	sourceId: number;
+	sourceId: Id;
 	readonly source: Source;
-	fileName: string;
-	sourceKey: string;
+	proposalDataFileId: Id;
```
bickelj, line 18: I think `fileId` would be sufficient here. The qualifier `proposalData` does not differentiate it from a file contained with a proposal. It could be qualified by `bulkUploadFileId` but then again we're inside a `BulkUploadTask` so that's redundant. :shrug:

bickelj (commented):
Overall this looks like an improvement, thank you!

I doubt we need a map of S3 clients, see above.

I haven't tried it locally because I assume it will take several hours to figure out how to get a post to succeed. Perhaps I am on the cusp though and should keep trying.

src/types/BulkUploadTask.ts
slifty: This name is necessary; we're about to add an additional file ID (`attachmentsArchiveFileID`)

src/database/queries/bulkUploadTasks/updateById.sql
```diff
@@ -1,7 +1,5 @@
 UPDATE bulk_upload_tasks
 SET
-	file_size = coalesce(:fileSize, file_size),
```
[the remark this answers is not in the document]
bickelj: Agree that we should treat the file data as immutable.

src/database/migrations/0070-alter-bulk_upload_tasks-use-files.sql
```diff
@@ -0,0 +1,48 @@
+-- Add nullable file reference columns first
+ALTER TABLE bulk_upload_tasks
+ADD COLUMN proposal_data_file_id integer REFERENCES files (id)
+ON DELETE CASCADE;
```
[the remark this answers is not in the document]
bickelj: I don't think we should delete a document or its reference or its reference to its reference. If we want defense in depth, then we wouldn't cascade here. But if we don't actually ever delete then it is probably OK. And audit logs should capture the deletes if they accidentally happen.

src/s3.ts
slifty: @bickelj interesting!  I had no idea / thought a region was a namespace.

This means we don't actually need to store bucket_region at all.  I'll rip out this commit and replace it with one that removes that attribute from `File`

src/types/BulkUploadTask.ts
slifty: Added plural (this contains data for multiple proposals).  Data does stink, and maybe should be like `proposalsTable` (downside is sounds like SQL related) or `proposalsCsv`(downside is we may support tsv / xsl).

For now keeping it as data but renaming is incredibly easy so please don't hold back on continuing to offer iterative feedback.  This is worth getting right.

src/s3.ts
slifty: All right.....

So what we learned is that Digital Ocean actually does couple bucket access to not just region but S3 endpoint.  It seems this means we do need to have multiple clients, but even worse we will also need to store the endpoint associated with a given bucket.

So I believe the following is true

* We don't need region-specific S3 clients, but I believe we will need endpoint-specific ones (unfortunately).
* We should normalize bucket metadata (e.g. File should have a s3_bucket_Id) which references an s3_buckets table that has a name, endpoint, and region) <-- this wouldn't have corresponding endpoints, though the bucket ID + information would be attached to a File entity (really only for internal use, but I don't see the harm in exposing it)

With those changes I believe we'll be in good standing from a data integrity / resolution PoV.

I'm going to make an issue to capture this an implement it as a separate PR / pull the S3 stuff out of this one.

slifty:
Now that the bucket stuff for S3 is being normalized inside an internal s3_bucket object, this PR needs to be rebased on that and tweaked to use that logic instead.

https://github.com/PhilanthropyDataCommons/service/pull/1902

Issue #1904, opened 2025-09-08
Cache user metadata in `users` table, expose it in `/users` endpoint


bickelj:
Following up on #1870, this ticket is to actually do the needful.

slifty:
This is resolved!

PR #1907, opened 2025-09-08
Cache user name on authenticated requests

bickelj:
This change allows the `/users` endpoint to return the name from a user's most recent JWT. Whenever a user authenticates, the table is updated with the `name` value from the JWT. The existing audit tables track updates to names.

Issue #1904 Cache user metadata


bickelj:
@slifty I rebased on main, renamed that migration, and tested it out locally. I guess this is a situation where re-review is not needed due to not changing anything else.

bickelj:
Doh. OpenAPI.

PR #1935, opened 2025-09-16
Single logout via a small Swagger-UI plugin

### LINKED ISSUE #1915: Single logout from OpenAPI docs
When logging in to PDC from the OpenAPI docs, a full PDC session is created, but when logging out from OpenAPI docs, the logout is partial. The session with Keycloak is still active such that logging in again grants a fresh JWT from the session and doesn't require a full login process. In general, whatever opened a channel should be responsible for closing a channel, and this applies to the PDC login/logout session as well. This ticket is to figure out how to make log out fully log out of PDC from the OpenAPI docs.

Issue #1915 Single logout from OpenAPI docs

bickelj:
To allow Keycloak logout from the OpenAPI documentation, this uses an OpenAPI plugin as a separate module. The reason for clean separation is inclusion of the "dom" module, which has no place in the node project except to be added to the `swagger-ui` browser code. Without this change, the logout only applies to the OpenAPI docs page and the PDC Keycloak session would remain. With this change, the logout really logs out of PDC Keycloak entirely.

Resolves #1915


slifty:
@bickelj did we end up moving this to it's own plugin outside of github?

bickelj:
While this would be great, it is definitely lower priority than other issues, and we still need to overcome `jest`'s limitations (or its upstream library limitations) to be do the plugin correctly. No need for this PR to hang around.

PR #1940, opened 2025-09-17
Lint Docker Compose script

bickelj:
This commit adds `dclint` for the single Docker Compose script. It also fixes (or adds exceptions) for the issues found by the linter.

Issue https://github.com/PhilanthropyDataCommons/service/issues/1960 Lint Docker Compose script


slifty (changes requested):
We shouldn't introduce commits that break CI, so this should either (1) introduce dclint in a way that does not identify any issues or (2) introduce dclint alongside the necessary fixes.

package.json
```diff
@@ -12,7 +12,8 @@
 		"format:eslint": "eslint ./src --fix || true",
 		"format:prettier": "prettier . --write",
 		"format:sqlfluff": "./venv/bin/sqlfluff fix src/database/",
-		"lint": "npm run lint:eslint && npm run lint:prettier && npm run lint:tsc && npm run lint:openapi && npm run lint:sqlfluff",
+		"lint": "npm run lint:eslint && npm run lint:prettier && npm run lint:tsc && npm run lint:openapi && npm run lint:sqlfluff && npm run lint:docker-compose",
+		"lint:docker-compose": "dclint ./compose-ci.yml",
```
slifty, line 16: I see there is a `.dclintrc`

If possible, it would be great to configure which files dclint is configured to lint in that location rather than as a CLI parameter.

slifty (approved):
Hooray for linting!

bickelj:
I had to resolve a merge conflict in `package.json` and `package-lock.json` when rebasing on main but it appears all of that is OK now.

Issue #1947, opened 2025-09-18
Update docker image


bickelj:
Bitnami images appear out of support, now the build fails. Find a new one.

E.g.
```
/usr/bin/docker buildx build --cache-from type=registry,ref=ghcr.io/philanthropydatacommons/service:buildcache --cache-to type=registry,ref=ghcr.io/philanthropydatacommons/service:buildcache,mode=max --iidfile /home/runner/work/_temp/docker-actions-toolkit-HjCA1e/build-iidfile-2ab1145f09.txt --platform linux/amd64,linux/arm64 --attest type=provenance,mode=max,builder-id=https://github.com/PhilanthropyDataCommons/service/actions/runs/17813577272/attempts/1 --tag ghcr.io/philanthropydatacommons/service:20250917-430ad4f --tag ghcr.io/philanthropydatacommons/service:latest --metadata-file /home/runner/work/_temp/docker-actions-toolkit-HjCA1e/build-metadata-423762fa08.json --push .
#0 building with "builder-20e36ee3-dbd1-4114-b2d2-535ffa390c39" instance using docker-container driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 444B done
#1 DONE 0.0s

#2 [auth] bitnami/node:pull token for registry-1.docker.io
#2 DONE 0.0s

#3 [linux/arm64 internal] load metadata for docker.io/bitnami/node:22-debian-12
#3 ERROR: docker.io/bitnami/node:22-debian-12: not found

#4 [linux/amd64 internal] load metadata for docker.io/bitnami/node:22-debian-12
#4 CANCELED
------
 > [linux/arm64 internal] load metadata for docker.io/bitnami/node:22-debian-12:
------
Dockerfile:1
--------------------
   1 | >>> FROM bitnami/node:22-debian-12
   2 |     
   3 |     RUN adduser --home /opt/philanthropy-data-commons --uid 902 \
--------------------
ERROR: failed to build: failed to solve: bitnami/node:22-debian-12: failed to resolve source metadata for docker.io/bitnami/node:22-debian-12: docker.io/bitnami/node:22-debian-12: not found
```

Reproduced locally:
```
service$ sudo docker build .
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  501.8MB
[...]

PR #1948, opened 2025-09-18
Use the official node docker image

bickelj:
Bitnami is not going to support all the images they used to anymore. See https://github.com/bitnami/charts/issues/35164

The plain (Debian) bookworm image was 1.5GiB (very big) and the slim version of the same image didn't come with `curl` which we use in the `compose-ci.yml` file to run some scripts. So install `curl` in the slim image. Also, due to warnings about UIDs under 1000, update the user ID to 1002 instead of 902.

Issue #1947 Update docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
-FROM bitnami/node:22-debian-12
+FROM docker.io/node:22-bookworm-slim
 
-RUN adduser --home /opt/philanthropy-data-commons --uid 902 \
+RUN apt-get update && apt-get install -y curl
+RUN adduser --home /opt/philanthropy-data-commons --uid 1002 \
     --disabled-login web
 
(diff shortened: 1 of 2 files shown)


bickelj:
I was able to confirm this image works locally using the `compose-ci.yml` script:

1. `sudo docker build .`
2. editing `compose-ci.yml` to use the literal image hash from step (1) in place of `ghcr.io/philanthropydatacommons/service:latest`
3. Running `compose-ci.yml` as usual: `sudo docker compose -f compose-ci.yml down --remove-orphans --volumes` then  `sudo docker compose -f compose-ci.yml up --remove-orphans`

slifty (approved):
Good changes overall!

Also it does make sense for us to use official node images instead of the new bitnami secure ones

PR #1951, opened 2025-09-18
Add a GH action to test Docker image on PRs

### LINKED ISSUE #1949: Include a `docker buildx` command in PR CI
At present, a docker image is only built and pushed on a merge to main. The positive side of the present approach is we only want to publish docker images that include tested and good code. The negative side is when `docker build` or `docker buildx` fails for whatever reason, we only find out on a merge to main (as recently happened). Another negative is that the image is not exactly tested (perhaps that is a separate issue). This ticket is to include a `docker buildx` command to test building the docker image in every PR, not to push that image anywhere, but simply to test that the image will be likely to build successfully.

bickelj:
Without this commit, we don't find out if the `docker buildx` command succeeds or fails until a merge to main. While it is still possible for the command to fail on merge to main with this change, we are much more likely to get earlier feedback (at PR time) as to whether the Docker image will be successfully built.

Fixes #1949 Include a docker buildx command in PR CI See also Issue #1947 for more context.


bickelj:
Tada. It reproduces the issue seen in #1947 but in a PR instead of on merge to main. I would probably merge this first (it should only fail on PRs, not on main) and then see what happens on a rebase of #1948. @slifty what do you think?

.github/workflows/test-docker-build.yml
```diff
@@ -0,0 +1,40 @@
+name: Test building a docker image
```
slifty, line 1: Let's just call this `Test docker build` (aligns with the file name and I think it's cleaner; this will be rendered in UX).

Also we might consider having this just be part of the `test.yml` file which would mean on CI it could show up as `Test (docker build)`

Then the name of the could just be `docker-build` since the job is already a subset of test ("discard" feels redundant / implied in that context, which is part of why I think that context might be worth considering!)

bickelj:
Rebased on main, we'd now expect the build to succeed because it's using a new successful docker image that we yoloed on main first.

PR #1959, opened 2025-09-19
Update node to 24

bickelj:
Node 22 is projected to cease being Active LTS on 2025-10-21 while node 24 is projected to become Active LTS on 2025-10-28. Updating this commit and merging it around that (latter) time would be appropriate.


slifty:
@bickelj I support this PR being open!  (I also support adding intentional wiggle room to the window of the policy if we wanna merge it earlier)

Won't be able to review it today, but will next week

Issue #1722, opened 2025-06-19
Consider an actual date, time, and duration library


bickelj:
While we only use dates, times, or durations in a few places, it is important to make them clear and correct. Consider luxon or temporal-polyfill.

PR #1977, opened 2025-09-26
Reduce unnecessary polling in graphile workers

bickelj:
A test of once per five minutes did not prevent a bulk upload from
being processed nearly immediately.

Issue #1266 Graphile-worker runs UPDATE statements very frequently

### DIFF
--- a/src/jobQueue.ts
+++ b/src/jobQueue.ts
@@ -1,4 +1,5 @@
 import { Logger, quickAddJob, run, runMigrations } from 'graphile-worker';
+import { TEN_MINUTES_IN_MS } from './constants/time';
 import { copyBaseFields, processBulkUploadTask } from './tasks';
 import { getLogger } from './logger';
 import { db } from './database/db';
@@ -11,7 +12,6 @@ import type {
 const logger = getLogger(__filename);
 
 const CONCURRENT_JOB_COUNT = 5;
-const POLL_INTERVAL_MS = 1000;
 
 enum JobType {
 	PROCESS_BULK_UPLOAD = 'processBulkUploadTask',
@@ -43,7 +43,13 @@ export const startJobQueue = async (): Promise<void> => {
 		pgPool: db.pool,
 		concurrency: CONCURRENT_JOB_COUNT,
 		noHandleSignals: false,
-		pollInterval: POLL_INTERVAL_MS,
+		/*
+		 * The poll interval is for scheduled (not on-demand) jobs combined with idle workers. See
+		 * https://worker.graphile.org/docs/faq#if-we-have-jobs-that-are-scheduled-in-the-futurefailed-will-workers-continuously-poll-to-run-those-jobs-or-will-the-listennotify-mechanism-be-used-for-that
+		 * Every poll interval times the number of workers, a SQL update statement runs.
+		 * If the need arises for frequent scheduled jobs, update this value appropriately.
+		 */
+		pollInterval: TEN_MINUTES_IN_MS,
 		taskList: {
 			processBulkUploadTask,
 			copyBaseFields,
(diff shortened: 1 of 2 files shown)


bickelj:
I tested this by setting `log_statement='mod'` on my pg instance, starting up PDC, looking for the graphile worker UPDATE statements, waiting a minute, then doing a bulk upload, noticing that it completed right away without any of those UPDATE statements, then saw the later UPDATE statements come through. A bulk upload still happens immediately. Please try it yourself, too.

src/jobQueue.ts
```diff
@@ -11,7 +11,11 @@ import type {
 const logger = getLogger(__filename);
 
 const CONCURRENT_JOB_COUNT = 5;
-const POLL_INTERVAL_MS = 1000;
+/** The poll interval is for scheduled (not on-demand) jobs combined with idle workers. See
+ * https://worker.graphile.org/docs/faq#if-we-have-jobs-that-are-scheduled-in-the-futurefailed-will-workers-continuously-poll-to-run-those-jobs-or-will-the-listennotify-mechanism-be-used-for-that
+ * Every poll interval times the number of workers, a SQL update statement runs.
+ */
+const POLL_INTERVAL_MS = 600_000;
```
slifty, line 18: It took me a moment to translate this number into what it actually means -- what do you think of adding  `ONCE_PER_MINUTE_MS = 600_000` to our `time.ts` constants list?

Truly if that seems silly no problem, though if not that it might not hurt to add `// once per minute` to this line just for human legibility.

slifty (approved):
Looks good to me!  One tiny thought around legibility but truly only if you agree with it.

Might be worth noting in the commit message that you determined once per minute was plenty; also, I imagine this means we couldn't have scheduled crons that run more frequently than once-per-minute forevermore, right?

bickelj:
> Looks good to me! One tiny thought around legibility but truly only if you agree with it.
> 
> Might be worth noting in the commit message that you determined once per minute was plenty; also, I imagine this means we couldn't have scheduled crons that run more frequently than once-per-minute forevermore, right?

I wouldn't say we couldn't have scheduled crons that run more frequently than once per ten minutes forevermore, I'd say that if the use case arises where we need to schedule a job more frequently than once per ten minutes, we could decrease the interval when implementing that feature.

slifty:
@bickelj noted!  Realizing we should probably document that current state somewhere it might be worth doing.

What I mean / meant with that comment is that if someone took this code base and set it up, and used the graphile CLI to schedule a something the granularity of triggering that scheduled event would be 10m which might be unexpected behavior without documentation.

src/jobQueue.ts
slifty: (Marking this as resolved to hide my shame)

bickelj:
> @bickelj noted! Realizing we should probably document that current state somewhere it might be worth doing.
> 
> What I mean / meant with that comment is that if someone took this code base and set it up, and used the graphile CLI to schedule a something the granularity of triggering that scheduled event would be 10m which might be unexpected behavior without documentation.

Is a comment below the current comment sufficient? I mean, if you notice a problem with a graphile worker, you go to the graphile worker config, right? That was my first step in #1266 a long time ago. At that time I didn't know the impact of decreasing the frequency, so I didn't try it then. See updated PR.

slifty:
@bickelj I think so, given the fact that we don't currently have any documentation related to graphile APIs as it relates to this project.

If we ever make documentation related to the various async tasks the PDC supports it would make sense to include that in there!

src/jobQueue.ts
bickelj: Meh, it's too many numerals. #1722 would help.

Issue #731, opened 2024-01-23
Report preflight errors via Bulk Upload API


reefdog:
When querying the status of a Bulk Upload item, if there are any file-level preflight errors, those should be reported to the API consumer.

Here is an example of how this would appear in the UI.

![Bulk Upload - Failed Preflight](https://github.com/PhilanthropyDataCommons/service/assets/4731/ff055d4d-32ff-4816-bbdb-0d6022c824ce)

Note that while this UI demonstrates multiple pre-flight errors, if this wasn't the case, we would still display it the same. (Just with a single bullet point.)

PR #1981, opened 2025-09-29
Bulk upload task errors are now available

bickelj:
When errors are encountered during bulk upload, they are saved, and when the bulk upload task is retrieved via API call, the same logs are presented in a `logs` array.

Issue #731 Report Bulk Upload preflight errors via API


bickelj:
It looks like I still have some logger statements to remove here, perhaps also unneeded changes to the bulk upload functions.

bickelj:
Part of #1986 is to make sure that a past failed bulk upload is improved by our changes. To that end, I tested the above changes locally with that failed bulk upload, and on a `GET /tasks/bulkUploads` here is the entry:

```json
{
  "entries": [
    {
      "id": 94,
      "logs": [
        {
          "details": {
            "name": "Error",
            "message": "proposal_title is not a valid BaseField short code."
          },
          "isError": true,
          "createdAt": "2025-10-01T14:38:01.800221+00:00",
          "bulkUploadTaskId": 94
        }
      ],
      "funder": {
        "name": "System Placeholder Funder",
        "createdAt": "2025-02-18T19:55:50.462212+00:00",
        "shortCode": "pdc",
        "isCollaborative": false,
        "keycloakOrganizationId": null
      },
      "source": {
        "id": 2, /* ... */
```

There is a single log entry with a `details` key, and under that `details` is a `message` that a person can use to fix the bulk upload CSV.

src/tasks/processBulkUploadTask.ts
```diff
@@ -243,6 +245,60 @@ const loadTaskRunnerAuthContext = async (): Promise<AuthContext> => ({
...
+/**
+ * A value for initializing and incrementing the recursion guard
+ * @see getDetailsFromError
+ */
+const ONE = 1;
+/**
+ * A limit for the recursion guard below
+ * @see getDetailsFromError
+ */
+const ONE_THOUSAND = 1000;
```
slifty, line 257: I believe the intention of no magic numbers is for the named variable to reference the point of the value, so this should be `MAX_RECURSION = 1000`

src/tasks/processBulkUploadTask.ts
```diff
@@ -243,6 +245,60 @@ const loadTaskRunnerAuthContext = async (): Promise<AuthContext> => ({
...
+		};
+	if (err instanceof Error) {
+		if (err.cause !== undefined && err.cause !== null) {
+			return {
+				/* eslint-disable-next-line @typescript-eslint/no-misused-spread --
+				Errors are special, this is probably the only way to get them to serialize. */
+				...err,
+				message: err.message,
+				name: err.name,
+				cause: getDetailsFromError(err.cause, recursionGuard + ONE),
```
slifty, line 282: Instead of a recursion counter, let's have this accept an `errorStack` Array<LoggableError>, and ensure the serializee has not appeared on the serialization stack so far (thus preventing recursion loops).  This is more precise and I think a bit more intuitive.

I'd recommend naming this function in terms of log serialization (e.g. `serializeErrorForBulkUploadLogs`).  Alternatively we might want to consider creating a `LoggableError` class which has a `serializeForLogs` function and a `cause` which is typed as being a `LoggableError` (though I suppose if the error came from a library we still want to try to serialize it... so in that sense we'd just keep it as this more-general function, ok, sorry!)

Regardless, if you're up for the `serialize` verb that would make it clear, and I think the `serializedStack` would be an improved recursion method (plus then we can ditch the whole `const ONE` thing)

tsconfig.dev.json
```diff
@@ -2,6 +2,8 @@
 	"extends": "@tsconfig/node22/tsconfig.json",
 	"compilerOptions": {
 		"noEmit": true,
+		"target": "es2022",
```
slifty, line 5: Can you say more about this change / the need for it in the commit message?

I ask because the target is normally provided by our base tsconfig ([and appears to actually already be `es2022`](https://github.com/tsconfig/bases/blob/main/bases/node22.json)) -- which means that as we update that base this is going to eventually become a stale override, making it a non-ideal change.

Meanwhile the lib override removes a bunch of things that would otherwise be included.

src/tasks/processBulkUploadTask.ts
slifty: I hear you re the equity check, but in this case my understanding is we actually are trying to figure out if a literal specific object is the thing that has been seen before (which is where infinite recursion would happen, I think?)

I believe this is is the rare case where objA === objB is actually the check we want, so basically `if(processedObjects.includes(err) { return; }` I *believe* would actually work nicely.

bickelj: @slifty Right, the doubt was whether I could do a reference equality instead of value equality. Qwen3 is right that a `Set` would work because it is shallow reference equality, not value.

bickelj: Regarding naming, I hesitate to name it `serialize` because it neither serializes nor do I want it to serialize. I want a useful, serializable object instead of the :dog::poop: JS `Error`. So perhaps `getSerializableObjectFromError`? I'm OK if this is used outside the context of BulkUpload as well. At that point we'd bring it out of the bulk upload class and into a more general area.

slifty: > getSerializableObjectFromError
I like that a lot!

> he doubt was whether I could do a reference equality instead of value equality

Sorry that's what I was trying to say -- https://stackoverflow.com/questions/18037350/how-to-test-reference-equality-in-javascript

My understanding is that === for objects is reference equality in javascript, not value equality (and *usually* this actually trips people up because what they want is value, but in our case we DO want reference)

bickelj: Right, right. Set seems to work great. I did a quick test with my object of choice, very special "object" it turns out, `Error`, and it did the expected.

bickelj: I had trouble yesterday trying to actually create the recursion within an error. Doesn't seem straightforward but maybe it's worth trying again.

bickelj:
@slifty On my development workstation I am trying `88a95d3a21c7af133960162930485f618c4de2c1` with the mock AWS to see if it does better here.

src/tasks/getBulkUploadLogDetailsFromError.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @throws TypeError when JSON.stringify throws a (non-cause) circularity error.
+ * @returns A serializable object with at least a `message` property.
+ */
+export const getBulkUploadLogDetailsFromError = (
+	err: unknown,
+	previouslySeenErrors: Set<unknown> = new Set(),
+): BulkUploadLogDetails => {
+	if (previouslySeenErrors.has(err)) return { message: CYCLE_ERROR_MESSAGE };
+	previouslySeenErrors.add(err);
+	let objectFromError: {
```
slifty, line 19: Rather than defining a mutable object here what do you think of simply having return statements that are explicitly crafted within each type-guard-block?

This would mean, for instance, something like:

```
	if (err instanceof Error) {
	  const cause = (err.cause !== undefined && err.cause !== null) ? getBulkUploadLogDetailsFromError(
				err.cause,
				previouslySeenErrors,
			) : undefined;
	  return {
			message: err.message,
			name: err.name,
			cause,
		};
	}
	```

File: src/tasks/getBulkUploadLogDetailsFromError.ts [hunk 5]
```diff
@@ -0,0 +1,59 @@
...
+		message: string;
+		[key: string]: unknown;
+	} = {
+		message: 'An unknown error occurred',
+	};
+	if (err instanceof Error) {
+		objectFromError = {
+			/* eslint-disable-next-line @typescript-eslint/no-misused-spread --
+			Errors are special, this is probably the only way to get them to serialize. */
+			...err,
```
slifty, line 29: My understanding of what you're doing here is that ultimately for things that Extend error we want to capture their custom attributes (which this spread does).

That said, this lint rule exists to prevent `...err`, and so disabling it while saying error is special doesn't convey the depth of the situation.  In reality the rule is there BECAUSE error is special (and expanding it this way results in unexpected outcomes and won't capture things like message / name).  You do account for this behavior in the next two lines but it's still hard to fully understand what's going on as written.

Would you be open to first converting the error into something that *is* expandable, by something like this:

```
export interface ExpandableError {
  name: string;
  message: string;
  stack?: string;
  [key: string]: unknown;
}

const objectFromError = (err: Error): ExpandableError =>
Object.fromEntries(
  Object.getOwnPropertyNames(err).map(key => [key, err[key]])
);
```

I believe this would copy message + name + any other custom properties exist on the error object.

Then we could use that expanded object and decorate with `cause` (if it exists).

src/tasks/getBulkUploadLogDetailsFromError.ts
```diff
@@ -0,0 +1,59 @@
...
+			);
+		}
+	} else {
+		objectFromError = {
+			name: 'ProgrammingError',
+			message: 'A programming error occurred: errors should be of type Error',
+		};
+	}
+	// One last step to verify that this is actually serializable, try serializing it!
+	JSON.stringify(objectFromError);
```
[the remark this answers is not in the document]
slifty: I'd rather take out the "will throw an error" and instead gracefully handle that case by returning a serialized object that signals an unexpected type (similar to what we already have).

Basically: if passed an error, return the log details for that error; if passed a non-error, return the static object saying "this should have been an error" -- then there's no need for that last check, right?

src/tasks/getBulkUploadLogDetailsFromError.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param previouslySeenErrors A Set of previously seen errors to break cycles.
+ * @throws TypeError when JSON.stringify throws a (non-cause) circularity error.
+ * @returns A serializable object with at least a `message` property.
+ */
+export const getBulkUploadLogDetailsFromError = (
+	err: unknown,
+	previouslySeenErrors: Set<unknown> = new Set(),
+): BulkUploadLogDetails => {
+	if (previouslySeenErrors.has(err)) return { message: CYCLE_ERROR_MESSAGE };
+	previouslySeenErrors.add(err);
```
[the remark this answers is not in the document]
kfogel: *Somebody*'s itching to re-implement this in Rust, I see.

src/tasks/getBulkUploadLogDetailsFromError.ts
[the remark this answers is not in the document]
bickelj: So far, it looks like the main library that Amazon provides does actually throw Error, so it is OK to do this. Time will tell if we are swallowing too many other possibilities here.

slifty (approved):

PR #1995, opened 2025-10-02
Use `aws-sdk-client-mock` for s3 testing

### LINKED ISSUE #1994: Move to `aws-sdk-client-mock` for tests involving S3
We were simulating S3 server responses via nock / mocked http calls.  This worked... but it was brittle and as I've been trying to add new functionality to the bulk uploader it is resulting in complications (#1858).

This exists and seems well supported: https://www.npmjs.com/package/aws-sdk-client-mock

slifty:
This PR moves us to using `aws-sdk-client-mock` for S3 mocks in the bulk uploader.

This is a lot cleaner and more robust, which will come in handy particularly as we ad support for file uploads.

Resolves #1994


src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -1,6 +1,8 @@
 import path from 'node:path';
-import nock from 'nock';
-import { requireEnv } from 'require-env-variable';
+import fs from 'node:fs';
+import { mockClient } from 'aws-sdk-client-mock';
+import { GetObjectCommand, S3Client } from '@aws-sdk/client-s3';
+import { sdkStreamMixin } from '@smithy/util-stream';
```
slifty, line 5: This needs to be included as a direct dependency too

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -139,22 +106,28 @@ describe('processBulkUploadTask', () => {
...
+
 		await processBulkUploadTask(
 			{
 				bulkUploadId: bulkUploadTask.id,
 			},
 			getMockJobHelpers(),
 		);
-		expect(requests.getRequest.isDone()).toEqual(true);
+
+		expect(s3Mock.commandCalls(GetObjectCommand).length).toBeGreaterThan(0);
```
slifty, line 130: I think this can be more specific here (or rather, I think this can be `.toEqual(1)`

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -111,26 +94,10 @@ const createTestBaseFields = async (): Promise<void> => {
...
-		bulkUploadFilePath,
-	);
-	return {
-		getRequest,
-	};
-};
-
 describe('processBulkUploadTask', () => {
+	beforeEach(() => {
+		s3Mock.reset();
```
bickelj, line 99: To avoid test interaction in a world where all tests can run concurrently, maybe one `s3Mock` per test instead of one per file. I have no idea how expensive that is, though, and for now it's OK.

PR #1997, opened 2025-10-02
Fix `file` type for base_fields in our schema

### LINKED ISSUE #1996: Fix `file` enum in the database
We added support for BaseFields of type `file` but the database didn't get a migration.  This means if you attempt to create a base field of type `file` you'll get a server error.

This is just a bugfix / not an API change (we document the type already, there's just a bug at the DB level).

Related to #1689

slifty:
This PR fixes a mistake that made it so that base fields could not ACTUALLY have a `file` type.

Resolves #1996

### DIFF
--- a/src/database/migrations/0073-alter-base_fields-add-file.sql
+++ b/src/database/migrations/0073-alter-base_fields-add-file.sql
@@ -1,0 +1,15 @@
+CREATE TYPE updated_field_type AS ENUM (
+	'string',
+	'number',
+	'phone_number',
+	'email',
+	'url',
+	'boolean',
+	'currency',
+	'file'
+);
+ALTER TABLE base_fields
+ALTER COLUMN data_type TYPE updated_field_type USING
+data_type::text::updated_field_type;
+DROP TYPE field_type CASCADE;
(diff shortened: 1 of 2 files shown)


bickelj (approved):
LGTM.

Migrated fine on my machine, didn't lose types, I see the file type added, and I see a test in here providing evidence that it works.

src/database/migrations/0073-alter-base_fields-add-file.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TYPE updated_field_type AS ENUM (
```
kfogel, line 1: Now I know that your "time to go multi-line" limit is either seven items or 80 columns, @slifty (comparing the one-item-per-line approach here with the all-on-one-line approach in [0054-update-field_type-enum.sql](https://github.com/PhilanthropyDataCommons/service/blob/main/src/database/migrations/0054-update-field_type-enum.sql)).

kfogel: I meant "your" in a very particular way that encompasses that kind of outsourcing, don't worry.

PR #2003, opened 2025-10-02
Lint pino

bickelj:


bickelj (commented):
Hmm, several of these are false positives because the `logger` is not a pino logger.

slifty (approved):

Issue #1943, opened 2025-09-17
Handle non-name cases more gracefully


slifty:
We recently added the idea of a user name populated via JWT (#1907) but that implementation had an assumption that names were always present.  We recently learned that isn't actually the case (e.g. the specification of Keycloak JWT allows an undefined name)

While it's true that it's better for name to exist, it seems like incorrect behavior for API calls that otherwise have nothing to do with a name to fail.

Some options are:

(1) Make name a nullable column, and only populate it if name is present in the JWT
(2) Make name required but populate it with some kind of default in the event of a null name
(3) Keep the current solution which is to maintain the requirement and rely on the specific keycloak user having the field

I would suggest `1` is the most correct design. This means client code would handle the "user-with-no-name case" in whatever way is most appropriate for the client use case.

bickelj:
I agree that 2 or 1 is better than the current situation. Mitigation by adding data is there for now, but there is always the chance (approaching 1.0 as users and time increase) it will happen again.

PR #2046, opened 2025-10-15
Relax User `name` constraint from JWT

bickelj:
Sometimes a machine user may not have a `name` in the JWT. Allow that situation without breaking API calls.

Issue #1943 Handle non-name cases more gracefully


bickelj:
I tested locally with a new Keycloak client having a service account without first or last name and then calling the service using that service account to `GET /users`. This covers pretty much all of it.

```
$ curl -s 'http://localhost:3001/users' -H "Authorization: Bearer $(curl -s -k 'https://keycloak.test/realms/pdc/protocol/openid-connect/token' -d 'grant_type=client_credentials' -d 'client_id=pdc-ots-ingest' -d 'client_secret=[redacted]' | jq .access_token | tr -d '"')" | jq
{
  "entries": [
    {
      "createdAt": "2025-10-15T15:08:19.089252+00:00",
      "permissions": {
        "funder": {},
        "changemaker": {},
        "opportunity": {},
        "dataProvider": {}
      },
      "keycloakUserId": "bd5be3f8-8b18-4dba-a9f8-a4d345e669d8",
      "keycloakUserName": null
    }
  ],
  "total": 19
}
```

bickelj:
Ah yes, I remember a test passing when it should fail.

```typescript

describe('requireAuthentication (unit)', () => {
	it('calls next with an UnauthorizedError when no `name` is in the JWT', () => {
		// Because `requireAuthentication` is used to verify that `auth` exists, it
		// would not make sense to make `req` an `AuthenticatedRequest` here.
		const req = getMockRequest() as JwtRequest;
		const res = getMockResponse();
		const mockAuthExp = Math.round(new Date().getTime() / 1000) + 3600;
		req.auth = {
			sub: '3cbe4293-3dcc-463f-8749-488e189aae5a',
			exp: mockAuthExp,
			role: { isAdministrator: false },
		};
		let argToNext: unknown = null;
		requireAuthentication(req, res, (err: unknown) => {
			argToNext = err;
		});
		expect(argToNext).toBeInstanceOf(UnauthorizedError);
	});
});
```

This should actually fail but it passes.

PR #2053, opened 2025-10-17
Lock uuid major version

slifty:
This PR will prevent major version dependabot bumps for this library; starting in v12 uuid has dropped commonJS support, and we are not yet ready to prioritize an ESM migration.

This will close out #1911


bickelj (approved):
LGTM

PR #2075, opened 2025-10-22
Add `createdByUser` to `BulkUpload`

### LINKED ISSUE #2037: Decorate BulkUploads with full User entities
We determined that the current information contained in a User object(namely username) should be accessible to anyone with the access to that user's ID. To start, we want to decorate the bulkUpload, who currently returns the keycloakID of the user who created it, with the actual User object. Further down the line we'd like to decorate all objects returning keycloakId with the users who created them.

slifty:
This PR:

1. updates our `BulkUpload` entity to include a `createdByUser` attribute, exposing basic user information for the bulk upload creator.
2. Updates our user queries to only provide permission data as appropriate for a given auth context.

Resolves #2037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.25.5 2025-10-22
+
+### Added
+
+- `BulkUpload` entities now have a `createdByUser` attribute.
+
 ## 0.25.4 2025-10-22
 
 ### Fixed
--- a/src/__tests__/bulkUploadTasks.int.test.ts
+++ b/src/__tests__/bulkUploadTasks.int.test.ts
@@ -336,6 +336,7 @@ describe('/tasks/bulkUploads', () => {
 				.send(testData)
 				.expect(201);
 			const after = await loadTableMetrics('bulk_upload_tasks');
+			const expectedCreatedByUser = await loadTestUser();
 
 			expect(before.count).toEqual(0);
 			expect(result.body).toEqual({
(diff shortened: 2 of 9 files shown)


src/types/AuthContext.ts
```diff
@@ -11,6 +11,10 @@ interface AuthContext {
 	};
 }
 
+type AuthIdentityAndRole = Pick<AuthContext, 'role'> & {
```
slifty: Is `role` part of identity (could this just be `AuthIdentity`?)

bickelj: The role can be added and removed, so IDK, they're two separate things. What was wrong with AuthContext? Doh, I see, this is the shallow AuthContext not the full AuthContext. So `ShallowAuthContext` perhaps? Oh, not really, because the user name is part of `User` and that's a level deeper. Hmm. `AuthSubjectAndGroup` not better. Hmm.

bickelj (approved):
As an admin in my local Keycloak instance, I can see the user is properly decorated. No red flags in the code below.

I wanted to test the non-admin-user scenario to verify that I do not see all the gory user details when I'm not an admin (but do see the name) but due to local configuration experimentation putting my Keycloak in a weird state, I think it's best if I mash Approve since this is a much needed change and appears correct!

src/types/AuthContext.ts
bickelj: How about `PublicAuthContext`?

slifty: For context this isn't for public use -- this is for internal use, just that DB operations only need to know the user ID and role(s) of the current auth context (the DB operations don't need to know other `User` attributes such as the permissions object, since the queries are responsible for checking the authorization internally).

Basically this is just the `Authentication` part of the AuthContext, not the `Authorization` part, since that is what is needed for database operations.

Basically my question was more philosophical: are externally assigned roles part of the Authentication portion of Auth (which then translates to various Authorizations).

Side note when I asked an LLM it suggest AuthorizationContext which made me lol because that would be such a confusing name.

Issue #1613, opened 2025-04-16
Implement audit logging (HTTP request level, 3 of 3)


bickelj:
Issue https://github.com/PhilanthropyDataCommons/service/issues/1559 ended up being associated with the first increment of work on audit logging using database triggers, the stuff originally thought of in https://github.com/PhilanthropyDataCommons/service/issues/1355. Implementation of table triggers to capture changes was straightforward but it would have been awkward to integrate user context into every log there, so we decided to have multiple levels of audit logs rather than relying solely on database triggers. The advantage of triggers on tables is it captures data changes made outside the application. The advantage of audit logging from our generated functions is flexibility over what does and does not get logged. The goal in this ticket is to augment the lower-level and mid-level audit logs with essentially a copy of what pino-http generates, HTTP logs. This ticket is lower priority and may or may not be done in conjunction with the first two level, may or may not be done with a SQL audit log table, may or may not be done at all.

bickelj:
As of this week, I think @slifty and I agreed that this would best be accomplished with the use of a log-file aggregation tool. These logs exist currently as part of the node process' logs though they are not easily accessible. Things we could forward to the log-file aggregation tool would include these HTTP request logs, node/pino logs in general, postgresql logs (log files), Keycloak logs, and maybe even the two audit logs referenced above.

bickelj:
Not that it matters a great deal, but it appears GlitchTip hosted platform uses Digital Ocean as well. It may offer some latency benefits, I guess, if our hosted datacenters are near theirs: https://glitchtip.com/documentation/hosted-architecture

bickelj:
First attempt at getting GlitchTip up and running via `pino-sentry` seems to have worked:

<img width="1920" height="928" alt="Image" src="https://github.com/user-attachments/assets/d58c582b-1b51-4b31-a606-beb7bc4d14d2" />

bickelj:
I changed the tee-shirt size to `M` because it looks straightforward to point all the database and app instances to it. For example, in the managed database UI:

<img width="1671" height="713" alt="Image" src="https://github.com/user-attachments/assets/6df0de44-2fbd-46a7-8a42-03eda4f9a829" />

But not tiny or small because OpenSearch/Elasticsearch configuration and experimentation with actual data and further tuning is needed. There is no way to just say "I want 20GiB of logs" straightaway, you have to fiddle with it.

As Qwen3.6-35B-A3B succinctly put it:

⚖️ The Fundamental Conflict

[...]

PR #2081, opened 2025-10-24
Enable GlitchTip via `pino-sentry`

bickelj:
Issue #1613 Implement audit logging (HTTP request level, 3 of 3)


bickelj:
Closing until we find actual solutions to the problem. GlitchTip is cool but isn't the solution to the log aggregation problem, which to me is more important than the HTTP logging problem.

Issue #1827, opened 2025-08-05
update/correct baseField shortcode on PUT?


jmergy:
I'm sure I was the one that did this but is there a way through a baseField put to update the shortcode? We have one that has a bad shortcode and I just want to update it. 

Example: 

<img width="834" height="517" alt="Image" src="https://github.com/user-attachments/assets/5754cbd2-6a54-414d-b8b9-5be00e1ae2f4" />

if the put could update it, that would be great.

jmergy:
I've updated this to kind of take it out of the pool for now


{
  "label": "TBD",
  "category": "technical",
  "dataType": "string",
  "createdAt": "2025-07-31T14:37:45.611983+00:00",
  "shortCode": "oo",
  "description": "TBD",
  "localizations": {},
  "valueRelevanceHours": 4380,
  "sensitivityClassification": "forbidden"
}

bickelj:
I could be mistaken but I don't think we have a rename capability yet.

bickelj:
For those that can see it, [here is a conversation](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/admin.20interface/near/229066) about the capability.

For those that cannot, here is a summary from Devstral:

```quote
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮                       
│  Based on the conversation, here are the key considerations, opinions, and potential issues regarding the mutability of the shortCode field:                        │                       
│                                                                                                                                                                     │                       
│   1. Current State & Technical Limitations:                                                                                                                         │                       
│      ● The shortCode is now used as a primary key for base fields.                                                                                                  │                       
│      ● Currently, updating the shortCode is not supported, which seems to be a technical limitation rather than an absolute rule.                                   │                       
│      ● Jesse Bickel suggested using PATCH with the old short code in the URL and new short code in the body, but noted this would require complex handling like     │                       
│        PostgreSQL cascading updates and 301 redirects.                                                                                                              │                       
│   2. Design Considerations & Intent:                                                                                                                                │                       
│      ● Justin Reese raised the question of whether editing short codes is intended to be supported eventually.                                                      │                       
│      ● Jesse Bickel agreed that if the API doesn't support editing now, neither should the UI suggest it does.                                                      │                       
│      ● Dan Schultz emphasized the REST API design challenges, but agreed mutability could be possible.                                                              │                       
[...]

PR #2112, opened 2025-11-11
Cascade updates to `base_field.short_code`

bickelj:
On tables that reference base field short codes, update the column automatically when the referenced short code changes. This maintains referential integrity in the database when short codes change.

This migration does not make such changes available at the API level nor does it make such changes safe at the social level.

Issue #1827


bickelj:
I merely rebased and signed the existing commit in the most recent push from [26a3e20](https://github.com/PhilanthropyDataCommons/service/commit/26a3e209468dd81def17b2b4264f67c77af0bfa4) to [0fce105](https://github.com/PhilanthropyDataCommons/service/commit/0fce105ff7f5406e9899e1dc423d1d6b3402e831).

PR #2125, opened 2025-11-19
Support partial editing of applicationFormFields

### LINKED ISSUE #2065: Support editing of application form fields
We want certain components of an application form field to be editable -- specifically the instructions and label.

slifty:
This PR adds a `PATCH` endpoint for `/applicationFormFields` which allows the editing of `label` and `description` of a given form field.

In future we might allow editing of other aspects but this narrow set is intentional because it (A) addresses specific use cases and (B) does not have the type of deeper implications that editing field type might create.

Resolves #2065

### DIFF
--- a/src/database/queries/applicationFormFields/updateById.sql
+++ b/src/database/queries/applicationFormFields/updateById.sql
@@ -1,0 +1,6 @@
+UPDATE application_form_fields
+SET
+	label = update_if(:labelWasProvided, :label, label),
+	instructions = update_if(:instructionsWasProvided, :instructions, instructions)
+WHERE id = :applicationFormFieldId
+RETURNING application_form_field_to_json(application_form_fields) AS object;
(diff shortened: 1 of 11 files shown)


src/database/queries/applicationFormFields/updateById.sql
```diff
@@ -0,0 +1,6 @@
+UPDATE application_form_fields
+SET
+	label = coalesce(:label, label),
+	instructions = coalesce(:instructions, instructions)
```
bickelj: The label and instructions are the only nullable fields but `coalesce` means we cannot set them to null. I think it's OK because we can use an empty string as a workaround in the rare case where we want to go from a string of some kind to null.

bickelj:
I was able to try it locally successfully, too!

src/database/queries/applicationFormFields/updateById.sql
slifty: Oh snap that's awkward... what do we even do 😭

(I'll see if I can think of an elegant solution)

PR #2139, opened 2025-11-25
Convert patch queries to use nullable pattern

### LINKED ISSUE #1472: Add ability to reset a nullable field back to null
Specifically, when running `PATCH /changemakers/:id`, it is currently possible to go from a `null` Keycloak organization UUID to a new UUID, from an existing UUID to a different UUID, but not back to `null` from any given UUID. The difficulty lies in distinguishing "I mean to set this field to null" and "I do not mean to set this field at all." See #1468.

slifty:
This PR updates our `update` queries to use `update_if` instead of `coalesce`.  This change allows explicit `null` values to be set in an update (before null was treated as though it meant `undefined` since pg converts both `null` and `undefined` to SQL null)

Resolves #1472

I have rebased this against #2164 and so it shouldn't be merged until after that PR is merged.

### DIFF
--- a/src/__tests__/applicationFormFields.int.test.ts
+++ b/src/__tests__/applicationFormFields.int.test.ts
@@ -264,6 +264,66 @@ describe('/applicationFormFields', () => {
 			});
 		});
 
+		it('successfully updates label to null', async () => {
+			const testUser = await loadTestUser();
+			const systemUser = await loadSystemUser(db, null);
+			const systemUserAuthContext = getAuthContext(systemUser);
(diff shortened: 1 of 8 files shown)


bickelj (approved):
LGTM and this is worth merging as-is.

I tried it only on the `PATCH /changemakers` endpoint and it seemed to work. The error that comes back when trying to set `name` or `taxId` to null is a 500 error while ideally it would be a 400 error but again this is minor compared to not being able to reset the `keycloakOrganizationId` to `null` which this does allow.

PR #2164, opened 2025-12-09
Disable AJV coersion

### LINKED ISSUE #2158: Disable ajv coercion
We have ajv validators which will coerce things -- e.g. if a number is expected, but a string is provided and that string contains a valid number AJV will replace the string with the parsed number.

I think we maybe benefitted from this in the past, but ultimately it's (1) implicit magic and (2) it means that when `null` is passed as PUT / PATCH to a non-nullable string value it is being converted to an empty string instead of the request being rejected, which is unexpected / inconsistent behavior (the user wanted to set to null, but instead it was set to "").

It may be that when working on this we find out why we had coercion enabled, in which case i'll update the issue and add a comment to prevent re-discovery in the future.

slifty:
This PR disables AJV's built in coercion and applies coercion of our own in a few specific locations.

This is a fairly large change to the code base, but I think it's overall better.  It is now more explicit and our API is stricter about inputs and does not magically coerce client input.

The need for this change came because AJV was converting `null` to `""` in certain cases, making it impossible to update a nullable string field to null.  This, combined with some issues related to AJV's schema-to-typescipt-type mapping was resulting in a situation the use of coalesce was becoming unpredictable enough to warrant this change.

Resolves #2158
Related to #1472

### DIFF
--- a/src/ajv.ts
+++ b/src/ajv.ts
@@ -3,9 +3,7 @@ import ajvKeywords from 'ajv-keywords';
 import addFormats from 'ajv-formats';
 import * as tags from 'language-tags';
 
-export const ajv = new Ajv({
-	coerceTypes: true,
-});
+export const ajv = new Ajv();
 
 ajv.addKeyword({
 	keyword: 'isValidLanguageTag',


src/coercion/__tests__/coerceQuery.unit.test.ts
```diff
@@ -0,0 +1,124 @@
...
+		expect(result.value).toBe('');
+	});
+
+	test('coerces multiple query params with different types', () => {
+		const query = {
+			page: '1',
+			count: '10',
+			search: 'test',
+			active: 'true',
+			ids: ['1', '2', '3'],
```
hminsky2002: Should we be testing for nested objects as well? I don't see that coming up for query parameters, but if we're testing arrays, wondering if we should also add a test for nested objects

src/middleware/requireChangemakerPermission.ts
```diff
@@ -28,9 +29,7 @@ const requireChangemakerPermission =
 			next();
 			return;
 		}
-		const {
-			params: { changemakerId },
-		} = req;
+		const { changemakerId } = coerceParams(req.params);
```
hminsky2002: noting here that we are still reading the req object directly in requireDataProviderPermission and requireFunderPermission. Is that because these are string-only values that are not nullable? I know we are calling coerceParams in handlers that previously read dataProviderShortCode directly from the req object

src/coercion/coerceParams.ts
```diff
@@ -0,0 +1,24 @@
+import type { Request } from 'express';
+
+type CoercedParamValue = string | number | boolean;
```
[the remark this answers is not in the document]
slifty: I know it's redundant but I don't think the two things are actually coupled concepts.

One captures the possible values that a parameter can be coerced to (currently just these three, though maybe some day this will expand); the other captures the possible values that a queryString value can be coerced to.

Which I suppose is a very long way of saying "we don't always mean for these values to be the same"!

PR #2166, opened 2025-12-10
Add `ChangemakerFieldValue`

### LINKED ISSUE #1321: Support "external field values"
We need a way to submit organizational data outside of the context of a proposal.  This will be done through the concept of external field values, although the name may change as we implement.

This is related to #1088, though not all external field values will necessarily come from a data provider (e.g. in some cases it may come from a changemaker directly)

slifty:
This PR adds two new entities with related `POST` endpoints:

1. `ChangemakerFieldValue` entity with a related `POST` endpoint.  This entity exists to allow data to be created directly for a given changemaker outside of the scope of a specific project / proposal.
2. `ChangemakerFieldValueBatch` entity which exists to provides more context around the "how" and "why" a given set of changemaker field values was inserted into the system.

I intentionally did not include DELETE and PATCH endpoints for this PR in order to keep the scope narrow, but we will want to think through the use cases there and likely implement them in the future.

Resolves #1321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.27.0 2025-12-22
+
 ### Added
 
 - `ChangemakerFieldValueBatch` entity type now exists to group changemaker field values that were imported together from an external source.
 - `POST /changemakerFieldValueBatches` endpoint now exists to create changemaker field value batches.
+- `ChangemakerFieldValue` entity type now exists to represent field values directly associated with changemakers.
+- `POST /changemakerFieldValues` endpoint now exists to create changemaker field values.
 
 ## 0.26.0 2025-10-30
 
--- a/src/handlers/changemakerFieldValuesHandlers.ts
+++ b/src/handlers/changemakerFieldValuesHandlers.ts
@@ -1,0 +1,126 @@
+import { HTTP_STATUS } from '../constants';
+import {
+	db,
+	createChangemakerFieldValue,
+	loadBaseField,
+	loadChangemaker,
+	loadChangemakerFieldValueBatch,
+} from '../database';
+import {
+	isAuthContext,
+	isWritableChangemakerFieldValue,
+	Permission,
+	BaseFieldCategory,
+	BaseFieldSensitivityClassification,
+} from '../types';
+import {
+	FailedMiddlewareError,
+	InputValidationError,
+	InputConflictError,
+	NotFoundError,
+	UnprocessableEntityError,
+} from '../errors';
+import { authContextHasChangemakerPermission } from '../authorization';
+import { fieldValueIsValid } from '../fieldValidation';
+import type { Request, Response } from 'express';
+
+const postChangemakerFieldValue = async (
+	req: Request,
+	res: Response,
+): Promise<void> => {
+	if (!isAuthContext(req)) {
+		throw new FailedMiddlewareError('Unexpected lack of auth context.');
+	}
+
+	const body = req.body as unknown;
+	if (!isWritableChangemakerFieldValue(body)) {
+		throw new InputValidationError(
+			'Invalid request body.',
+			isWritableChangemakerFieldValue.errors ?? [],
+		);
+	}
+
+	const { changemakerId, baseFieldShortCode, batchId, value, goodAsOf } = body;
+
+	// Verify changemaker exists
+	await loadChangemaker(db, req, changemakerId).catch((error: unknown) => {
+		if (error instanceof NotFoundError) {
+			throw new InputConflictError('The changemaker does not exist.', {
+				entityType: 'Changemaker',
+				entityId: changemakerId,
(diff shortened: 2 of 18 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.27.0 2025-12-22
+
+### Added
+
+- `ChangemakerFieldValueBatch` entity type now exists to group changemaker field values that were imported together from an external source.
```
slifty: I landed on this name but I will be very curious if it conveys intention naturally or if it's confusing.

hminsky2002 (approved):
Looks great, one note I have is verifying permissions for sources in the POST changemakerfieldvaluebatch -- just want to make sure we're not accidentally letting users see sources they shouldn't be able to. But approving because all tests are comprehensive and passing locally, and design seems correct!

src/handlers/changemakerFieldValueBatchesHandlers.ts
```diff
@@ -0,0 +1,58 @@
...
+		throw new InputValidationError(
+			'Invalid request body.',
+			isWritableChangemakerFieldValueBatch.errors ?? [],
+		);
+	}
+
+	const { sourceId, notes } = body;
+
+	// Verify source exists
+	await loadSource(db, req, sourceId).catch((error: unknown) => {
```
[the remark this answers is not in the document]
slifty: @hminsky2002 The answer should be yes (this is why we pass auth to the query) I'll confirm, though, and if we don't I'll update the query to include that.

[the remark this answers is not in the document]
slifty: The answer is no!

It also turns out this issue manifests in the bulkUploadTasksHandlers as well.  This is a bug, but if it's OK I'm going to capture it in a new issue and reference the permissions issues I'm working on now, since it will impact those as well.

CHANGELOG.md
slifty: Totally makes sense -- the issue I would raise regarding group is I think it implies semantic association beyond simply the process by which the data was inserted; ultimately the point here is that this entity is meant to describe the insertion operation that led to a given set of data being added.

I think what we're about to do *is* a batch insertion (the batching happens external to the API, but it's still a batch insert operation).  On the flip side... imagine a use case where someone is editing org data in exchange app and they edit a single entity.  The API in this PR would require exchange to create a batch operation => singular new field value, which seems a bit silly.  That is an issue regardless of the name, though,  and the need for it is that source association comes from the batch / isn't applied to the individual field value.

---

I think the bottom line for right now is that this is probably "good enough" and we can migrate to a better name in future once this data is entered and we have a chance to reflect in the new year.

src/openapi/components/schemas/ChangemakerFieldValue.json
```diff
@@ -0,0 +1,66 @@
...
+	"type": "object",
+	"properties": {
+		"id": {
+			"type": "integer",
+			"readOnly": true,
+			"example": 1
+		},
+		"changemakerId": {
+			"type": "integer",
+			"readOnly": true,
```
bickelj: I think this `readOnly` flag (or something else prevents it from showing up in the OpenAPI example schema on post:

image

src/handlers/changemakerFieldValuesHandlers.ts
```diff
@@ -0,0 +1,126 @@
...
+				});
+			}
+			throw error;
+		},
+	);
+
+	// Verify base field is organization category
+	if (baseField.category !== BaseFieldCategory.ORGANIZATION) {
+		throw new UnprocessableEntityError(
+			`Base field ${baseField.shortCode} has category '${baseField.category}' but must have category 'organization' for changemaker field values.`,
```
bickelj: This error should be thrown but the message should be changed.

Example when I tried to post:
 "message": "Base field proposal_primary_contact_title has category 'project' but must have category 'organization' for changemaker field values."

This message could be (mis)interpreted that something is wrong with the base field rather than something wrong with the post. Instead, clarify that `Values for ${baseField.shortCode} must be posted to /applicationForms as proposal field values because this field is of category '{baseField.category}'. Only 'organization' field values may be posted here.` (Or something like that)

src/openapi/paths/changemakerFieldValues.json
```diff
@@ -0,0 +1,65 @@
+{
+	"post": {
+		"operationId": "addChangemakerFieldValue",
+		"summary": "Creates a new field value for a changemaker.",
+		"description": "Creates a new field value record for a changemaker. Multiple values can exist for the same changemaker and field combination.",
```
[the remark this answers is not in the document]
slifty: Fixed I think!

slifty:
Since I can't reply to a root comment directly without making another root comment...  I will do so!

@bickelj Agreed that we may well want to provide a batch / bulk upload for this data too in a future iteration

slifty:
I believe everything is good to go!  Enabling auto-merge.

Issue #2170, opened 2025-12-29
Update npmjs.com API key


bickelj:
[A recent publication of the PDC SDK to npmjs.com failed](https://github.com/PhilanthropyDataCommons/service/actions/runs/20471144986/job/58826537703#step:13:19):
> Error: NpmCallError: Call to "npm publish" exited with non-zero exit code 1
> npm notice Publishing to https://registry.npmjs.org/ with tag latest and public access
> npm notice Access token expired or revoked. Please try logging in again.
> npm error code E404

Given the recent supply chain attacks on npmjs.com it seems [MS has shortened the default token validity duration and more](https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/).

bickelj:
Despite the merge of #2172, still getting this:

> npm error 422 Unprocessable Entity - PUT https://registry.npmjs.org/@pdc%2fsdk - Error verifying sigstore provenance bundle: Failed to validate repository information: package.json: "repository.url" is "", expected to match "https://github.com/PhilanthropyDataCommons/service" from provenance

PR #2173, opened 2025-12-30
Remove the "git+" prefix from repo URL

bickelj:
Issue #2170 Fix our NPM publishing action


slifty (approved):
We looked into this and it's actually better to use https regardless!  We think this will also end add a link back to the repo in the npm.

Issue #2161, opened 2025-12-09
Apply new permission model to existing entities


slifty:
In #2160 we talk about implementing the API for the permission model described in #1785

This issue captures the work of actually applying that permission model in various queries.

This will be a series of PRs for various endpoints (it may end up making sense to make sub-issues as that work is being done, in order to better track it).

PR #2187, opened 2026-01-05
Implement new permission grant endpoints

### LINKED ISSUE #2160: Implement API for a new permission model
In #1785 we laid out a vision for a more generalized and granular permission model.  This issue captures the work of defining and implementing the API for managing that model.

This does not capture the work of actually applying the permission model (e.g. actually using the permissions for other operations), nor does it capture the work of removing the old model.

This issue will involve:

1. Model definition + implementation
2. Endpoint definition + implementation
3. API Documentation

slifty:
This PR adds a new endpoints for a more robust and flexible permission system.

It allows the creation and deletion of grants in the new system, though it does not actually apply the new system yet (which is to say: admins can create new permission grants but they won't do anything until #2161).

Resolves #2160


src/openapi/paths/permissionGrant.json
```diff
@@ -0,0 +1,89 @@
...
+				"description": "The permission grant.",
+				"content": {
+					"application/json": {
+						"schema": {
+							"$ref": "../components/schemas/PermissionGrant.json"
+						}
+					}
+				}
+			},
+			"401": {
```
hminsky2002: This also returns a 400 when id is invalid i believe

hminsky2002 (commented):
Looking really good, from what @slifty and I discussed on Friday the 23rd, I am on board with the model here. I am wondering if we need a unique constraint on the permissions (for scope/entity/user fields) to avoid an issue of duplicates/losing the 'true' permission entity.

PR #2190, opened 2026-01-06
Use `ChangemakerFieldValues` for gold data

### LINKED ISSUE #2137: Include `ChangemakerFieldValue` data in changemaker "gold" data
After we have #1321 as an entity collecting organization attributes, include these attributes in the "gold" data queries that return changemakers (`GET /changemakers`).

slifty:
This PR updates our system to incorporate `ChangemakerFieldValues` when calculating gold data for a `Changemaker`.

It also introduces the concept of `FieldValueBase` which reflects the common elements of any type of `FieldValue`.

Resolves #2137


hminsky2002 (approved):
Tests are passing here and this seems like a reasonable extension of our current initialization query. I do think @bickelj's insight would be valuable here, but approving this from a technical standpoint!

PR #2193, opened 2026-01-07
Use gold schema for tests

slifty:
This PR is not ready for review but I wanted to show it off for now.

It moves our tests to use a gold schema instead of running migrations for every test.  This is faster out of the box, but also it allows us to run our integration tests in parallel (since the migration library had a limitation forcing us to run in band).

What once took 5 minutes will now take... much less than that (TBD let's see)!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,12 @@ jobs:
           --health-timeout 5s
           --health-retries 5
     steps:
+      - name: Install PostgreSQL client tools
+        run: |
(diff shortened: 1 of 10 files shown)


slifty:
It looks like a test is failing -- I'm betting something about our mocking is leaking between parallel tests.

This PR is currently somewhat in "proof of concept" mode, so what I think I'll do is clean up the gold data portion of this implementation and then make sure our tests support parallelization.  I may break that into two PRs, or maybe keep it as one with mutliple commits.

PR #2198, opened 2026-01-13
Support bulk upload to existing application form

### LINKED ISSUE #2063: Support bulk upload to existing application form
Slightly different from #1880, let the user submit an application form first, and then make a bulk upload with reference to that form.

hminsky2002:
This PR refactors the bulk upload process to no longer create an application form as part of the bulk upload. Rather, it now takes an existing application form id and verifies the data against the form. As part of this PR, bulk uploads that exist before the migration are backfilled to be associated with a system application form.

See the corresponding front-end pr: https://github.com/PhilanthropyDataCommons/front-end/pull/1310

Closes #2063

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.29.0 2025-01-23
 
 ### Added
 
--- a/src/openapi/api.json
+++ b/src/openapi/api.json
@@ -3,7 +3,7 @@
 	"info": {
 		"title": "Philanthropy Data Commons API",
 		"description": "An API for a common data platform to make the process of submitting data requests to funders less burdensome for changemakers seeking grants.",
-		"version": "0.28.0",
+		"version": "0.29.0",
 		"license": {
 			"name": "GNU Affero General Public License v3.0 only",
 			"url": "https://spdx.org/licenses/AGPL-3.0-only.html"


src/database/initialization/bulk_upload_task_to_json.sql
```diff
@@ -49,12 +55,14 @@ BEGIN
     'id', bulk_upload_task.id,
     'sourceId', bulk_upload_task.source_id,
     'source', source_json,
+    'applicationFormId', bulk_upload_task.application_form_id,
```
slifty: Make sure in this commit that it's formatted properly (I think tabs is what we're using)

src/database/migrations/0081-alter-bulk_upload_tasks-add-application_form_id.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE bulk_upload_tasks
+ADD COLUMN application_form_id integer REFERENCES application_forms (id);
```
slifty: We want this to be not null, I think -- basically, going forward we're saying a bulk upload task MUST have an application form ID.  By making this not null our types are more constrained (making them easier to work with).

To accomplish this, we will need this migration to create a `system_application_form` (I don't think this already exists, but if it does we can just use that).  We've seen this pattern already for Source, which had a similar issue (this may have actually been for bulk upload!)

I think this would also require the cration of a system opportunity.  System funder already exists.

You might want to simply inject these values as a migration + commit that is made *before* this one.  Note that system entities have some boilerplate and we would want to be consistent in creating it for the new types:

1. DB operation / utility for use in TypeScript
2. DB function for use in SQL

Issue #2197, opened 2026-01-08
Add documentation for using API to attach files to proposal versions


yshng:
### Summary

From Zulip [#PDC > basefield files @ 💬](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/basefield.20files/near/245629)

@kfogel asks:

Do we have any documentation/guidance on how to use the API to upload a file (i.e., as the value of a baseField whose type is file) and tie it to a given proposalVersion? The end-to-end process for doing this is not entirely clear from the Swagger documentation.

### Problem / Opportunity

Multi-step "recipes" for using API are not documented, including this particular case.

### Proposed solution

Add such documentation.

yshng:
I've run into issues uploading a file with `api-test` (EntityTooSmall error), so I'm blocked right now on testing out the full workflow documented in the above linked PR

PR #2202, opened 2026-01-26
Add documentation for file uploads

yshng:
Provides documentation requested in #2197

Includes:
* How to use bulk upload to upload proposals and attachments
* Generally applicable instructions for uploading files and adding them to the database
* Specific case of attaching a file to a ProposalVersion which prompted this documentation

### DIFF
--- a/docs/ADD_FIELDVALUE_ATTACHMENT.md
+++ b/docs/ADD_FIELDVALUE_ATTACHMENT.md
@@ -1,0 +1,151 @@
+# Uploading attachments to field values
+
+This details the steps for uploading files as attachments to field
+values using the PDC API. Note that PDC is unable to act as storage
+for any other files except those which will be used as field value
+attachments (or as part of the Bulk Upload process).
+
+If you would like to use the Swagger UI to make requests and view their
+attendant schema, you can do so at api.philanthropydatacommons.org. You will
+need to click Authorize, and then Authorize again on the pop-up to log in.
+
+## Process Overview
+
+Uploading documents to the database via the PDC API is a multistep process.
+
+1. Register file metadata to generate a [presigned post
+   URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html)
+   and file ID
+2. Use `curl` to upload the file
+3. Use the revelant POST endpoint to associate the file ID with your desired
+   field value
+
+### Create a presigned post URL
+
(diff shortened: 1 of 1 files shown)


docs/UPLOAD.md
```diff
@@ -0,0 +1,238 @@
...
+- If you do not have attachments, use `null`. Howevever, you will get
+  get an error if you use a base field which has datatype file and you
+  do not include an attachment archive.
+- Click "Execute"
+- You should receive a response with code 201 that shows the correct
+  files and information for the bulk upload. This does not mean the
+  upload was successful, only that this is the form of the upload that
+  will be attempted.
+
+**2. Confirm bulk upload and proposal**
```
kfogel: The more I read, the more I wonder if maybe there's some key thing *I'm* not understanding here.  I'll ask some questions when we talk.

docs/ADD_FIELDVALUE_ATTACHMENT.md
```diff
@@ -0,0 +1,139 @@
+# Uploading attachments to field values
+
+This details the steps for uploading files as attachments to field values using
+the PDC API.
```
hminsky2002: Perhaps it's worth an extra level of emphasis here that we don't allow file uploads as storage or any other purpose besides as attachments for field values (or the csvs for the bulk uploads).

docs/ADD_FIELDVALUE_ATTACHMENT.md
```diff
@@ -0,0 +1,139 @@
...
+	-F "key=<UUID>" \
+	-F "Policy=<extremely-long-alphanumeric-string>" \
+	-F "X-Amz-Signature=<longish-string>" \
+	-F "Content-Type=text/csv" \
+	-F "file=@/path/to/sampleproposal.csv"
+```
+
+### Associate file ID with a field value
+
+- Make sure you have the file IDs of the files you'd like to include.
```
hminsky2002: Worth noting that you can use the `GET files` route to verify both that the file upload was successful and to retrieve the file id if you don't note it down after upload

hminsky2002 (changes requested):
This looks good! I left two suggestions for places that could be clarified but this is a thorough walk through of the process. In an ideal world we could now have one other person with as little context as possible to step through this, but that may not be possible at this moment.

hminsky2002 (approved):
Looks great! Needs to be rebased and I'm not 100% sure why CI is failing, but approving these docs!

Issue #2162, opened 2025-12-09
Remove old permission model endpoints


slifty:
Once #1785 is complete we will want to remove the old permission endpoints from the system.

PR #2225, opened 2026-02-06
Migrate `Funder` permissions to new system

slifty:
This PR continues our migration to the new permission system, for the `Funder` entity / routes.

Related to #2161 
Related to #2162

This PR should be merged after #2224


src/openapi/components/schemas/User.json
```diff
@@ -33,17 +24,17 @@
 						}
 					}
 				},
-				"funder": {
+				"opportunity": {
```
bickelj: Is this change right?

PR #2247, opened 2026-02-13
Changemaker field value permissions

### LINKED ISSUE #2245: Add `ChangemakerFieldValue` grant permissions
We're about to add support for conditions in permissions (#2231) and we know already that we want this to primarily be used for conditions on field value access.

In order to do this comprehensively we will want to be able to apply them to `ChangemakerFieldValues` as well as `ProposalFieldValues`.

This means we should add permission support for this entity type (right now it's inherited from Changemaker overall)

slifty:
This PR adds permission checks using the new permission system for accessing `ChangemakerFieldValues`.

Resolves #2245

### DIFF
--- a/src/database/initialization/changemaker_to_json.sql
+++ b/src/database/initialization/changemaker_to_json.sql
@@ -104,22 +104,18 @@ BEGIN
 				ON cfvb.source_id = s.id
 			WHERE cfv.changemaker_id = changemaker.id
 				AND bf.category = 'organization'
-				AND bf.sensitivity_classification != 'forbidden'
 				AND cfv.is_valid
 				-- Guard against possible removal of NON NULL constraint on users table:
 				AND u.keycloak_user_id IS NOT NULL
 				-- Guard against the valid-but-not-really-valid-here system user:
 				AND u.keycloak_user_id != system_keycloak_user_id()
-				-- Check permission: field must be public OR user has view permission
-				AND (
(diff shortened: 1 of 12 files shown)


src/database/initialization/changemaker_to_json.sql
```diff
@@ -104,22 +104,18 @@ BEGIN
...
-					OR has_changemaker_permission(
-						auth_context_keycloak_user_id,
-						auth_context_is_administrator,
-						cfv.changemaker_id,
-						'view',
-						'changemakerFieldValue'
-					)
+				-- Check permission to view this changemaker field value:
+				AND has_changemaker_field_value_permission(
+					auth_context_keycloak_user_id,
```
hminsky2002: One thing to question is if there's any level of inheritance that should apply to these (i.e are we certain that change maker permission should not be considered in granting permission here?) I feel like practically speaking, if someone has edit permissions on the change maker, they should have access to editing the field values, and I know part of the point of this system is to be as graunlar as possible, but i forsee a world where someone who was granted changemaker permission was also implicitly expected to be able to view all field values.

hminsky2002: Reasonable enough, I suppose this gives us the tools to implement whatever level of permission-ing we want. The onus is then on the UI to fit the permission-ing to it's liking. You've convinced me on the matter of extend-ability!

hminsky2002 (approved):

PR #2259, opened 2026-02-18
Improve the OpenAPI spec permission grant example

bickelj:
Without this change, some critical keys in the JSON request body do not appear in the example POST request body. With this change, it is easier to see the required request body keys because they're present.


slifty:
@bickelj is there more you wanted to include in this PR before I review? (I know you were fighting with the linter)

bickelj:
I think the spirit of this attempt lives on but this particular PR can be closed.

PR #2261, opened 2026-02-19
Add fieldType for clarfiying rendering policy on applicationFormField

### LINKED ISSUE #1847: [Enhancement]: add display/render hinting for fields on applicationForm
### Summary

Right now, and historically, PDC has not been all that concerned with rendering the data for people to see in displays. But, now that we have Exchange and hoping for other applications down the road not to mention integrations into existing platforms, etc. 

### Problem / Opportunity

It might make sense to add an optional attribute to the fields on the applicationForm. This would be helpful, if included, so any rendering of the proposal data against that application form could be pulled from the data. 

This is coming up with Changemaker Direct. 

Image

### Proposed solution

Rather than having to design the form for data entry here and maintain rendering or PDC data in Exchange or other platforms, it would be ideally nicer to just defer to data parameters on the field specific to the applicationForm on whether to render it as a single text line, multiple line, or whatever. It feels like it would be similar to what we are doing with the position attribute. 

Image

Could we add an optional parameter at that level to signal preferred rendering? 

### Acceptance criteria

_No response_

### Priority

Medium

hminsky2002:
This commit adds an enum, fieldType, to applicationFormField, which is meant to store some simple information about how the form field should be rendered in the UI. Note that it does not contain deeper information, such as in the case of a dropdown, what the options of the dropdown should be, but rather servers as a general guideline.

Closes #1847

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Added
-
-- `applicationFormField` now has an `inputType` field to provide rendering hints for UIs rendering the fields.
-
 ### Changed
 
 - Permission grants for users no longer require the user to exist in the PDC database. Grants can now reference any Keycloak user UUID, matching the existing behavior for user group grants.
@@ -27,6 +23,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Permission checks for userGroup-based grants now correctly ignore expired ephemeral user group associations.
+
+## 0.32.0 2026-02-20
+
+### Added
+
+- `applicationFormField` now has an `inputType` field to provide rendering hints for UIs rendering the fields.
 
 ## 0.31.0 2026-02-12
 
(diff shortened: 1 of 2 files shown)


src/__tests__/applicationFormFields.int.test.ts
```diff
@@ -12,6 +12,7 @@ import { createTestFunder, createTestOpportunity } from '../test/factories';
 import { getAuthContext, loadTestUser } from '../test/utils';
 import { mockJwt as authHeader } from '../test/mockJwt';
 import {
+	ApplicationFormFieldType,
```
slifty: I think this will be clearer with a more specific name (to help differentiate this from the application form field's base field's type)

We talked about maybe `ApplicationFormFieldInputType` to underscore that this is about the input UX, not the data.

src/openapi/components/schemas/ApplicationFormField.json
```diff
@@ -36,6 +36,25 @@
...
+				"boolean",
+				"date",
+				"date_time",
+				"select",
+				"multi_select",
+				"file",
+				"email",
+				"phone",
+				"hidden",
+				null
```
slifty: I don't believe null should be part of the enum itself.

Issue #2265, opened 2026-02-25
Make external IDP integration instructions easier to navigate


yshng:
Document should tell reader what services there are instructions for and direct them to easily find their service.

PR #2266, opened 2026-02-25
Restructure external IDP integration docs

yshng:
Addresses #2265


bickelj:
Whoops, I didn't realize auto-merge was enabled, forgot to check. Oh well, a new PR can address the above : )

yshng:
@bickelj Oops! Yes I will followup tomorrow and address this. My initial thought is that that step may already be included in the other service instructions because they go back and forth between the external IdP and Keycloak. Of course, I will double-check this thoroughly and correct as needed. 

Thanks for reviewing - apologies that my diffs were so dismal here. In hindsight I realize the correct order of operations for producing better diffs (or editing the commits after the fact).

PR #2309, opened 2026-04-02
Add `PUT /permissionGrants` endpoint

### LINKED ISSUE #2308: Support editing of specific permission grants
There's no real reason not to allow editing of a specific permission grant, now that they are rich objects.  From a user perspective this is going to be expected.

This means adding support for PUT or PATCH for permission grants (I lean towards PUT for now since it's more straightforward and the elements of a permission grant are so intertwined to the data being sent)

slifty:
This PR adds a `PUT` endpoint to permissionGrants.

Resolves #2308

### DIFF
--- a/src/database/operations/permissionGrants/updatePermissionGrant.ts
+++ b/src/database/operations/permissionGrants/updatePermissionGrant.ts
@@ -1,0 +1,38 @@
+import { generateCreateOrUpdateItemOperation } from '../generators';
+import type {
+	Id,
+	PermissionGrant,
+	WritablePermissionGrant,
+} from '../../../types';
+
+const updatePermissionGrant = generateCreateOrUpdateItemOperation<
+	PermissionGrant,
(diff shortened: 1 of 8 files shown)


src/database/operations/permissionGrants/updatePermissionGrant.ts
```diff
@@ -0,0 +1,38 @@
...
+} from '../../../types';
+
+const updatePermissionGrant = generateCreateOrUpdateItemOperation<
+	PermissionGrant,
+	WritablePermissionGrant,
+	[permissionGrantId: Id]
+>(
+	'permissionGrants.updateById',
+	[
+		'granteeType',
```
hminsky2002: Yep this covers everything. I think this is preferable, if there's certain values we want to guide the user away from changing that can be a UI level decision. But I think having this general functionality is very much what we want!

Issue #1275, opened 2024-10-22
Build out PR deploy previews


jasonaowen:
It should be possible to create a deploy preview for each new pull request to aid in review and testing.

[DigitalOcean has an example of how to do so](https://github.com/digitalocean/app_action?tab=readme-ov-file#launch-a-preview-app-per-pull-request), although I expect we'd have to figure out some wrinkles like the test database.

Issue https://github.com/PhilanthropyDataCommons/deploy/issues/118 Migrate to DigitalOcean Apps
Issue #622 Change deploy model

bickelj:
While working on #1410 I noticed we can create and delete databases within an existing cluster. I think that would be the path forward for deploy previews, create and delete dbs in the test cluster. Another thing to consider is creating and destroying some S3 buckets. We'd have to carefully create and push docker images that are tagged with `preview` or `beta` or `test` and also get destroyed on PR merge. Another thing is the PDC Keycloak expected URIs. Some kind of call to Keycloak to add and remove those might be desired, or perhaps a simple wildcard in the test environment would work.

But I think this is all back-burner for now.

bickelj:
The dependency tree for a service instance is like this:
```mermaid
flowchart TD
    P(PDC Service) --> K(Keycloak)
    K --> KDB(Keycloak DB)
    P --> PDB(PDC DB)
    P --> S3(S3 Bucket)
```

The tricky or sticky parts, the last of which I was reminded of above:
* Creating databases
* Destroying databases
* Creating a Keycloak instance
* Creating a throwaway service image such that it is appropriately separated from production images (and not accidentally deployed somewhere else)

The assumption currently is that we would use DO API calls to create and destroy components as needed.

Options for databases from most isolation to least:
* Separate VM for each DB
* Single VM with separate PG instances for each DB
* Single VM with shared PG instance, separate DB within PG (may require a code change)

I think we'd want the last option if possible: one single DB VM to serve all branches and all DBs. If the code doesn't support it, then a small PG instance launched per branch. We could also use a managed PG instance in place of the VM above.

So the postgres axes are:
* VM vs Managed
* Multiple PGs vs shared PG

Options for Keycloak from most isolation to least:
* Separate VM for each Keycloak
* Keycloak Apps instead of on a VM
* Single VM with separate Keycloak instances (perhaps containers)
* Use the test or prod Keycloak instance for authentication

[...]

PR #2313, opened 2026-04-03
Drop and create PR deploy preview databases

bickelj:
Deploy previews of the service/framework back-end need their own database instances. For most environments we isolate data by database cluster but for the deploy preview environment we isolate by database instance within a cluster (PostgreSQL instance).

This action drops and creates deploy preview databases on a single shared database cluster. It uses the `pdc` database as a template which (as of this writing) is a snapshot of the sandbox database on April 3, 2026.

Further lifecycle changes to this script are expected. In other words, we may not want to run this action on every PR event but condition it on other events and/or separate actions' outcomes.

Issue #1275 Build out PR deploy previews

### DIFF
--- a/.github/workflows/preview-db.yml
+++ b/.github/workflows/preview-db.yml
@@ -1,0 +1,41 @@
+name: Drop and create preview DBs
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, closed]
+
+env:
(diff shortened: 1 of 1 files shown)


bickelj:
Closing to see that the database is in fact destroyed on PR close.

PR #2315, opened 2026-04-06
Add date validation to date fields

### LINKED ISSUE #285: Format and validate dates
One recently imported proposal has this entry:

> Proposal Date: `03-01-05`

Does that mean March 1, 2005? January 3, 2005? January 5, 2003? Something else entirely?

We should parse incoming dates, normalize them before storing, and reject ambiguous or invalid dates. That'd allow us to accept input like `Fri Mar 31 2023`, which is unambiguous, but always present a consistent format: `2023-03-31`.

yshng:
Adds date validation, allowing either `date` or `date-time` (timezone required) per [ajv-formats](https://github.com/ajv-validator/ajv-formats)

Resolves #285

### DIFF
--- a/docs/APPLICATION_RENDERING_HINTS.md
+++ b/docs/APPLICATION_RENDERING_HINTS.md
@@ -61,6 +61,7 @@ logic based on `baseField.dataType`. For instance:
 - `url` → URL input
 - `phone_number` → telephone input
 - `file` → file upload control
+- `date` or `date_time` → date picker
 
 ## API Usage
 
--- a/src/__tests__/baseFields.int.test.ts
+++ b/src/__tests__/baseFields.int.test.ts
@@ -230,6 +230,70 @@ describe('/baseFields', () => {
 			expect(after.count).toEqual(1);
 		});
 
+		it('supports creation of date-type base fields', async () => {
+			const db = getDatabase();
+			const before = await loadTableMetrics(db, 'base_fields');
+			const result = await request(app)
+				.put('/baseFields/shorts')
+				.type('application/json')
+				.set(adminUserAuthHeader)
+				.send({
+					label: '🏷️',
+					description: '😍',
+					dataType: BaseFieldDataType.DATE,
+					category: BaseFieldCategory.PROJECT,
+					valueRelevanceHours: null,
+					sensitivityClassification: BaseFieldSensitivityClassification.PUBLIC,
+				})
+				.expect(200);
+			const after = await loadTableMetrics(db, 'base_fields');
+			expect(before.count).toEqual(0);
+			expect(result.body).toMatchObject({
+				label: '🏷️',
+				description: '😍',
+				shortCode: 'shorts',
+				dataType: BaseFieldDataType.DATE,
+				category: BaseFieldCategory.PROJECT,
+				valueRelevanceHours: null,
+				sensitivityClassification: BaseFieldSensitivityClassification.PUBLIC,
+				localizations: {},
+				createdAt: expectTimestamp(),
+			});
+			expect(after.count).toEqual(1);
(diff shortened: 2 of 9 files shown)


slifty (changes requested):
w00t!  Nice contribution!

A few things we will need to add to this PR before merging!

1. OpenAPI specification needs to add `date` as an allowed value for BaseFieldType; we may also have other documentation items that need updating.

2. The schema will need to be updated too (see https://github.com/PhilanthropyDataCommons/service/blob/main/src/database/migrations/0054-update-field_type-enum.sql for an example).

3. We will want a test that ensures that a base field of type `date` can be created (this would help validate that (2) above is properly implemented).

4. Be sure to add a changelog entry!

Separately from those changes, there is a question worth posing: do we want date and date time to be combined into a single type for our API (as opposed to having two distinct `date` and a `date-time` types).  Is that the current intention, or is this meant as `date` as "a valid date might also have a time but you should ignore it if so" that seems reasonable -- this seems fine, but we'll want to be sure to document that.

That said, if we wanted a user building a form to be able say "this base field really is meant to be a date, no time" (e.g. "organizationFoundedOn") or "this really is meant to have a time" (e.g. "nextScheduledBoardMeeting") we would want distinct types.

I'm curious what use cases there are for dateTime, honestly... Which is to say, the current implementation of "it's a date, but we'll allow a date time" seems good, and keeping the current name is fine with me if you prefer it!

hminsky2002 (changes requested):

docs/APPLICATION_RENDERING_HINTS.md
```diff
@@ -61,6 +61,7 @@ logic based on `baseField.dataType`. For instance:
 - `url` → URL input
 - `phone_number` → telephone input
 - `file` → file upload control
+- `date` or `date_time` → date picker
```
hminsky2002, line 64: oh very nice catch to update the application rendering hints!

slifty (approved):
This looks great!  Thank you for the iteration :D

src/__tests__/baseFields.int.test.ts
```diff
@@ -230,6 +230,70 @@ describe('/baseFields', () => {
...
+					category: BaseFieldCategory.PROJECT,
+					valueRelevanceHours: null,
+					sensitivityClassification: BaseFieldSensitivityClassification.PUBLIC,
+				})
+				.expect(200);
+			const after = await loadTableMetrics(db, 'base_fields');
+			expect(before.count).toEqual(0);
+			expect(result.body).toMatchObject({
+				label: '🏷️',
+				description: '😍',
```
hminsky2002, line 285: This isn't a comment on this PR, just the test file in general, but it strikes me that maybe we should try to be a little more intentional with our test inputs, or standardize on using semantically related fluff data when we're mocking stuff. Then again, maybe that would detract from the main point of the test (in this case, the data type, not the label or description).

hminsky2002 (approved):
Looks great to me, nice work @yshng ! One final thing I'd ask is just if you've had a chance to try a bulk upload with a date/date-time field and been able to verify that validation works as expected. Good practice, but in this case the tests really do just provide full coverage.

src/__tests__/baseFields.int.test.ts
slifty: @hminsky2002 I think things will be better when more areas are using test factories (the point of the test factory is "reasonable defaults are in place, and if there is something you actually care about for the test you pass that explicitly")

yshng:
@hminsky2002 I hadn't tried the bulk uploader with it, thanks for the tip. I will make sure to do that going forward (unless you think I should do it now too, which I totally can)

PR #2318, opened 2026-04-07
Add changemaker search

### LINKED ISSUE #1647: Search/filter for changemakers
On Exchange, users need to be able to search or filter for changemakers.

See PhilanthropyDataCommons/service#1641 for access permissions around changemaker data.

See PhilanthropyDataCommons/service#1642 about an organization "profile".

See https://github.com/PhilanthropyDataCommons/service/issues/1618 for general discussion about search.

slifty:
This PR improves our `GET /changemakers` endpoint to support search parameters.

Resolves #1647

### DIFF
--- a/src/database/migrations/0099-alter-changemakers-add-name_search.sql
+++ b/src/database/migrations/0099-alter-changemakers-add-name_search.sql
@@ -1,0 +1,7 @@
+ALTER TABLE changemakers
+ADD COLUMN name_search tsvector
+GENERATED ALWAYS AS (to_tsvector('english', name)) STORED;
+
+CREATE INDEX idx_changemakers_name_search
+ON changemakers
+USING gin (name_search);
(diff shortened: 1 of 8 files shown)


src/database/migrations/0099-alter-changemakers-add-name_search.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE changemakers
+ADD COLUMN name_search tsvector
+GENERATED ALWAYS AS (to_tsvector('english', name)) STORED;
```
hminsky2002: This seems cool! To clarify, the text search vector is used for exact string matching, or does it do soft matching as well (ignoring case, best effort matching)? I think safer to do exact matching for now, but i imagine we'll need to get smarter with it moving forward. Not an issue for rn though!

hminsky2002 (approved):
Looks comprehensive, question about text search in postgres, but approving!

src/database/migrations/0099-alter-changemakers-add-name_search.sql
hminsky2002: Neat. Cheers to system libraries!

PR #2322, opened 2026-04-07
Include changemaker data with proposals

slifty:
This PR adds a `changemakers` array to `Proposal` entities.

Resolves #2321


yshng (approved):
Looks good to me!

PR #2324, opened 2026-04-08
Use gold database for tests

### LINKED ISSUE #2167: Don't run migrations for every test
Right now our test suite runs migrations for every test (ensuring the tests is working on a clean DB state / avoiding contamination across tests).  A faster way to achieve this would be to have the suite run migrations once at the start of running tests, store the resulting database state into a loadable format, and re-load that state directly for each test.

slifty:
This PR changes how our tests run, so that instead of using a test db schema it actually uses separate databases for each test runner (and a gold database).  It also adds some guardrails to ensure tests never run against a "real" database.

For devs this means we now need to populate test db environment variables.

Resolves #2167

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,18 @@ PGUSER=${PGUSER}
 PGPASSWORD=${PGPASSWORD}
 PGDATABASE=${PGDATABASE}
 
+# Test Database Settings
+# Integration tests create and drop temporary databases, which requires the
+# CREATEDB privilege. To limit blast radius, tests use a dedicated PostgreSQL
+# user that is separate from the application user. All TEST_PG* variables must
+# be set to run integration tests.
+# The test user must have the CREATEDB privilege:
+#   CREATE USER pdc_test WITH PASSWORD 'pdc_test' CREATEDB;
+TEST_PGHOST=${TEST_PGHOST}
+TEST_PGPORT=${TEST_PGPORT}
+TEST_PGUSER=${TEST_PGUSER}
+TEST_PGPASSWORD=${TEST_PGPASSWORD}
+
 # `AUTH_SERVER_ISSUER` serves two purposes related to authentication via JSON Web Tokens (JWTs):
 # 1. The "issuer" of a Bearer JWT in the Authorization header needs to match this value.
 # 2. The public key of the JWT issuer is found via
--- a/jest.config.int.js
+++ b/jest.config.int.js
@@ -1,4 +1,9 @@
 var config = require('./jest.config.base.js');
 config.testMatch = ['**/?(*.)+(int).(spec|test).[jt]s?(x)'];
 config.setupFilesAfterEnv = ['<rootDir>/src/test/integrationSuiteSetup.ts'];
+config.globalSetup = '<rootDir>/src/test/integrationGlobalSetup.ts';
+config.globalTeardown = '<rootDir>/src/test/integrationGlobalTeardown.ts';
+// Each worker uses up to 11 PostgreSQL connections (10 pool + 1 admin).
+// Cap at 9 workers to stay within PostgreSQL's default max_connections of 100.
+config.maxWorkers = 9;
 module.exports = config;
--- a/src/test/harnessFunctions.ts
+++ b/src/test/harnessFunctions.ts
@@ -1,67 +1,79 @@
+import { closeDatabase, createAndSetDatabase } from '../database';
+import { TEST_DATABASE_PREFIX } from '../constants';
 import {
-	createDatabase,
-	getDatabase,
-	initializeDatabase,
-	migrate,
-	setDatabase,
-} from '../database';
-import type { TinyPg } from 'tinypg';
+	assertTestDatabaseName,
+	createTestAdminClient,
+	getTestPgConfig,
+	GOLD_DATABASE_NAME,
+} from './testDatabase';
+import type { Client } from 'pg';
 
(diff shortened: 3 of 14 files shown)


src/test/integrationGlobalSetup.ts
```diff
@@ -0,0 +1,63 @@
...
+		'S3_REGION',
+		'S3_ENDPOINT',
+	);
+
+	assertTestDatabaseName(GOLD_DATABASE_NAME);
+	const adminClient = createTestAdminClient();
+	await adminClient.connect();
+	await adminClient.query(`DROP DATABASE IF EXISTS ${GOLD_DATABASE_NAME}`);
+	await adminClient.query(`CREATE DATABASE ${GOLD_DATABASE_NAME}`);
+	await adminClient.end();
```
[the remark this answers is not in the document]
slifty: we control this (it's a constant we reference), so I'm not worried about injection here.

src/test/integrationGlobalTeardown.ts
```diff
@@ -0,0 +1,16 @@
...
+	assertTestDatabaseName,
+	createTestAdminClient,
+	GOLD_DATABASE_NAME,
+} from './testDatabase';
+
+export default async (): Promise<void> => {
+	assertTestDatabaseName(GOLD_DATABASE_NAME);
+	const adminClient = createTestAdminClient();
+	await adminClient.connect();
+	await adminClient.query(`DROP DATABASE IF EXISTS ${GOLD_DATABASE_NAME}`);
```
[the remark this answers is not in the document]
slifty: we control this (it's a constant we reference), so I'm not worried about injection here.

src/test/harnessFunctions.ts
```diff
@@ -1,67 +1,70 @@
...
+		throw new Error(
+			'Admin client not initialized. Call initializeWorker first.',
+		);
+	}
+	const workerDbName = getWorkerDatabaseName();
+	assertTestDatabaseName(workerDbName);
+	await adminClient.query(`DROP DATABASE IF EXISTS ${workerDbName}`);
+	await adminClient.query(
+		`CREATE DATABASE ${workerDbName} TEMPLATE ${GOLD_DATABASE_NAME}`,
+	);
```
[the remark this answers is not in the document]
slifty: Same as other places: injection from our own constants isn't a threat

src/test/testDatabase.ts
```diff
@@ -0,0 +1,78 @@
...
+		'TEST_PGUSER',
+		'TEST_PGPASSWORD',
+	);
+	return {
+		host: TEST_PGHOST,
+		port: Number(TEST_PGPORT),
+		user: TEST_PGUSER,
+		password: TEST_PGPASSWORD,
+		database: TEST_ADMIN_DATABASE,
+	};
```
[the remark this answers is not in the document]
slifty: Meh -- if someone misconfigures they will get an error.  I think validating beyond that is overkill.

slifty:
Look at that: 2m tests down from 6m in CI.

Locally tests ran in *36 seconds* (it depends on how many cores you have locally)

.env.example
```diff
@@ -21,6 +21,18 @@ PGUSER=${PGUSER}
 PGPASSWORD=${PGPASSWORD}
 PGDATABASE=${PGDATABASE}
 
+# Test Database Settings
+# Integration tests create and drop temporary databases, which requires the
+# CREATEDB privilege. To limit blast radius, tests use a dedicated PostgreSQL
+# user that is separate from the application user. All TEST_PG* variables must
+# be set to run integration tests.
+# The test user must have the CREATEDB privilege:
+#   CREATE USER pdc_test WITH PASSWORD 'pdc_test' CREATEDB;
```
hminsky2002, line 30: I wonder if it might be worth an extra flag for this as required, because this reads a smidge as "optional" to me. Then again, might just be me! Maybe the entire file speaks as required.

jest.config.int.js
```diff
@@ -1,4 +1,9 @@
 var config = require('./jest.config.base.js');
 config.testMatch = ['**/?(*.)+(int).(spec|test).[jt]s?(x)'];
 config.setupFilesAfterEnv = ['<rootDir>/src/test/integrationSuiteSetup.ts'];
+config.globalSetup = '<rootDir>/src/test/integrationGlobalSetup.ts';
+config.globalTeardown = '<rootDir>/src/test/integrationGlobalTeardown.ts';
+// Each worker uses up to 11 PostgreSQL connections (10 pool + 1 admin).
+// Cap at 9 workers to stay within PostgreSQL's default max_connections of 100.
+config.maxWorkers = 9;
```
hminsky2002, line 8: blazing fast local run with this! i noticed one test had failed due to a socket hang up, but that seemed like a nondeterministic failure, and on re run all tests passed.

hminsky2002 (approved):
All tests ran successfully in under a minute locally, marked improvement over the usual flow. The gains in github actions seem huge too. Great work @slifty !

.env.example
slifty: It says `must` a few times in the documentation -- what would you suggest changing specifically to make it more explicit?

jest.config.int.js
slifty: If the socket hangups become common we should make an issue / flag it (we don't want flaky test runs) but for now it sounds like we don't need to block on it.

.env.example
slifty: (For now I'll hit merge since approved, but we can improve later!)

PR #2327, opened 2026-04-09
Add `createdby` field to more entities

### LINKED ISSUE #1860: Add `createdBy` field to more entities
While doing something else, I noticed we don't have `createdBy` on several proposal-related fields. It may be nice to have this data. We have some audit logs so for audit purposes this is low priority.

* Opportunities
* Sources
* Funders
* Data Providers
* Changemakers

yshng:
This PR adds a new `createdBy` attribute to `Opportunity,` `Funder,` `Changemaker,` `DataProvider,` and `Source.` This doesn't affect required fields from end users since this is populated based on auth context, but will add to the returned object.

Resolves #1860

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `createdBy` field to funders, sources, opportunities, changemakers, and data providers.
 - `BaseField` can now be of type `date` and `date_time`.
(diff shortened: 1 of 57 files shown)


slifty:
Mentioned this in zulip but to document it here: this is a large PR (hooray!) and it would be super helpful for reviewing + git history if the changes related to each particular entity could be isolated in an individual commit.

yshng:
@slifty Your wish is Claude's command! Everything should be split up now.

slifty (approved):
Looks great!  Thank you!

PR #2328, opened 2026-04-09
Fix bundle counts

### LINKED ISSUE #895: Fix bundle counts / totals
Several of our endpoints that return collections return a `Bundle` which includes a total number of entries (not just the amount returned).  As we begin to add things like filters and user authentication this total is going to start to get a little confusing.

Currently it returns the true total number of entries in the database, NOT the total that would be available based on the filters provided.  For instance, this would expose the total number of bulk uploads across ALL users, rather than just the total number of uploads for a given user.

I think this is probably an incorrect implementation.

slifty:
This PR fixes a long-standing bug where collection endpoints would return a generally inaccurate `total` value.

Resolves #895


src/database/queries/proposals/selectWithPagination.sql
```diff
@@ -1,66 +1,84 @@
...
+			)
+		GROUP BY proposals.id
+		ORDER BY proposals.id DESC
+	),
+
+	entry_count AS (
+		SELECT count(*) AS total FROM candidate_entries
+	),
+
+	paginated_entries AS (
```
hminsky2002, line 69: Yep this seems like the key fix. Elegant!

hminsky2002 (approved):
Tests pass locally and everything is done according to the new pattern. Woohoo!

slifty:
Oh good catch by copilot -- fixing that up now.

slifty:
I am not going to manually hit "resolve" on the 500 messages from copilot.

PR #2331, opened 2026-04-14
Ignore major eslint versions

slifty:
Eslint has a large network of ecosystem dependencies, and it often takes months for those dependencies to update to support major changes to the core eslint package.

Rather than get stuck with all lint dependencies being frozen when eslint has a major bump, let's manually perform those changes once the ecosystem has caught up.


hminsky2002 (approved):
Indeed the right call!

PR #2343, opened 2026-04-14
Ignore major `@eslint/*` dependency bumps

slifty:
This PR expands the `eslint` major dep ignore to include other official eslint packages (which tend to be co-released / require the now-ignored packages).


hminsky2002 (approved):
Sure sure, follows the same logic as #2331. Though this does mean that we need to keep track of these major version releases ourselves!

PR #2354, opened 2026-04-16
Use `Id` consistently

### LINKED ISSUE #2015: Use the `Id` schema (and type) in more places
`Id` and `integer` are similar, but Id provides some extra rules (e.g. >0, < MAX_INT).

We should use it consistently in our OpenAPI specification as well as in our types.

slifty:
This PR improves our internal and documented types to use the `Id` schema / type consistently.  It also fixes a documentation error where base field localizations were saying short codes were numbers.

Resolves #2015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
 
+### Fixed
+
+- `baseFieldShortCode` path parameters in base field localization endpoints were incorrectly typed as `integer` instead of `shortCode`.
+
 ## 0.33.0 2026-04-14
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
 
+### Changed
+
+- All ID fields in the OpenAPI specification now reference a shared `id` schema with constraints (minimum: 1, maximum: 2147483647) instead of using bare `integer` types.
+
 ### Fixed
 
 - `baseFieldShortCode` path parameters in base field localization endpoints were incorrectly typed as `integer` instead of `shortCode`.
--- a/src/openapi/api.json
+++ b/src/openapi/api.json
@@ -65,6 +65,9 @@
 			"BulkUploadTaskLogDetails": {
 				"$ref": "./components/schemas/BulkUploadLogDetails.json"
 			},
+			"id": {
+				"$ref": "./components/schemas/id.json"
+			},
 			"keycloakOrganizationId": {
 				"$ref": "./components/schemas/keycloakOrganizationId.json"
 			},
--- a/src/openapi/components/schemas/ApplicationFormFieldPermissionGrant.json
+++ b/src/openapi/components/schemas/ApplicationFormFieldPermissionGrant.json
@@ -9,8 +9,7 @@
 					"const": "applicationFormField"
 				},
 				"applicationFormFieldId": {
-					"type": "integer",
-					"description": "The ID of the application form field.",
+					"$ref": "./id.json",
 					"example": 500
 				},
 				"scope": {
--- a/src/openapi/components/schemas/BulkUploadPermissionGrant.json
+++ b/src/openapi/components/schemas/BulkUploadPermissionGrant.json
@@ -9,8 +9,7 @@
 					"const": "bulkUpload"
 				},
 				"bulkUploadTaskId": {
-					"type": "integer",
-					"description": "The ID of the bulk upload task.",
+					"$ref": "./id.json",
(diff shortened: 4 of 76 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
 
+### Fixed
+
+- `baseFieldShortCode` path parameters in base field localization endpoints were incorrectly typed as `integer` instead of `shortCode`.
```
bickelj, line 21: This is only in the OpenAPI doc, right? Not the actual code. In other words, maybe the comment here can clarify that the code didn't change.

src/openapi/components/schemas/BulkUploadPermissionGrant.json
```diff
@@ -9,9 +9,7 @@
 					"const": "bulkUpload"
 				},
 				"bulkUploadTaskId": {
-					"type": "integer",
-					"description": "The ID of the bulk upload task.",
-					"example": 800
```
bickelj, line 14: I like to keep a distinct-per-entity but consistent-for-same-entity example around, but that's me being nitpicky.

src/openapi/paths/changemakerFiscalSponsor.json
```diff
@@ -37,11 +37,11 @@
...
 							"properties": {
 								"fiscalSponseeChangemakerId": {
-									"type": "integer",
+									"allOf": [{ "$ref": "../components/schemas/id.json" }],
 									"example": "43"
 								},
 								"fiscalSponsorChangemakerId": {
-									"type": "integer",
+									"allOf": [{ "$ref": "../components/schemas/id.json" }],
 									"example": "42"
```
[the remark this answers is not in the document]
bickelj: Good call!

src/openapi/api.json
```diff
@@ -65,6 +65,9 @@
 			"BulkUploadTaskLogDetails": {
 				"$ref": "./components/schemas/BulkUploadLogDetails.json"
 			},
+			"id": {
+				"$ref": "./components/schemas/id.json"
```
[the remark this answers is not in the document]
bickelj: It's not max uint 32, it's max int 32 due to PostgreSQL's handling of [numeric types](https://www.postgresql.org/docs/18/datatype-numeric.html).

bickelj (approved):
Overall I see nothing objectionable and it is most certainly an improvement, thus the approval. Copilot has some handy comments. And if we don't define the Id as being maximum (signed) int32 value, I think we should clarify it to be that because PostgreSQL is what generates these (https://www.postgresql.org/docs/18/datatype-numeric.html). It would also be nice to keep some of the example values unless they conflict with other example values. The nice thing about the example values is that OpenAPI uses them to generate bigger examples and if we don't specify them then we may get incongruent example bodies (such as ApplicationFormFieldID=2 then inside the ApplicationForm it has ID=3).

src/openapi/components/schemas/ApplicationFormFieldPermissionGrant.json
```diff
@@ -9,9 +9,7 @@
 					"const": "applicationFormField"
 				},
 				"applicationFormFieldId": {
-					"type": "integer",
-					"description": "The ID of the application form field.",
```
yshng, line 13: Curious why the removal of all these description lines

yshng (approved):
Not sure why the descriptions have been removed from some of these fields, but otherwise looks good to me!

CHANGELOG.md
slifty: Correct

src/openapi/components/schemas/BulkUploadPermissionGrant.json
slifty: Neat I didn't know that pattern -- I wonder if there is a way to make this semantically explicit (e.g. defining `EXAMPLE_BULK_UPLOAD_TASK_ID` somewhere and having it used for all possible references to that kind of thing...

Going to take a glance.

src/openapi/components/schemas/ApplicationFormFieldPermissionGrant.json
slifty: I will do a glance -- I do think they're not particularly meaningful descriptions `applicationFormFieldId is the Id of the application form field` is really just re-arranging the words of the attribute name!

src/openapi/components/schemas/BulkUploadPermissionGrant.json
slifty: My hopeful idea is not actually a thing, unfortunately.

Adding the examples back now!

We might want to consider moving the examples away from where properties are defined and instead having a schema level `example` that does everything at once, but that's a separate idea.

slifty: In the interest of not expanding this PR too much what I think I'm going to do is use the `allOf` (keeping the examples as written), and then open a PR or issue for doing that example refactoring.

src/openapi/components/schemas/ApplicationFormFieldPermissionGrant.json
slifty: Going to keep the descriptions removed for now (we may want to do a larger pass on our descriptions overall), just because I think they really did tend to be redundant.

PR #2356, opened 2026-04-16
Support extensions in phone number validation

### LINKED ISSUE #2151: Support `ext` in phone numbers
The `phone` type's validator does not appear to support the concept of an extension.  We should add that support to the validator.

yshng:
This adds support for extensions in phone numbers by validating the base number only, without the extension, by stripping off any sequence that begins with a letter from the end of a phone number (which can only contain punctuation, whitespace, and numbers). The stored base field value is unchanged. 

Overall validation documentation will be added in a separate PR to follow this one, to cover decisions made thus far. 

Resolves #2151 Support ext in phone numbers

### DIFF
--- a/src/__tests__/fieldValidation.unit.test.ts
+++ b/src/__tests__/fieldValidation.unit.test.ts
@@ -58,19 +58,16 @@ describe('field value validation against BaseFieldDataType', () => {
 	});
 	test('validate a valid phone number as PHONE_NUMBER', () => {
 		expect(
-			fieldValueIsValid('18005555555', BaseFieldDataType.PHONE_NUMBER),
+			fieldValueIsValid('+18005555555', BaseFieldDataType.PHONE_NUMBER),
 		).toBe(true);
 		expect(
 			fieldValueIsValid('+1(800)-555-5555', BaseFieldDataType.PHONE_NUMBER),
 		).toBe(true);
 		expect(
-			fieldValueIsValid('800-555-5555', BaseFieldDataType.PHONE_NUMBER),
-		).toBe(true);
-		expect(
-			fieldValueIsValid('800-555-5555', BaseFieldDataType.PHONE_NUMBER),
+			fieldValueIsValid('+1800-555-5555', BaseFieldDataType.PHONE_NUMBER),
 		).toBe(true);
 	});
-	test('validate an invalid phone number as PHONE_NUMBER', () => {
+	test('validate non-phone number strings as PHONE_NUMBER', () => {
 		expect(
 			fieldValueIsValid(
 				'112345678901234567890',
@@ -83,6 +80,68 @@ describe('field value validation against BaseFieldDataType', () => {
 		expect(fieldValueIsValid('      ', BaseFieldDataType.PHONE_NUMBER)).toBe(
 			false,
 		);
+	});
+	test('validate otherwise valid phone number without +country code as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid('18005555555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
+		expect(
+			fieldValueIsValid('8005555555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
+	});
+	test('validate a valid phone number with no extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid('+18005555555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+	});
+	test('validate a phone number with visual markers as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid('+1 (800) 555-5555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+		expect(
+			fieldValueIsValid('+1800-555-5555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+	});
+	test('validate a valid phone number with ;ext= extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid(
+				'+1 (800) 555-5555;ext=100',
+				BaseFieldDataType.PHONE_NUMBER,
+			),
+		).toBe(true);
+	});
+	test('validate a valid phone number with comma extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid(
+				'+1 (800) 555-5555,100',
+				BaseFieldDataType.PHONE_NUMBER,
+			),
+		).toBe(true);
+		expect(
+			fieldValueIsValid('+1800-555-5555,,2', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+	});
+	test('validate an invalid phone number with valid extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid('(800) 555-5555,,090', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
+		expect(
+			fieldValueIsValid(
+				'1(800)555-55555;ext=5',
+				BaseFieldDataType.PHONE_NUMBER,
+			),
+		).toBe(false);
+	});
+	test('validate a valid phone number with invalid extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid('+18005555555;;5', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
+		expect(
+			fieldValueIsValid('+18005555555x254', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
+		expect(
+			fieldValueIsValid('+18005555555,', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
 	});
 	test('validate a valid boolean as BOOLEAN', () => {
 		expect(fieldValueIsValid('true', BaseFieldDataType.BOOLEAN)).toBe(true);
--- a/src/fieldValidation.ts
+++ b/src/fieldValidation.ts
@@ -34,12 +34,34 @@ const isString = ajv.compile({
 	type: 'string',
 });
 
+const splitPhoneNumberAndExtensionCandidates = (
+	value: string,
+): [baseNumber: string, extension: string | undefined] => {
+	const match = /^(?<baseNumber>.+?)(?:(?:,+|;ext=)(?<extension>.*))?$/v.exec(
+		value,
+	);
+	return [match?.groups?.baseNumber ?? '', match?.groups?.extension];
+};
+
+const isValidBasePhoneNumber = (candidate: string): boolean =>
+	candidate.trim() === candidate &&
+	validator.isMobilePhone(candidate, 'any', { strictMode: true });
+
+const isValidExtension = (candidate: string): boolean =>
+	/^\d+$/v.test(candidate);
+
 // The validator package has only one phone number validator function,
 // 'isMobilePhone.' but the PDC does not currently have any requirement
 // for a phone number to be a mobile number, nor does the function seem
 // to discriminate on landline numbers.
-const isPhoneNumberString = (value: string): boolean =>
-	validator.isMobilePhone(value);
+const isPhoneNumberString = (value: string): boolean => {
+	const [baseNumberCandidate, extensionCandidate] =
+		splitPhoneNumberAndExtensionCandidates(value);
+	return (
+		isValidBasePhoneNumber(baseNumberCandidate) &&
+		(extensionCandidate === undefined || isValidExtension(extensionCandidate))
+	);
+};
 
 const isCurrencyWithCodeString = (value: string): boolean => {
 	const [currency, code] = value.split(' ');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add extension support in phone number validation.
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
 


src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -67,7 +67,19 @@ describe('field value validation against BaseFieldDataType', () => {
...
+		).toBe(true);
+		expect(
+			fieldValueIsValid('(800) 555-5555x100', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+		expect(
+			fieldValueIsValid('1800-555-5555EXT2', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+		expect(
+			fieldValueIsValid(
+				'800-555-5555    cow20',
```
yshng: As implemented right now, we'll allow anything to follow the number, so long as the number itself is valid

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -67,7 +67,19 @@ describe('field value validation against BaseFieldDataType', () => {
 			fieldValueIsValid('800-555-5555', BaseFieldDataType.PHONE_NUMBER),
 		).toBe(true);
 		expect(
-			fieldValueIsValid('800-555-5555', BaseFieldDataType.PHONE_NUMBER),
```
yshng, line 70: same test as above

src/fieldValidation.ts
```diff
@@ -38,8 +38,19 @@ const isString = ajv.compile({
...
+// We validate just the phone number part if a number comes with an
+// extension.
+const isPhoneNumberString = (value: string): boolean => {
+	const letterIndex = value.search(/[a-zA-z]/v);
+	const NO_LETTER_FOUND = -1;
+	if (letterIndex === NO_LETTER_FOUND) {
+		return validator.isMobilePhone(value);
+	}
+	const START_INDEX = 0;
+	const baseNumber = value.slice(START_INDEX, letterIndex).trimEnd();
```
yshng: trimming whitespace off the end only, as extensions may be preceded by a space. However, can't slice based on spaces because there can be a range of number of spaces in a valid telephone number, per https://github.com/validatorjs/validator.js/blob/master/src/lib/isMobilePhone.js. So split before first letter and trim any spaces off the end. If there are leading spaces, the number is still invalid, in line with existing validation which doesn't do any whitespace trimming.

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -67,7 +67,19 @@ describe('field value validation against BaseFieldDataType', () => {
 			fieldValueIsValid('800-555-5555', BaseFieldDataType.PHONE_NUMBER),
```
bickelj: I missed the news that we label '800-555-5555' as valid. I think the point of our `isValid` flag is strict validation, so I am surprised we allow this. Because we allow anything in the string and we're trying to help out machine clients here. Only the `+18005555555` would be right, right?

yshng: @bickelj 
Strict mode was not previously on, but I can turn it on, which would require the `+` I believe. All the other punctuation is allowed by validator.js: https://github.com/validatorjs/validator.js/blob/master/src/lib/isMobilePhone.js

Then users can expect the `+` and normalize the numbers on their own?

bickelj: TLDR: I retract my objection to the phone number part. We can still be strict with extensions.

My late-in-the-afternoon-drive-by comment yesterday is mildly related to this pull request in the sense that strict may help. If we required a strict phone number such as `+18005551212` (no spaces, no hyphens, no parentheses) on the left side, then it would be easier for (a) extension validation to split on the first space and (b) users of the phone number "isValid" attribute to split on the first space.

Using an existing library is good, though, so I can't object to that. The deciding factor is whether the telephone number can be used straightaway in HTML such that a phone browser can click it to make a call. Looking at https://html.spec.whatwg.org/multipage/input.html#telephone-state-(type=tel) and https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/tel those are for telephone inputs, not links. The link thing is kinda-sorta standard https://stackoverflow.com/questions/1164004/how-to-mark-up-phone-numbers#answer-1467262 which leads to https://www.rfc-editor.org/rfc/rfc5341 and https://www.rfc-editor.org/rfc/rfc3966. So [URI Syntax](https://www.rfc-editor.org/rfc/rfc3966#section-3) says `   extension            = ";ext=" 1*phonedigit`.

The real question, of course, is not what a purported standards document says, but what the de facto standard is among mobile browsers. [Oh dear](https://stackoverflow.com/questions/9482633/how-do-i-include-extensions-in-the-tel-uri#answer-67810744) it seems they do not do a great job with the semicolon. Still, I think we could only allow either the "official" ...`;ext=\d+$` or the unofficial "pause" character(s?) ...`,\d+$` patterns.

Sorry to bikeshed here :sweat_smile: but it seems inevitable for such a thing.

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -83,6 +95,12 @@ describe('field value validation against BaseFieldDataType', () => {
 		expect(fieldValueIsValid('      ', BaseFieldDataType.PHONE_NUMBER)).toBe(
 			false,
 		);
+		expect(
```
bickelj, line 85: This and the below test should be separate tests so that we can get finer-grained test-failure feedback.

src/fieldValidation.ts
bickelj: I don't think we should trim, I think the `isValid` flag being false can be a clue to the changemaker/dataprovider/funder that they have trailing whitespace in their data. We're not going to reject the data, but on the front-end/UI user side we want to drop in the data without needing to process it when it `isValid`.

bickelj (changes requested):
A couple of tests should be split out instead of combined. I think we can be  strict on what a valid extension looks like, allowing both the RFCs' format(s) and whatever works in a phone browser, but not more than that. If others think the looser format should be marked `isValid`, I am not going to die on this hill 😆

src/fieldValidation.ts
yshng: I see what you mean about not having to process the data, does this mean, for example, that dates will be presented as full RFC formatted timestamps on the front-end/UI though? What's the difference between pulling out relevant parts of the date and pulling out the base number given a clear rule for separating it? I think this will shake out more clearly once we figure out what we want with respect to overall validation below.

src/__tests__/fieldValidation.unit.test.ts
yshng: I would guess machine users will still want to display link text even if they are providing a `href` `tel:` attribute, in which case they will be happy to have the conventional punctuation, should the original data contain it to begin with. If we went with that for the base number, then it doesn't make sense to validate the extension strictly with the pause character since users will not be plopping the data straight into a `tel` attribute, though they may pop it straight in between `a` tags.

bickelj: I think machine users want to directly plop the data into a `tel:` link when it `isValid`. I was wrong about strict vs non-strict with respect to base number in the RFCs/browsers. I mean I now think we were already strict with base numbers and these can in fact be dropped into a `tel:` link and work fine. I think we want the extension to have the same property: dropping it into a `tel:` link should "just work" for front-end consumers of the data.

src/__tests__/fieldValidation.unit.test.ts
yshng: I have changed this to accept only extensions formatted as:

1. (optional space) string of letters, string of digits (only in earlier commit)
2. one or more commas, string of digits 
3. one semicolon, string of digits

If there's anything more than that following the base number, including whitespace, the data is invalid.

src/fieldValidation.ts
yshng: No longer trimming whitespace, see above comment.

yshng:
Okay, after some processing on my part, I have now basically implemented what @bickelj  said, including strict mode on telephone numbers (must lead with `+` country code). That should be easy enough to remove if we don't want it. I did include an extra commit so there is an improved version where `ext` type letter strings are allowed, with only one leading space. But the final commit does not allow extensions of that type. Will squash upon approval if we go with the latter.

src/fieldValidation.ts
```diff
@@ -38,8 +38,38 @@ const isString = ajv.compile({
 // 'isMobilePhone.' but the PDC does not currently have any requirement
 // for a phone number to be a mobile number, nor does the function seem
 // to discriminate on landline numbers.
-const isPhoneNumberString = (value: string): boolean =>
-	validator.isMobilePhone(value);
+const isPhoneNumberString = (value: string): boolean => {
+	// Get a valid extension string out if one exists, consisting of an
```
slifty: We aren't consistent about this (we tend to not have comments overall for better or worse), but this function already has a comment block above it; could we incorporate this comment into that comment block instead of having have the function documentation broken up by the parameter definition?

All that said -- I think that we could remove the need for a comment entirely (and the later comments) by leaning more on an up-front regex that splits the string along "{ phone candidate }{ extension candidate }" lines, and then uses helper functions to validate each candidate string in isolation. 

Specifically

- Step 1: split the string into (phone number candidate, extension candidate) using a regex.
- Step 2: validate the phone number candidate => via named helper function
- Step 3: validate the extension candidate (including "no extension) => via named helper function
- Step 4: return true if both candidates are valid

For instance I would something like this is self documenting and isolates the steps:

```

const splitPhoneNumberAndExtensionCandidates = (
        value: string,                                                                                                                                        
  ): [baseNumber: string, extension: string | undefined] => {
       const { baseNumber, extension } =                                                                                                                     
                value.match(/^(?<baseNumber>.+?)(?:(?:,+|;)(?<extension>.*))?$/)?.groups ?? {};   
        return [baseNumber ?? '', extension];                                                                                                                 
  };

  const isValidBasePhoneNumber = (candidate: string): boolean =>       
        candidate.trim() === candidate &&
        validator.isMobilePhone(candidate, 'any', { strictMode: true });

  const isValidExtension = (candidate: string): boolean =>
        /^\d+$/.test(candidate);                                                                                                                              

  const isPhoneNumberString = (value: string): boolean => {
        const [baseNumberCandidate, extensionCandidate] =
                splitPhoneNumberAndExtensionCandidates(value);                                                                                                
        return (                                                       
                isValidBasePhoneNumber(baseNumberCandidate) &&
                (extensionCandidate === undefined || isValidExtension(extensionCandidate))                                                                    
        );
  };   
```

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -58,16 +58,24 @@ describe('field value validation against BaseFieldDataType', () => {
...
+	});
+	test('validate a valid phone number with extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid(
+				'+1 (800) 555-5555;100',
+				BaseFieldDataType.PHONE_NUMBER,
+			),
 		).toBe(true);
 		expect(
-			fieldValueIsValid('800-555-5555', BaseFieldDataType.PHONE_NUMBER),
```
bickelj, line 70: I think this old test can be kept (with the inserted `+` of course) and also the new one that replaced it.

src/__tests__/fieldValidation.unit.test.ts
bickelj: I think each of these `expect`s can be in their own `test` block. There is not much overhead for unit tests, so we can get the finer-grained feedback when each fails.

src/fieldValidation.ts
```diff
@@ -38,8 +38,38 @@ const isString = ajv.compile({
...
+	// extension separator and a string of digits
+
+	// Possible extension separators are:
+	// 1. one or more commas
+	// 2. a semicolon
```
bickelj: The commas are a "pause" so I think it's OK to keep that as-is, but for the semi-colon, since that's RFC-related, the semi-colon should match this:
`   extension            = ";ext=" 1*phonedigit`

So either of these would be valid:
* `+18005551212,,123`
* `+18005551212;ext=123`

But this would be invalid:
* `+18005551212;123`

Anyway, I feel I have given enough feedback on this one and I'm curious about others' opinions 😅

src/fieldValidation.ts
```diff
@@ -38,8 +38,38 @@ const isString = ajv.compile({
...
+
+	const match = /(?:,+|;)\d+/v.exec(value);
```
slifty: match is a generic variable name which is a missed opportunity to name it something more meaningful.  You also could define the regular expression here in a named constant which gives it meaning as well.

This may be moot if you refactor based on the above feedback, but even then if you're dealing with matches and patterns it's nice to use meaningful names

bickelj (approved):
@yshng I think this is an improvement overall and I have given plenty of feedback, so I am mashing Approve. That said, I do have some more feedback above, the most important of which is to hear feedback from other teammates.

That means, ahem, @slifty @hminsky2002

src/fieldValidation.ts
yshng: Sorry that’s poor reading comprehension on my part! @bickelj

src/fieldValidation.ts
```diff
@@ -38,8 +38,38 @@ const isString = ajv.compile({
...
+	}
+
+	const [ext] = match;
+	const { index: extStartIndex } = match;
+
+	const START_INDEX = 0;
+	const baseNumber = value.slice(START_INDEX, extStartIndex);
+	const afterBaseNumber = value.slice(extStartIndex);
+
+	// Make sure no additional nonsense
```
slifty: Let's see if we can't cover this kind of thing with the use of a good regex up front

slifty (changes requested):
I have a proposal for a more self-documenting implementation that also leans into regex / isolation of concerns via functions in order to avoid as much special case detection in the validator.

Importantly though, we'll need a changelog entry for this!  There are two breaking changes: (1) some strings that were valid before are no longer valid due to the new use of strict mode, and (2) some strings that were not valid before are now valid due to the support for extensions.

We also must document the expectation of what a valid phone number means for clients, specifically that they now need to split these strings by { one or more commas or a semicolon } and treat them as two halves.

src/fieldValidation.ts
yshng: Yes, this makes sense, and I will do this. Validation documentation to come in follow-up PR.

slifty: > Validation documentation to come in follow-up PR.

Understood -- I forgot we didn't already have that (if we did have it already, then we would want to update that documentation to note the new feature as part of the commit the feature was added!  But like you note: we don't so won't!)

hminsky2002 (commented):
I think @slifty has the right idea with breaking the validation up in to self-describing functions, and catching up a bit here think that having the validation scheme documented explicitly somewhere is a really key idea! And I can make an issue to link that information explicitly on the bulk-up loader interface. 

I remember when we first set this up, in lieu of maintaining our own regex scheme for validating the phone number, the philosophy was that it was better to invoke a library that was responsible for maintaining the definition of the type. I think as it's been laid out here it's reasonable to mix and match, but it does get me thinking that a good sign of a validation scheme is that someone has already made a validator that supports it. I did some research and it seems like the most used existing phone number validators are `validator` (woohoo) and google's `[libphonenumber](https://github.com/google/libphonenumber)`. Both of those are more permissive for validating phone number than we're going here with the extension handling. 

Still I think that we aren't making too much of a reach here, and the core phone number validation we're using is still solid. Just re emphasizing that we should try to avoid becoming arbiters of any kind of protocol definition.

yshng:
See #2359 for validation documentation.

src/fieldValidation.ts
```diff
@@ -34,12 +34,34 @@ const isString = ajv.compile({
 	type: 'string',
 });
 
+const splitPhoneNumberAndExtensionCandidates = (
+	value: string,
+): [baseNumber: string, extension: string | undefined] => {
+	const match = /^(?<baseNumber>.+?)(?:(?:,+|;ext=)(?<extension>.*))?$/v.exec(
+		value,
+	);
+	return [match?.groups?.baseNumber ?? '', match?.groups?.extension];
```
hminsky2002, line 43: To me, it would make more sense for base number to coalesce into an undefined value than the empty string, and then we just reject it inside the isValidBasePhoneNumber return value. I know we're separating the stripping from the validation here, but still seems like if match.groups.baseNumber is undefined, maybe the whole function should just return undefined or null.

hminsky2002 (approved):
Nice work! Looks good to me overall, and incorporates @slifty and @bickelj 's feedback effectively in my opinion. Left a note about my thoughts on the organization of the functions inside the validator, but take em or leave em, not dealbreaking to me!

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -84,6 +81,55 @@ describe('field value validation against BaseFieldDataType', () => {
...
+		).toBe(false);
+		expect(
+			fieldValueIsValid('8005555555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(false);
+	});
+	test('validate a valid phone number with no extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid('+1 (800) 555-5555', BaseFieldDataType.PHONE_NUMBER),
+		).toBe(true);
+		expect(
```
bickelj, line 101: I still think this should be in a `test('validate a hyphenated phone number with no extension') {`...

src/__tests__/fieldValidation.unit.test.ts
```diff
@@ -84,6 +81,55 @@ describe('field value validation against BaseFieldDataType', () => {
...
+		).toBe(true);
+	});
+	test('validate a valid phone number with valid extension as PHONE_NUMBER', () => {
+		expect(
+			fieldValueIsValid(
+				'+1 (800) 555-5555;ext=100',
+				BaseFieldDataType.PHONE_NUMBER,
+			),
+		).toBe(true);
+		expect(
```
bickelj, line 120: Likewise this should be in a `test('validate a hyphenated phone number with comma extention') {`

bickelj (approved):
I still think that unit tests should, where practical, have a single assertion. Another way of thinking of it is that each assertion that can be separated easily should be separated into its own test. There are more places than the one I suggested.

That said, I think the yak is almost completely shorn at this point and we would be better off merging this and we'll be OK.

bickelj:
@yshng ~~Oh, a changelog entry is also still needed.~~ I see it now.

slifty (approved):
Thank you for the iteration!

src/__tests__/fieldValidation.unit.test.ts
yshng: I didn't realize that's what you were trying to call out specifically - happy to add.

src/__tests__/fieldValidation.unit.test.ts
yshng: I see the tests as building on each other. Since we've cleared the base number hurdle without extensions, I think we can stop separating them, but in this iteration, separate the type of extensions.

src/fieldValidation.ts
yshng: @hminsky2002  I get what you mean. I ended up leaving it returning an empty string because that's what `isMobilePhone()` wants in the end, so adding an `undefined` type didn't seem worth it on balance (perhaps you had a way of implementing it that didn't cause this snag?). More substantively, having an extension really is optional in a way that base number is not, so it would make sense to be okay with `undefined` there, but not with base number - we simply don't trigger ext validation when there is no extension.

yshng: Oh I see, do you mean just have `splitPhone...` return `false` ? - I do prefer keeping separation of concerns and not have a function potentially return a `boolean` or `array.` Maybe I still misunderstand what you meant.

PR #2362, opened 2026-04-21
Add permissions to `/sources` endpoints

### LINKED ISSUE #2361: Add `Source` permissions
We want to incorporate the permission model for the `Source` entity type.

Right now sources are visible to all users, and can be used by any user when creating data / selecting a source.

After this change we will have:

1. Users are not able to view sources unless they have `VIEW` permission.
2. Users are not able to create sources for a given organizational entity unless they have `CREATE` permission for the source scope on that entity.

The less straightforward decision is when can a user *use* a source?

`USE` is not a verb, and it isn't clear it should be.  That said, we could ensure that (1) the user has the right to view the source, and (2) the user has the right to create whatever entity is being created in the context of the source's organization.  For instance, if I have the right to view Foo Foundation's sources, and I have the permission to `create Proposal` entities in the context of Foo Foundation, then I would be allowed to create a proposal with Foo Foundation as a source.

I'm going to use that logic for now.

### LINKED ISSUE #2364: Add a new `reference` verb
Right now we have:

- View
- Create
- Edit
- Delete

But we're seeing in #2361 that we might want users to be able to see certain types of entity without being able to "use" or "reference" that entity in other places.

Adding a "reference" verb will cover that case.  It's immediately needed for source but I do expect it to also be useful in other cases.

slifty:
This PR applies the permission grant system to the creation + viewing of `Source` entities.

Prior to this change all sources were visible to all authenticated users.  After this change, a user needs to be given explicit visibility of a given source to see it, and they need to be given explicit permission to reference a given source when creating new data in the system.

This PR also adds a new `reference` verb to cover that second case.  I explicitly do NOT grant any users `reference` rights since prior to this PR the state was overly permissive (for the most part we don't want users to be able to reference our existing sources today / that should be manually set in the cases we want that permission.)

Resolves #2361
Resolves #2364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - All ID fields in the OpenAPI specification now reference a shared `id` schema with constraints (minimum: 1, maximum: 2147483647) instead of using bare `integer` types.
+- `GET /sources` and `GET /sources/:sourceId` now filter results by `view | source` permission; sources are no longer universally visible to authenticated users.
+- Source creation now requires `create | source` scope (previously required `edit | funder`, `edit | changemaker`, or `edit | dataProvider`). Existing permission grants with the parent entity's scope have been migrated to also include `source` scope, and new `create | source` grants have been created for users who previously had `edit` verb on the parent entity.
+- `POST /proposalVersions`, `POST /tasks/bulkUploads`, and `POST /changemakerFieldValueBatches` now require a `reference | source` grant on the supplied source. Grants can be made directly on the source or inherited from the source's funder, changemaker, or data provider. This is a breaking change: previous behavior allowed any authenticated user to supply any source, so existing users will need new `reference | source` grants before they can use these endpoints.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add extension support in phone number validation.
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
+- Added a new `reference` verb to the permission grant verb set. Reference is intended to gate whether a user may cite an entity as a pointer in data they are creating, separate from `view` (which only grants read access) and `create` (which grants the ability to author new instances of an entity type).
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.34.0 2026-04-23
+
 ### Added
 
 - Add extension support in phone number validation.
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -242,8 +242,9 @@ context key).
 |        |                    | Create bulk upload tasks for the funder's opportunities           |
 | edit   | opportunity        | Create or update application forms and fields for the funder      |
 | edit   | funder             | Create or update changemaker-proposal associations                |
-|        |                    | Create sources associated with the funder                         |
 | edit   | proposal           | Create or update proposal versions for the funder's proposals     |
+| view   | source             | View the funder's sources                                         |
+| create | source             | Create sources associated with the funder                         |
 | manage | funder             | View, send, and respond to funder collaborative invitations       |
 |        |                    | View collaborative members for the funder                         |
 
@@ -252,16 +253,17 @@ context key).
 Permissions granted against a changemaker (using the changemaker's `id` as the
 context key).
 
-| Verb | Scope              | What It Enables                                                    |
-| ---- | ------------------ | ------------------------------------------------------------------ |
-| view | changemaker        | View changemaker field values for the changemaker                  |
-| view | proposal           | View proposals associated with the changemaker                     |
-|      |                    | View proposal versions associated with the changemaker             |
-|      |                    | View changemaker-proposal associations for the changemaker         |
-| view | proposalFieldValue | View proposal field values for the changemaker's proposals         |
-| edit | changemaker        | Create or update changemaker field values                          |
-|      |                    | Create sources associated with the changemaker                     |
-| edit | proposal           | Create or update proposal versions for the changemaker's proposals |
+| Verb   | Scope              | What It Enables                                                    |
+| ------ | ------------------ | ------------------------------------------------------------------ |
+| view   | changemaker        | View changemaker field values for the changemaker                  |
+| view   | proposal           | View proposals associated with the changemaker                     |
+|        |                    | View proposal versions associated with the changemaker             |
+|        |                    | View changemaker-proposal associations for the changemaker         |
+| view   | proposalFieldValue | View proposal field values for the changemaker's proposals         |
+| edit   | changemaker        | Create or update changemaker field values                          |
+| edit   | proposal           | Create or update proposal versions for the changemaker's proposals |
+| view   | source             | View the changemaker's sources                                     |
+| create | source             | Create sources associated with the changemaker                     |
 
 ### Opportunity Permissions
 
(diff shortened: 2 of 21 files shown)


src/database/initialization/has_source_permission.sql
```diff
@@ -0,0 +1,72 @@
...
+			OR (
+				pg.context_entity_type = 'dataProvider'
+				AND pg.data_provider_short_code = s.data_provider_short_code
+				AND 'source' = ANY(pg.scope)
+			)
+			-- Inherited from changemaker with source scope
+			OR (
+				pg.context_entity_type = 'changemaker'
+				AND pg.changemaker_id = s.changemaker_id
+				AND 'source' = ANY(pg.scope)
```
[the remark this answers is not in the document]
slifty: Permission inheritence by scope is somewhat moot here since source only allows a `source` scope at this point anyway...

But I'll still make this change, and I suspect this will be revisited when we explore the "who can use a scope for relationships" issue.

CHANGELOG.md
```diff
@@ -11,10 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
+- Sources are now subject to permission checks. Viewing sources requires `view | source` permission and creating sources requires `create | source` permission. Permissions can be granted directly on a source or inherited from the source's funder, changemaker, or data provider context.
 
 ### Changed
 
 - All ID fields in the OpenAPI specification now reference a shared `id` schema with constraints (minimum: 1, maximum: 2147483647) instead of using bare `integer` types.
+- `GET /sources` and `GET /sources/:sourceId` now filter results by `view | source` permission; sources are no longer universally visible to authenticated users.
+- Source creation now requires `create | source` scope (previously required `edit | funder`, `edit | changemaker`, or `edit | dataProvider`). Existing permission grants with the parent entity's scope have been migrated to also include `source` scope, and new `create | source` grants have been created for users who previously had `edit` verb on the parent entity.
```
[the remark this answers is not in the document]
slifty: This is accurate -- I'm waiting on guidance about whether visibility is actually something we plan to restrict before moving forward in either direction.

Once I have clarity there then I'm either going to need to make "visibility is required for use" explicit, or look into some other requirement.

CHANGELOG.md
```diff
@@ -11,10 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `POST /proposalVersions`, `POST /tasks/bulkUploads`, and `POST /changemakerFieldValueBatches` now require a `reference | source` grant on the supplied source. Grants can be made directly on the source or inherited from the source's funder, changemaker, or data provider. Existing grants that previously implied the ability to use sources in these flows have been migrated to include the `reference` verb so that users retain that ability.
```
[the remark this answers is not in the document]
slifty: Ah right -- I explicitly removed that migration but forgot to fix the changelog.

src/database/migrations/0108-add-reference-to-permission_grant_verb_t.sql
```diff
@@ -0,0 +1,13 @@
...
+	'edit',
+	'delete',
+	'manage',
+	'reference'
+);
+ALTER TABLE permission_grants
+ALTER COLUMN verbs TYPE updated_permission_grant_verb_t [] USING
+verbs::text []::updated_permission_grant_verb_t [];
+DROP TYPE permission_grant_verb_t CASCADE;
+ALTER TYPE updated_permission_grant_verb_t RENAME TO permission_grant_verb_t;
```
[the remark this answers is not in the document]
slifty: Huh... I didn't think that was actually possible which is why we used this approach... If so, that's much better.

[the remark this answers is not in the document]
slifty: *ah ha* -- when adding a value we can use ALTER TYPE but when removing we need to use this approach!  Updating this migration.

CHANGELOG.md
bickelj, line 8: I think this PR should bump the minor version within the PR.

CHANGELOG.md
```diff
@@ -9,13 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add extension support in phone number validation.
 - Add changelog to documents that auto-publish to WordPress.
 - `GET /funders` now supports the `isCollaborative` query parameter to filter funders by their collaborative status.
+- Sources are now subject to permission checks. Viewing sources requires `view | source` permission and creating sources requires `create | source` permission. Permissions can be granted directly on a source or inherited from the source's funder, changemaker, or data provider context.
+- Added a new `reference` verb to the permission grant verb set. Reference is intended to gate whether a user may cite an entity as a pointer in data they are creating, separate from `view` (which only grants read access) and `create` (which grants the ability to author new instances of an entity type).
```
bickelj, line 17: I think we decided that `create` implicitly grants `reference` on whichever entity one was granted `create` access. So one can (coarsely) grant `create`, say, to a funder's sources, or one can (finely) grant `reference` instead to all funder's sources or an individual source.

And therefore this text could be clarified to reflect that, or maybe could be simplified to say "Added a new `reference` verb, see the permissions document for details."

docs/PERMISSIONS.md
```diff
@@ -176,18 +176,24 @@ The PDC permission system uses four concepts:
...
-| delete | Delete data                                       |
-| manage | Manage permission grants associated with the data |
+| Verb      | Description                                         |
+| --------- | --------------------------------------------------- |
+| view      | Read access to data                                 |
+| create    | Create new data                                     |
+| edit      | Modify existing data                                |
+| delete    | Delete data                                         |
+| manage    | Manage permission grants associated with the data   |
+| reference | Use an entity as a pointer in data you are creating |
```
bickelj, line 186: We agreed that, if possible, we do not wish to expand the verb list. And we also agreed that in this case it makes sense to add this verb. It is because the permission we wish to grant doesn't neatly fit the other verbs and also it is a permission that can be applied in other contexts too.

src/database/initialization/has_source_permission.sql
```diff
@@ -0,0 +1,71 @@
...
+			)
+			-- Inherited from changemaker
+			OR (
+				pg.context_entity_type = 'changemaker'
+				AND pg.changemaker_id = s.changemaker_id
+			)
+		)
+		WHERE s.id = has_source_permission.source_id
+			AND has_source_permission.permission = ANY(pg.verbs)
+			AND has_source_permission.scope = ANY(pg.scope)
```
bickelj, line 47: At first I thought this `ANY` was looking for whether there was any verb or any scope, but no, it is looking for whether the exact permission and exact scope are within the list of granted verbs and granted scopes, respectively.

All good.

src/database/migrations/0108-add-source-scope-to-existing-permissions.sql
bickelj, line 1: On the one hand this file preserves some backward compatibility with regard to permissions. On the other hand, it hides some of the changes to permissions. Since we do not have very many grants in the system it is probably OK either way. I think I prefer this approach of preserving existing grants, so huzzah for backward compatibility!

bickelj (approved):
Dan and I spoke about this PR during office hours and we agreed that a new verb `reference` makes the most sense rather than attempting to re-use another verb.

Having skimmed this PR, I find nothing objectionable. (The reason I skim PRs now instead of actually trying them or diving deeply is I feel as though I was told to stop doing that, not because I think it's best practice to only read code as if that were sufficient to review code.)

I made a [meta-comment in Zulip for those who can see it](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/permissions/near/258886) regarding the danger of the permissions system as a whole, but that does not take away from the awesome work in this PR.

CHANGELOG.md
slifty: Currently this does describe the the implementation (you need explicit reference) -- which is in line with other permissions.  Generally we have been leaning into "explicit vs implicit" as much as possible for *first passes at least*.

I do still agree this would be reasonable to relax (e.g. "create gives reference rights") but since that could be handled separately and this implementation isn't incorrect, I'd rather stay "conservative" / maintain the explicit-is-preferred design philosophy since adding a verb in itself is already such a big change.

The "extra consideration" I'd want to take is if there are other places where implicit would make sense, and if we should make those changes too (for instance: right now on lots of entities a user could have edit rights without view rights, which is an interesting and probably silly concept.)

slifty:
All righty -- tested this locally a bit more, was able to create a reference permission and validate that it allowed a non-admin user to use the source, and also did not allow to use non-available (or non-existent) sources.

Issue #2368, opened 2026-04-23
Add support for "all" scopes and "all" verbs in permission grants


slifty:
We have a few use cases where we want to be able to give a given permission verb to "all" entities related to a given entity, or to give "all" verbs.

For example when someone creates an entity they should be given all permissions on all related to that entity.

I can think of two reasons an "all" verb / scope is useful as opposed to simply creating a permission with all the available verbs / scopes at the time of insert:

1. Convenience <-- this is not a reason in itself, but it's worth mentioning
2. Conveying intent.  This one matters because as we expand the permission system we will have additional scopes / entities.  As we add these we might not be able to necessarily assume that a user was intended to have all future permissions / scopes going forward simply by looking at a list that happens to have all current entries.  Are they the owner, or are they just well permissioned?  On the other hand, an "all" verb conveys the intent that they are supposed to be treated as an owner / admin of an entity and have their permissions expanded.

For the verb case "all" could also be "admin" or "own"; for scope I think it does need to be "all"

---

And finally from a verb perspective... we could simply have `manage` mean "all access to all scopes".  This might be the most reasonable approach, because manage does *ALREADY* mean that a user has the ability to grant all permissions on all entities to themselves, so functionally speaking a user with manage permissions on an entity has the capacity for full access to it and all child entities.

bickelj:
I like this idea, yes. Getting more nitpicky, if we replace "given all permissions on all related to that entity" with "given all permissions on all underneath that entity" I fully agree.

It would also be nice to extend this concept to users and groups, e.g. grant view permission on this entity to all users.

slifty:
Agreed yes, that is more accurate!

For the groups / users we have #1774 to catch that need (very much needed, agreed!)

Also I added an "edit" thinking about `manage` -- what do you think there?

bickelj:
We just spoke and I agree that `manage` can mean "all permissions at and under this entity". Save "all" for perhaps "scope" and/or other stuff.

slifty:
Ok just talked with @bickelj and I think this summarizes the plan:

1. `manage` as a verb already makes sense to be treated as admin / owner (no need for a rename, an no need for an `all` verb).  A user with manage permission can give themselves any permission for that entity OR subordinate entity.

2. `all` as a scope may make sense but let's do the manage change first and see what is still needed after that!

EDIT: ha jinx @bickelj

slifty:
[...]

PR #2370, opened 2026-04-23
Treat `manage` permission as "all"

slifty:
This PR expands `manage` so that it implicitly provides all verbs for the items it has been granted to.

I want to call out that "manage" does NOT affect permission inheritance (e.g. "manage opportunities of funder A" does not translate to "manage opportunities AND proposals associated with those opportunities"); an entity will still only have the permission applied to it according to scope.

Related to #2368

### DIFF
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -182,12 +182,20 @@ The permission system supports the following verbs:
 | create    | Create new data                                     |
 | edit      | Modify existing data                                |
 | delete    | Delete data                                         |
-| manage    | Manage permission grants associated with the data   |
+| manage    | Full permissions, including permission management   |
 | reference | Use an entity as a pointer in data you are creating |
 
 Note: The current implementation uses `edit` for both creation and modification
 operations in most contexts. This is a known semantic mismatch with the intended
 meaning of the verb.
+
+A `manage` grant is treated as satisfying any verb check on the same scope.
+A grant of `manage | proposal` on a funder context, for example, lets the
+grantee view, create, edit, delete, and reference proposals for that funder
+without naming each verb separately. `manage` does not expand the scope set;
+a grantee holding `manage | proposal` does not gain access to `funder`-,
+`opportunity`-, or `source`-scoped data unless those scopes are also
+included in the grant.
 
 The `reference` verb is separate from `view` and `create` so that permission to
 see an entity does not automatically imply permission to cite it from elsewhere,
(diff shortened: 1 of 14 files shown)


docs/PERMISSIONS.md
```diff
@@ -182,13 +182,22 @@ The permission system supports the following verbs:
...
 meaning of the verb.
 
+Because a grantee with `manage` can grant themselves any permission on the
+same scope, a `manage` grant is treated as satisfying any verb check on that
+scope. A grant of `manage | proposal` on a funder context, for example, lets
+the grantee view, create, edit, delete, and reference proposals for that
+funder without naming each verb separately. `manage` does not expand the
+scope set; a grantee holding `manage | proposal` does not gain access to
+`funder`-, `opportunity`-, or `source`-scoped data unless those scopes are
+also included in the grant.
```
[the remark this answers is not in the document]
slifty: Honestly the "because" is not useful to document; we don't need to frame a specific motivation (which, to this comment's point, may or may not be true) we can just document the decision and impact.

src/database/initialization/has_data_provider_permission.sql
```diff
@@ -14,14 +14,18 @@ BEGIN
...
+	-- via direct user grant or group membership. A granted 'manage' verb
+	-- satisfies any verb check.
 	SELECT EXISTS (
 		SELECT 1
 		FROM permission_grants pg
 		WHERE pg.context_entity_type = 'dataProvider'
 			AND pg.data_provider_short_code
 				= has_data_provider_permission.data_provider_short_code
-			AND has_data_provider_permission.permission = ANY(pg.verbs)
+			AND (
```
hminsky2002: Seeing all these having to be updated manually makes me wonder if there's any way we could generalize these further / abstract out some of the logic. I guess one advantage of our permissions system is that we have this level of control over how we handle each entities permissions checks. Still, if we weren't backed up by Claude here it seems like these kinds of 'across-the-board' updates would be prone to missing one (or more) queries.

hminsky2002 (approved):
Seems reasonable to me! Overall a small change to the codebase and executed correctly. Not that this PR prohibits inheritance, but its my intuition that users will want a 'just give X access to everything' option sooner rather than later.

src/database/initialization/has_data_provider_permission.sql
slifty: Inheritance is hard to generalize (the relationship a given entity has with other entities is unique to that entity's place in the ERD, and so it does have to be explicitly codified on a per-entity basis, I think).  So while it might be somehow possible, I'm not sure.

That said, I'm going to see what it would look like to DRY up the concept of "manage satisfies all verbs" with a SQL function of some type.  If that looks good I might ask for a quick re-review; if not I might just merge this as is :D

hminsky2002 (approved):
Haha this is awesome! Exactly what I was thinking. Also sets us up with a nice sql pattern for these types of across the board changes in the future.  One day we'll be able to reduce our entire codebase to a single generator. Thanks @slifty!

PR #2373, opened 2026-04-24
Fix function initializations

slifty:
This PR fixes our deployments by properly removing old sql functions before defining new ones.

### DIFF
--- a/src/database/initialization/has_changemaker_field_value_permission.sql
+++ b/src/database/initialization/has_changemaker_field_value_permission.sql
@@ -1,4 +1,6 @@
-CREATE OR REPLACE FUNCTION has_changemaker_field_value_permission(
+SELECT drop_function('has_changemaker_field_value_permission');
+
+CREATE FUNCTION has_changemaker_field_value_permission(
 	user_keycloak_user_id uuid,
 	user_is_admin boolean,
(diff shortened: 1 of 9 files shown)


yshng (approved):
Looks good to me!

hminsky2002 (approved):

bickelj (commented):
I don't think it will work:
{"length":241,"name":"error","severity":"ERROR","code":"42P13","hint":"Use DROP FUNCTION has_opportunity_permission(uuid,boolean,integer,permission_grant_verb_t,permission_grant_entity_type_t) first.","file":"pg_proc.c","line":"486","routine":"ProcedureCreate"},"data":null}},"msg":"Database failed to initialize"}

bickelj:
In other words, I think you need the list of args in the call to `DROP FUNCTION`. But we'll see.

Oh right, this is our (custom) function we usually call first, nevermind. It works great! :sweat_smile:

hminsky2002:
@bickelj where does that message come from out of useful curiosity? To test this i just ran npm run migrate and npm run start and verified the success of those.

bickelj:
@hminsky2002 That was from the deploy logs in the DO cloud interface.

PR #2375, opened 2026-04-27
Use `bundle` for `GET /baseFields`

### LINKED ISSUE #1545: Use `bundle` for `GET /baseFields`
This is the only collection endpoint that does not use the bundle / pagination system.  We should change that to make it consistent.

yshng:
Use `bundle` for `GET /baseFields` in the same manner as all other `GET` endpoints. Also removes old `loadBaseFields` operation. 

Breaking change.

Resolves #1545


bickelj:
I have only skimmed the code, not really read it, but I didn't see anything alarming there.

The more alarming thing is the breaking change without a fallback for clients expecting the full list of base fields. This includes internal tools in the data-scripts and front-end repos and more importantly Exchange and external integrations. In this case it seems worthwhile to provide backward compatibility with the old call deprecated so that clients have a chance to migrate without coordinating several parties releasing at the exact same moment. One way that could be done, I guess, is to add a handler that allows the old-style call (without pagination parameters) which calls the paginated back-end with max int32 as the max records, and then unwraps the response so that the body is the old style.

yshng:
@bickelj I was wondering about that - thanks for the suggestion of how to implement some backwards compatibility - will come back with something in this vein.

bickelj:
Both @kfogel and @slifty recalled and confirmed a decision to make it a breaking change. To reduce the negative impact [I announced the change to current users of the API](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/API.20Breaking.20Change.20Notifications/near/259195) as we decided today in a meeting.

bickelj:
Here is (or soon to be "was") the preview env, for some reason the action couldn't post the comment: https://pdc-service-preview-2375-wcak9.ondigitalocean.app/

I opened #2382 to attempt to overcome the probable permissions issue.

I am curious, though, if this was a one-off error or if it consistently happens.

bickelj:
Surprise! Even without #2382 preview environments worked on a second try. Perhaps #2382 is unneeded.

src/__tests__/baseFields.int.test.ts
```diff
@@ -468,8 +652,14 @@ describe('/baseFields', () => {
 					sensitivityClassification: BaseFieldSensitivityClassification.PUBLIC,
 				})
 				.expect(200);
-			const baseFields = await loadBaseFields(db);
-			expect(baseFields[0]).toEqual({
+			const baseFields = await loadBaseFieldBundle(
+				db,
+				null,
+				DEFAULT_SENSITIVITY_FILTER,
```
slifty: Does this test work if we don't pass any filter at all? I don't think we need to reference `DEFAULT_SENSITIVITY_FILTER` here at all so ideally we wouldn't, since I don't think this test is about default filters.  Rather, we should just define the filter we intend for the test to work.

src/database/operations/baseFields/loadBaseFieldBundle.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { ExpandedParameterFilter } from '../../parameters';
+import type {
+	BaseField,
+	BaseFieldSensitivityClassification,
+} from '../../../types';
+
+const loadBaseFieldBundle = generateLoadBundleOperation<
+	BaseField,
+	[
+		sensitivityFilter: ExpandedParameterFilter<BaseFieldSensitivityClassification>,
```
slifty, line 11: Generators don't allow you to specify default values for parameters, which is a limitation and forces us to either:

1. improve the generators for that logic (arguably out of scope of this PR)
2. change the internal API for this call so that a filter must always be provided
3. make it nullable and update the "default" behavior to happen inside of the SQL.

I do think you made the right call with (2) but wanted to highlight that internal change.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `GET /baseFields` is now paginated in the same manner as all other `GET` endpoints. This is a breaking change. The endpoint will now expect pagination parameters (`_count`, `_page`) and return `{ entries , total }` where `entries` is an array of base fields,
```
slifty: Correction here: `_count` and `_page` are not "expected" but rather are allowed -- if they aren't provided they should have default values / that wouldn't result in an error.

slifty (changes requested):
We should fix the changelog language to be accurate; I do think it'd be more explicit if our test didn't reference a default defined in the production code (which could change for business reasons in future), but instead was just a value defined inline in the test's function call

src/__tests__/baseFields.int.test.ts
yshng: Thanks for catching this - I changed it to check the PUT response and db, with filters passed inline.

slifty (approved):

PR #2376, opened 2026-04-28
Add `any` scope to permission grant system

### LINKED ISSUE #2368: Add support for "all" scopes and "all" verbs in permission grants

slifty:
This PR adds the `any` scope to the permission system which allows the creation of permissions that apply to all types of related entity to a given context.

Resolves #2368

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added `any` to the permission grant entity type set. When `any` is included in a grant's `scope` array, that grant satisfies any scope check on its context (e.g., `view | any` on a funder context grants view access for funder-, opportunity-, proposal-, and proposal-field-value-scoped data on that funder, including scope types added in the future). `any` is also recognized as a context entity type for forward compatibility but is not yet accepted by the API as a context.
+
 ### Changed
 
 - The `manage` permission verb now satisfies any verb check on the scope to which it is granted. A grantee with `manage` on a given scope no longer needs the other verbs listed alongside it to perform view, create, edit, delete, or reference operations at that scope. Scope matching is unchanged: `manage` does not grant access to scopes that are not explicitly included in the grant.
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -197,6 +197,15 @@ a grantee holding `manage | proposal` does not gain access to `funder`-,
 `opportunity`-, or `source`-scoped data unless those scopes are also
 included in the grant.
 
+The `any` scope is treated as satisfying any scope check on the same context.
+A grant of `view | any` on a funder context, for example, lets the grantee
+view funder-, opportunity-, proposal-, and proposal-field-value-scoped data
+for that funder without naming each scope separately. The `any` scope is
+useful for "owner" or "admin"-style grants that should automatically extend
+to scope types added in the future. Combining `any` with `manage` (e.g.,
+`manage | any`) effectively grants full administrative access on the given
+context.
+
 The `reference` verb is separate from `view` and `create` so that permission to
 see an entity does not automatically imply permission to cite it from elsewhere,
 and permission to create entities in one context does not automatically imply
--- a/src/__tests__/permissionGrants.int.test.ts
+++ b/src/__tests__/permissionGrants.int.test.ts
@@ -2,8 +2,15 @@ import request from 'supertest';
 import { app } from '../app';
 import {
 	createOrUpdateFunder,
+	createProposal,
 	createSource,
 	getDatabase,
+	hasChangemakerPermission,
+	hasDataProviderPermission,
+	hasFunderPermission,
+	hasOpportunityPermission,
+	hasProposalPermission,
+	hasSourcePermission,
 	loadTableMetrics,
 	removeSource,
 } from '../database';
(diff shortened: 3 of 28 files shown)


bickelj (commented):
I don't see anything objectionable when skimming the code.

If I understand correctly, this doesn't only lay groundwork for a future `any` scope but actually implements some of it.

~~@slifty Would you be willing to rebase and push so we can see it in a preview environment?~~ Oops, there it is, 30 mins ago :blush:

slifty:
@bickelj this completely implements the `any` scope -- it also lays some groundwork for an `any` context overall.

bickelj:
I tried this call against the preview environment (getting my JWT from another OpenAPI docs instance) and got an unexpected response:

```bash
$ curl -X POST 'https://pdc-service-preview-2376-s39x8.ondigitalocean.app/permissionGrants'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhb...KKfgg' -d '{
        "granteeType": "user",
        "granteeUserKeycloakUserId": "9f16a4e6-acfe-4048-82dd-d8a2d14effd0",
        "contextEntityType": "funder",
        "funderShortCode": "aspen_ridge_philanthropic_trust",
        "scope": ["all"],
        "verbs": ["view"]
}'
{"name":"InputValidationError","message":"Invalid request body.","details":[{"instancePath":"","schemaPath":"#/required","keyword":"required","params":{"missingProperty":"contextEntityType"},"message":"must have required property 'contextEntityType'"}]}
```

It looks like I included `contextEntityType` in there...

bickelj:
Locally I get a much better answer:

```json
{
  "message": "Invalid request body.",
  "details": [
    {
      "instancePath": "/scope/0",
      "schemaPath": "#/properties/scope/items/enum",
      "keyword": "enum",
      "params": {
        "allowedValues": [
          "funder",
          "changemaker",
          "dataProvider",
          "opportunity",
          "proposal",
          "proposalVersion",
          "applicationForm",
          "applicationFormField",
          "proposalFieldValue",
          "source",
          "bulkUpload",
          "changemakerFieldValue",
          "any"
        ]
      },
      "message": "must be equal to one of the allowed values"
    }
  ]
}
```

And when updating it to `any` another helpful result:

```json
{
  "message": "Scope value \"any\" is not valid for context entity type \"funder\". Allowed values: funder, opportunity, proposal, proposalFieldValue.",
  "details": []
}
```

I diffed the OpenAPI-generated `curl` call with the one I was using. I was missing the `Content-Type` header `  -H 'Content-Type: application/json' \`. When including it with `curl` I get the above. I'm still not sure how what I'm asking differs from the integration test.

src/__tests__/permissionGrantAnyScope.int.test.ts
```diff
@@ -0,0 +1,295 @@
...
+
+describe('`any` scope semantics', () => {
+	it('satisfies any scope check on a funder grant', async () => {
+		const db = getDatabase();
+		const testUser = await loadTestUser(db);
+		const testUserAuthContext = getAuthContext(testUser);
+		const adminAuthContext = getAuthContext(testUser, true);
+		const funder = await createTestFunder(db, testUserAuthContext);
+
+		await createPermissionGrant(db, adminAuthContext, {
```
bickelj: I think I'm trying to do this exact invocation via REST API call but getting an error. The funder exists, the user exists, but I get a 400 error:

```json
{
  "message": "Scope value \"any\" is not valid for context entity type \"funder\". Allowed values: funder, opportunity, proposal, proposalFieldValue.",
  "details": []
}
```

Here is the (redacted) curl equivalent from OpenAPI:

```bash
curl -X 'POST' \
  'http://localhost:3001/permissionGrants' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer REDACTED' \
  -H 'Content-Type: application/json' \
  -d '{
        "granteeType": "user",
        "granteeUserKeycloakUserId": "REDACTED",
        "contextEntityType": "funder",
        "funderShortCode": "aspen_ridge_philanthropic_trust",
        "scope": ["any"],
        "verbs": ["view"]
}'
```

slifty: The issue was these tests were testing the operator rather than the endpoint; that isn't inherently wrong... but this exposes why we do it the way we do in other cases!

Fixed the issue (`any` was not allowed at the handler validation layer), and updated the tests to correct the right place.

slifty:
@bickelj all righty!  Fixed, tests are better, and ready for re-review!

I manually created an `view : any` permission on a funder and indeed it allowed viewing of related funder entities.

bickelj (approved):
This works now:

```bash
$ curl -X 'POST'   'https://utilities.philanthropydatacommons.org/pdc-service-preview-2376-24eal/permissionGrants' -H 'Content-type: application/json'   -H 'accept: application/json'   -H 'Authorization: Bearer REDACTED' -d '{
        "granteeType": "user",
        "granteeUserKeycloakUserId": "REDACTED",
        "contextEntityType": "funder",
        "funderShortCode": "aspen_ridge_philanthropic_trust",
        "scope": ["any"],
        "verbs": ["view"]
}' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   628  100   364  100   264    576    417 --:--:-- --:--:-- --:--:--   993
```

Returned this:

```json
{
  "id": 96,
  "scope": [
    "any"
  ],
  "verbs": [
    "view"
  ],
  "createdAt": "2026-05-05T15:09:04.711717+00:00",
  "createdBy": "REDACTED",
  "conditions": null,
  "granteeType": "user",
  "funderShortCode": "aspen_ridge_philanthropic_trust",
  "contextEntityType": "funder",
  "granteeUserKeycloakUserId": "REDACTED",
  "granteeKeycloakOrganizationId": null
}
```

Issue #2385, opened 2026-04-30
OpenAPI docs needs relative URLs for all endpoints


bickelj:
Similar to #2383, in a preview environment that is proxied via an appended path, the wrong URL is accessed by the OpenAPI docs site.

For example, when visiting https://utilities.philanthropydatacommons.org/pdc-service-preview-2384-f8b4v/#/Base%20Fields/getBaseFields and trying the endpoint it tries to visit the incorrect baseFields endpoint.

Incorrect:
https://utilities.philanthropydatacommons.org/baseFields

Correct:
https://utilities.philanthropydatacommons.org/pdc-service-preview-2384-f8b4v/baseFields

[...]

PR #2386, opened 2026-04-30
Set an OpenAPI server URL of `/`

bickelj:
This is a long-shot attempt at fixing relative URLs in SwaggerUI.

Issue #2385 OpenAPI docs needs relative URLs for all endpoints


bickelj:
Does not work.

PR #2399, opened 2026-05-06
Grant permissions on entity creation

### LINKED ISSUE #2317: Grant permissions on item creation
When creating new entities we want to make sure the creator gets an appropriate permission grant applied to that entity (basically... all the permissions on that new entity).

I believe this happens in some cases, but it should happen in *all* cases

slifty:
This PR updates the API to grant mange permissions on created entities for the creating user.

Resolves #2317 
Relies on #2401

### DIFF
--- a/src/handlers/proposalVersionsHandlers.ts
+++ b/src/handlers/proposalVersionsHandlers.ts
@@ -1,5 +1,6 @@
 import { HTTP_STATUS } from '../constants';
 import {
+	createPermissionGrant,
 	createProposalFieldValue,
 	createProposalVersion,
 	getDatabase,
@@ -11,6 +12,7 @@ import {
 	loadProposalVersion,
 } from '../database';
 import {
+	getSelfManageGrantFragment,
 	isAuthContext,
 	isWritableProposalVersionWithFieldValues,
 	isId,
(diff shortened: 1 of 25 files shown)


src/types/PermissionGrant.ts
```diff
@@ -444,14 +447,38 @@ const getConditionsForScope = (
 	scopeKey: PermissionGrantEntityType,
 ): readonly PermissionGrantCondition[] => scopeConditions[scopeKey];
 
+// Returns the grantee + scope + verbs portion of a permission grant that
```
bickelj: Perhaps this comment should be a jsdoc.

src/types/PermissionGrant.ts
```diff
@@ -444,14 +447,38 @@ const getConditionsForScope = (
...
+// gives the auth user `manage` against `any` scope. Spread into a grant
+// alongside `contextEntityType` and the entity-specific key field.
+const getSelfManageGrantPartial = (
```
slifty: Realizing that since `Partial` has a specific meaning in TypeScript we might want to call this `Base` (or `Fragment` but I'm leaning `Base`)

src/handlers/proposalVersionsHandlers.ts
```diff
@@ -179,45 +181,61 @@ const postProposalVersion = async (
...
-					);
-					return proposalFieldValue;
-				}),
-			);
-			return {
-				...proposalVersion,
-				fieldValues: proposalFieldValues,
-			};
-		});
+		const committedProposalVersion = await db.transaction(
```
bickelj, line 184: Nice name change from final to committed!

bickelj (approved):
Having skimmed the code and talking through it live with Dan during office hours I see that this is an excellent and praiseworthy addition to the code. I recall suggesting the idea of "owner" but that comes with its own problems such as folks joining and leaving, etc. Separating permissions into its own area makes sense. Also I see that this creates permissions rows as part of the transaction that creates the data rows, which is good. We don't want a row to succeed without its associated permission grant. Thank you for this!

PR #2401, opened 2026-05-06
Return correct HTTP status on `PUT`

### LINKED ISSUE #2400: Expose `insert` vs `update` for upsert DB operations.
We have some PUT operations which might update or might create an entity.

We're about to want to be able to know when an entity was created (vs updated) -- for instance, when an entity is created we need to create permission grants as part of #2317.

This could be a purely internal concept, but it could also be something we expose via the API in the form of an optional `.wasInserted` attribute on the resulting entity.  It could also be something we expose via the API in the form of returning a meta object on POST rather than the actual entity, e.g.

```
{
   entity: foo,
   wasInserted: true,
}
```

^ this would obviously be a breaking change

or we could use an HTTP response header `X-WasInserted: true` which is also a reasonable approach!

slifty:
This PR fixes an issue where some of our endpoints returned `200` even when an entity was created.

In doing so, it also adds internal support for knowing whether an upsert query resulted in an insert or an update.

Resolves #2400


src/database/queries/changemakers/insertOrSelectOne.sql
```diff
@@ -14,8 +14,12 @@ ON CONFLICT (tax_id, name)
...
-) AS object;
+-- xmax is 0 for fresh inserts and non-zero when a row was reached via
+-- ON CONFLICT DO UPDATE; we expose that to distinguish create vs update.
+RETURNING
+	changemaker_to_json(
+		changemakers,
+		:authContextKeycloakUserId,
+		FALSE
+	) AS object,
+	(xmax = 0) AS "wasInserted";
```
bickelj: What is this wizardry!?

https://www.postgresql.org/docs/current/ddl-system-columns.html

Hmm.

bickelj: > xmax
>
>    The identity (transaction ID) of the deleting transaction, or zero for an undeleted row version. It is possible for this column to be nonzero in a visible row version. That usually indicates that the deleting transaction hasn't committed yet, or that an attempted deletion was rolled back.

bickelj: Interesting note here:

Image

bickelj (commented):
I wonder if there's a more official or better way than relying on internal postgresql implementation column `xmax = 0`. I doubt it but hoping...

src/database/queries/changemakers/insertOrSelectOne.sql
slifty: I meant to explicitly call this out AND flag it for "plz weigh in on this approach" -- I did see one alternative which I'll give a try + share in case it's preferred.

slifty: It seems like the alternative is a CTE, such as:

```
  WITH inserted AS (                                                                                                                                          
      INSERT INTO funders (                                                                                                                                   
          short_code,                                                                                                                                         
          name,                                                                                                                                               
          keycloak_organization_id,                                 
          is_collaborative,
          created_by
      ) VALUES (
          :shortCode,                                                                                                                                         
          :name,
          :keycloakOrganizationId,                                                                                                                            
          :isCollaborative,                                         
          :authContextKeycloakUserId
      )
      ON CONFLICT (short_code) DO NOTHING
      RETURNING funders, TRUE AS was_inserted                                                                                                                 
  ),
  updated AS (                                                                                                                                                
      UPDATE funders                                                
      SET
          name = :name,
          keycloak_organization_id = :keycloakOrganizationId,                                                                                                 
          is_collaborative = :isCollaborative
      WHERE short_code = :shortCode                                                                                                                           
          AND NOT EXISTS (SELECT 1 FROM inserted)                   
      RETURNING funders, FALSE AS was_inserted                                                                                                                
  )
  SELECT                                                                                                                                                      
      funder_to_json(row) AS object,                                
      was_inserted AS "wasInserted"                                                                                                                           
  FROM (
      SELECT funders AS row, was_inserted FROM inserted                                                                                                       
      UNION ALL                                                     
      SELECT funders AS row, was_inserted FROM updated                                                                                                        
  ) AS combined;
  ```
  
  Which means these queries would double as both an insert and update query.

bickelj: @slifty While that is more verbose, I think using the officially-supported language makes it more durable and clearer. Would all that go right here into each and every `.sql` or is it  within each function?

bickelj (approved):
I'm approving whichever way @slifty thinks balances out DRY, elegance, durability, etc. So I'm OK with the `xmax` thing, we'd need a note to avoid partitioning until we confirm it works with that, though. I'm also OK with the CTE despite its ugliness if it is going to be guaranteed to work across a range of situations (e.g. partitioning). Also if there's a better way, that's good too.

src/database/queries/changemakers/insertOrSelectOne.sql
slifty: Update!

PSQL 14 introduced and PSQL 17 expanded the concept of MERGE which is for just this use case, hooray!

It's still *slightly* duplicative, but far less verbose than the CTE and I think considered the modern / official tool for the job.

This does mean we require psql 17+ for local development which is a change.
sqlfluff doesn't support the syntax yet -- in the past I know we have been willing to appease the linter but this is a step too far, so I'd update the config to have it ignore those files.

slifty:
@bickelj I updated the `.sql` files to use `MERGE INTO` instead, which I think is a cleaner approach and is the modern way to do upserts.  It means we require PSQL 17+ but we were already there in our deployments.

bickelj (approved):
The lesser of three evils here, it seems. I am OK with any of them but a preference toward either CTE or MERGE.

Issue #2189, opened 2026-01-06
Use factories in tests


slifty:
We recently started the pattern of `factories` for generating test data; this will DRY our tests where possible and make it easier for us to know exactly what is being tested in a given test.  This issue captures the need to make this improvement overall.

This may sound inconsequential but we do a good job of testing this project, and writing tests represents a fair portion of the effort of adding a new feature.  Streamlining this will increase our throughput overall.

PR #2402, opened 2026-05-07
Use a factory for test `BaseFields`

slifty:
This PR updates our tests to use a factory for `BaseFields`

Related to #2189

### DIFF
--- a/src/__tests__/changemakers.int.test.ts
+++ b/src/__tests__/changemakers.int.test.ts
@@ -19,6 +19,7 @@ import {
 	loadTableMetrics,
 } from '../database';
 import {
+	createTestBaseField,
 	createTestChangemaker,
 	createTestDataProvider,
 	createTestFile,
@@ -43,7 +44,6 @@ import {
 } from '../test/mockJwt';
 import {
 	BaseFieldDataType,
-	BaseFieldCategory,
 	BaseFieldSensitivityClassification,
 	PermissionGrantEntityType,
 	PermissionGrantGranteeType,
@@ -73,32 +73,17 @@ const setupTestContext = async (db: TinyPg) => {
 	const systemUserAuthContext = getAuthContext(systemUser, true);
 	const testUser = await loadTestUser(db);
 	const testUserAuthContext = getAuthContext(testUser);
-	const baseFieldEmail = await createOrUpdateBaseField(db, null, {
-		label: 'Fifty one fifty three',
+	const baseFieldEmail = await createTestBaseField(db, null, {
 		shortCode: 'fifty_one_fifty_three',
-		description: 'Five thousand one hundred fifty three.',
 		dataType: BaseFieldDataType.EMAIL,
-		category: BaseFieldCategory.ORGANIZATION,
-		valueRelevanceHours: null,
-		sensitivityClassification: BaseFieldSensitivityClassification.RESTRICTED,
 	});
-	const baseFieldPhone = await createOrUpdateBaseField(db, null, {
-		label: 'Fifty three ninety nine',
+	const baseFieldPhone = await createTestBaseField(db, null, {
 		shortCode: 'fifty_three_ninety_nine',
-		description: 'Five thousand three hundred ninety nine.',
 		dataType: BaseFieldDataType.PHONE_NUMBER,
-		category: BaseFieldCategory.ORGANIZATION,
-		valueRelevanceHours: null,
-		sensitivityClassification: BaseFieldSensitivityClassification.RESTRICTED,
 	});
-	const baseFieldWebsite = await createOrUpdateBaseField(db, null, {
-		label: 'Fifty four seventy one 5471',
+	const baseFieldWebsite = await createTestBaseField(db, null, {
 		shortCode: 'fifty_four_seventy_one',
-		description: 'Five thousand four hundred seventy one.',
 		dataType: BaseFieldDataType.URL,
-		category: BaseFieldCategory.ORGANIZATION,
-		valueRelevanceHours: null,
-		sensitivityClassification: BaseFieldSensitivityClassification.RESTRICTED,
 	});
 	const firstChangemaker = await createChangemaker(db, testUserAuthContext, {
 		name: 'Five thousand one hundred forty seven reasons',
@@ -1024,15 +1009,9 @@ describe('/changemakers', () => {
 					firstFunderSourceId: funderSourceId,
 				} = await setupTestContext(db);
 
-				const forbiddenBaseField = await createOrUpdateBaseField(db, null, {
-					label: 'Forbidden Field',
-					shortCode: 'forbiddenField',
-					description: 'This field is forbidden',
+				const baseField = await createTestBaseField(db, null, {
 					sensitivityClassification:
 						BaseFieldSensitivityClassification.RESTRICTED,
-					dataType: BaseFieldDataType.STRING,
-					valueRelevanceHours: null,
-					category: BaseFieldCategory.ORGANIZATION,
 				});
 				const opportunity = firstFunderOpportunity;
 				const proposal = await createProposal(db, systemUserAuthContext, {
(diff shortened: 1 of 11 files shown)


src/__tests__/proposalVersions.int.test.ts
```diff
@@ -1639,25 +1557,12 @@ describe('/proposalVersions', () => {
...
-				sensitivityClassification:
-					BaseFieldSensitivityClassification.RESTRICTED,
 			});
-			await createOrUpdateBaseField(db, null, {
-				label: 'Org Name',
-				description: 'The organization name',
+			await createTestBaseField(db, null, {
 				shortCode: 'orgName',
-				dataType: BaseFieldDataType.STRING,
-				category: BaseFieldCategory.ORGANIZATION,
```
bickelj, line 1657: Is this category important not to lose? Apparently not since tests pass, but perhaps we lose something. OK, I see the default is ORGANIZATION, so no.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -674,28 +652,19 @@ describe('/proposalVersions', () => {
 				opportunityId: opportunity.id,
 				name: null,
 			});
-			const forbiddenBaseField = await createOrUpdateBaseField(db, null, {
-				label: 'Forbidden Field',
-				description: 'This field should not be used in proposal versions',
-				shortCode: 'forbiddenField',
```
bickelj, line 680: shortCode lost here.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -587,18 +564,19 @@ describe('/proposalVersions', () => {
 				opportunityId: opportunity.id,
 				name: null,
 			});
-			await createTestBaseFields(db);
+			const firstNameBaseField = await createTestBaseField(db, null);
+			const lastNameBaseField = await createTestBaseField(db, null);
 			const firstNameField = await createApplicationFormField(db, null, {
 				applicationFormId: applicationForm.id,
-				baseFieldShortCode: 'firstName',
```
bickelj, line 593: shortCode lost here.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -587,18 +564,19 @@ describe('/proposalVersions', () => {
...
+				baseFieldShortCode: firstNameBaseField.shortCode,
 				position: 1,
 				label: 'First Name',
 				instructions: 'Please enter the first name of the applicant.',
 				inputType: null,
 			});
 			const lastNameField = await createApplicationFormField(db, null, {
 				applicationFormId: applicationForm.id,
-				baseFieldShortCode: 'lastName',
```
bickelj, line 601: shortCode lost here.

src/test/factories/createTestBaseField.ts
```diff
@@ -17,10 +17,11 @@ const createTestBaseField = async (
 	authContext: AuthContext | null,
 	overrideValues?: Partial<InternallyWritableBaseField>,
 ): Promise<BaseField> => {
+	const id = uuidv4();
```
bickelj, line 20: While a uuidv4 is certainly better than every base field being the same, it does erase traceability. I mean that if I were in a situation where I am troubleshooting a failing test, I cannot easily tell which code generated which base field. The exception to this is where the shortCode or description is uniquely described within the test that calls this.

src/test/factories/createTestBaseField.ts
```diff
@@ -17,10 +17,11 @@ const createTestBaseField = async (
...
 	const defaultValues: InternallyWritableBaseField = {
-		shortCode: `test_field_${uuidv4()}`,
-		label: 'Test Field',
-		description: 'A test field for integration tests',
+		shortCode: `test_field_${id}`,
+		label: `Test Field ${id}`,
+		description: `A test field for integration tests (${id})`,
 		category: BaseFieldCategory.ORGANIZATION,
```
bickelj, line 25: This should probably be `PROJECT` and then tests that use `ORGANIZATION` can specify. Or vice versa.

src/database/initialization/__tests__/proposal_field_value_to_json.int.test.ts
```diff
@@ -271,14 +257,10 @@ describe('/proposal_field_value_to_json', () => {
 		});
 
 		// Create a file base field
-		const fileBaseField = await createOrUpdateBaseField(db, null, {
-			label: 'File Upload',
-			description: 'A field for file uploads',
+		const fileBaseField = await createTestBaseField(db, null, {
 			shortCode: 'fileUpload3',
```
bickelj, line 261: Example where `shortCode` is **not** lost, nice, and I see a lot more of these.

src/__tests__/proposalVersions.int.test.ts
```diff
@@ -587,18 +564,19 @@ describe('/proposalVersions', () => {
 				opportunityId: opportunity.id,
 				name: null,
 			});
-			await createTestBaseFields(db);
+			const firstNameBaseField = await createTestBaseField(db, null);
```
bickelj, line 567: Category of `PROJECT` is lost here unless we default to `PROJECT`.

src/__tests__/changemakers.int.test.ts
```diff
@@ -1799,15 +1769,8 @@ describe('/changemakers', () => {
 			const systemUserAuthContext = getAuthContext(systemUser, true);
 			const testUser = await loadTestUser(db);
 			const testUserAuthContext = getAuthContext(testUser);
-			const baseField = await createOrUpdateBaseField(db, null, {
-				label: 'Organization Website',
-				shortCode: 'org_website_priority_test',
```
bickelj, line 1804: `shortCode` lost here. Perhaps it's OK since an unusual type of `URL` is set.

src/__tests__/changemakers.int.test.ts
```diff
@@ -1734,16 +1713,7 @@ describe('/changemakers', () => {
...
-				label: 'Organization Mission',
-				shortCode: 'org_mission_cfv_test',
-				description: 'The mission of the organization.',
-				dataType: BaseFieldDataType.STRING,
-				category: BaseFieldCategory.ORGANIZATION,
-				valueRelevanceHours: null,
-				sensitivityClassification:
-					BaseFieldSensitivityClassification.RESTRICTED,
-			});
+			const baseField = await createTestBaseField(db, null);
```
bickelj, line 1716: There is fully defaults, nothing to distinguish this one.

bickelj (approved):
Looks good enough to me. In the unusual circumstance of a failing test, I like to be able to see which data came from where. This change preserves that capability. I wonder if there's a way, such as `expect.getState().currentTestName` to put the test name into the shortCode long description by default, such that it would be even easier to see the provenance of the test that created the base field.

Regardless, this is a welcome improvement and can be merged as-is. I made a few notes about a handful of places where a tiny bit of info may be lost or a decision about a default could be changed.

slifty:
@bickelj added https://github.com/PhilanthropyDataCommons/service/issues/2405 which I think is worth doing systemically across all types of factory (and shouldn't be too tough)

src/__tests__/proposalVersions.int.test.ts
slifty: Picking this randomly to reply to since I know this is one of several similarly raised questions!

In general my philosophy was "if the value is not actually needed for the test / if we aren't relying on that value specifically for the test assertions (as opposed to needing to reference *a* value), don't specify it".

So in this case (and the other lost values) -- those values didn't actually have significance for the test's assertions.  We still name the test object with something meaningful just to make the test itself more legible / make it easier for us to know the intention of when the object was going to be used.

This also means *even if something is the default* it should be specified in cases where that value is required for a test assertion.

Cases where it matters: the business logic being tested relies on a specified value; the value is being changed so we want to be able to know with certainty the original value.
^ I tried to make sure to keep things in those cases!

src/test/factories/createTestBaseField.ts
slifty: Opened #2405 to address this in a future iteration (I swear future won't mean never)

PR #2403, opened 2026-05-07
Use a factory for test `Users`

slifty:
This PR updates tests to use a factory for `Users`

Related to #2189

### DIFF
--- a/src/__tests__/users.int.test.ts
+++ b/src/__tests__/users.int.test.ts
@@ -1,26 +1,15 @@
 import request from 'supertest';
-import { v4 as uuidv4 } from 'uuid';
 import { app } from '../app';
-import {
-	getDatabase,
-	createOrUpdateUser,
-	loadSystemUser,
-	loadTableMetrics,
-} from '../database';
+import { getDatabase, loadSystemUser, loadTableMetrics } from '../database';
 import { loadTestUser } from '../test/utils';
+import { createTestUser } from '../test/factories';
 import { expectTimestamp } from '../test/asymettricMatchers';
 import {
 	mockJwt as authHeader,
 	mockJwtWithAdminRole as authHeaderWithAdminRole,
 } from '../test/mockJwt';
-import { keycloakIdToString, stringToKeycloakId } from '../types';
-import type { TinyPg } from 'tinypg';
-
-const createAdditionalTestUser = async (db: TinyPg) =>
-	await createOrUpdateUser(db, null, {
-		keycloakUserId: stringToKeycloakId('123e4567-e89b-12d3-a456-426614174000'),
-		keycloakUserName: 'Call me Ishmael',
-	});
+import { keycloakIdToString } from '../types';
+import type { User } from '../types';
 
 describe('/users', () => {
 	describe('GET /', () => {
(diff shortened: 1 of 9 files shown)


src/test/utils.ts
```diff
@@ -41,12 +41,6 @@ export const getMockedUser = (): User => ({
 	createdAt: '',
 });
 
-export const createTestUser = async (db: Pick<TinyPg, 'sql'>): Promise<User> =>
```
bickelj, line 44: Since this is being replaced, can `loadTestUser` be removed or replaced too? I am guessing not, I forget the circumstances here.

src/__tests__/users.int.test.ts
```diff
@@ -110,33 +99,10 @@ describe('/users', () => {
...
-						keycloakUserName: 'Alice',
-						createdAt: expectTimestamp(),
-					},
-					{
-						keycloakUserId: uuids[10],
-						keycloakUserName: 'Alice',
-						createdAt: expectTimestamp(),
-					},
-				],
+				entries: [10, 11, 12, 13, 14].reverse().map((index) => ({
```
bickelj: I like the old code here.

src/__tests__/users.int.test.ts
```diff
@@ -31,7 +20,7 @@ describe('/users', () => {
 		it('returns the user associated with the requesting user', async () => {
 			const db = getDatabase();
 			const testUser = await loadTestUser(db);
```
bickelj, line 22: Should these `loadTestUser` calls also be replaced by `createTestUser`?

bickelj (changes requested):
There is one test assertion that I cannot abide 😆 (thus the request changes).

Otherwise this is OK, but I think even more from `utils.ts` could be replaced.

Oh, and same comment from #2402, I wonder if the name of the generated user could include the name of the test whence it came.

src/__tests__/users.int.test.ts
slifty: Agreed -- will move back to something more explicit

src/__tests__/users.int.test.ts
slifty: I wondered that too -- I *think* so but didn't want to shave that yak right here, since I forget if there was anything "special" about the test user somehow in the test setup.

If it's OK, I'll keep this PR as is on that regard and circle back with another pass later / make an issue to investigate!

bickelj: Sounds good to me.

src/__tests__/users.int.test.ts
```diff
@@ -111,31 +100,11 @@ describe('/users', () => {
...
-						keycloakUserId: uuids[11],
-						keycloakUserName: 'Alice',
-						createdAt: expectTimestamp(),
-					},
-					{
-						keycloakUserId: uuids[10],
-						keycloakUserName: 'Alice',
-						createdAt: expectTimestamp(),
-					},
+					createdUsers[14],
```
slifty, line 103: @bickelj I think this is better! (it may not look it, but this is actually *more* specific than before because now createdAt must match the actual expected value not just "any timestamp")

bickelj (approved):
Requested change made. Much better!

src/__tests__/users.int.test.ts
bickelj: Very nice.

src/__tests__/users.int.test.ts
slifty: https://github.com/PhilanthropyDataCommons/service/issues/2406

PR #2404, opened 2026-05-07
Use a factory for test `Funders`

slifty:
This PR updates our test to use a factory for `Funder` entities.

Related to #2189

### DIFF
--- a/src/middleware/__tests__/requireFunderPermission.int.test.ts
+++ b/src/middleware/__tests__/requireFunderPermission.int.test.ts
@@ -2,11 +2,11 @@ import { requireFunderPermission } from '../requireFunderPermission';
 import { InputValidationError, UnauthorizedError } from '../../errors';
 import {
 	getDatabase,
-	createOrUpdateFunder,
 	createPermissionGrant,
 	loadSystemUser,
 } from '../../database';
 import { getAuthContext, getMockedUser, loadTestUser } from '../../test/utils';
+import { createTestFunder } from '../../test/factories';
 import { getMockRequest, getMockResponse } from '../../test/mockExpress';
 import {
 	PermissionGrantEntityType,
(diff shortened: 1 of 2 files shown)


src/middleware/__tests__/requireFunderPermission.int.test.ts
```diff
@@ -102,12 +97,7 @@ describe('requireFunderPermission', () => {
 			const systemUserAuthContext = getAuthContext(systemUser, true);
 			const testUser = await loadTestUser(db);
 			const testUserAuthContext = getAuthContext(testUser);
-			const funder = await createOrUpdateFunder(db, testUserAuthContext, {
-				shortCode: 'test_funder_with_perm',
-				name: 'Permitted Funder',
-				keycloakOrganizationId: null,
-				isCollaborative: false,
-			});
+			const funder = await createTestFunder(db, testUserAuthContext);
```
bickelj: At least one attribute that preserves the distinction between this funder being intended to have permission vs not would be good.

src/middleware/__tests__/requireFunderPermission.int.test.ts
```diff
@@ -143,12 +133,7 @@ describe('requireFunderPermission', () => {
 			const systemUserAuthContext = getAuthContext(systemUser, true);
 			const testUser = await loadTestUser(db);
 			const testUserAuthContext = getAuthContext(testUser);
-			const funder = await createOrUpdateFunder(db, testUserAuthContext, {
-				shortCode: 'test_funder_view_only',
-				name: 'View Only Funder',
-				keycloakOrganizationId: null,
-				isCollaborative: false,
-			});
+			const funder = await createTestFunder(db, testUserAuthContext);
```
bickelj: Again preserving at least one attribute that signals intent from the test would be good.

bickelj (approved):
This is OK. Same comment from #2402 and #2403: it would be nice to auto-include a test name. Also if there are two funders it is nice to include a distinguishing attribute, though I think in these cases there is only one funder so that probably doesn't matter.

src/middleware/__tests__/requireFunderPermission.int.test.ts
slifty: This comment triggered a neuron: rather than having intent be something the caller has to think about, maybe a factory method takes an `intent` attribute which can be incorporated into default values as appropriate on a per-object-type-basis.

That allows us as test writers to still have the value of "if this attribute is being explicitly set it is for the purposes of an assertion" but also still have the benefit of "if this test fails you can know from the data which object you're working with".

For now I will *at least* signal intent with a well named variable; then later we can either (A) update the tests to override values, or (B) add in this new concept of `intent` as a factory param in this and other places.

PR #2417, opened 2026-05-18
Add `authenticatedUsers` grantee type

### LINKED ISSUE #1774: Support "universal" permission grants
We have the concept of user permissions and user group permissions, but what if someone wants to grant a type of permission to ALL users?

We need to support this for the planned implementation of #1677 but there are other possible use cases (e.g. a changemaker wants to allow anybody to view a given proposal's data).

slifty:
This PR expands the permission system making it possible to grant a permission to all authenticated users.

Resolves #1774

### DIFF
--- a/src/database/initialization/grantee_permits_user.sql
+++ b/src/database/initialization/grantee_permits_user.sql
@@ -1,0 +1,30 @@
+SELECT drop_function('grantee_permits_user');
+
+-- Returns TRUE when the grant's grantee covers `user_keycloak_user_id` —
+-- either because the grant is to all authenticated users, directly to that
+-- user, or to a user group the user currently belongs to.
+CREATE FUNCTION grantee_permits_user(
+	grantee_type permission_grant_grantee_type_t,
+	grantee_user_keycloak_user_id uuid,
+	grantee_keycloak_organization_id uuid,
+	user_keycloak_user_id uuid
+) RETURNS boolean AS $$
+	SELECT grantee_type = 'authenticatedUsers'
+		OR (
+			grantee_type = 'user'
+			AND grantee_user_keycloak_user_id
+				= grantee_permits_user.user_keycloak_user_id
+		)
+		OR (
+			grantee_type = 'userGroup'
+			AND EXISTS (
+				SELECT 1
+				FROM ephemeral_user_group_associations euga
+				WHERE euga.user_keycloak_user_id
+						= grantee_permits_user.user_keycloak_user_id
+					AND euga.user_group_keycloak_organization_id
+						= grantee_permits_user.grantee_keycloak_organization_id
+					AND NOT is_expired(euga.not_after)
+			)
+		);
+$$ LANGUAGE sql STABLE;
--- a/src/database/initialization/has_application_form_permission.sql
+++ b/src/database/initialization/has_application_form_permission.sql
@@ -46,25 +46,11 @@ BEGIN
 			AND scope_set_permits_scope(
 				pg.scope, has_application_form_permission.scope
 			)
-			AND (
-				pg.grantee_type = 'authenticatedUsers'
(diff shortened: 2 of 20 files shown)


src/database/initialization/has_source_permission.sql
```diff
@@ -52,7 +52,8 @@ BEGIN
 				pg.scope, has_source_permission.scope
 			)
 			AND (
-				(
+				pg.grantee_type = 'authenticatedUsers'
```
hminsky2002: seems fine. unless i'm mistaken this is the only change in our logic right? everything else is just ensuring that we are compliant with this / reminds us that this is a check?

hminsky2002 (approved):
Looks good, unless I'm missing anything just an additional permissions check in our chain. I will assume that this has been caught in all the right places, as this is another one of those 'add everywhere changes'. Could be an opportunity to put some of this logic in a function? But i know this is a downstream blocker, and this does bring in the new functionality.

src/database/initialization/has_source_permission.sql
slifty: I'll see if there's a handy way to refactor the user check; it might not be easy or wise (sql functions can limit the query optimizer's ability to do its thing), but worth an exploration.

slifty: Added this refactor as a separate commit -- `STABLE` seems to mean it will not have optimizer implications, and the resulting checks are much cleaner.

slifty:
@hminsky2002 I know you approved, but with the consolidation of grantee check logic into a function I'd love your eyes on that commit too.

src/database/initialization/grantee_permits_user.sql
```diff
@@ -0,0 +1,30 @@
...
+			grantee_type = 'user'
+			AND grantee_user_keycloak_user_id
+				= grantee_permits_user.user_keycloak_user_id
+		)
+		OR (
+			grantee_type = 'userGroup'
+			AND EXISTS (
+				SELECT 1
+				FROM ephemeral_user_group_associations euga
+				WHERE euga.user_keycloak_user_id
```
hminsky2002, line 23: "euga" is a great onomatopoeia.

src/database/initialization/has_application_form_permission.sql
```diff
@@ -46,25 +46,11 @@ BEGIN
...
-						SELECT 1
-						FROM ephemeral_user_group_associations euga
-						WHERE euga.user_keycloak_user_id
-							= has_application_form_permission.user_keycloak_user_id
-							AND euga.user_group_keycloak_organization_id
-								= pg.grantee_keycloak_organization_id
-							AND NOT is_expired(euga.not_after)
-					)
-				)
+			AND grantee_permits_user(
```
hminsky2002, line 49: oh it's so gorgeous to see the code get replaced with a function.

src/database/initialization/grantee_permits_user.sql
```diff
@@ -0,0 +1,30 @@
...
+	grantee_keycloak_organization_id uuid,
+	user_keycloak_user_id uuid
+) RETURNS boolean AS $$
+	SELECT grantee_type = 'authenticatedUsers'
+		OR (
+			grantee_type = 'user'
+			AND grantee_user_keycloak_user_id
+				= grantee_permits_user.user_keycloak_user_id
+		)
+		OR (
```
hminsky2002, line 18: yep this is the right check, pass in both ids. Just to clarify we're not setting ourselves up for any sort of failure by checking it's a user grantee_type first right? Since the two values are mutually exclusive?

hminsky2002 (approved):
Woohoo for reusable code! More importantly, if and when the permissions check needs to be extended, i look forward to the PR that only has to change `grantee_permits_user.sql`

PR #2419, opened 2026-05-19
Use factory for `Source` in tests

slifty:
This PR adds a `sourceFactory` test utility.

Related to #2189

### DIFF
--- a/src/test/factories/createTestSource.ts
+++ b/src/test/factories/createTestSource.ts
@@ -1,0 +1,70 @@
+import { v4 as uuidv4 } from 'uuid';
+import { createChangemaker, createSource } from '../../database';
+import type { TinyPg } from 'tinypg';
+import type {
+	AuthContext,
+	Changemaker,
+	Id,
+	Source,
+	WritableSource,
+} from '../../types';
+
+type ObjectWithOrganizationKey =
+	| { changemakerId: Id }
+	| { funderShortCode: string }
+	| { dataProviderShortCode: string };
+
+let defaultChangemakerPromise: Promise<Changemaker> | null = null;
+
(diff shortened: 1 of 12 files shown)


src/test/factories/createTestSource.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { TinyPg } from 'tinypg';
+import type {
+	AuthContext,
+	Changemaker,
+	Id,
+	Source,
+	WritableSource,
+} from '../../types';
+
+type ObjectWithOrganizationKey =
```
hminsky2002, line 12: I believe this is a shortcoming of the Source type, not this type, but it seems like we should type these values as ShortCode, given that we have a type for it. That would make this a larger refactor that doesn't need to be scoped in here, but for consistency it feels like we should do that at some point.

src/test/factories/createTestSource.ts
```diff
@@ -0,0 +1,70 @@
...
+
+let defaultChangemakerPromise: Promise<Changemaker> | null = null;
+
+const getOrCreateDefaultChangemaker = async (
+	db: TinyPg,
+	authContext: AuthContext | null,
+): Promise<Changemaker> => {
+	defaultChangemakerPromise ??= createChangemaker(db, authContext, {
+		taxId: `default_test_source_changemaker_${uuidv4()}`,
+		name: 'Default Test Source Changemaker',
```
hminsky2002, line 25: Good to have these self-identify as the default values! Smart way to avoid ever confusing them for ones we explicitly defined.

hminsky2002 (approved):
Seems consistent with the other factory PRs, tests pass and I think it's fairly clean. Nice work @slifty !

PR #2420, opened 2026-05-19
Use factory for `Proposal`

slifty:
This PR adds a new `Proposal` factory test utility

Related to #2189

### DIFF
--- a/src/test/factories/createTestProposal.ts
+++ b/src/test/factories/createTestProposal.ts
@@ -1,0 +1,44 @@
+import { v4 as uuidv4 } from 'uuid';
+import { createProposal } from '../../database';
+import { createTestOpportunity } from './createTestOpportunity';
+import type { TinyPg } from 'tinypg';
+import type {
+	AuthContext,
+	Opportunity,
+	Proposal,
+	WritableProposal,
+} from '../../types';
+
+let defaultOpportunityPromise: Promise<Opportunity> | null = null;
+
+const getOrCreateDefaultOpportunity = async (
+	db: TinyPg,
+	authContext: AuthContext | null,
+): Promise<Opportunity> => {
+	defaultOpportunityPromise ??= createTestOpportunity(db, authContext);
+	return await defaultOpportunityPromise;
+};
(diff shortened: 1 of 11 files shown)


src/test/factories/createTestProposal.ts
```diff
@@ -0,0 +1,44 @@
...
+	db: TinyPg,
+	authContext: AuthContext | null,
+	overrideValues?: Partial<WritableProposal>,
+): Promise<Proposal> => {
+	const opportunityId =
+		overrideValues?.opportunityId ??
+		(await getOrCreateDefaultOpportunity(db, authContext)).id;
+	const defaultValues: WritableProposal = {
+		opportunityId,
+		externalId: `test_proposal_${uuidv4()}`,
```
hminsky2002, line 36: Maybe worth adding that this is a default value? I guess because the uuid is generated at runtime that kind of loses its value.

src/test/factories/createTestProposal.ts
```diff
@@ -0,0 +1,44 @@
...
+const getOrCreateDefaultOpportunity = async (
+	db: TinyPg,
+	authContext: AuthContext | null,
+): Promise<Opportunity> => {
+	defaultOpportunityPromise ??= createTestOpportunity(db, authContext);
+	return await defaultOpportunityPromise;
+};
+
+const resetTestProposalFactory = (): void => {
+	defaultOpportunityPromise = null;
```
hminsky2002, line 23: this is probably overkill but considering the pattern being repeated here, would it make sense to have another test generator that takes in a type and generates a tuple of a value and a reset function? (default_TYPE,reset_default_TYPE()).

hminsky2002 (approved):
Approved, with a light suggestion for a potential DRY move!

src/test/factories/createTestProposal.ts
slifty: It's an interesting idea, and it's now repeated a few times.  It's also even possible there's already a library for it...  I'll make an issue.

PR #2424, opened 2026-05-21
Use Opportunity factory consistently

slifty:
This PR fixes some instances where we weren't using our Opportunity factory in tests

Related to #2189


hminsky2002 (approved):
LGTM

PR #2425, opened 2026-05-21
Support `TerminologySets`

### LINKED ISSUE #2330: [Enhancement]: add attribute to opportunity for source_record_name to capture contributors record naming
### Summary

Long discussion leading up to this but there is a need to render the source record naming from the source system in PDC. We want to be able to capture this and make it available for related applications and proposals. 

### Problem / Opportunity

_No response_

### Proposed solution

_No response_

### Acceptance criteria

_No response_

### Priority

High

slifty:
This PR introduces a new entity type, `TerminologySet`, which allows funders to define new labels for specific PDC entity types (Opportunity, Proposal, and Application Form) in the context of a particular opportunity.

The PR includes CRUD endpoints, documentation, and permission support.

Resolves #2330


bickelj:
I found this in the DO logs:
> May 22 17:48:51  {"level":50,"time":1779472131777,"pid":3,"hostname":"service-8b8646bdb-778px","source":"/opt/philanthropy-data-commons/server/dist/scripts/migrate.js","msg":"Found a non-consecutive migration ID on file: '0111-add-terminologySet-to-permission_grant_entity_type_t.sql'"}

slifty:
> I found this in the DO logs:
> 
> > May 22 17:48:51  {"level":50,"time":1779472131777,"pid":3,"hostname":"service-8b8646bdb-778px","source":"/opt/philanthropy-data-commons/server/dist/scripts/migrate.js","msg":"Found a non-consecutive migration ID on file: '0111-add-terminologySet-to-permission_grant_entity_type_t.sql'"}

This is a WIP for a reason :D

src/types/Funder.ts
```diff
@@ -34,6 +36,10 @@ const writableFunderSchema: JSONSchemaType<WritableFunder> = {
 		isCollaborative: {
 			type: 'boolean',
 		},
+		defaultTerminologySetId: {
+			type: 'integer',
```
bickelj: Can this make use of `idSchema` or is that not appropriate since it can be null or undefined?

src/types/Opportunity.ts
```diff
@@ -27,6 +30,10 @@ const writableOpportunitySchema: JSONSchemaType<WritableOpportunity> = {
 		funderShortCode: {
 			...shortCodeSchema,
 		},
+		terminologySetId: {
```
bickelj, line 37: Likewise here, can this use `idSchema`?

src/database/migrations/0113-add-terminology_sets.sql
```diff
@@ -0,0 +1,22 @@
...
+	short_code_t NOT NULL REFERENCES funders (short_code) ON DELETE CASCADE,
+	name text NOT NULL,
+	opportunity_label text,
+	opportunities_label text,
+	application_form_label text,
+	application_forms_label text,
+	proposal_label text,
+	proposals_label text,
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+	created_by uuid NOT NULL REFERENCES users (keycloak_user_id) ON DELETE CASCADE,
```
bickelj, line 13: I'm not sure about either `ON DELETE CASCADE` here. Perhaps we simply shouldn't delete users or funders and then all is OK.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -119,6 +120,20 @@ erDiagram
     string name
     uuid keycloakOrganizationId
     boolean isCollaborative
+    int defaultTerminologySetId FK
+    datetime createdAt
+    uuid createdBy FK
+  }
+  TerminologySet {
+    int id PK
+    string funderShortCode FK
```
bickelj, line 129: At first glance there appears to be a circular reference (`TerminologySet.funderShortCode` and `Funder.defaultTerminologySetId`) but it is not really circular because there can be many TerminologySets created.

I wonder, though, if we even need the TerminologySet to be associated with a `funderShortCode`. Why should the definition of the terminology need to be associated with the funder? We have the user who created it, and perhaps the user can grant to the funder's organization the permissions to manage it. And we already have the `defaultTerminologySetId` on `funder` and also a `terminologySetId` on opportunity.

src/__tests__/terminologySets.int.test.ts
```diff
@@ -0,0 +1,784 @@
...
+				db,
+				testUserAuthContext,
+				{
+					funderShortCode: funder.shortCode,
+					name: 'Reference-protected vocab',
+				},
+			);
+
+			// Caller can CREATE opportunities on funder, but has no reference
+			// permission on the terminology set.
```
bickelj, line 503: This fits our "no or very few implicit grants" permissions model but this is awkward.

docs/PERMISSIONS.md
```diff
@@ -268,6 +268,8 @@ context key).
 | edit   | proposal           | Create or update proposal versions for the funder's proposals                        |
 | view   | source             | View the funder's sources                                                            |
 | create | source             | Create sources associated with the funder                                            |
+| view   | terminologySet     | View the funder's terminology sets                                                   |
```
bickelj, line 271: I am trying to imagine a case where a funder would like some other party to be able to view the funder but then **not view** the preferred terminology of that funder.

bickelj (commented):
When reading code (not actually trying it) I don't see anything truly alarming but I have some questions in-line.

src/types/Funder.ts
slifty: Oh good call -- I think we should be able to do both (use the schema AND mark it as nullable)

docs/PERMISSIONS.md
slifty: We could say that viewing a funder means viewing all of the funder's terminology sets, but it is starting to stray from the grant permission design philosophy.

Also remember this also covers the inverse (you can view the terminology set but not the funder)

bickelj: Oh, yes, that last capability makes sense.

bickelj: Except, if we are using `funder_short_code` in the terminology set is that attribute filtered out on `GET`?

slifty: No -- that isn't the behavior across any of our FK relationships (nor should it be).

If you have the permission to view an entity you can view all of its attributes.

(now, you wouldn't see the actual OBJECT unless you have permission!)

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: Yeah this is valid third normal form for sure; it isn't circular because default is nullable.

We do need a schema-level (not just permission) association between terminology sets and funders because terminology sets are a by funders.  Users are not a proxy for organization at the schema level, since a user's association with any given organization is permission based rather than schema based.

It might help to think in terms of UX:

1. When a funder creates an opportunity, they can choose a terminology set to use within that opportunity.
2. The dropdown would show a list of all terminology sets associated with the funder (that they have permission to view).
3. The default value of that dropdown would be driven by `defaultTerminologySetId`.

The user makes a selection, hits submit, and the opportunity is created.

slifty: I've updated the API to remove any actual "use" of the default -- it is now just information for clients to leverage if they want.

bickelj: The situation I am thinking of is this:
1. A funder creates a terminology set with very commonly used terminology and sets it as default for its opportunities.
2. Another funder cannot see that terminology set so creates an identical terminology set and sets it as default for its opportunities.

It seems to me that terminology (language) may be used by a particular organization but is not always particular to that organization.

We tend to avoid ownership in other cases so the alternative would be to use permissions (or lack thereof) rather than ownership. Like this:
1. All terminology sets are always viewable by anyone.
2. Authenticated users (or perhaps only users associated with funders) can create terminology sets.
3. There is no funder associated with a terminology set.
4. The user who creates the terminology set can manage it. Alternative: terminology sets are immutable.

The downside to this setup is change over time: a funder could not modify its existing terminology set and have that edit apply to all the things without going and touching all the things.

Therefore, having thought through it a bit more, it is not so bad to have the funder associated, with the exception that the funder's name will be visible to those who have access to that terminology set, when a funder may want to be a secret funder.

docs/PERMISSIONS.md
bickelj: Rather than thinking in terms of the technical implementation (FK relationships) I'm thinking more in terms of UX: if you can view the terminology set but not the funder, we can't truly do that if the funder name is part of the terminology set.

src/__tests__/funders.int.test.ts
```diff
@@ -371,6 +374,7 @@ describe('/funders', () => {
 				name: 'The Foundation Foundation',
 				keycloakOrganizationId: '0de87edc-be40-11ef-8249-0312f1b87538',
 				isCollaborative: false,
+				defaultTerminologySetId: null,
```
bickelj, line 377: For a moment I thought this would require setting `defaultTerminologySetId` when creating a funder, but no, I verified in the preview environment that I could PUT a funder like this:

```json
{
  "name": "Example Foundation",
"isCollaborative": false
}
```

Furthermore the test is for a GET 😅 

Nice.

bickelj (approved):
My objections have been settled, having skimmed the code this looks like a great setup that will help a lot.

I was able to create a terminology set in the preview environment like this (omitting some of the terminology I didn't need):

```json
{
  "funderShortCode": "test_funder",
  "name": "RFP Vocabulary",
  "opportunityLabel": "Request for Proposal",
  "proposalsLabel": "Funding Requests"
}
```

I did not try the permissions part but I trust the automated tests.

PR #2434, opened 2026-05-28
Update development realm.json with organizations scope

chrisjkuch:
This PR adds the organizations scope to the Keycloak realm.json and enables other settings for running Exchange integration tests against PDC in CI

Exchange checks out the service repo and spins up the dev docker compose stack in CI for running integration tests. As we are switching to multi-org users, we needed to sort out the right OIDC scope for Exchange to request and realized that the realm was not configured properly. @bickelj provided an updated export from prod Keycloak that revealed a pluralization difference (organization vs organizations) and the unexpected necessity of `organizations:*` rather than just `organizations` in order to get everything in the JWT (see https://github.com/PhilanthropyDataCommons/auth/issues/110).

Beyond the rename, we flip to JSON-type labels (so we don't end up with a string blob) and we add the org ids to the token so that the uuid appears in userinfo.

### DIFF
--- a/keycloak/realm.json
+++ b/keycloak/realm.json
@@ -555,13 +555,13 @@
 				"profile",
 				"roles",
 				"basic",
-				"email"
+				"email",
+				"organizations"
 			],
 			"optionalClientScopes": [
 				"address",
 				"phone",
 				"offline_access",
-				"organization",
 				"microprofile-jwt"
 			]
 		},
@@ -610,13 +610,13 @@
 				"profile",
 				"roles",
 				"basic",
-				"email"
+				"email",
+				"organizations"
 			],
 			"optionalClientScopes": [
 				"address",
 				"phone",
 				"offline_access",
-				"organization",
 				"microprofile-jwt"
 			]
 		},
@@ -652,13 +652,13 @@
 				"profile",
 				"roles",
 				"basic",
-				"email"
+				"email",
+				"organizations"
 			],
 			"optionalClientScopes": [
 				"address",
 				"phone",
 				"offline_access",
-				"organization",
 				"microprofile-jwt"
 			]
 		},
@@ -693,13 +693,13 @@
 				"profile",
 				"roles",
 				"basic",
-				"email"
+				"email",
+				"organizations"
 			],
 			"optionalClientScopes": [
 				"address",
 				"phone",
 				"offline_access",
-				"organization",
 				"microprofile-jwt"
 			]
 		},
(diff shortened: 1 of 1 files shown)


keycloak/realm.json
bickelj: We actually have `organizations` in the `defaultDefaultClientScopes` rather than `defaultOptionalClientScopes`.

chrisjkuch: Amended + updated

bickelj: I'm not sure about the remainder, but the way its actually configured, I think it is default on all the rest too. Checking now...

bickelj: Yes, I confirmed that all clients that have it list it under "default" rather than "optional".

keycloak/realm.json
```diff
@@ -616,7 +616,7 @@
 				"address",
 				"phone",
 				"offline_access",
-				"organization",
+				"organizations",
```
bickelj: This (and the rest) should go up into the `defaultClientScopes` .

bickelj:
Here's what I see in an export:

```json
    {
      "id": "SOME-REDACTED-UUID",
      "name": "organizations",
      "description": "All the organizations",
      "protocol": "openid-connect",
      "attributes": {
        "include.in.token.scope": "true",
        "display.on.consent.screen": "true",
        "gui.order": "",
        "consent.screen.text": ""
      },
      "protocolMappers": [
        {
          "id": "SOME-OTHER_REDACTED-UUID",
          "name": "organizations",
          "protocol": "openid-connect",
          "protocolMapper": "oidc-organization-membership-mapper",
          "consentRequired": false,
          "config": {
            "introspection.token.claim": "true",
            "multivalued": "true",
            "userinfo.token.claim": "true",
            "addOrganizationAttributes": "true",
            "id.token.claim": "true",
            "lightweight.claim": "false",
            "access.token.claim": "true",
            "claim.name": "organizations",
            "jsonType.label": "JSON",
            "addOrganizationId": "true"
          }
        }
      ]
    }
```

keycloak/realm.json
```diff
@@ -1489,17 +1489,20 @@
...
 						"access.token.claim": "true",
-						"claim.name": "organization",
-						"jsonType.label": "String",
-						"multivalued": "true"
+						"userinfo.token.claim": "true",
+						"claim.name": "organizations",
+						"jsonType.label": "JSON",
+						"multivalued": "true",
+						"addOrganizationId": "true",
+						"addOrganizationAttributes": "false"
```
bickelj: We have this as `true`.

keycloak/realm.json
chrisjkuch: Okay, should now be in double default throughout

keycloak/realm.json
chrisjkuch: Done!

keycloak/realm.json
chrisjkuch: Done, and lightweight.claim was also added

bickelj (approved):
Thank you!

PR #2445, opened 2026-06-02
Consistently use component responses

### LINKED ISSUE #2444: Avoid redundant openapi documentation
Our openapi specifications sometimes use references but it can be inconsistent.  For example we have some `responses` but they aren't always used when defining status code responses.

We're about to take on #2421, and it would be great to have this reference consistency in place ahead of that change.

slifty:
This PR makes our use of component responses more consistent in our API specification.  This will help us as we make improvements to status code consistency in implementation as well.

Resolves #2444


slifty:
Copilot made 22 of the same comment (sigh) and it is also wrong -- `description` can given alongside a reference in OpenAPI 3.1

hminsky2002 (approved):
Yep looks reasonable. Lot of file changes, but just the one functional change

bickelj (approved):
Looks fine to me! Thank you!

Do we not use DatabaseError? I can imagine we wouldn't want to surface too gory of details but I suppose we still can within InternalServerError if we want.

PR #2446, opened 2026-06-02
Document autopublication process separately

kfogel:


bickelj (commented):
Right now the README.md shows up automatically at https://github.com/PhilanthropyDataCommons/service/tree/main/docs in the GH interface. I think this would prevent it from showing up. Is that the idea?

kfogel:
> Right now the README.md shows up automatically at https://github.com/PhilanthropyDataCommons/service/tree/main/docs in the GH interface. I think this would prevent it from showing up. Is that the idea?

No, not at all -- at least, that's not what I intended!

The purpose of this change is to separate out documentation about the technical process by which our documentation is auto-published to WordPress from documentation about "what is in this directory right here".  Both the old and the new versions of `README.md` have `nopublish: true` at the top, but that signal just says not to publish this file on the WordPress site -- I don't think it has any significance for GitHub's rendering of the README, or at any rate, if it *does* have that significance, this commit does not change what we were doing before in that regard.

(Secondarily, I just noticed now that the documentation about the autopublication process -- that is, the material that this commit moves into the separate new file `DOCS_AUTOPUBLICATION.md` -- does not even *mention* the use of the [Markdown Renderer for GitHub](https://wordpress.org/plugins/markdown-renderer-for-github/) plugin on the WordPress side.  Adding that would be a substantive change to the documentation here, so I'm not trying to include it in this PR, but I'm noting it because it seems pretty important and like it would probably be another PR coming soon.)

docs/README.md
```diff
@@ -2,97 +2,12 @@
...
-Not related to WordPress metadata. When set to true, it prevents a file from
-being sent to WordPress at all. See the top of this document for example.
+This directory and its subdirectories contain documentation about
+various aspects of the Philanthropy Data Commons framework, data
+entities, permissions system, etc.
+
+Much of this documentation is automatically published to the
+[PhilanthropyDataCommons.org](https://www.PhilanthropyDataCommons.org)
+web site (and is automatically converted from Markdown to HTML along
+the way).  See [DOCS_AUTOPUBLICATION.md](DOCS_AUTOPUBLICATION.md) for
```
bickelj: Our format style doesn't like the two spaces here.

kfogel: @bickelj I'll make that change now, and separately take up the argument about the two-spaces-after-a-sentence-ending period elsewhere.

bickelj (approved):
LGTM

docs/README.md
kfogel: Okay, branch updated (and [writing-style question taken up in Zulip](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/documentation/near/261934)).

PR #2448, opened 2026-06-04
Use stable functions

### LINKED ISSUE #2329: Update permission check functions to be `STABLE`
This allows psql to be more optimized when running these queries / lowers the odds of a full table scan.

slifty:
This PR updates our functions to be `STABLE` where appropriate.  This exposed an issue where `NOT_MATERIALIZED` was ironically resulting in 2x processing, so it also updates our CTEs to be materialized.

Resolves #2329


hminsky2002 (approved):
Looks reasonable to me!

PR #2453, opened 2026-06-12
Correct OpenAPI specifications

### LINKED ISSUE #2451: OpenAPI spec deviates from real payload signatures
This happens in a few different places, I'll try to enumerate them with links:

- [`Organization.funder`](https://github.com/PhilanthropyDataCommons/service/blob/f10e2984a986587baca19552bec1fd472c53544d/src/openapi/components/schemas/Organization.json#L15-L19) is currently mis-typed as `dataProvider`, should be `funder`. 

- [`Source.relatedEntityId`](https://github.com/PhilanthropyDataCommons/service/blob/f10e2984a986587baca19552bec1fd472c53544d/src/openapi/components/schemas/Source.json#L28) is listed in `required` but never defined as a property.

- [`PresignedPost.fields`](https://github.com/PhilanthropyDataCommons/service/blob/f10e2984a986587baca19552bec1fd472c53544d/src/openapi/components/schemas/PresignedPost.json#L7-L16) modeled as a fixed object with only key. It actually carries the dynamic S3 POST fields (policy, signature, x-amz-*, …) and should have `additionalProperties: true`

I haven't done an exhaustive inventory, but it appears that in some places the type definition is correct even though the spec is wrong (e.g., for [Organization](https://github.com/PhilanthropyDataCommons/service/blob/f10e2984a986587baca19552bec1fd472c53544d/src/types/Organization.ts#L5-L10))

Relates to #2353, #1725, #701, #1011, #1012

(For context, I'm trying to set up our python PDC client to auto-generate pydantic models from the OpenAPI json, and am having to do some workarounds to get it to work properly).

slifty:
There were a series of cases where the OpenAPI specification did not align with our actual implementations.

Some of these were identified by users, others were identified by an audit.

Resolves #2451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Corrected the OpenAPI `Organization.funder` schema to reference the `Funder` schema instead of `DataProvider`. Documentation-only; no behavior change.
+- Corrected the OpenAPI `Source` schema to no longer list the non-existent `relatedEntityId` as a required property. Documentation-only; no behavior change.
+- Corrected the OpenAPI `PresignedPost.fields` schema to allow the dynamic S3 POST form fields (e.g. `policy`, `x-amz-*`) via `additionalProperties`. Documentation-only; no behavior change.
+- Corrected the OpenAPI `BaseField` schema to mark `shortCode` as `readOnly` (it used the misspelled, ignored `read-only` keyword). Documentation-only; no behavior change.
+- Corrected the OpenAPI `BaseFieldLocalization` schema: `description` and `createdAt` are no longer nullable, `createdAt` is now `readOnly` and required, and `baseFieldShortCode` is now `readOnly`, matching the actual payload. Documentation-only; no behavior change.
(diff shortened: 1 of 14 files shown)


src/openapi/components/schemas/Organization.json
```diff
@@ -13,7 +13,7 @@
 			"nullable": true
 		},
 		"funder": {
-			"$ref": "./DataProvider.json",
+			"$ref": "./Funder.json",
```
bickelj, line 16: Zoinks

src/openapi/components/schemas/User.json
```diff
@@ -7,7 +7,7 @@
 			"format": "uuid"
 		},
 		"keycloakUserName": {
-			"type": "string",
+			"type": ["string", "null"],
 			"description": "The full name of the user (described in OIDC 1.0 'name' standard claim) as of most recent logged-in service activity.",
 			"example": "John Doe"
```
bickelj, line 12: I wonder if this is `readOnly` too.

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
```
bickelj, line 12: If the SDK is built using the OpenAPI spec it would behoove us to update the version here.

bickelj (approved):
I quickly skimmed the changes here without trying them (for example, in the deploy preview environment) and without comparing to internal types but I see no red flags. This looks like a lot of good corrections, thank you!

I think we want a version increase so the SDK and its users will benefit.

PR #2460, opened 2026-06-18
Use tables instead of row-booleans for permission checks

### LINKED ISSUE #2450: Optimize permission based queries
We're using SQL / PSQL functions to determine permissions in queries.  This, unfortunately, means the query planner does not have the ability to avoid full table scans and queries often become O(N^2) or worse.

The result is that queries can take 30s with as few as 800 permissions rows.

I believe there are two improvements we can make to get down to <10ms

1. Add indexes <-- we always knew this would be important, but now that we have real data it's more important
2. Move from functions that return boolean values at the row level to functions that return tables (e.g. instead of `has_changemaker_permission` in a WHERE we do something like `changemakers_with_permission` in a JOIN)

slifty:
This PR shifts our sql-based permission checks from row-level psql functions that return a boolean into sql functions that will return a complete result set.  This is significantly faster (e.g. 300ms => 2ms) even without using indexes and also allows the query planner to inline the function internals and incorporate it into the larger query plan.

I'll be adding indexes in a future PR.

Resolves #2450 

Some before vs after metrics to pair with this PR:

```
  ┌─────────────────┬───────────────────────────────────────┬───────┬───────────┬───────────────┬────────────────┬─────────┐
  │     Entity      │              Inheritance              │ Rows  │ Permitted │ Before (main) │ After (branch) │ Speedup │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ funder          │ flat                                  │   100 │        80 │       10.2 ms │         1.3 ms │      8× │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ dataProvider    │ flat                                  │    50 │        40 │        8.9 ms │         1.1 ms │      8× │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ changemaker     │ flat                                  │ 4,000 │       700 │        480 ms │         2.6 ms │    187× │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ opportunity     │ ← funder                              │ 2,000 │     1,660 │        443 ms │         4.3 ms │    103× │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ applicationForm │ ← opportunity ← funder                │ 2,000 │     1,660 │        530 ms │         3.6 ms │    147× │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ source          │ ← funder/dataProvider/changemaker     │ 1,998 │     1,573 │      1,902 ms │         3.7 ms │    519× │
  ├─────────────────┼───────────────────────────────────────┼───────┼───────────┼───────────────┼────────────────┼─────────┤
  │ proposal        │ ← opportunity ← funder, + changemaker │ 4,000 │     3,380 │      4,027 ms │         5.6 ms │    722× │
  └─────────────────┴───────────────────────────────────────┴───────┴───────────┴───────────────┴────────────────┴─────────┘
  ```


bickelj (approved):
I see good things here, based on a skim of the code. Our tests have complete or near-complete coverage of permissions scenarios and these tests are passing, which gives me the confidence to mash "Approve" without trying it manually myself. Fantastic, thank you!

PR #2464, opened 2026-06-23
Add query parameters for related entities on GET permissions

### LINKED ISSUE #2447: [Enhancement]: Add filtering to permission grant queries
### Summary

It would be great if we could filter permission grant queries on specific criteria. For example, as a changemaker, I want to know whether I have manage permissions for my organization, and to see individual permission grants applied to any of my organization's proposals if so.

### Problem / Opportunity

`/permissionGrants` only takes `_page` and `_count`, so getting this information requires paginating through all permission grants.

#2329 also makes this pagination for users slow.

Related: #1976 and #2418 

### Proposed solution

Add filters to `GET /permissionGrants` by context entity (`changemakerId`, `funderShortCode`, `dataProviderShortCode`, `proposalId`) and optionally grantee (`granteeType`, `verb`).

### Acceptance criteria

`/permissionGrants` accepts filter params, e.g., `changemakerId`, `granteeType`, `verb`. An unmatched filter should return empty.

### Priority

Low

hminsky2002:
This adds optional filter query parameters, mirroring the existing `GET /proposals` filtering pattern on Context entity, `changemaker`, `funder`, `dataProvider`, `proposal` and Grantee, `granteeType` and `verb`. This means adding a dataprovider parameters extractor, a granteeType, and a verb extractor method.

Closes #2447


bickelj:
@hminsky2002 I'm not sure what happened with the preview env but I expect if you push again it should succeed.

src/handlers/permissionGrantsHandlers.ts
```diff
@@ -207,10 +215,22 @@ const getPermissionGrants = async (
 	const db = getDatabase();
 	const paginationParameters = extractPaginationParameters(req);
 	const { limit, offset } = getLimitValues(paginationParameters);
+	const { changemakerId } = extractChangemakerParameters(req);
+	const { funderShortCode } = extractFunderParameters(req);
+	const { dataProviderShortCode } = extractDataProviderParameters(req);
+	const { proposalId } = extractProposalParameters(req);
+	const { granteeType } = extractGranteeTypeParameters(req);
+	const { verb } = extractVerbParameters(req);
```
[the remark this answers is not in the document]
hminsky2002: reasonable point from AI!

src/handlers/permissionGrantsHandlers.ts
```diff
@@ -207,10 +217,35 @@ const getPermissionGrants = async (
...
+
+	const contextEntityFilters = [
+		changemakerId,
+		funderShortCode,
+		dataProviderShortCode,
+		proposalId,
+	].filter((filterValue) => filterValue !== undefined);
+	if (contextEntityFilters.length > MAXIMUM_CONTEXT_ENTITY_FILTERS) {
+		throw new InputValidationError(
+			'Only one of "changemaker", "funder", "dataProvider", or "proposal" may be provided.',
```
bickelj: The way this looks in OpenAPI spec:

Image

The message is clear but the escaping is a bit distracting.

src/database/queries/permissionGrants/selectWithPagination.sql
```diff
@@ -8,6 +8,31 @@ WITH
...
+				ELSE permission_grants.proposal_id = :proposalId
+			END
+			AND CASE
+				WHEN :granteeType::permission_grant_grantee_type_t IS NULL THEN TRUE
+				ELSE permission_grants.grantee_type = :granteeType
+			END
+			AND CASE
+				WHEN :verb::permission_grant_verb_t IS NULL THEN TRUE
+				ELSE :verb::permission_grant_verb_t = any(permission_grants.verbs)
+			END
```
bickelj, line 35: Both Qwen3.6-35B-A3B and GLM-5.2 flagged a potential performance issue in the above block.

Qwen:
> The new selectWithPagination.sql uses CASE WHEN :param::type IS NULL THEN TRUE ELSE column = :param END for each filter. This pattern prevents PostgreSQL from using indexes on the filtered columns because the expression is non-sargable. Consider dynamic SQL (building the query string based on which parameters are provided) or a WHERE (param IS NULL OR column = param) pattern instead. This is a performance concern for large tables.

GLM:
> Correct: = any(permission_grants.verbs) is the right way to match "grants that include this verb", and the test filters by verb, matching grants that include the verb confirms a [view, edit] grant matches verb=edit while a [manage]-only grant does not.
>
> Optional scale note: array-containment can't use a btree index, so if this filter becomes hot, a GIN index on permission_grants.verbs would help. Tiny consistency nit: granteeType casts only in the WHEN branch (relies on column type inference in ELSE), whereas verb casts in both branches — harmless, just FYI.

I wouldn't have mentioned it except two different models mentioned something here. I'm not sure how germane it is.

Do you have a copy of the demo database on which to try this change?

src/__tests__/proposals.int.test.ts
```diff
@@ -1520,6 +1520,12 @@ describe('/proposals', () => {
 			const grants = await loadPermissionGrantBundle(
 				db,
 				getAuthContext(systemUser, true),
+				undefined,
```
bickelj, line 1523: This might be more related to the bundle function generator than your call here, but is there any way to avoid explicit `undefined`s? Does limit and offset have to be the last args or could the new optional args be the last ones and therefore not need the `undefined`?

bickelj (commented):
I tried this change in the preview environment and the filters seemed to work just fine and as designed. This is a wonderful change but given the recent performance concern and two models flagging a potential performance issue, I would like to double-check that performance is as expected using a demo-like database. I didn't notice any big performance hit when trying the preview environment, but I don't recall a performance issue being found in that one before.

@hminsky2002 I'll send you a database backup in case you don't already have it.

src/handlers/permissionGrantsHandlers.ts
hminsky2002: Hmm, perhaps better to drop the quotes, i think it's probably still clear.

src/database/queries/permissionGrants/selectWithPagination.sql
hminsky2002: I have a copy of the demo database now, thank you very much!

src/__tests__/proposals.int.test.ts
hminsky2002: My understanding is that that's defined by the generator. I agree that five consecutive undefined-s do not read very well. Conversely, I do think that the current model allowing us to slot in the values here did make it easier for this PR. Perhaps some insight from @slifty would be helpful?

src/database/queries/permissionGrants/selectWithPagination.sql
hminsky2002: After running some queries locally and running a couple permissions queries, I didn't notice anything outlandish. I asked my own llm (claude opus 4.8) about it and it spat back at me a whole spiel. Admittedly, my understanding here is week. I guess it's that we are not currently using indexes in this table, and that both the qwen and glm changes would compile to the same operations in postgres anyway?

> 
> ## Benchmarked the CASE-WHEN filter concern — no change needed
> 
> I restored our demo database (PostgreSQL 18) and benchmarked the filter block directly. TL;DR: the "non-sargable" concern doesn't apply to how this service executes queries, the suggested OR-pattern rewrite is plan-identical, and at current scale the filters are noise either way.
> 
> ### 1. The premise doesn't apply to our wire protocol
> 
> We create TinyPg without `use_prepared_statements` (`src/database/db.ts`), so every query is sent as an *unnamed* statement — Postgres plans it fresh with the real parameter values each time (a custom plan) and constant-folds the CASE expressions away. Running the PR's actual filter block against the demo DB with `plan_cache_mode = force_custom_plan`, the plan shows all six CASE conjuncts reduced to a single plain predicate:
> 
> ```
> Filter: (changemaker_id = 1)
> ```
> 
> The five unused (NULL) filters vanished entirely. Only under `force_generic_plan` — a mode this app never triggers, since generic plans require *named* prepared statements — does the opaque `CASE WHEN ($3 IS NULL) …` chain survive into the plan.
> 
> ### 2. At scale, the folded predicate uses indexes — and the OR rewrite is plan-identical
> 
> On a synthetic 1M-row table with a btree index on `changemaker_id`:
> 
> | Pattern | Plan mode | Measured result |
> |---|---|---|
> | CASE (this PR) | custom | Index Only Scan, `Index Cond: (changemaker_id = 42)`, **0.078 ms** |
> | CASE | generic | full scan (`Rows Removed by Filter: 333300` × 3 workers), **19.4 ms** |
> | OR (suggested rewrite) | custom | Index Only Scan, **0.038 ms** — identical shape |
> | OR | generic | same full scan (`Filter: (($1 IS NULL) OR …)`), **14.2 ms** — same weakness |
> 
> Both patterns are sargable under custom plans and non-sargable under generic plans. Switching buys nothing; dynamic SQL would likewise buy nothing while abandoning our file-based tinypg query convention.
> 
> ### 3. The concern is moot at current scale anyway
> 
> The demo DB has **682** `permission_grants` rows, and the table has **no indexes on any of the six filtered columns** (only the PK and two grantee-identity partial indexes) — so even a perfectly sargable predicate seq-scans today. The endpoint's measured cost is dominated by the `permitted_permission_grant_ids()` auth join, not the filters (~13–16 ms total on demo data, filters or not).
> 
> ### 4. The GIN note is correct but premature
> 
> Verified: `'edit' = ANY(verbs)` cannot use any index (parallel seq scan, 31.5 ms at 1M rows). A GIN index only becomes usable if the operator is rewritten to containment — `verbs @> ARRAY['edit']` did use the GIN index (Bitmap Index Scan). That's the right move *if* verb filtering ever becomes hot on a large table, done together with btree indexes on the other filter columns — a future scaling task, not this PR.
> 
> ### Bottom line
> The pattern matches the existing `GET /proposals` house style and, under the app's actual protocol, executes as an index-friendly plain equality (demonstrated via `EXPLAIN ANALYZE` on the real demo database and a 1M-row stress table). If permission grants ever grow by orders of magnitude, the remedy is indexes, not query-shape changes.
>

bickelj: Huzzah, fight LLM with LLM! Sounds like a false positive. Sorry. Thanks for taking a look and having another LLM do the same.

bickelj (approved):
LGTM, thank you again for looking at the potential performance regression even though it turned out to not be an issue!

src/database/queries/permissionGrants/updateById.sql
```diff
@@ -18,7 +18,7 @@ SET
 	changemaker_field_value_id = :changemakerFieldValueId::integer,
 	terminology_set_id = :terminologySetId::integer,
 	scope = :scope::permission_grant_entity_type_t [],
-	verbs = :verbs::permission_grant_verb_t [],
+	verbs = :verb s::permission_grant_verb_t [],
```
bickelj: Is this the issue?

hminsky2002: agh yes, good catch thank you! somehow the sqlfluff version i have does that on lint. need to debug that.

PR #2465, opened 2026-06-23
Link to glossary from ERD

kfogel:
This is a useful change in its own right, but the real motivation for it is to test that autopublication to the PDC web site works as per https://github.com/philanthropyDataCommons/meta/issues/170.


bickelj (approved):
LGTM!

PR #2467, opened 2026-06-25
Add handling for explicit changemaker column in bulk uploader

### LINKED ISSUE #2466: Add explicit single identifier for changemaker in bulk upload
Service-side portion of issue that was brought up in [front-end issue #1389](https://github.com/PhilanthropyDataCommons/front-end/issues/1389)

As identified by @jmergy, the bulk uploader will create a duplicate changemaker instead of reusing an existing one because the service matches changemakers on both tax ID and name. This is currently by design, since distinct orgs can share a tax ID, but for simplicity purposes we'd like a way to assure that an existing changemaker will be used on upload that only relies on a single field. After discussing, we've decided that the best approach is to add an additional field, `pdc_changemaker_id`, which will always be considered the top authority on the changemaker that should be related in the bulk-upload.

This issue was brought up in the front-end but is a service change. It will be worth documenting this explicitly in the front-end bulk uploader interface as well!

cc @kfogel @slifty

hminsky2002:
This commit adds another 'special' column for bulk uploads, `pdc_changemaker_id`, which is considered the highest truth to identify a changemaker in a bulk upload. This is to allow there to be a single source for identifying the related changemaker in a bulk upload, as opposed to being matched on name AND tax id.

Closes #2466

### DIFF
--- a/src/tasks/processBulkUploadTask.ts
+++ b/src/tasks/processBulkUploadTask.ts
@@ -21,6 +21,7 @@ import {
 	createProposalFieldValue,
 	createProposalVersion,
 	loadBulkUploadTask,
+	loadChangemaker,
 	loadOrCreateChangemaker,
 	updateBulkUploadTask,
 	loadSystemUser,
@@ -30,10 +31,12 @@ import {
 import {
 	TaskStatus,
 	getSelfManageGrantFragment,
+	isId,
 	isProcessBulkUploadJobPayload,
 	BaseFieldDataType,
 	PermissionGrantEntityType,
 } from '../types';
+import { InputValidationError } from '../errors';
 import { fieldValueIsValid } from '../fieldValidation';
 import { allNoLeaks } from '../promises';
 import { SINGLE_STEP } from '../constants';
@@ -52,6 +55,7 @@ import type {
 
 const CHANGEMAKER_TAX_ID_SHORT_CODE = 'organization_tax_id';
 const CHANGEMAKER_NAME_SHORT_CODE = 'organization_name';
+const CHANGEMAKER_PDC_ID_SHORT_CODE = 'pdc_changemaker_id';
 const SINGLE_ENTRY = 1;
 
 const downloadFileDataToTemporaryStorage = async (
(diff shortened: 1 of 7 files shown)


src/tasks/processBulkUploadTask.ts
```diff
@@ -594,8 +605,27 @@ export const processBulkUploadTask = async (
 				const {
 					[changemakerNameIndex]: changemakerName,
 					[changemakerTaxIdIndex]: changemakerTaxId,
+					[changemakerPdcIdIndex]: changemakerPdcIdValue,
 				} = record;
-				if (changemakerTaxId !== undefined) {
+				const changemakerPdcId = changemakerPdcIdValue?.trim();
```
bickelj, line 611: Elsewhere I don't think we trim (in order to alert the caller to a potential bug in the CSV generation process) but I when looking at this I think it is OK to have this quality-of-life trim.

src/tasks/processBulkUploadTask.ts
```diff
@@ -594,8 +605,27 @@ export const processBulkUploadTask = async (
...
+				if (changemakerPdcId !== undefined && changemakerPdcId !== '') {
+					const changemakerId = Number(changemakerPdcId);
+					if (!isId(changemakerId)) {
+						throw new Error(
```
bickelj: Something more specific than `Error` would be good, like `InputValidationError` or even a fresh new Error.

bickelj (changes requested):
While I haven't actually tested it yet, I think this looks like a great quality of life improvement that will prevent extra changemakers from appearing on bulk upload! One small change needed around throwing an Error.

src/tasks/processBulkUploadTask.ts
hminsky2002: yep, it's a good call!

hminsky2002:
@bickelj should be ready for your re-review, relevant error now being thrown!

bickelj (approved):
I intended to do a more thorough test of this in the preview environment but events lead me to mash approve without doing so.

Make sure to test it in the sandbox before announcing availability, per usual.

Thanks for this change!

PR #2474, opened 2026-06-28
Improve performance of `proposal_to_json`

### LINKED ISSUE #2473: Address `GET /proposals` slowness
We recently made an improvement to how permission queries are written; though it was a step in the right direction it resulted in a gnarly query plan when loading proposals (particularly for users with access to a lot of proposal field values).

The root issue is a { waves hands at SQL }.  Basically: I was under the impression that the function queries would be expanded into their callers, and then things like `WHERE foo.bar = ...` would be used by the query planner to narrow the candidate entities BEFORE checking permissions.  This is the case in some places, and that's why it appears to be a valid approach.

Unfortunately in other places (e.g. when loading proposals), the planner goes the other direction: calculate the list of ALL possible values you can see, and then narrow by conditions.  This becomes incredibly expensive.

It appears we can update our permission SQL functions to take the specific set of objects that we want to narrow based on a user's permission.  I'm going to apply this approach specifically to the `proposal_field_values` permission checks to test the pattern.  Ultimately we should have a common pattern for the other permission check functions too.

It's also possible that I'll find a better solution while working on the problem.

slifty:
This PR addresses an issue where the query planner would not effectively narrow the potential list of proposal field value IDs when loading field value permissions in the context of proposal collections.

Resolves #2473

### DIFF
--- a/src/database/initialization/1_changemaker_to_json.sql
+++ b/src/database/initialization/1_changemaker_to_json.sql
@@ -26,6 +26,16 @@ BEGIN
 	WHERE NOT shallow;
 
 	-- Combine ProposalFieldValues and ChangemakerFieldValues, select gold per base field
+	WITH changemaker_proposal_field_value_ids AS (
+		SELECT pfv.id
+		FROM proposal_field_values pfv
+		INNER JOIN proposal_versions pv
+			ON pfv.proposal_version_id = pv.id
+		INNER JOIN changemakers_proposals op
+			ON pv.proposal_id = op.proposal_id
+		WHERE op.changemaker_id = changemaker.id
+	)
+
 	SELECT jsonb_agg(field_value_json)
 	INTO field_values_json
 	FROM (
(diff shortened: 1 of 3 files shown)


bickelj:
@slifty Could you paste some before-and-after query plans that you're seeing?

bickelj (approved):
I see nothing objectionable, let's merge.

Caveat: I don't think this resolves #2473 entirely because this is only one of many queries that have performance issues in the demo environment.

bickelj:
I can say that calling `curl https://utilities.philanthropydatacommons.org/pdc-service-preview-2474-ojgm4/proposals?_page=1&_count=2147483647` consistently takes near 800ms while the same call against the sandbox API consistently takes near 2000ms. I alternated curl calls back and forth > 5 times to get this.

While the databases and environments are not exactly commensurate they are probably close enough because the preview environments are copied from a past snapshot of sandbox. The preview env has slightly less data due to data accrual since the snapshot. On the other hand the preview env probably adds a millisecond or two due to proxying per request. But more than halving the request time is unlikely to be a coincidence.

hminsky2002 (approved):
Seems reasonable to me! I will admit to taking the sql in `permitted_proposal_field_value ids among`at "tests are passing", but the principle behind the change seems sound. Otherwise seems like it is being applied in the right places!

Issue #2475, opened 2026-06-28
Reconsider N+1 query approach for deep entities


slifty:
Right now we return entities with "related" entities included.  This is convenient for clients, but the current implementation results in N+1 queries (loading related entities and attaching them).

I'd like to see a solution where (1) we run one query per entity type collected, and (2) a client can specify which types of entities they want collected.  

Whether entities are still returned with nested values or results are returned as shallow references plus a URI => entity mapping for related entities is up for discussion (I'd err on the second, as it bypasses concerns around potential recursive references).

slifty:
This is now resolved for both changemakers and proposals, which were the two entities that truly needed a refactoring due to their reliance on field values.  I'm tempted to close this since that was the driving demand for this issue; we have reconsidered, and we have identified the correct short term refactor and applied that refactor to the two most pressing entities.  The refactor should arguably be applied to all entities, however.

Also, I do expect that future API design change that will allow a caller to specify which "deep objects" we want to return in a result set.  I consider that to be a *related* concept, but not part of this issue.

All this is to say I'm going to mark this as closed + done.  At our next planning session we can decide, when reviewing, whether we want to make individual issues for the remaining entities that still use n+1 internally.

Issue #2023, opened 2025-10-09
Upgrade synthetic IDs to bigint


bickelj:
Given that we expect production usage with potentially large datasets, it would benefit everybody to get ahead of `int` exhaustion in tables where we use a synthetic auto-incrementing integer, particularly `proposal_field_values`. If we add, for example, 10 million proposals with 200 fields each, we would be (almost) at the limit of 2^31-1. https://www.postgresql.org/docs/17/datatype-numeric.html

The complicating factor here, of course, is JSON representation, which can only accurately represent an integer (without quoting) far smaller than 2^63-1. We might need to add a check in PostgreSQL when that limit is hit and validate in our APIs. Still, 2^53-1 is a much better situation than 2^31-1.

PR #2477, opened 2026-06-29
Batch proposal serialization

slifty:
Our json serializer ran on individual entities which was resulting in a ton of internal + duplicate calls.  This change moves towards a more efficient model where entity sets are serialized together.  We're starting with proposals, but will move into lower proposal entities as well.

Related to #2473
Related to #2475

### DIFF
--- a/src/database/initialization/1_proposal_field_value_file_to_json.sql
+++ b/src/database/initialization/1_proposal_field_value_file_to_json.sql
@@ -1,0 +1,17 @@
+SELECT drop_function('proposal_field_value_file_to_json');
+
+-- The file a proposal field value points at, or null. A field value references
+-- a file when its base field is a file field and its value is the file's id;
+-- the file is only revealed to the version's own creator.
+CREATE FUNCTION proposal_field_value_file_to_json(
+	proposal_field_value proposal_field_values,
+	is_file_field boolean,
+	version_created_by uuid
+) RETURNS jsonb AS $$
+	SELECT file_to_json(files.*)
+	FROM files
+	WHERE proposal_field_value_file_to_json.is_file_field
+		AND proposal_field_value.value ~ '^[0-9]+$'
+		AND files.id = proposal_field_value.value::integer
+		AND files.created_by = proposal_field_value_file_to_json.version_created_by;
+$$ LANGUAGE sql STABLE;
--- a/src/database/initialization/1_assert_proposal_field_value_not_forbidden.sql
+++ b/src/database/initialization/1_assert_proposal_field_value_not_forbidden.sql
@@ -1,0 +1,26 @@
+SELECT drop_function('assert_proposal_field_value_not_forbidden');
+
+-- Raises if the field value belongs to a forbidden base field, so a forbidden
+-- value is never serialized. Read paths exclude forbidden fields via the
+-- permitted_* functions; this guards the paths that serialize a row directly
+-- (the field value insert), which do not filter by permission.
+CREATE FUNCTION assert_proposal_field_value_not_forbidden(
+	proposal_field_value proposal_field_values
+) RETURNS void AS $$
+BEGIN
+	IF EXISTS (
+		SELECT 1
+		FROM application_form_fields
+		INNER JOIN base_fields
+			ON base_fields.short_code = application_form_fields.base_field_short_code
+		WHERE application_form_fields.id
+			= proposal_field_value.application_form_field_id
+			AND base_fields.sensitivity_classification = 'forbidden'
+	) THEN
+		RAISE EXCEPTION
+			'Refusing to serialize forbidden proposal_field_value (%)',
+			proposal_field_value.id
+			USING ERRCODE = '22023'; -- invalid_parameter_value
+	END IF;
+END;
+$$ LANGUAGE plpgsql STABLE;
--- a/src/database/initialization/1_build_proposal_field_value_result.sql
+++ b/src/database/initialization/1_build_proposal_field_value_result.sql
@@ -1,0 +1,41 @@
+SELECT drop_function('build_proposal_field_value_result');
+
+-- Gathers a single proposal field value's children (application form field,
+-- file) and assembles the result with the proposal_field_value_to_json shape.
+-- For the single-row endpoints; the proposals read path gathers field values in
+-- bulk instead. Guards against forbidden fields, since callers may serialize a
+-- row that has not been permission-filtered (the field value insert).
+CREATE FUNCTION build_proposal_field_value_result(
+	proposal_field_value proposal_field_values
+) RETURNS jsonb AS $$
+DECLARE
+	is_file_field BOOLEAN;
+	application_form_field_json JSONB;
+	version_created_by UUID;
+BEGIN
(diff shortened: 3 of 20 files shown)


bickelj:
Again while not exactly commensurate environments, a good sign comes from comparing the time it takes to get proposals in the preview environment versus the sandbox environment.

Example call for preview `time curl -X 'GET'   'https://utilities.philanthropydatacommons.org/pdc-service-preview-2477-d5z2o/proposals?_page=1&_count=2147483647' ... > /dev/null` consistently gives less than a second, usually around 800ms give or take, while the same call to sandbox consistently gives more than a second, usually around 1200ms give or take. So that is in the right direction.

src/database/initialization/1_proposal_field_value_file_to_json.sql
```diff
@@ -0,0 +1,17 @@
+SELECT drop_function('proposal_field_value_file_to_json');
+
+-- The file a proposal field value points at, or null. A field value references
+-- a file when its base field is a file field and its value is the file's id;
+-- the file is only revealed to the version's own creator.
```
bickelj, line 5: Why?

1. I thought we had a policy not to use `created_by` for permissions, only for provenance.
2. If the user has access to the proposal data, why wouldn't the file be included?

bickelj:
The size of this change makes it difficult to review quickly. It also is outside my postgres knowledge. Until I can get a bit more time to look, I will paste what my current LLM said. I don't think the first "critical" point is accurate (or else tests would fail), nor is the third "critical" critical, but the remainder may be worth considering.

<details><summary>Qwen3.6 30B A3B review</summary>
Here is a review of the provided SQL and TypeScript test changes, categorized as requested.

### 🔴 Critical Show-Stoppers

1. **`src/database/queries/proposals/insertOne.sql`: Invalid Array Constructor Cast**
   ```sql
   array(SELECT inserted_proposal::proposals FROM inserted_proposal)
   ```
   PostgreSQL does not allow casting a composite row alias to its type inside an `array()` constructor. This will throw a syntax/runtime error (e.g., `ERROR: cannot cast record to type proposals`).
   **Fix:** Remove the cast. The CTE already returns the correct composite type:
   ```sql
   array(SELECT inserted_proposal FROM inserted_proposal)
   ```

2. **`src/database/initialization/1_proposal_field_value_file_to_json.sql`: Integer Overflow Risk**
   ```sql
   AND files.id = proposal_field_value.value::integer
   ```
   The regex `proposal_field_value.value ~ '^[0-9]+$'` ensures the string contains only digits, but it does not enforce length. If a value like `'9999999999'` is stored, the regex passes, but the cast to `integer` will throw `integer out of range`, crashing the request.
   **Fix:** Cast to `bigint` (or validate length):
   ```sql
   AND files.id = proposal_field_value.value::bigint
   ```

3. **`src/database/initialization/5_build_proposals_results.sql`: Non-Deterministic `ORDER BY` with NULLs**
   ```sql
   ORDER BY pfv.position, pfv.id DESC
   ```
   If `position` is nullable (which is common in application form fields), PostgreSQL will place `NULL`s first. This can cause non-deterministic ordering for field values that share an ID (though IDs are unique, position grouping is expected). Frontend consumers may rely on strict ordering.
   **Fix:** Explicitly handle NULLs:
   ```sql
   ORDER BY pfv.position NULLS LAST, pfv.id DESC
   ```

---

### 🟡 Minor Nice-to-Haves

1. **Redundant Joins in `5_build_proposals_results.sql`**
   The `field_value_json` CTE joins `application_form_fields` and `base_fields` twice: once via the `application_form_field_json` CTE, and again directly to get `bf.data_type`. Since `application_form_field_to_json` likely already incorporates base field metadata, you could either extract `data_type` from the pre-joined JSON or remove the redundant joins to simplify the query plan.

2. **Permission Function Consistency (`selectById.sql`)**
   The single-row queries for `proposalFieldValues` and `proposalVersions` still use `permitted_*_ids` (unbounded) instead of the new `permitted_*_ids_among`. For consistency and potential planner improvements, consider passing a single-element array to `_among`:
   ```sql
   AND proposal_field_values.id IN (
       SELECT permitted_field_values.id
       FROM permitted_proposal_field_value_ids_among(
           :authContextKeycloakUserId,
           :authContextIsAdministrator,
           'view',
           'proposalFieldValue',
           ARRAY[proposal_field_values.id]
       ) AS permitted_field_values
   );
   ```

3. **Array Deduplication in `build_proposals_results`**
   `ARRAY(SELECT ...)` from joined CTEs may contain duplicate IDs if the underlying relationships produce them. While `= ANY()` handles duplicates gracefully, adding `DISTINCT` prevents redundant permission lookups and keeps array payloads lean:
   ```sql
   ARRAY(SELECT DISTINCT pfv.id FROM proposal_field_values pfv ...)
   ```

4. **CTE Materialization in `build_proposals_results`**
   The function is `LANGUAGE sql STABLE`. In complex `LANGUAGE sql` functions, the PostgreSQL planner sometimes unnests CTEs instead of materializing them, which can degrade performance. Consider adding `MATERIALIZED` to `target_versions` and `permitted_field_values` if they are expected to grow large:
   ```sql
   target_versions AS MATERIALIZED ( ... )
   ```

5. **Test Coverage for Guard Integration**
   The test for forbidden fields now calls `assert_proposal_field_value_not_forbidden` directly. While valid, consider adding a test case for `build_proposal_field_value_result` to ensure the guard behaves correctly when integrated into the full builder (e.g., verifying it doesn't mask other validation errors or incorrectly swallow exceptions).

6. **`PERFORM` vs Direct Exception in Builders**
   `build_proposal_field_value_result` and `build_proposal_version_result` use `PERFORM assert_...` before proceeding. This is functionally correct, but inlining the check or using `RAISE EXCEPTION ... USING ERRCODE` directly in the builder could slightly reduce PL/pgSQL function call overhead on hot paths. (Likely negligible, but worth noting for extreme QPS scenarios.)

### ✅ Overall Assessment
The architectural shift to **Shape vs. Result Builder** is well-executed and aligns with PostgreSQL best practices for bulk serialization. The separation of table-free JSON assembly (`LANGUAGE sql IMMUTABLE`) from data gathering (`LANGUAGE plpgsql STABLE`) correctly leverages the planner's inlining capabilities while keeping permission logic bounded. The primary risks are the two runtime errors noted above (array cast and integer overflow), which should be fixed before merging. The rest are performance/hygiene refinements.

</details>

src/database/initialization/1_proposal_field_value_file_to_json.sql
bickelj: It is subtle but there it is, verifying that the same user created the pfv and the f.

bickelj (approved):
Given the urgency and that all the tests seem to be passing (and no alarming changes to the tests), I am mashing that approve button, trusting the tests and my earlier experiments with performance vs the preview environment. Take or leave whatever the LLM above gave.

slifty:
Updated a thing based on the LLM (some of those didn't seem to fit perfectly; we can revisit later, and the bigint one is something I think will have other instances e.g. via #2023 )

PR #2478, opened 2026-06-30
Add control columns to bulk uploader

### LINKED ISSUE #2476: Reference pdc_changemaker_id as control data in Bulk Upload, to avoid validation problem w.r.t. ApplicationForm
In #2466 we implemented a special behavior in Bulk Upload based on the new optional `pdc_changemaker_id` column: if the column is present, then treat the Changemaker corresponding to that row as the one identified by the value in that column.  This is fine in theory, but in practice, no ApplicationForm has that field, so the Bulk Upload CSV will fail validation (as seen [here](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/Bulk.20Uploads/near/263468)).  Unlike all the other columns, which correspond to fields in the ApplicationForm and thence to base fields, this one isn't *really* a base field, and even if we were to have such a base field some day -- who knows, some foundations' applications might start at least optionally referencing PDC changemaker IDs -- we certainly can't assume that every ApplicationForm is going to have it.  Uploaders need to be able to use the `pdc_changemaker_id` at will, without regard to whether the Opportunity's ApplicationForm has it.

**Solution** (designed in a phone call with @hminsky2002, @slifty, and me just now):

Make a new class of column available in the Bulk Upload CSV: "control data".  In other contexts this is sometimes called "pragma" information, or maybe "keywords", "instructions", "directives", "commands", "overrides", etc.  Basically, it's meta-information that changes *how* Bulk Upload processes the information in the CSV.

It will be indicated in the column name with a "control:" prefix before the field name, so in this case the column name would be:

```
control:pdc_changemaker_id
```

We're open to a different prefix, if anyone has any inspired ideas.  The separator will be colon, because that's not a permitted character in a base field shortcode, which can only be made of alphanumerics and underscores.

(This is informationally equivalent to having a separate CSV with optional control data -- e.g., it would have the same number of rows as the main Bulk Upload CSV, and only have values in the cells where control data is being offered.  But it's a *lot* more human-friendly to just put the control data in the main CSV and use a prefix to distinguish those columns, so that's what we went with.)

This will allow the Bulk Upload process to immediately recognize control columns and treat each one in its appropriate way, while exempting them from validation against the ApplicationForm as needed.

Note that this allows us to make `pdc_changemaker_id` a base field some day.  The whole point of the `control:` prefix is that we can use terms without interference from base field semantics.  For example, if a Bulk Upload CSV were to have *both* `pdc_changemaker_id` and  `control:pdc_changemaker_id` as columns, that would be fine -- there is no inherent contradiction there.  Obviously we would have to decide what the behavior should be if both have values present and those values are not equal, but that's fine -- it's a "denotational semantics" vs "operational semantics" distinction.  (Right now, we are also able to determine the operational semantics for `control:pdc_changemaker_id` because `pdc_changemaker_id` is *not* actually a base field and thus no conflict can arise.)

For the future (not for this issue): Dan mentioned that we might want to consider allowing (or even encouraging?) Changemaker name and Tax ID to be presented as "control" data as well.  That's an interesting idea, and worth further discussion, but it doesn't affect this issue and we won't block on it.

hminsky2002:
Closes #2476

### DIFF
--- a/src/tasks/__tests__/processBulkUploadTask.int.test.ts
+++ b/src/tasks/__tests__/processBulkUploadTask.int.test.ts
@@ -37,6 +37,7 @@ import {
 	expectObject,
 	expectObjectContaining,
 	expectString,
+	expectStringContaining,
 	expectTimestamp,
 } from '../../test/asymettricMatchers';
 import {
@@ -128,12 +129,6 @@ const createTestBaseFields = async (db: TinyPg): Promise<void> => {
 		description: 'Just a file we want to attach.',
 		shortCode: 'favorite_file',
 		dataType: BaseFieldDataType.FILE,
-	});
-	await createTestBaseField(db, null, {
-		label: 'PDC Changemaker ID',
-		description:
-			'The PDC id of an existing changemaker to attach this proposal to.',
-		shortCode: 'pdc_changemaker_id',
 	});
 };
 
@@ -978,7 +973,7 @@ describe('processBulkUploadTask', () => {
 		const { applicationFormId } = await createTestApplicationForm(
 			db,
 			systemUserAuthContext,
-			['proposal_submitter_email', 'organization_name', 'pdc_changemaker_id'],
+			['proposal_submitter_email', 'organization_name'],
 		);
 		const bulkUploadTask = await createTestBulkUploadTask(
 			db,
@@ -1013,13 +1008,21 @@ describe('processBulkUploadTask', () => {
 		);
 		// No new changemaker is created and the existing name is left untouched,
 		// even though the uploaded row carried a different organization_name.
-		expect(changemakerBundle.total).toBe(1);
-		expect(changemakerBundle.entries).toEqual([
-			expectObjectContaining({
-				id: existingChangemaker.id,
-				name: 'Original Changemaker Name',
-			}),
-		]);
+		expect(changemakerBundle).toEqual({
+			entries: [
+				{
+					createdAt: expectTimestamp(),
+					createdBy: systemUser.keycloakUserId,
+					taxId: '11-1111111',
+					id: existingChangemaker.id,
+					name: 'Original Changemaker Name',
+					keycloakOrganizationId: null,
+					fields: [],
+					fiscalSponsors: [],
+				},
+			],
+			total: 1,
+		});
 
 		const changemakerProposalBundle = await loadChangemakerProposalBundle(
 			db,
@@ -1029,10 +1032,19 @@ describe('processBulkUploadTask', () => {
 			NO_LIMIT,
 			NO_OFFSET,
 		);
-		expect(changemakerProposalBundle.total).toBe(1);
-		expect(changemakerProposalBundle.entries).toEqual([
-			expectObjectContaining({ changemakerId: existingChangemaker.id }),
-		]);
+		expect(changemakerProposalBundle).toEqual({
+			entries: [
+				{
+					createdAt: expectTimestamp(),
+					id: expectNumber(),
+					changemakerId: existingChangemaker.id,
+					proposalId: expectNumber(),
+					changemaker: expectObject(),
+					proposal: expectObject(),
+				},
+			],
+			total: 1,
+		});
 
 		const updatedBulkUploadTask = await loadBulkUploadTask(
 			db,
@@ -1057,12 +1069,7 @@ describe('processBulkUploadTask', () => {
 		const { applicationFormId } = await createTestApplicationForm(
 			db,
 			systemUserAuthContext,
-			[
-				'proposal_submitter_email',
-				'organization_name',
-				'organization_tax_id',
-				'pdc_changemaker_id',
-			],
+			['proposal_submitter_email', 'organization_name', 'organization_tax_id'],
 		);
 		const bulkUploadTask = await createTestBulkUploadTask(
 			db,
@@ -1097,10 +1104,21 @@ describe('processBulkUploadTask', () => {
 		);
 		// The tax id matched nothing, but no new changemaker is created because the
 		// explicit pdc_changemaker_id takes precedence.
-		expect(changemakerBundle.total).toBe(1);
-		expect(changemakerBundle.entries).toEqual([
-			expectObjectContaining({ id: existingChangemaker.id }),
-		]);
+		expect(changemakerBundle).toEqual({
+			entries: [
+				{
+					createdAt: expectTimestamp(),
+					createdBy: systemUser.keycloakUserId,
+					taxId: '11-1111111',
+					id: existingChangemaker.id,
+					name: 'Original Changemaker Name',
+					keycloakOrganizationId: null,
+					fields: [],
+					fiscalSponsors: [],
+				},
+			],
+			total: 1,
+		});
 
 		const changemakerProposalBundle = await loadChangemakerProposalBundle(
 			db,
@@ -1110,9 +1128,19 @@ describe('processBulkUploadTask', () => {
 			NO_LIMIT,
 			NO_OFFSET,
 		);
-		expect(changemakerProposalBundle.entries).toEqual([
-			expectObjectContaining({ changemakerId: existingChangemaker.id }),
-		]);
+		expect(changemakerProposalBundle).toEqual({
+			entries: [
+				{
+					createdAt: expectTimestamp(),
+					id: expectNumber(),
+					changemakerId: existingChangemaker.id,
+					proposalId: expectNumber(),
+					changemaker: expectObject(),
+					proposal: expectObject(),
+				},
+			],
+			total: 1,
+		});
 	});
 
 	it('fails the task when pdc_changemaker_id is not a valid id', async () => {
@@ -1125,7 +1153,7 @@ describe('processBulkUploadTask', () => {
 		const { applicationFormId } = await createTestApplicationForm(
 			db,
 			systemUserAuthContext,
-			['proposal_submitter_email', 'organization_name', 'pdc_changemaker_id'],
+			['proposal_submitter_email', 'organization_name'],
 		);
 		const bulkUploadTask = await createTestBulkUploadTask(
 			db,
@@ -1176,7 +1204,7 @@ describe('processBulkUploadTask', () => {
 		const { applicationFormId } = await createTestApplicationForm(
 			db,
 			systemUserAuthContext,
-			['proposal_submitter_email', 'organization_name', 'pdc_changemaker_id'],
+			['proposal_submitter_email', 'organization_name'],
 		);
 		const bulkUploadTask = await createTestBulkUploadTask(
 			db,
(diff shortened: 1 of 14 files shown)


src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -1210,6 +1199,240 @@ describe('processBulkUploadTask', () => {
...
+			db,
+			testAuthContext,
+			bulkUploadTask.id,
+		);
+		expect(updatedBulkUploadTask.status).toEqual(TaskStatus.COMPLETED);
+
+		const proposalBundle = await loadProposalBundle(
+			db,
+			testAuthContext,
+			undefined,
```
bickelj, line 1291: Shouldn't there be a way to invoke this function without literally writing `undefined` as an argument?

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -1210,6 +1199,240 @@ describe('processBulkUploadTask', () => {
...
+		);
+		const fieldValues = proposalBundle.entries
+			.flatMap((proposal) => proposal.versions)
+			.flatMap((version) => version.fieldValues)
+			.sort((a, b) => a.position - b.position);
+		expect(fieldValues).toEqual([
+			expectObjectContaining({
+				position: 0,
+				value: 'foo@example.com',
+				applicationFormField: expectObjectContaining({
```
bickelj: Instead of `expectObjectContaining`, which is a loose assertion, I think it would be better to start with a strict assertion and if it becomes necessary to relax it later, relax it later at that time.

src/tasks/processBulkUploadTask.ts
```diff
@@ -648,54 +729,60 @@ export const processBulkUploadTask = async (
...
-							},
-						);
-						await createPermissionGrant(transactionDb, taskAuthContext, {
-							...getSelfManageGrantFragment(taskAuthContext),
-							contextEntityType: PermissionGrantEntityType.PROPOSAL_FIELD_VALUE,
-							proposalFieldValueId: proposalFieldValue.id,
-						});
-						return proposalFieldValue;
-					}),
+							const proposalFieldValue = await createProposalFieldValue(
```
bickelj, line 785: A lot of the lines around here look similar or the same, is this a new formatting rule that changed it? Oh, new function signature increased indentation.

src/tasks/processBulkUploadTask.ts
```diff
@@ -122,43 +143,93 @@ const loadHeaderRowFromCsv = async (csvPath: string): Promise<string[]> => {
...
+	controlColumns.forEach(({ csvIndex, key }) => {
+		const value = record[csvIndex]?.trim() ?? '';
+		if (value !== '') {
+			controlValues.set(key, value);
+		}
+	});
+	return controlValues;
+};
+
 const assertCsvMatchesApplicationForm = async (
```
bickelj: Now that this is no longer a "throw an error when invalid or return void" but returns a thing, I think a new name like `validate` or something along those lines would make more sense than `assert`.

src/tasks/processBulkUploadTask.ts
```diff
@@ -648,54 +729,60 @@ export const processBulkUploadTask = async (
...
 
 				await allNoLeaks(
-					record.map<Promise<ProposalFieldValue>>(async (fieldValue, index) => {
-						const { [index]: applicationFormField } = applicationFormFields;
-						if (applicationFormField === undefined) {
-							throw new Error(
-								'There is no form field associated with this column',
-							);
-						}
+					dataColumns.map<Promise<ProposalFieldValue>>(
```
bickelj: A `map` on columns instead of rows seems like a radical change, yet somehow it works? How about for a test file with two rows?

bickelj (changes requested):
A few minor notes are in band. The only thing that gives me pause is the iteration/map on columns rather than rows. If it's really rows, it should be renamed. If it's columns, something may be wrong there.

Qwen3.6 (A3B) had some feedback from which I'm cherry-picking here:

Based on the provided `AGENTS.md` guidelines and the `git show` diff, here is a structured code review. I've focused on correctness, adherence to project conventions, and potential runtime pitfalls.

BEGIN LLM SLOP

### Issues that could cause runtime failures, incorrect behavior, or break tests/linting.

...

**B. `record[csvIndex] ?? ''` vs `record[csvIndex]` Inconsistency**
In `extractControlValues`, you correctly use `record[csvIndex]?.trim() ?? ''` to handle short rows. However, in `processBulkUploadTask`, you use:
```typescript
const changemakerName = changemakerNameCsvIndex === undefined ? undefined : record[changemakerNameCsvIndex];
const changemakerTaxId = changemakerTaxIdCsvIndex === undefined ? undefined : record[changemakerTaxIdCsvIndex];
```
If a CSV row has fewer columns than the header (a common real-world scenario), `record[csvIndex]` will be `undefined`. This is fine for names/tax IDs, but later in the same loop, you do:
```typescript
const fieldValue = record[csvIndex] ?? '';
```
This inconsistency is acceptable, but be aware that `undefined` values for names/tax IDs will bypass validation and skip fallback logic. Ensure your downstream logic explicitly handles `undefined` for these fields, or normalize them to `''` early to match the rest of the pipeline.

...

### Improvements for readability, maintainability, or strict convention adherence.

...

**B. Error Message Standardization**
The error messages are clear, but consider extracting them to constants or using a centralized error factory if the project has one. For example:
```typescript
throw new Error(`unknown control key: ${key}`);
```
Could be:
```typescript
throw new InputValidationError(`unknown control key: ${key}`, [...]);
```
Since you're already throwing `InputValidationError` for the PDC ID, consider using it here too for consistency.

**C. `RECOGNIZED_CONTROL_KEYS` Extensibility**
You've hardcoded `CONTROL_KEY_PDC_CHANGEMAKER_ID` in `RECOGNIZED_CONTROL_KEYS`. The commit message hints at future control keys (`changemaker name`, `tax id`). Consider making `RECOGNIZED_CONTROL_KEYS` a `const` array or object that can be easily extended, or document that new control keys should be added here. This prevents accidental duplication and makes future additions obvious.

...

**F. Whitespace Trimming in Control Keys**
`classifyHeaderRow` trims whitespace around control keys: `const key = label.slice(CONTROL_COLUMN_PREFIX.length).trim();`. This is good for user-friendliness, but consider logging a warning if whitespace was found, as it indicates a malformed CSV. Alternatively, document this behavior clearly in the changelog or user-facing docs.

END LLM SLOP

src/tasks/processBulkUploadTask.ts
hminsky2002: yeah i also noticed that when reviewing, thought it was strange but seems to be an artifact of prettier on the function handle!

src/tasks/__tests__/processBulkUploadTask.int.test.ts
slifty: Not easily with the current function signature (since the later parameters are needed).

We can / should look at updating our generators to use option objects instead of parameters though, which I personally strongly prefer for (1) this reason and (2) because then you have key / value pairs that make it very clear what each value actually means.

src/tasks/processBulkUploadTask.ts
slifty: +1 to this!

`assert` is used *for cases when an error will be thrown*, if it's just returning a bool we tend to  use `isValidFoo`

src/tasks/processBulkUploadTask.ts
```diff
@@ -122,43 +143,93 @@ const loadHeaderRowFromCsv = async (csvPath: string): Promise<string[]> => {
...
+	});
+};
+
+export const extractControlValues = (
+	record: string[],
+	controlColumns: ControlColumn[],
+): Map<string, string> => {
+	const controlValues = new Map<string, string>();
+	controlColumns.forEach(({ csvIndex, key }) => {
+		const value = record[csvIndex]?.trim() ?? '';
```
slifty: A bit torn on whether we want to trim here (which is implicit support for whitespace) vs explicitly rejecting whitespace with a clear error so users can be directed to format correctly / as expected.

I would probably opt for the explicit requirement + explicit guidance if that requirement is violated in that particular way, but understood if you prefer this approach / that's fine too!

slifty (commented):
This looks really great, thank you Harry -- I see @bickelj has comments so I'll let him issue the final review; I added one thought about whitespace support vs explicit rejection

src/tasks/processBulkUploadTask.ts
hminsky2002: I think explicit requirement is consistent with what we do in other places, as with base field validation. Avoids the scenario where a user actually meant that white space, for whatever reason (and more broadly us storing data that the user didn't technically give us). I'll make the change!

hminsky2002: I added an explicit error when the control column has whitespace and log that as a bulkupload error (in the bulkupload error logs). A little bit of a departure from how we normally do things, but I think it does a nice job of predicting a common error while still not modifying user data! Would take feedback on the implementation there though!

src/tasks/processBulkUploadTask.ts
hminsky2002: Reasonable point, but I think it still just works map runs per row (excluding the column header row). Added a multi-row test!

src/tasks/processBulkUploadTask.ts
```diff
@@ -122,43 +143,103 @@ const loadHeaderRowFromCsv = async (csvPath: string): Promise<string[]> => {
...
+			);
+		}
+		if (seenKeys.has(key)) {
+			throw new Error(
+				`CSV has more than one "${CONTROL_COLUMN_PREFIX}${key}" control column.`,
+			);
+		}
+		seenKeys.add(key);
+	});
+};
```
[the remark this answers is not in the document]
bickelj: I agree with this suggestion, use a more specific Error class.

src/tasks/processBulkUploadTask.ts
```diff
@@ -55,9 +55,30 @@ import type {
...
+
+export interface ControlColumn {
+	csvIndex: number;
+	key: string;
+}
+
+export interface ClassifiedColumns {
+	dataColumns: DataColumn[];
+	controlColumns: ControlColumn[];
+}
```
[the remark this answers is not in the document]
bickelj: Good call.

src/tasks/__tests__/processBulkUploadTask.int.test.ts
bickelj: Named parameters for the win.

src/tasks/processBulkUploadTask.ts
```diff
@@ -122,43 +143,103 @@ const loadHeaderRowFromCsv = async (csvPath: string): Promise<string[]> => {
...
+	recordNumber: number,
+): Map<string, string> => {
+	const controlValues = new Map<string, string>();
+	controlColumns.forEach(({ csvIndex, key }) => {
+		const { [csvIndex]: value } = record;
+		if (value === undefined || value === '') {
+			return;
+		}
+		if (value !== value.trim()) {
+			throw new InputValidationError(
```
bickelj, line 196: Nice!

src/tasks/processBulkUploadTask.ts
bickelj: OK great, I see your test file.

Maybe my beef is with the name, then. If this is a list of columns and I iterate over it, shouldn't I get one column per iteration? But I'm getting a row. I see that we now distinguish between data and control so it kind of makes sense. But this is the one place where it is difficult to read.

bickelj (changes requested):
'Tis so close, I think the copilot/claude review had two worthy comments. I also see a naming thing.

src/tasks/processBulkUploadTask.ts
bickelj: It in fact **is** iterating over columns, because we are already honed in on a row. My bad.

bickelj (approved):
This is looking good, thanks for the iteration!

src/tasks/processBulkUploadTask.ts
```diff
@@ -603,18 +689,28 @@ export const processBulkUploadTask = async (
...
-					[changemakerTaxIdIndex]: changemakerTaxId,
-					[changemakerPdcIdIndex]: changemakerPdcIdValue,
-				} = record;
-				const changemakerPdcId = changemakerPdcIdValue?.trim();
-				if (changemakerPdcId !== undefined && changemakerPdcId !== '') {
+				const controlValues = extractControlValues(
+					record,
+					controlColumns,
+					recordNumber,
+				);
```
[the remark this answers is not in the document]
hminsky2002: Makes sense I suppose, though this isn't really something introduced by this PR.

src/tasks/processBulkUploadTask.ts
```diff
@@ -122,43 +143,105 @@ const loadHeaderRowFromCsv = async (csvPath: string): Promise<string[]> => {
...
+	const { dataColumns } = classifiedColumns;
 	const expectedLabels = applicationForm.fields.map((field) => field.label);
 
-	if (csvLabels.length !== expectedLabels.length) {
+	if (dataColumns.length !== expectedLabels.length) {
 		throw new Error(
-			`CSV has ${csvLabels.length} columns but the application form has ${expectedLabels.length} fields.`,
+			`CSV has ${dataColumns.length} data columns but the application form has ${expectedLabels.length} fields.`,
 		);
 	}
```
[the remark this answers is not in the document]
hminsky2002: Yeah good idea.

Issue #2479, opened 2026-06-30
Address `GET /changemakers` slowness


slifty:
This is a related issue to #2473 and #2475 

We have some inefficient patterns in our SQL for complex objects (particularly those related to field values, which have a lot of logic to determine which to use in a given space).  This becomes a particular issue for the `GET /changemakers` endpoint.

slifty:
This is now complete for now -- the gold data function *is still expensive* but at this point the API seems to be responsive enough.

PR #2489, opened 2026-07-01
Remove N+1 issue for `GET /changemakers`

slifty:
This PR updates our changemaker serialization to remove the N+1 effect when loading changemaker field values.

Related to #2475 
Related to #2479

### DIFF
--- a/src/database/queries/changemakers/insertOne.sql
+++ b/src/database/queries/changemakers/insertOne.sql
@@ -1,16 +1,22 @@
-INSERT INTO changemakers (
-	tax_id,
-	name,
-	keycloak_organization_id,
-	created_by
-) VALUES (
-	:taxId,
-	:name,
-	:keycloakOrganizationId,
-	:authContextKeycloakUserId
-)
-RETURNING changemaker_to_json(
-	changemakers,
+WITH
+	inserted_changemaker AS (
+		INSERT INTO changemakers (
+			tax_id,
+			name,
+			keycloak_organization_id,
+			created_by
+		) VALUES (
+			:taxId,
+			:name,
+			:keycloakOrganizationId,
+			:authContextKeycloakUserId
+		)
+		RETURNING *
+	)
+
+SELECT serialized_changemaker.object
+FROM build_changemakers_results(
+	array(SELECT inserted_changemaker::changemakers FROM inserted_changemaker),
 	:authContextKeycloakUserId,
 	FALSE
-) AS object;
+) AS serialized_changemaker;
--- a/src/database/initialization/1_assert_changemaker_field_value_not_forbidden.sql
+++ b/src/database/initialization/1_assert_changemaker_field_value_not_forbidden.sql
@@ -1,0 +1,23 @@
+SELECT drop_function('assert_changemaker_field_value_not_forbidden');
+
+-- Raises if the field value belongs to a forbidden base field, so a forbidden
+-- value is never serialized. Read paths exclude forbidden fields via the
+-- permitted_* functions; this guards the paths that serialize a row directly.
+CREATE FUNCTION assert_changemaker_field_value_not_forbidden(
+	changemaker_field_value changemaker_field_values
+) RETURNS void AS $$
+BEGIN
+	IF EXISTS (
+		SELECT 1
+		FROM base_fields
+		WHERE base_fields.short_code
+			= changemaker_field_value.base_field_short_code
+			AND base_fields.sensitivity_classification = 'forbidden'
+	) THEN
+		RAISE EXCEPTION
+			'Refusing to serialize forbidden changemaker_field_value (%)',
+			changemaker_field_value.id
+			USING ERRCODE = '22023'; -- invalid_parameter_value
+	END IF;
+END;
+$$ LANGUAGE plpgsql STABLE;
--- a/src/database/initialization/1_build_changemaker_field_value_result.sql
+++ b/src/database/initialization/1_build_changemaker_field_value_result.sql
@@ -1,0 +1,42 @@
+SELECT drop_function('build_changemaker_field_value_result');
+
+-- Gathers a single changemaker field value's children (base field, batch, file)
+-- and assembles the result with the changemaker_field_value_to_json shape.
+-- Guards against forbidden fields, since callers may serialize a row that has
+-- not been permission-filtered (the field value insert).
+CREATE FUNCTION build_changemaker_field_value_result(
+	changemaker_field_value changemaker_field_values,
+	auth_context_keycloak_user_id uuid DEFAULT NULL,
+	auth_context_is_administrator boolean DEFAULT FALSE
+) RETURNS jsonb AS $$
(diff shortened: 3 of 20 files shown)


src/database/queries/changemakers/insertOne.sql
```diff
@@ -1,16 +1,22 @@
...
+			name,
+			keycloak_organization_id,
+			created_by
+		) VALUES (
+			:taxId,
+			:name,
+			:keycloakOrganizationId,
+			:authContextKeycloakUserId
+		)
+		RETURNING *
```
bickelj, line 14: The star gave me pause.

We pass the full result through the serializer function so that gives the function the ability to filter out whatever columns it likes, right?

The situation I'm thinking of is change to the changemaker table over time, add/drop columns.

bickelj (approved):
I have not reviewed all the code in utmost detail here, but I lean on our tests. I see no tests changed. I see tests pass. And I see no obvious red flags in the code.

I compared the JSON result of a `GET /changemakers?_page=1&_count=2000000` from the preview environment to the same GET in the sandbox environment. They were the same with the exception of download URIs (expected).

I briefly compared the `time` of the same `GET /changemakers` call (with output piped to `/dev/null`) alternating between the preview environment and the sandbox environment. The calls to the preview environment were always faster than the calls to the sandbox environment (N=10), even comparing the best case of the sandbox environment (2714ms) to the worst of the preview environment (1561ms).

In all this is very nice, I like, great success, etc.

While I was doing the above I asked an LLM to review the changes as well. It was with the context of the previous review of the previous similar PR (#2477) and I asked it to ignore the invalid cast thing since our tests would pass that. Small Qwen models such as this usually give a handful of false positives but frequently flag legitimate issues. Take or leave whichever feedback you find useful. Obviously the "critical" `integer` cast issue is also not critical since all our IDs are `integer` (the model didn't see that context in this case). I also told it to consider this is a performance issue change.

<details><summary>Qwen3.6 30B A3B review of only the commit messages and files that were changed</summary>

Here is the review of the changemakers serialization changes, categorized as requested. I've acknowledged your note about the array constructor and skipped repeating it since your tests confirm it works in your environment.

### 🔴 Critical Show-Stoppers

1. **`src/database/initialization/2_changemaker_field_value_file_to_json.sql`: Integer Overflow Risk**
   ```sql
   AND files.id = changemaker_field_value.value::integer;
   ```
   The regex `changemaker_field_value.value ~ '^[0-9]+$'` only validates that the string contains digits, not that it fits within a 32-bit signed integer (`2,147,483,647`). If a value like `'3000000000'` is stored, the regex passes, but the cast throws `integer out of range`, crashing the request.
   **Fix:** Cast to `bigint` (or validate length):
   ```sql
   AND files.id = changemaker_field_value.value::bigint;
   ```

2. **`src/database/initialization/2_permitted_changemaker_field_value_ids_among.sql`: Unbounded Join in Inheritance Branch**
   ```sql
   INNER JOIN permitted_changemaker_ids(
       permitted_changemaker_field_value_ids_among.user_keycloak_user_id,
       ...
   ) AS permitted_changemakers
       ON permitted_changemakers.id = cfv.changemaker_id
   ```
   This function is designed to bound work to `filter_ids`, but this branch calls `permitted_changemaker_ids` (unbounded). While the final `WHERE cfv.id = ANY(filter_ids)` filters the output, PostgreSQL's planner does not always push the `ANY` predicate deep enough to prevent scanning the entire changemaker permission set. Under high concurrency or with many changemakers, this can cause severe performance degradation or timeouts.
   **Fix:** If `permitted_changemaker_ids` can be expressed as a CTE or bounded variant, use it here. Otherwise, add `MATERIALIZED` to the CTE that calls it to force the planner to evaluate it once and reuse it, or refactor the inheritance logic to filter changemakers first.

### 🟡 Minor Nice-to-Haves

1. **Redundant `users` Join in `build_changemakers_results`**
   The `gold_winner` CTE joins `users u` in both the Proposal and Changemaker `UNION ALL` branches to filter unauthenticated users:
   ```sql
   INNER JOIN users u ON u.keycloak_user_id = build_changemakers_results.auth_context_keycloak_user_id
   ```
   This is redundant because `permitted_proposal_field_value_ids_among` and `permitted_changemaker_field_value_ids_among` already handle authentication. Removing these joins will reduce join overhead and simplify the query plan.

2. **Boolean `ORDER BY` Clarity in `gold_winner`**
   ```sql
   combined.source_data_provider_short_code IS NOT NULL AND combined.source_id != system_source_id() DESC,
   ```
   While valid PostgreSQL, boolean expressions in `ORDER BY` can be opaque to readers. The intent is clear ("data provider > system source"), but wrapping it in a `CASE` or extracting it as a computed column in the `combined` CTE would improve readability and allow the planner to potentially use indexes on the boolean result.

3. **PL/pgSQL Builders Called Inside `LANGUAGE sql STABLE`**
   `build_changemakers_results` (SQL) calls `build_proposal_field_value_result` and `build_changemaker_field_value_result` (PL/pgSQL). This is functionally correct, but PostgreSQL does not inline PL/pgSQL functions in SQL functions, meaning each winner triggers a separate PL/pgSQL context switch. For extreme QPS, consider moving the guard/permission logic into SQL CTEs or using `LANGUAGE sql` builders if the logic permits. (Likely negligible for current scale, but worth noting.)

4. **`gold_winner` `DISTINCT ON` with `UNION ALL` Materialization**
   The `gold_winner` CTE runs `DISTINCT ON` over a `UNION ALL` of two large, permission-filtered result sets. If the page size is large or changemakers have many field values, this sorts the entire union before picking winners. Adding `MATERIALIZED` to the `combined` subquery can sometimes help the planner avoid redundant sorts, though the current approach is acceptable for typical pagination sizes.

5. **`build_changemakers_results` `fiscal_sponsor_json` `INNER JOIN users`**
   ```sql
   INNER JOIN users u ON u.keycloak_user_id = build_changemakers_results.auth_context_keycloak_user_id
   ```
   If `auth_context_keycloak_user_id` is `NULL`, this `INNER JOIN` yields zero rows, correctly filtering out sponsors for unauthenticated users. However, it's cleaner to use a `LEFT JOIN` with a `WHERE u.keycloak_user_id IS NOT NULL` or handle the null check at the application layer to avoid surprising join semantics.

6. **Array Cast Consistency (Optional Cleanup)**
   Since your tests pass, the explicit casts like `updated_changemaker::changemakers` are working. However, PostgreSQL infers composite types from CTEs automatically. Removing the explicit cast (`array(SELECT updated_changemaker FROM updated_changemaker)`) improves readability and reduces maintenance surface without changing behavior.

### ✅ Overall Assessment
The architectural pattern (shapes + result builders + bounded permission lookups) is consistently applied and correctly handles the "gold field value" selection with deterministic tie-breaking. The single-pass serialization design effectively eliminates the N+1 query problem for changemakers. The integer overflow in the file resolution function is the only true show-stopper; the rest are performance hygiene and readability refinements.

</details>

PR #2491, opened 2026-07-01
Paginate list entries before serializing them

### LINKED ISSUE #2490: Pagination converts every previous page of proposals to json before returning
https://github.com/PhilanthropyDataCommons/service/blob/1d0b141e593f3a339de5b868d472d736cfb231b7/src/database/queries/changemakersProposals/selectWithPagination.sql#L32-L42

I noticed that the fetch times were increasing linearly as I paginated through the changemaker proposals endpoint, and tracked the cause to this query.

By calling `changemaker_proposal_to_json` _inside_ the `SELECT`, we apply it to every page of results regardless of whether we return it or not.

We should instead move it to a CTE or subquery so that we only jsonify the page we care about, e.g., 

```sql
paginated_entries AS (
    SELECT changemaker_proposal_to_json(page.*::changemakers_proposals, ...) AS object
    FROM (
        SELECT candidate_entries.*
        FROM candidate_entries
        ORDER BY id DESC
        LIMIT :limit OFFSET :offset
    ) AS page
)
```

slifty:
This PR updates our pagination queries to paginate *before* serialization, since postgres will apply the serialization on the entire potential result set before pagination (resulting in later pages performing a lot of work for no reason / linear growth)

Resolves #2490

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Paginated list endpoints no longer slow down as you request later pages.
+
 ## 0.40.0 2026-06-26
 
 ### Added
--- a/src/database/queries/applicationFormFields/selectWithPagination.sql
+++ b/src/database/queries/applicationFormFields/selectWithPagination.sql
@@ -25,14 +25,20 @@ WITH
 		SELECT count(*) AS total FROM candidate_entries
 	),
 
-	paginated_entries AS (
-		SELECT
-			application_form_field_to_json(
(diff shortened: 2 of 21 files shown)


hminsky2002 (approved):
This all looks reasonable to me, and coverage looks complete. Approving for throughput purposes! 

But I will say that this once again seems like an opportunity to DRY up our sql code (if the select functions all share a nearly identical logic block when selecting paginated entries). Not a required change, but would clean up future changes of this ilk.

bickelj:
No objection, though I cannot reproduce the original issue on demo or sandbox with the linear speed increase on further pages. Perhaps that is too small a dataset to reproduce it.

slifty:
I *was* able to reproduce on demo, but it was specifically for `/changemakersProposals` not `/proposals`

PR #2493, opened 2026-07-02
Serialize changemaker proposals in a single pass

### LINKED ISSUE #2492: Address `GET /changemakerProposals` slowness
We have been working on fixing our most egregious `N+1` queries and started with Proposals (#2473) and Changemakers (#2479).  We also have to fix `GET /changemakerProposals`

See #2475

slifty:
This PR applies our new collection patterns (to remove `N+1` queries) to the `GET /changemakersProposals` endpoint.

Resolves #2492

### DIFF
--- a/src/database/initialization/1_changemaker_proposal_to_json.sql
+++ b/src/database/initialization/1_changemaker_proposal_to_json.sql
@@ -2,45 +2,15 @@ SELECT drop_function('changemaker_proposal_to_json');
 
 CREATE FUNCTION changemaker_proposal_to_json(
 	changemaker_proposal changemakers_proposals,
-	auth_context_keycloak_user_id uuid DEFAULT NULL,
-	auth_context_is_administrator boolean DEFAULT FALSE
-)
-RETURNS jsonb AS $$
-DECLARE
-  proposal_json JSONB;
-  changemaker_json JSONB;
-BEGIN
-  SELECT serialized_proposal.object
-  INTO proposal_json
-  FROM build_proposals_results(
-    ARRAY(
-      SELECT proposals
-      FROM proposals
-      WHERE proposals.id = changemaker_proposal.proposal_id
-    ),
-    auth_context_keycloak_user_id,
-    auth_context_is_administrator
-  ) AS serialized_proposal;
-
-  SELECT serialized_changemaker.object
-  INTO changemaker_json
-  FROM build_changemakers_results(
-    ARRAY(
-      SELECT changemakers
-      FROM changemakers
-      WHERE changemakers.id = changemaker_proposal.changemaker_id
-    ),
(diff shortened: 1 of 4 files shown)


src/database/initialization/1_changemaker_proposal_to_json.sql
```diff
@@ -2,45 +2,15 @@ SELECT drop_function('changemaker_proposal_to_json');
...
+) RETURNS jsonb AS $$
+	SELECT jsonb_build_object(
+		'id', changemaker_proposal.id,
+		'changemakerId', changemaker_proposal.changemaker_id,
+		'changemaker', changemaker,
+		'proposalId', changemaker_proposal.proposal_id,
+		'proposal', proposal,
+		'createdAt', changemaker_proposal.created_at
+	);
+$$ LANGUAGE sql IMMUTABLE;
```
bickelj, line 16: I love that postgresql has this capability (marking it `IMMUTABLE`).

src/database/initialization/6_build_changemakers_proposals_results.sql
```diff
@@ -0,0 +1,49 @@
...
+-- result builders). Volatile, not STABLE: the embedded proposal reads its
+-- changemakers from changemakers_proposals, so the insert path needs a fresh
+-- snapshot to see the row it just inserted.
+CREATE FUNCTION build_changemakers_proposals_results(
+	changemakers_proposals changemakers_proposals [],
+	auth_context_keycloak_user_id uuid DEFAULT NULL,
+	auth_context_is_administrator boolean DEFAULT FALSE
+) RETURNS TABLE (id int, object jsonb) AS $$
+	WITH input_entries AS (
+		SELECT cp.*
```
bickelj, line 13: For a moment I thought "star: BAAAD, columns GOOOD" but that oversimplifies. Here we are building an intermediate result so it's OK. The higher abstraction function will filter.

src/database/queries/changemakersProposals/insertOne.sql
```diff
@@ -1,12 +1,18 @@
...
+WITH
+	inserted_entry AS (
+		INSERT INTO changemakers_proposals (
+			changemaker_id,
+			proposal_id
+		) VALUES (
+			:changemakerId,
+			:proposalId
+		)
+		RETURNING *
```
bickelj, line 10: Again I had an internal knee-jerk reaction to `*` here but I think our pattern is to actually want all of them.

bickelj (approved):
Even though this is a smaller version of the previous similar changes, I did not dive so deep as to carefully understand each and every line, instead I'm leaning on our automated test passage and noticing that there were no changes to tests here.

I did some `curl` calls against the endpoint in demo vs the preview environment here and the preview here won in both absolute times and in consistency.

PR #2506, opened 2026-07-21
Include proposalId in proposalFieldValue

### LINKED ISSUE #1993: Include proposalId in Changemaker fields objects
Currently, a Changemaker object's `fields` objects has `proposalVersionId`, but not `proposalId`. Having `proposalId` would allow Exchange to hyperlink/look up to the proposal that a field value is from for a user.

hminsky2002:
This commit adds proposalId as a field in the proposalFieldValue object when it is queried from the database, a convenience for retrieving the related proposal.

Closes #1993

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.41.0 2026-07-24
+
 ### Fixed
 
 - Paginated list endpoints no longer slow down as you request later pages.
 
 ### Added
 
+- Proposal field value objects now include a `proposalId` attribute, so a value in a changemaker's `fields` (or anywhere else a proposal field value appears) can be traced back to its proposal without an extra call to `GET /proposalVersions/{proposalVersionId}`.
 - Added a `GET /funders/{memberFunderShortCode}/collaboratives` endpoint that lists funder collaborative memberships for a given member funder.
 - Bulk upload CSVs may now include `control:`-prefixed columns (e.g. `control:pdc_changemaker_id`). Control columns carry meta-information that changes how a row is processed; they are exempt from application-form validation and are not stored as base fields.
 
--- a/src/database/initialization/1_build_proposal_field_value_result.sql
+++ b/src/database/initialization/1_build_proposal_field_value_result.sql
@@ -12,6 +12,7 @@ DECLARE
 	is_file_field BOOLEAN;
 	application_form_field_json JSONB;
 	version_created_by UUID;
+	version_proposal_id INTEGER;
 BEGIN
 	PERFORM assert_proposal_field_value_not_forbidden(proposal_field_value);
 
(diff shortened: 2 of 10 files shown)


bickelj:
I tried `GET /changemakers` in the preview environment, and I see the `proposalId` next to the `proposalVersionId` in the response, as requested in #1993.

```json
//...
          "id": 39841,
          "file": null,
          "value": "",
          "isValid": true,
          "goodAsOf": null,
          "position": 42,
          "createdAt": "2026-02-09T18:06:57.692268+00:00",
          "proposalId": 868,
          "proposalVersionId": 1037,
//...
```

src/database/initialization/1_build_proposal_field_value_result.sql
```diff
@@ -25,8 +26,8 @@ BEGIN
 	WHERE application_form_fields.id
 		= proposal_field_value.application_form_field_id;
 
-	SELECT proposal_versions.created_by
-	INTO version_created_by
+	SELECT proposal_versions.created_by, proposal_versions.proposal_id
```
bickelj, line 29: Oh, nice that this doesn't require an additional join. I was kind of expecting one but this makes sense!

src/openapi/components/schemas/ProposalFieldValue.json
```diff
@@ -4,6 +4,11 @@
 		{
 			"type": "object",
 			"properties": {
+				"proposalId": {
+					"allOf": [{ "$ref": "./id.json" }],
+					"readOnly": true,
+					"example": 2911
```
bickelj, line 10: This is nice that it's a unique value. I previously strove for prime integers and then tried to consistently use the same prime integer for the same entity ID but it looks like they have become inconsistent in this regard. It's not documented anywhere either, so neither us nor our agents could figure that out.

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -547,6 +548,7 @@ describe('processBulkUploadTask', () => {
 									isValid: true,
 									goodAsOf: null,
 									position: 1,
+									proposalId: 2,
```
bickelj, line 551: Another eventual-cleanup-task this reminds me of is that we should capture the proposal ID on create and then expect that value here rather than assuming the DB sequence will be in this or that state. This is OK though.

src/tasks/__tests__/processBulkUploadTask.int.test.ts
```diff
@@ -1489,6 +1497,7 @@ describe('processBulkUploadTask', () => {
 				isValid: true,
 				goodAsOf: null,
 				position: 0,
+				proposalId: expectNumber(),
```
bickelj: Another eventual cleanup task, or perhaps for this instance should be cleaned up here: any number is too loose an expectation. It should be the correct proposal ID, shouldn't it? Or is that outside the scope of what this test attempts to validate? I suppose the latter and it would be hard to predict in this case.

CHANGELOG.md
bickelj, line 8: To be useful to downstream consumers, I think an SDK version update is warranted here.

bickelj (approved):
This looks good to me overall. I have some stream-of-consciousness quibbles about our tests but those are not directly related to this change. One suggestion is an actual version change in the openAPI spec as well as the changelog such that the change is included in the SDK package.

src/tasks/__tests__/processBulkUploadTask.int.test.ts
hminsky2002: No i think you're right actually, in this case we're expecting a specific id (and it is a failure if the wrong id is returned). As opposed to just verifying a proposal was made (which i think is when an expect number is warranted)

PR #2507, opened 2026-07-21
Limit changemaker attributes for non-authenticated clients

### LINKED ISSUE #2459: Limit `Changemaker` attributes for non-authenticated clients
`Changemakers` in the pdc are fully public right now.  We should be explicit about which attributes are visible in an unauthenticated context.

I think only `id` `name` and `TIN` and `fields` (with PUBLIC type base fields) are intended to be public right now.  Others attributes might be acceptable to include but I would encourage us to opt towards narrow and add attributes as use cases come up.

hminsky2002:
Closes #2459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- A changemaker now carries the `keycloakOrganizationId`, `createdAt`, and `createdBy` attributes, and its `fiscalSponsors`, only when the requester holds `view` permission on that changemaker. Every other requester, including an unauthenticated one, receives `id`, `name`, and `taxId`, plus `fields` where the endpoint returns it at all. This applies wherever a changemaker appears, not just at `/changemakers`: the changemakers embedded in a proposal, in a source, and in another changemaker's `fiscalSponsors` are narrowed the same way. Permission is evaluated per changemaker, so a single array may hold both shapes.
+- `fields` remains an empty array for unauthenticated requests; it is kept in the response shape to support future visibility of `public` base field values.
+
 ## 0.42.0 2026-08-25
 
 ### Added
--- a/src/__tests__/changemakers.int.test.ts
+++ b/src/__tests__/changemakers.int.test.ts
@@ -55,7 +55,7 @@ import {
 	stringToKeycloakId,
 } from '../types';
 import type { TinyPg } from 'tinypg';
-import type { AuthContext } from '../types';
+import type { AuthContext, Id, KeycloakId } from '../types';
 
 const insertTestChangemakers = async (db: TinyPg, authContext: AuthContext) => {
 	await createChangemaker(db, authContext, {
@@ -67,6 +67,23 @@ const insertTestChangemakers = async (db: TinyPg, authContext: AuthContext) => {
 		taxId: '22-2222222',
 		name: 'Another Inc.',
 		keycloakOrganizationId: '57ceaca8-be48-11ef-8c91-5732d98a77e1',
+	});
+};
+
+const grantChangemakerView = async (
+	db: TinyPg,
+	changemakerId: Id,
+	granteeUserKeycloakUserId: KeycloakId,
+) => {
+	const systemUser = await loadSystemUser(db, null);
+	const systemUserAuthContext = getAuthContext(systemUser, true);
+	await createPermissionGrant(db, systemUserAuthContext, {
+		granteeType: PermissionGrantGranteeType.USER,
+		granteeUserKeycloakUserId,
+		contextEntityType: PermissionGrantEntityType.CHANGEMAKER,
+		changemakerId,
+		scope: [PermissionGrantEntityType.CHANGEMAKER],
+		verbs: [PermissionGrantVerb.VIEW],
 	});
 };
 
@@ -182,8 +199,33 @@ const setupTestContext = async (db: TinyPg) => {
 
 describe('/changemakers', () => {
 	describe('GET /', () => {
-		it('does not require authentication', async () => {
-			await request(app).get('/changemakers').expect(200);
+		it('does not require authentication and returns only public attributes', async () => {
+			const db = getDatabase();
+			const testUser = await loadTestUser(db);
+			const testUserAuthContext = getAuthContext(testUser);
+			await insertTestChangemakers(db, testUserAuthContext);
+			await request(app)
+				.get('/changemakers')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).toStrictEqual({
+						total: 2,
+						entries: [
+							{
+								id: 2,
+								taxId: '22-2222222',
+								name: 'Another Inc.',
+								fields: [],
+							},
+							{
+								id: 1,
+								taxId: '11-1111111',
+								name: 'Example Inc.',
+								fields: [],
+							},
+						],
+					});
+				});
 		});
 
 		it('returns an empty Bundle when no data is present', async () => {
@@ -193,11 +235,12 @@ describe('/changemakers', () => {
 			});
 		});
 
-		it('returns changemakers present in the database', async () => {
+		it('returns full attributes only for changemakers the user has view permission on', async () => {
 			const db = getDatabase();
 			const testUser = await loadTestUser(db);
 			const testUserAuthContext = getAuthContext(testUser);
 			await insertTestChangemakers(db, testUserAuthContext);
+			await grantChangemakerView(db, 1, testUser.keycloakUserId);
 			await request(app)
 				.get('/changemakers')
 				.set(authHeader)
@@ -210,10 +253,6 @@ describe('/changemakers', () => {
 								id: 2,
 								taxId: '22-2222222',
 								name: 'Another Inc.',
-								keycloakOrganizationId: '57ceaca8-be48-11ef-8c91-5732d98a77e1',
-								createdAt: expectTimestamp(),
-								createdBy: testUser.keycloakUserId,
-								fiscalSponsors: [],
 								fields: [],
 							},
 							{
@@ -249,7 +288,7 @@ describe('/changemakers', () => {
 					_page: 2,
 					_count: 5,
 				})
-				.set(authHeader)
+				.set(adminUserAuthHeader)
 				.expect(200)
 				.expect((res) => {
 					expect(res.body).toEqual({
@@ -335,7 +374,7 @@ describe('/changemakers', () => {
 			});
 			const response = await request(app)
 				.get(`/changemakers?proposal=1`)
-				.set(authHeader)
+				.set(adminUserAuthHeader)
 				.expect(200);
 			expect(response.body).toEqual({
 				total: 1,
@@ -382,7 +421,7 @@ describe('/changemakers', () => {
 			});
 			const response = await request(app)
 				.get(`/changemakers`)
-				.set(authHeader)
+				.set(adminUserAuthHeader)
 				.expect(200);
 			expect(response.body).toEqual({
 				total: 1,
@@ -417,7 +456,7 @@ describe('/changemakers', () => {
 			});
 			const response = await request(app)
 				.get('/changemakers?_content=garden')
-				.set(authHeader)
+				.set(adminUserAuthHeader)
 				.expect(200);
 			expect(response.body).toEqual({
 				total: 1,
@@ -480,7 +519,7 @@ describe('/changemakers', () => {
 			});
 			const response = await request(app)
 				.get('/changemakers?_content=garden&proposal=1')
-				.set(authHeader)
+				.set(adminUserAuthHeader)
 				.expect(200);
 			expect(response.body).toEqual({
 				total: 1,
@@ -512,12 +551,26 @@ describe('/changemakers', () => {
 	});
 
 	describe('GET /:id', () => {
-		it('does not require authentication', async () => {
+		it('does not require authentication and returns only public attributes', async () => {
 			const db = getDatabase();
 			const testUser = await loadTestUser(db);
 			const testUserAuthContext = getAuthContext(testUser);
 			await insertTestChangemakers(db, testUserAuthContext);
-			await request(app).get('/changemakers/1').expect(200);
+			await request(app)
+				.get('/changemakers/1')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).toStrictEqual({
+						id: 1,
+						taxId: '11-1111111',
+						name: 'Example Inc.',
+						fields: [],
+					});
+					expect(res.body).not.toHaveProperty('keycloakOrganizationId');
+					expect(res.body).not.toHaveProperty('createdAt');
+					expect(res.body).not.toHaveProperty('createdBy');
+					expect(res.body).not.toHaveProperty('fiscalSponsors');
+				});
 		});
 
 		it('returns 404 when given id is not present', async () => {
@@ -529,6 +582,7 @@ describe('/changemakers', () => {
 			const testUser = await loadTestUser(db);
 			const testUserAuthContext = getAuthContext(testUser);
 			await insertTestChangemakers(db, testUserAuthContext);
+			await grantChangemakerView(db, 2, testUser.keycloakUserId);
 			await request(app)
 				.get('/changemakers/2')
 				.set(authHeader)
@@ -680,6 +734,11 @@ describe('/changemakers', () => {
 					isValid: false,
 					goodAsOf: null,
 				});
+				await grantChangemakerView(
+					db,
+					changemakerId,
+					getTestUserKeycloakUserId(),
+				);
 				await request(app)
 					.get(`/changemakers/${changemakerId}`)
 					.set(authHeader)
@@ -786,6 +845,11 @@ describe('/changemakers', () => {
 					isValid: true,
 					goodAsOf: null,
 				});
+				await grantChangemakerView(
+					db,
+					changemaker.id,
+					getTestUserKeycloakUserId(),
+				);
 				await request(app)
 					.get(`/changemakers/${changemaker.id}`)
 					.set(authHeader)
@@ -793,8 +857,10 @@ describe('/changemakers', () => {
 					.expect((res) => {
 						expect(res.body).toEqual({
 							...changemaker,
+							keycloakOrganizationId: '8b15d276-be48-11ef-a061-5b4a50e82d50',
 							createdAt: expectTimestamp(),
 							createdBy: getTestUserKeycloakUserId(),
+							fiscalSponsors: [],
 							fields: [changemakerEarliestValue],
 						});
 					});
@@ -884,6 +950,11 @@ describe('/changemakers', () => {
 					isValid: true,
 					goodAsOf: null,
 				});
+				await grantChangemakerView(
+					db,
+					changemaker.id,
+					getTestUserKeycloakUserId(),
+				);
 				await request(app)
 					.get(`/changemakers/${changemaker.id}`)
 					.set(authHeader)
@@ -891,8 +962,10 @@ describe('/changemakers', () => {
 					.expect((res) => {
 						expect(res.body).toEqual({
 							...changemaker,
+							keycloakOrganizationId: '8b15d276-be48-11ef-a061-5b4a50e82d50',
 							createdAt: expectTimestamp(),
 							createdBy: getTestUserKeycloakUserId(),
+							fiscalSponsors: [],
 							fields: [funderEarliestValue],
 						});
 					});
@@ -985,6 +1058,11 @@ describe('/changemakers', () => {
 						goodAsOf: null,
 					},
 				);
+				await grantChangemakerView(
+					db,
+					changemaker.id,
+					getTestUserKeycloakUserId(),
+				);
 				await request(app)
 					.get(`/changemakers/${changemaker.id}`)
 					.set(authHeader)
@@ -992,8 +1070,10 @@ describe('/changemakers', () => {
 					.expect((res) => {
 						expect(res.body).toEqual({
 							...changemaker,
+							keycloakOrganizationId: '8b15d276-be48-11ef-a061-5b4a50e82d50',
 							createdAt: expectTimestamp(),
 							createdBy: getTestUserKeycloakUserId(),
+							fiscalSponsors: [],
 							fields: [dataProviderNewestValue],
 						});
 					});
@@ -1060,6 +1140,11 @@ describe('/changemakers', () => {
 						BaseFieldSensitivityClassification.FORBIDDEN,
 				});
 
+				await grantChangemakerView(
+					db,
+					changemaker.id,
+					getTestUserKeycloakUserId(),
+				);
 				await request(app)
 					.get(`/changemakers/${changemaker.id}`)
 					.set(authHeader)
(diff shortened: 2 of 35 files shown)


src/types/Changemaker.ts
```diff
@@ -14,13 +14,17 @@ interface ShallowChangemaker {
...
+	readonly createdAt: string | undefined;
+	// Omitted when serialized for an unauthenticated caller.
+	readonly createdBy: KeycloakId | undefined;
 }
 
 interface Changemaker extends ShallowChangemaker {
-	readonly fiscalSponsors: ShallowChangemaker[];
+	// Omitted when serialized for an unauthenticated caller.
+	readonly fiscalSponsors: ShallowChangemaker[] | undefined;
 	readonly fields: Array<ProposalFieldValue | ChangemakerFieldValue>;
```
bickelj, line 28: I almost thought this should also be `| undefined` because it will always be an empty list in the unauthenticated case, but I'm not sure this is necessarily true because we may have public fields, is that right?

If there is a such thing as a truly public public field that can be returned to an unauthenticated user, it makes sense to keep it as is.

If there is no such thing as a truly public public field that can be returned to an unauthenticated user, it would make more sense to make it `| undefined`.

The reason is that this is already a breaking change to the JSON-shape contract, so it may as well be the most accurate shape in so breaking it.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `pdc_changemaker_id` is now expected as a `control:` column in a bulk-upload, rather than as a base field.
 - `GET /permissionGrants` now accepts optional filter query parameters: `changemaker`, `funder`, `dataProvider`, and `proposal` to filter by context entity, and `granteeType` and `verb` to filter by grantee. At most one context entity filter may be provided per request; providing more than one returns a `400`.
+- Unauthenticated requests to `GET /changemakers` and `GET /changemakers/{changemakerId}` now return only `id`, `name`, `taxId`, and `fields`; the `keycloakOrganizationId`, `createdAt`, `createdBy`, and `fiscalSponsors` attributes are omitted. `fields` now contains values of base fields whose sensitivity classification is `public` (previously always empty when unauthenticated).
```
bickelj: Is `public` supposed to mean the entire internet or does it only mean "authenticated users"?

hminsky2002: public meaning non-authenticated users, so the entire internet (which I believe was explicitly requested in the issue)

src/types/Changemaker.ts
hminsky2002: So my understanding is that even if there were fields that could be returned for a Changemaker, if a non-logged in user requested a changemaker there's a [is not null condition](https://github.com/PhilanthropyDataCommons/service/blob/44aa6a1ab755dd4ac7c75d4bea588a0e097d980b/src/database/initialization/5_build_changemakers_results.sql#L103) that tosses them out. I added in the same file a condition that forces through unauthenticated requests to see if there are any public fields to view. Anyway, even if there are no fields present that the user is allowed to see the structure returned by the database still returns an empty list. So i don't think we need undefined here.

bickelj: I reviewed #1621 and #1686 to refresh my understanding of base field classification of `public` vs `restricted` vs `forbidden`. I think you're right that `public` there is intended to (eventually) mean really really public whereas `restricted` is restricted to authenticated users. Hopefully that's the right interpretation. My question, I suppose, was at the higher level of what we want in future rather than what the current state is. In the end, I agree that the best course is to leave it an empty list, because that supports truly public `fields` if we want them.

bickelj: Oh and I see that the future is here with this PR, but I'll comment on that elsewhere.

CHANGELOG.md
bickelj: OK, cool, I see in the preview environment that the public base fields are returned, along with all the depth/accoutrements alongside them.

This is good, but raises a new question (especially given that the title of the referenced issue is to "limit") as to what attributes ought to be limited in the new, deep structure that's revealed.

For example, sandbox (without this change) looks like this:

```json
{
  "entries": [
    {
      "id": 92,
      "name": "The Files Organization",
      "taxId": "67-725625",
      "fields": [],
      "createdAt": "2026-02-09T18:06:57.692268+00:00",
      "createdBy": "00000000-0000-0000-0000-000000000000",
      "fiscalSponsors": [],
      "keycloakOrganizationId": null
    },
```

Whereas with this change, in [a preview environment for this change](https://utilities.philanthropydatacommons.org/pdc-service-preview-2507-a6qs6/#/Changemakers/getChangemakers) it looks like this:

```json
{
  "entries": [
    {
      "id": 92,
      "name": "The Files Organization",
      "taxId": "67-725625",
      "fields": [
        {
          "id": 39841,
          "file": null,
          "value": "",
          "isValid": true,
          "goodAsOf": null,
          "position": 42,
          "createdAt": "2026-02-09T18:06:57.692268+00:00",
          "proposalId": 868,
          "proposalVersionId": 1037,
          "applicationFormField": {
            "id": 2473,
            "label": "Fiscal Sponsor City",
            "position": 43,
            "baseField": {
              "label": "Fiscal Sponsor City",
              "category": "organization",
              "dataType": "string",
              "createdAt": "2025-07-28T23:01:49.557721+00:00",
              "shortCode": "fiscal_sponsor_city",
              "description": "Fiscal Sponsor City",
              "localizations": {},
              "valueRelevanceHours": 4380,
              "sensitivityClassification": "public"
            },
            "createdAt": "2025-08-01T17:24:43.172855+00:00",
            "inputType": null,
            "instructions": null,
            "applicationFormId": 48,
            "baseFieldShortCode": "fiscal_sponsor_city"
          },
          "applicationFormFieldId": 2473
        },
```

On the one hand that's a limit of attributes on the top level, but an expansion of attributes underneath. It's a question for the team that I don't expect you to be able to immediately answer here (unless this was already discussed and decided in a ticket that can be linked).

hminsky2002: @bickelj that..is an excellent point. Hmm. Even beyond a security concern, from a consumer standpoint that's a lot of service-specific data that a public consumer shouldn't want anything to do with .

hminsky2002: @slifty @jmergy do you guys have any thoughts on how we want to return public basefields? Me and @bickelj are hung up on this issue of this PR actually increasing the amount of metadata returned to an unauthenticated user, since previously we did not any public fields to unauthenticated users. Do we have thoughts on if we want to always return all data in a `public` base field, or still limit some of the pdc-specific metadata (as we are now doing with changemaker in this PR)?

slifty: Closing the loop here (sorry I posted in the issue!) -- I think the right approach is to keep the behavior as is for PUBLIC as part of this PR.

I do think `PUBLIC` field values were intended to be visible for unauthenticated users, but this opens some weird worm-cans in non-changemaker-contexts (e.g. if a proposal has a public field but the proposal is not visible... what do we think should happen?). 

I think the answer is going to be:

> Any field value for a public base field, regardless of type, grants the `VIEW` verb with the `{FieldValue}` scope for all permission grant contexts regardless of auth context. This does not grant view access for parent contexts / this doesn't mean a user can view the proposal / initiative / etc, but it does mean users with view access to that parent context will be able to view the affected field values"

... but like I said I think it would be good to treat that as a separate issue / PR from this one.

hminsky2002: @slifty @bickelj  Noting that the latest version that has been pushed up here reverts the change to allow unauthenticated users to see fields, and just scopes to the limited changemaker attributes!

src/database/initialization/1_changemaker_to_json.sql
```diff
@@ -4,20 +4,37 @@ CREATE FUNCTION changemaker_to_json(
 	changemaker changemakers,
 	fiscal_sponsors jsonb,
 	fields jsonb,
-	shallow boolean DEFAULT FALSE
+	shallow boolean DEFAULT FALSE,
+	auth_context_keycloak_user_id uuid DEFAULT NULL,
```
slifty: We have moved to these functions really being more "serializers" and less "logical business logic processors".

To that end I would suggest either:

1. Have this take in a boolean to signal "return the public serialization" or "return the full serialization"
2. Have an entirely separate function `1_changemaker_to_public_json.sql` <-- less ideal since now you are duplicating the serialization logic.  We shouldn't do this I think!

Within (1) I'd recommend checking out a way to implement this without having to redundantly serialize common attributes.   For instance something like

```
SELECT jsonb_build_object(
        'id', changemaker.id,
        'taxId', changemaker.tax_id,
        'name', changemaker.name
  ) || CASE
        WHEN changemaker_to_json.shallow THEN '{}'::jsonb
        ELSE jsonb_build_object(
                'fields', COALESCE(changemaker_to_json.fields, '[]'::jsonb)
        )
  END || CASE
        WHEN NOT permitted THEN '{}'::jsonb
        ELSE jsonb_build_object(
                'keycloakOrganizationId', changemaker.keycloak_organization_id,
                'createdAt', changemaker.created_at,
                'createdBy', changemaker.created_by
        )
  END || CASE
        WHEN changemaker_to_json.shallow OR NOT permitted THEN '{}'::jsonb
        ELSE jsonb_build_object(
                'fiscalSponsors',
                COALESCE(changemaker_to_json.fiscal_sponsors, '[]'::jsonb)
        )
  END;
```

src/openapi/components/schemas/Changemaker.json
```diff
@@ -8,7 +8,7 @@
 			"properties": {
 				"fiscalSponsors": {
 					"type": "array",
-					"description": "Fiscal sponsors (past and present) of this changemaker. Empty when unauthenticated. The response here will not include values for `fields` nor `fiscalSponsors` to avoid infinite recursion. Use the returned `id` to call $getChangemakerById for those details.",
+					"description": "Fiscal sponsors (past and present) of this changemaker. Omitted when the request is unauthenticated. The response here will not include values for `fields` nor `fiscalSponsors` to avoid infinite recursion. Use the returned `id` to call $getChangemakerById for those details.",
```
slifty: We talked about this on call but the actual rule should be "Omitted when the user does not have view permission" or something like that.

src/types/Changemaker.ts
```diff
@@ -14,13 +14,17 @@ interface ShallowChangemaker {
...
-	readonly createdBy: KeycloakId;
+	// Omitted when serialized for an unauthenticated caller.
+	readonly createdAt: string | undefined;
+	// Omitted when serialized for an unauthenticated caller.
+	readonly createdBy: KeycloakId | undefined;
 }
 
 interface Changemaker extends ShallowChangemaker {
-	readonly fiscalSponsors: ShallowChangemaker[];
+	// Omitted when serialized for an unauthenticated caller.
```
slifty: We aren't as limited by openAPI schema tools here -- we can have a true `PublicChangemaker` and leverage things like `Omit<>` to strip the non-public fields.

src/openapi/components/schemas/ShallowChangemaker.json
```diff
@@ -15,21 +15,22 @@
...
 		"createdBy": {
-			"description": "The keycloak user id of the PDC user that created this changemaker",
+			"description": "The keycloak user id of the PDC user that created this changemaker. Omitted when the request is unauthenticated.",
 			"type": "string",
 			"format": "uuid",
 			"readOnly": true
 		}
 	},
-	"required": ["id", "taxId", "name", "createdAt", "createdBy"]
+	"required": ["id", "taxId", "name"]
```
slifty: I was struggling a bit with this -- these *are* required in the "has permission" context.  It isn't incorrect, but I wanted to share the conclusions from the brainstorm session I had with claude:

```
  You can have the single description without the union. Define the schemas by composition and reference the composed one:

  PublicChangemaker          → id, taxId, name
  ProtectedChangemakerAttrs  → keycloakOrganizationId, createdAt, createdBy

  Changemaker = allOf: [PublicChangemaker, ProtectedChangemakerAttrs(optional), ...]

  PublicChangemaker gets exactly what you want — one schema-level description saying "the changemaker representation returned to callers without view permission on this changemaker" — and the
  per-property asides disappear. The response still refs Changemaker, so no union, no subset problem, and Redocly's no-unused-components: error (redocly.yaml) is satisfied because the allOf
  references it.

  This also fixes a name collision you already have. There are two orthogonal axes, and "Shallow" is currently carrying both:

  - depth — shallow (no children) vs. deep (fields, fiscalSponsors)
  - permission — public vs. permitted

  Naming the blocks separates them, and the four blocks land exactly on the four additive clauses from the SQL rewrite: always / when-deep / when-permitted / when-deep-and-permitted. Schema
  composition and serialization composition become the same shape, which is a good sign.

  One thing to decide, and one caveat.

  The decision: ShallowChangemaker.json is also the POST request body (src/openapi/paths/changemakers.json:58), which adds a third axis (read/write). Under any rename, give the request body its
  own schema rather than reusing a response block — otherwise the 2×2 becomes a 2×2×2 and the naming gets unmanageable.

  The caveat: if you split into two TS interfaces but leave the OpenAPI schema as one with optional fields, the two descriptions of the same wire shape drift, and nothing catches it. That's the
  actual argument for doing composition on both sides rather than union-in-TS plus optional-in-schema — keep them isomorphic so an edit to one has an obvious counterpart.

  If you want the TS types to be a union (PublicChangemaker | Changemaker) rather than composed interfaces, that's defensible too — just know you're buying readability, not safety, since there
  are no use sites to protect. That's a fine thing to buy for a type whose only job is description.
```

src/__tests__/changemakers.int.test.ts
```diff
@@ -1060,15 +1140,114 @@ describe('/changemakers', () => {
...
+							keycloakOrganizationId: '8b15d276-be48-11ef-a061-5b4a50e82d50',
 							createdAt: expectTimestamp(),
 							createdBy: getTestUserKeycloakUserId(),
+							fiscalSponsors: [],
+							fields: [],
+						});
+					});
+			});
+
+			it('does not return field values when unauthenticated, even when the base field is public', async () => {
```
bickelj-agent, line 1164: Issue #2459 lists `fields` (with `PUBLIC` base field values) as intended to be visible to unauthenticated clients, but this test asserts the opposite: unauthenticated requests get `fields: []` even when a public base field has a value. The CHANGELOG notes the `fields` shape is kept for "future visibility" of public values. Is closing #2459 here intentional, or should a follow-up issue track returning public base field values to unauthenticated callers?

src/database/initialization/5_build_changemakers_results.sql
```diff
@@ -51,14 +62,18 @@ CREATE FUNCTION build_changemakers_results(
 		SELECT
 			fs.fiscal_sponsee_changemaker_id AS changemaker_id,
 			jsonb_agg(
-				changemaker_to_json(sponsor.*, NULL, NULL, TRUE) ORDER BY sponsor.id
+				changemaker_to_json(
+					sponsor.*, NULL, NULL, TRUE,
```
bickelj-agent, line 77: Fiscal-sponsor attributes are now narrowed per-sponsor: a caller with `view` on the sponsee but not on a given sponsor now sees only `{id, taxId, name}` for that sponsor, whereas previously `keycloakOrganizationId`, `createdAt`, and `createdBy` were always included. This is consistent with the PR's narrowing model, but it is a behavior change for existing callers who had view on the sponsee — please confirm intended.

bickelj-agent (commented):
Reviewed as GLM-5.2 per local AGENTS.md. CI is green (tsc/eslint/prettier/sqlfluff/openapi/test). The serialization model is consistent: `changemaker_to_json` gains a `full_view` flag driven by `permitted_changemaker_ids('view','changemaker')`, and `organization_to_json` is safe because it already INNER JOINs on that same permission function before passing `full_view = TRUE`. Two notes below.

CHANGELOG.md
bickelj: Hurray! As an unauthenticated user, I verified this is true using the preview environment https://utilities.philanthropydatacommons.org/pdc-service-preview-2507-slfrr/

I also created the follow-up issue for our future selves to consider: https://github.com/PhilanthropyDataCommons/service/issues/2548

src/__tests__/changemakers.int.test.ts
bickelj: I think this is where we landed, see the above conversation at https://github.com/PhilanthropyDataCommons/service/pull/2507#discussion_r3623014898. I made the follow-up issue at https://github.com/PhilanthropyDataCommons/service/issues/2548.

src/database/initialization/5_build_changemakers_results.sql
bickelj: Hmm, trying it...

bickelj: Indeed.

In the sandbox I tried the following.

1. As an admin, I found changemaker 32 had a fiscal sponsor 35, and I added another fiscal sponsor 37.
2. As an admin, I granted my non-admin user `view` `any` access on both changemakers 32 and 37 (but **not** on 35).
3. As my non-admin user, here are the `fiscalSponsors` I see mixed types in the response array:

```json
  "fiscalSponsors": [
    {
      "id": 35,
      "name": "Renewable Cities Initiative",
      "taxId": "741852369"
    },
    {
      "id": 37,
      "name": "Winnipeg Water Warriors",
      "taxId": "987456321",
      "createdAt": "2024-04-25T19:25:48.707224+00:00",
      "createdBy": "00000000-0000-0000-0000-000000000000",
      "keycloakOrganizationId": null
    }
  ],
```

bickelj: Do we want a mixed array like that?

On the one hand, it meets the requirements of the linked issue, and we can think of it not as a separate type, because the type clearly says that some attributes may be undefined.

On the other hand, it is a bit surprising to find some elements contain attributes while others do not. We usually would expect a consistent result in an array.

In any case, the OpenAPI doc and example might should clearly indicate this possibility due to the unexpectedness.

src/database/initialization/1_changemaker_to_json.sql
hminsky2002: Noting here that after discussion we decided on taking the 'pass a boolean' approach

src/database/initialization/5_build_changemakers_results.sql
hminsky2002: Hmm. Personally I would lean more towards the mixed array type, because if we do include the values but set them to null or undefined, that may give the incorrect appearance of those values actually `_being_` undefined/null, which is actual incorrect information. But agreed that at minimum this should be documented in spec. I'm about to push that up but interested in alternative solutions.

src/openapi/components/schemas/ShallowChangemaker.json
hminsky2002: I think avoiding having a schema for readable/writable 'shallow; changemakers is definitely worth avoiding. Hmm. I threw Claude at generating that structure, changing the shallow to the public / protectedChangemakerAttributes interface, which I think looks pretty reasonable now!

src/openapi/components/schemas/Changemaker.json
```diff
@@ -8,7 +8,7 @@
 			"properties": {
 				"fiscalSponsors": {
 					"type": "array",
-					"description": "Fiscal sponsors (past and present) of this changemaker. Empty when unauthenticated. The response here will not include values for `fields` nor `fiscalSponsors` to avoid infinite recursion. Use the returned `id` to call $getChangemakerById for those details.",
+					"description": "Fiscal sponsors (past and present) of this changemaker. Omitted unless the caller holds `view` permission on this changemaker. Permission is evaluated per sponsor, so entries the caller cannot view carry only the public attributes while the rest carry all of them; a single array can hold both shapes. The entries never include values for `fields` nor `fiscalSponsors` to avoid infinite recursion. Use the returned `id` to call $getChangemakerById for those details.",
```
bickelj, line 11: 👍

src/types/Changemaker.ts
```diff
@@ -19,10 +22,13 @@ interface ShallowChangemaker {
 	readonly createdBy: KeycloakId;
 }
 
-interface Changemaker extends ShallowChangemaker {
-	readonly fiscalSponsors: ShallowChangemaker[];
+type ShallowChangemaker = PublicChangemaker &
+	Partial<ProtectedChangemakerAttributes>;
```
bickelj, line 26: I'm trying to figure out if the `Partial` is required or if it could be simply `PublicChangemaker & ProtectedChangemakerAttributes`. I don't know enough TypeScript. GLM-5.2 please throw me a bone.

src/database/initialization/1_source_to_json.sql
```diff
@@ -24,7 +24,20 @@ BEGIN
 
   -- Shallow changemaker (no children) to prevent infinite recursion
   -- (source -> changemaker -> fields -> batch -> source).
-  SELECT changemaker_to_json(changemakers.*, NULL, NULL, TRUE)
+  SELECT changemaker_to_json(
+    changemakers.*, NULL, NULL, TRUE,
+    source_to_json.auth_context_keycloak_user_id IS NOT NULL
+    AND EXISTS (
```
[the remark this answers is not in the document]
hminsky2002: Yep good bot, missed this with the rebase!

src/openapi/paths/changemakers.json
```diff
@@ -55,7 +55,7 @@
 			"content": {
 				"application/json": {
 					"schema": {
-						"$ref": "../components/schemas/ShallowChangemaker.json"
+						"$ref": "../components/schemas/NewChangemaker.json"
```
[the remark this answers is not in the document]
hminsky2002: Also agreed, this was a swing for ensuring 'most accurate' type descriptions here but maybe the safer thing to do would just add in the schema description a note that it differs slightly depending on the endpoint

src/types/Changemaker.ts
bickelj-agent: Yes, `Partial` is required. `ProtectedChangemakerAttributes` declares `createdAt: string` and `createdBy: KeycloakId` as non-optional, but `changemaker_to_json` omits both when `permitted` is FALSE (no view). `PublicChangemaker & ProtectedChangemakerAttributes` (no `Partial`) would type them as always-present strings, which is unsound at runtime; `Partial<...>` correctly marks them optional, matching 'absent rather than null' in the schema description. `keycloakOrganizationId` already carries `| undefined`, but `Partial` also makes the property itself optional, which is what the SQL emits in the no-view case. — GLM-5.2

src/openapi/paths/changemakers.json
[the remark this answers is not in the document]
bickelj-agent: Agreed — this is the important one for me. `NewChangemaker.json` is exactly the writable form of `Changemaker.json` (strip the `readOnly` props `id`, `createdAt`, `createdBy`, `fields`, `fiscalSponsors` and you are left with `taxId`, `name`, `keycloakOrganizationId`). `AGENTS.md` says to use the base schema with `readOnly` rather than a separate input schema to avoid SDK type drift, so I'd reference `Changemaker.json` here and drop `NewChangemaker` plus its `api.json` registration. — GLM-5.2

src/database/initialization/1_source_to_json.sql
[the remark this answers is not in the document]
bickelj-agent: Agreed. I checked the other `source_to_json` callers: `build_proposal_version_result`, `bulk_upload_task_to_json`, `changemaker_field_value_batch_to_json`, and `build_proposals_results` all thread `auth_context_*`; `build_initiative_field_value_result` is the only one that does not, so threading the auth params through it (and its query callers) brings it in line. — GLM-5.2

src/__tests__/changemakers.int.test.ts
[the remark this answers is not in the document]
bickelj-agent: The team took the second option: returning public base field values to unauthenticated callers is deferred to #2548 (see bickelj's resolution in the thread above), so closing #2459 via this PR is intentional. — GLM-5.2

bickelj-agent (commented):
Re-reviewed at `0d95dd3` after the force-push; serialization is sound and consistent across every `source_to_json` caller except the one Copilot flagged (initiative-field-value, acknowledged by the author). Answered bickelj's `Partial` question in-thread. One important issue: `NewChangemaker.json` duplicates the writable form of `Changemaker.json` and deviates from the OpenAPI convention in `AGENTS.md` (use the base schema with `readOnly`). — GLM-5.2

hminsky2002:
@bickelj and any robots that may be listening -- all relevant queries that return changemakers now check authorization and the same paths, but that really does increase the scope of the changes here, not sure if it's something we want to kick down the line a bit / keep the behavior limited to just the changemakers endpoint and document that somewhere.

bickelj (approved):
It wasn't obvious to me why we have four different levels of Changemaker types while there are only three levels of access but GLM-5.2 (and [Dan's earlier comment)](https://github.com/PhilanthropyDataCommons/service/pull/2507#discussion_r3758948335) explained there are two axes even if one quadrant is unused. That makes sense.

And Copilot's finding below is germane, so let's iterate a couple of more times : )

src/database/initialization/1_changemaker_to_json.sql
```diff
@@ -4,20 +4,31 @@ CREATE FUNCTION changemaker_to_json(
 	changemaker changemakers,
 	fiscal_sponsors jsonb,
 	fields jsonb,
-	shallow boolean DEFAULT FALSE
+	shallow boolean DEFAULT FALSE,
+	permitted boolean DEFAULT FALSE
```
slifty: We talked on call about the name of this parameter -- `permitted` is not as clear about intention here, vs something like `with_view_permission`.

I'd also throw out that all of our other serializers are built assuming view permission, so I would advocate for this having `DEFAULT TRUE` e.g. if a caller wants a narrower serialization it should specify that explicitly.  Either way works though.

src/database/initialization/5_build_changemakers_results.sql
```diff
@@ -17,6 +17,28 @@ CREATE FUNCTION build_changemakers_results(
 	WITH input_changemakers AS (
 		SELECT c.* FROM unnest(build_changemakers_results.changemakers) AS c
 	),
+	-- changemakers that serialize with the full attribute set for this user,
+	-- bounded to this page's changemakers and their fiscal sponsors
+	permitted_changemakers AS (
```
slifty: Let's rename this to `viewable_changemaker_ids` since that is more specific (and accurate) to what it contains.

src/database/initialization/5_build_changemakers_results.sql
```diff
@@ -51,14 +73,18 @@ CREATE FUNCTION build_changemakers_results(
...
+					pc.id IS NOT NULL
+				) ORDER BY sponsor.id
 			) AS fiscal_sponsors
 		FROM fiscal_sponsorships fs
 		INNER JOIN changemakers sponsor
 			ON sponsor.id = fs.fiscal_sponsor_changemaker_id
 		INNER JOIN users u
 			ON u.keycloak_user_id
 			= build_changemakers_results.auth_context_keycloak_user_id
+		LEFT JOIN permitted_changemakers pc ON pc.id = sponsor.id
```
slifty: Is it possible to just have this column be a calculated boolean right off the bat (e.g. "is_viewable")?  If so that could make things slightly more self documenting.

I don't know if this is possible!

src/database/initialization/5_build_proposals_results.sql
```diff
@@ -94,15 +94,34 @@ CREATE FUNCTION build_proposals_results(
 		FROM opportunities o
 		WHERE o.id IN (SELECT ip.opportunity_id FROM input_proposals ip)
 	),
+	-- bounded to the changemakers attached to this page's proposals
+	permitted_changemakers AS (
```
slifty: Let's give this the `viewable_changemaker_ids` treatment too while we're at it.

PR #2511, opened 2026-07-27
Add workflow to auto-add issues to project

slifty:
Adds the org caller workflow so new issues land on the project board automatically.

Uses the current org template — pinned to the `v1` tag and passing only `PDC_BOT_PRIVATE_KEY` rather than `secrets: inherit`. Supersedes the earlier stale branch, which pointed at `@main`.

Requires the pdc-bot app setup in PhilanthropyDataCommons/.github to be complete before runs will pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V72EH61bzDFnGMcnEfvnZY

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -1,0 +1,13 @@
+name: Add to project
+
+on:
+  issues:
+    types: [opened, transferred, reopened]
+
+permissions: {}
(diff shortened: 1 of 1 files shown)


bickelj (approved):
LGTM!

PR #2519, opened 2026-08-03
Add more dependabot groups

slifty:
This PR creates more dependabot groups in an effort to further balance overhead with safety.


bickelj (approved):
But does `jest` still separate itself?

slifty:
I believe it should, yes

Issue #2340, opened 2026-04-14
[Enhancement]: add new entity "Initiative" and connected relationships with other entities


jmergy:
### Summary

We'll want to have initiative and have it be able to handle baseFields and have it be able to be associated to changemakers, funders, data platforms, and proposals. Open to whatever else makes sense to enable on this for the initial provisioning. 

### Priority

Low

jmergy:
@slifty @bickelj would be great to get this in with the attributes for the source name on opportunities/proposals as best implemented.

jmergy:
@slifty @bickelj @kfogel This has come up again around the the Exchange migration work and in meetings this week with the PDC Ops around the need and how much that would be an improvement for incoming changemakers and funders.

jmergy:
@kfogel assuming this is something coming in the next few weeks...

slifty:
Documenting the shape of the entities based on the discussions we have had.

1. We're talking about two entities here: `Initiative` and `InitiativeFieldValues` which would both be new context entity types.
2. We wouldn't support bulk upload of initiatives at this stage (this could be augmented later).
3. We would not support `versions` of initiatives like we do with proposals.  Individual field values would have direct CRUD.
4. I believe we will want to incorporate initiative data into the changemaker gold data algorithm.

## Initiative

### Attributes

- id
- title
- changemaker ID
- createdAt
- createdBy

### Endpoints

- CRU `/initiatives`
- RU `/initiatives/:initiativeId`

## Initiative Field Value Attributes

- id
- baseFieldShortCode
- initiativeId
- value
- isValid
- goodAsOf
- sourceId
- createdAt
- createdBy

### Endpoints

- CRU `/initiatives/:initiativeId/fieldValues`
- RU `/initiatives/:initiativeId/fieldValues/:fieldValueId`

---

@jmergy above said that we would want to be able to associate an initiative with a funder or data provider but I would want to unpack that, I believe that only changemakers lead initiatives (we already have a system that allows a given org to act as both a changemaker AND a funder).

Proposal association does make sense (e.g. this proposal is associated with `Initiative Foo`), but would suggest we add that functionality once the initial initiative system is complete.

jmergy:
@slifty I think that only associating with changemakers, for now, is fine. We can see if we get pull from funders. I was just wanting to lock on anything restrictive from a design perspective.

jmergy:
@slifty I know we will need some additional baseFields on this and I will create them when it makes sense

[...]

Issue #2421, opened 2026-05-19
Endpoints should return 403 (or 404) instead 401 when credentials without proper authorization are provided


jayqi:
I recently saw an error

> Authenticated user does not have permission to perform this action.

and was surprised to see that it was an HTTP 401 status code instead of 403.

From [Mozilla's MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401):

> The HTTP 401 Unauthorized client error response status code indicates that a request was not successful because it lacks valid authentication credentials for the requested resource. This status code is sent with an HTTP WWW-Authenticate response header that contains information on the authentication scheme the server expects the client to include to make the request successfully.
> 
> A 401 Unauthorized is similar to the 403 Forbidden response, except that a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action.

My understanding here is that 403 is semantically the expected error code in this situation. Is the use of 401 intentional?

jayqi:
@slifty would be great to get a check on intentions here, because we need to know if Exchange should be checking for 403 or 401 errors.

slifty:
@jayqi good catch and apologies for the drift from HTTP spec; we'll resolve this.

I will also note that as we fix this, I believe we will find the intention to be as follows:

- 401: authentication required for this endpoint but not provided.
- 404: the item does not exist *OR* the user does not have the right to *view* the item.
- 403: the item does exist, the user has the right to view it, but the user does not have the right to perform the specified action.

^ Which is to say, plan to handle both 404 and 403 responses in the event of a lack of access.

slifty:
A related issue: #2138

PR #2532, opened 2026-08-04
Add new `Initiative` entity and endpoints

### LINKED ISSUE #2528: Add `Initiative` entity and endpoints
This issue captures the work for creating the new `Initiative` entity described in https://github.com/PhilanthropyDataCommons/service/issues/2340

It doesn't cover permissions or actual field values, it's just about the entity itself.

slifty:
This PR adds a new `Initiative` entity per the discussion in #2340.

Worth noting what is NOT in here:

- Permissions (this PR makes the endpoints admin only for now)
- InitiativeFieldValues

But we have issues to capture those.

Resolves #2528


src/types/Initiative.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const initiativePatchSchema: JSONSchemaType<InitiativePatch> = {
+	type: 'object',
+	properties: {
+		title: {
+			type: 'string',
+			minLength: 1,
+			/* eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion --
+			 * AJV's JSONSchemaType does not properly support nullable for patches.
+			 * See https://github.com/ajv-validator/ajv/issues/2163
```
slifty, line 42: This bugs me every time.

bickelj:
> lite review requested due to automatic review settings 

Eh?

src/__tests__/initiatives.int.test.ts
```diff
@@ -0,0 +1,342 @@
...
+} from '../test/mockJwt';
+
+describe('/initiatives', () => {
+	describe('GET /', () => {
+		it('requires authentication', async () => {
+			await request(app).get('/initiatives').expect(401);
+		});
+
+		it('requires administrator role', async () => {
+			await request(app).get('/initiatives').set(authHeader).expect(401);
```
bickelj, line 30: Wouldn't we expect 403 here? I forget.

src/__tests__/initiatives.int.test.ts
```diff
@@ -0,0 +1,342 @@
...
+				.get('/initiatives')
+				.set(authHeaderWithAdminRole)
+				.expect(200);
+			expect(response.body).toEqual({
+				entries: [initiativeA, initiativeB],
+				total: 2,
+			});
+		});
+
+		it('filters by changemaker when ?changemaker is provided', async () => {
```
bickelj: Surprised to see the ? here, clearer to say "when changemakerId is provided"

src/__tests__/initiatives.int.test.ts
```diff
@@ -0,0 +1,342 @@
...
+		});
+	});
+
+	describe('GET /:initiativeId', () => {
+		it('requires authentication', async () => {
+			await request(app).get('/initiatives/1').expect(401);
+		});
+
+		it('requires administrator role', async () => {
+			await request(app).get('/initiatives/1').set(authHeader).expect(401);
```
bickelj, line 107: Again, wouldn't this be a 403?

src/database/migrations/0118-create-initiatives.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE initiatives (
+	id int PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+	changemaker_id int NOT NULL REFERENCES changemakers (id) ON DELETE CASCADE,
+	title text NOT NULL,
```
bickelj, line 4: We're all in on postgres, I give up on my `varchar` crusade.

src/database/migrations/0118-create-initiatives.sql
```diff
@@ -0,0 +1,17 @@
...
+	created_at timestamp with time zone NOT NULL DEFAULT now(),
+	created_by uuid NOT NULL REFERENCES users (keycloak_user_id) ON DELETE CASCADE
+);
+
+COMMENT ON TABLE initiatives IS
+'Projects or programs led by a changemaker.';
+
+-- CONCURRENTLY cannot run inside the transaction that wraps each migration,
+-- and the table is empty here, so the build takes no meaningful lock.
+CREATE INDEX idx_initiatives_changemaker -- noqa: PG01
```
bickelj: Why the wonky comments here? Is there a lint rule that requires this?

src/openapi/components/schemas/Initiative.json
```diff
@@ -0,0 +1,32 @@
...
+			"readOnly": true
+		},
+		"createdBy": {
+			"description": "The keycloak user id of the PDC user that created this initiative",
+			"type": "string",
+			"format": "uuid",
+			"readOnly": true
+		}
+	},
+	"required": ["id", "changemakerId", "title", "createdAt", "createdBy"]
```
bickelj, line 31: Is the ID and createdAt really required?

src/openapi/components/schemas/InitiativePatch.json
```diff
@@ -0,0 +1,13 @@
...
+	"properties": {
+		"title": {
+			"type": "string",
+			"minLength": 1,
+			"example": "Clean Water for the Upper Valley"
+		}
+	},
+	"additionalProperties": false,
+	"minProperties": 1,
+	"required": []
```
bickelj, line 12: At the moment we may as well add the `title` here, right? Then at expansion time, remove it?

src/types/Initiative.ts
bickelj: Sigh.

bickelj (commented):
I left a few quibbles here and there. I also await GLM-5.2's review.

src/handlers/initiativesHandlers.ts
```diff
@@ -0,0 +1,121 @@
...
+		throw new InputValidationError('Invalid id parameter.', isId.errors ?? []);
+	}
+	if (!isInitiativePatch(req.body)) {
+		throw new InputValidationError(
+			'Invalid request body.',
+			isInitiativePatch.errors ?? [],
+		);
+	}
+
+	await loadInitiative(db, req, initiativeId);
```
bickelj-agent: Following `patchChangemaker`'s try/catch that converts `updateInitiative`'s `NoDataReturnedError` into a `NotFoundError` would drop this pre-flight `loadInitiative` query (and the small load→update race) for a missing row. — GLM-5.2

src/__tests__/initiatives.int.test.ts
slifty: You are correct!

src/__tests__/initiatives.int.test.ts
slifty: When we first made filters we didn't do `changemakerId` but instead `changemaker` -- this was done because we wanted to leave room for supporting URI based pointers (?changemaker = { the URI that represents the REST entity }).

I'll take a moment to see if we actually followed that pattern overall, and then we can also revisit if it is even the right choice at this point.  Stay tuned.

src/openapi/components/schemas/Initiative.json
slifty: Yeah -- this is an important signal since it means clients can trust that it will be there and don't have to cover the undefined case.  Since they are `readonly` that covers the way they aren't present in POST / PUT context.

src/__tests__/initiatives.int.test.ts
bickelj: Apparently, though, this is the established convention, says GLM-5.2

src/handlers/initiativesHandlers.ts
bickelj: This is a good call, agree.

src/database/migrations/0118-create-initiatives.sql
slifty: Yes -- updating to make that clearer

slifty: Actually: I think maybe this would *always* have to be disabled in migrations so instead I'll just update the global config to ignore the rule.

src/__tests__/initiatives.int.test.ts
slifty: Argh, must not have caught that during https://github.com/PhilanthropyDataCommons/service/issues/2421

It shouldn't be (that said, it's a middleware issue and an easy fix, I'll open a followup issue)

src/__tests__/initiatives.int.test.ts
slifty: Update: Fixing this test description for sure!

src/openapi/components/schemas/InitiativePatch.json
slifty: I'd prefer not to -- It isn't that "title" is required it's that *something* is required (which is already conveyed exactly as it should be via `minProperties`).  Adding it here would be semantically misleading IMHO

Unless I'm misunderstanding!

slifty:
@bickelj I think I addressed everything except the one thing I think we should leave (and the 401 vs 403 which is a bug but we should fix that consistently as a separate issue)

src/__tests__/initiatives.int.test.ts
bickelj: Yes, I agree that a follow-up makes more sense than fixing here.

src/openapi/components/schemas/InitiativePatch.json
bickelj: My comment had to do with evolution of this endpoint over time. For any given patch with N attributes, the above makes perfect sense. We're in a special case, though, where N=1, so perhaps the OpenAPI and generators would give better clues if we constrained explicitly to the 1 attribute at the moment, `title`. If/when more attributes are added to the initiative, we could relax it back to the more flexible definition.

bickelj: To be clear it's fine as-is.

bickelj-agent:
Thanks for the rebase — my earlier `patchInitiative` suggestion is addressed (pre-flight `loadInitiative` removed, `NoDataReturnedError` → `NotFoundError` try/catch matches `patchChangemaker`), and `tsc`/`eslint`/`openapi` lint all pass on the new head. I don't see any new important issues. — GLM-5.2

src/__tests__/initiatives.int.test.ts
bickelj: I created https://github.com/PhilanthropyDataCommons/service/issues/2547

bickelj (approved):
This looks good to me, thanks for the iteration!

slifty:
I have to say I love getting both the robot and the jesse LGTM 😂

PR #2535, opened 2026-08-04
Ignore major `typescript` version updates

slifty:
This PR updates dependabot to ignore major versions of `typescript` to prevent dependency conflicts.


hminsky2002 (approved):
Noting explicitly that the one of the dependencies we care a lot about is our ts-eslint, which is not compatible with ts7

bickelj:
I wish there were a "dependabot ignore this until ESLint and all its lackluster lackeys stop lagging"

slifty:
Gtihub Actions were being weird yesterday, I'll see if I can re-trigger them...

PR #2549, opened 2026-08-13
Add `InitiativeFieldValue` entity

### LINKED ISSUE #2530: Add `InitiativeFieldValue` entity and endpoints
In #2340 we describe a new `Initiative` entity which allows changemakers to define projects they are working on outside the context of a proposal.  That issue also describes a new `InitiativeFieldValue` entity which is where the actual project data is captured.

This issue is for creating the entity and its endpoints (without permission system integration)

slifty:
This PR adds a new `InitiativeFieldValue` entity type which allows base field values with an initiative in the system.

In this implementation I did decide to disallow organization base field data to be linked to an initiative -- this is better represented by a `ChangemakerFieldValue`.

Resolves #2530

### DIFF
--- a/src/database/initialization/1_assert_initiative_field_value_not_forbidden.sql
+++ b/src/database/initialization/1_assert_initiative_field_value_not_forbidden.sql
@@ -1,0 +1,24 @@
+SELECT drop_function('assert_initiative_field_value_not_forbidden');
+
+-- Raises if the field value belongs to a forbidden base field, so a forbidden
+-- value is never serialized. Read paths exclude forbidden fields themselves;
+-- this guards the paths that serialize a row directly (the field value insert
+-- and update), which do not filter by sensitivity.
+CREATE FUNCTION assert_initiative_field_value_not_forbidden(
+	initiative_field_value initiative_field_values
+) RETURNS void AS $$
+BEGIN
+	IF EXISTS (
+		SELECT 1
+		FROM base_fields
+		WHERE base_fields.short_code
+			= initiative_field_value.base_field_short_code
+			AND base_fields.sensitivity_classification = 'forbidden'
+	) THEN
+		RAISE EXCEPTION
+			'Refusing to serialize forbidden initiative_field_value (%)',
+			initiative_field_value.id
+			USING ERRCODE = '22023'; -- invalid_parameter_value
+	END IF;
+END;
+$$ LANGUAGE plpgsql STABLE;
--- a/.sqlfluff
+++ b/.sqlfluff
@@ -71,3 +71,5 @@ terminologySetId = 1
 isCollaborative = false
 granteeType = 'user'
 verb = 'view'
+initiativeId = 1
+initiativeFieldValueId = 1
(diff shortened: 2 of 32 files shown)


src/database/initialization/1_assert_initiative_field_value_not_forbidden.sql
```diff
@@ -0,0 +1,20 @@
+SELECT drop_function('assert_initiative_field_value_not_forbidden');
+
+CREATE FUNCTION assert_initiative_field_value_not_forbidden(
```
slifty, line 7: This is the pattern we use (a dedicated function based on the field value rather than the base field).

I suspect it might make more sense to have a single `assert_base_field_not_forbidden` since we're now repeating this logic for every field value type.  I don't think that change is in scope of this PR though.

bickelj-agent (commented):
Reviewed the `InitiativeFieldValue` entity end-to-end. It closely mirrors the established `changemaker`/`proposal` field-value patterns; defense-in-depth is consistent (handler `assertBaseFieldIsUsableByInitiatives` + DB triggers + `assert_initiative_field_value_not_forbidden` inside `build_initiative_field_value_result`), read paths filter forbidden base fields, cross-initiative access 404s via the `initiative_id` predicate, and PATCH recomputes `isValid` against the base field data type.

Static validation on this branch is clean: `tsc --noEmit`, ESLint, `redocly lint`/`bundle`, and `sqlfluff` all pass, migration `0119` is the correct next number, and all wiring is present (type/decorator/operations/queries/initialization exports, `api.json` schema+path registration, router mount, factory, CHANGELOG, ERD, `.sqlfluff` params). I could not run the integration suite locally (no matching DB credentials in this environment), but `initiativeFieldValues.int.test.ts` coverage looks thorough.

<details><summary>Minor nits (not blocking, mostly inherited from the changemaker pattern)</summary>

- `build_initiative_field_value_result` declares `auth_context_keycloak_user_id`/`auth_context_is_administrator` params and forwards them to `source_to_json`, but every call site (`insertOne`, `selectById`, `selectWithPagination`, `updateById`) invokes the function with a single argument, so the defaults (NULL/FALSE) always apply. `source_to_json` accepts those params but never uses them in its body, so the serialized `source` is identical regardless. No behavior bug, but a reader could mistakenly assume auth context is threaded into source serialization here. This mirrors `build_changemaker_field_value_result`, so it's an inherited convention.
- `selectWithPagination.sql` orders `BY id` ascending, whereas the changemaker analog orders `BY id DESC`. Ascending is reasonable for field values; noting only for awareness.
- The two `BEFORE INSERT OR UPDATE` triggers re-check `NEW.base_field_short_code`, but `updateById.sql` never changes that column, so the UPDATE branch is effectively unreachable. Harmless.

</details>

bickelj (approved):
A brief skim reveals no red flags. See [GLM-5.2's above nitpicks](https://github.com/PhilanthropyDataCommons/service/pull/2549#pullrequestreview-4962524159) if you wish.

slifty:
I took out the auth params from `source_to_json` <-- this actually is something we should update in other places too, since the `to_json` is increasingly being framed as JUST for serialization / doesn't need to know auth.

Rebased too.

Merging once CI passes!

PR #2551, opened 2026-08-18
Add Initiative Permissions

### LINKED ISSUE #2529: Add `Initiative` permissions
In #2528 we create a new entity type of `Initiative` -- we need to add permissions for this entity type.

Permissions for initiatives can be granted directly, but they can also be inherited from `Changemaker`.

slifty:
This PR adds `Initiative` entity support to our permission grant system.

Resolves #2529


src/handlers/initiativesHandlers.ts
```diff
@@ -102,32 +125,28 @@ const patchInitiative = async (req: Request, res: Response): Promise<void> => {
 		);
 	}
 
-	try {
-		const updatedInitiative = await updateInitiative(
-			db,
-			req,
-			req.body,
+	await loadInitiative(db, req, initiativeId);
```
bickelj-agent, line 130: An `edit | initiative` grant without `view`/`manage` returns 404 here, because `loadInitiative` runs `selectById` gated on the `view` scope before the edit check. `verb_set_permits_verb` lets only `manage` satisfy any verb, so an `edit`-only initiative grant can never reach the update path. Tests always pair VIEW+EDIT so this isn't surfaced. Worth either documenting that `edit` requires `view`, or checking edit-permission before the view-gated load.

_Reviewed by GLM-5.2_

bickelj-agent (commented):
Adds `initiative` to the permission grant system, modeled on the terminologySet precedent: split enum+column migrations, `permitted_initiative_ids` with changemaker inheritance, and a self-manage grant on creation. tsc, unit, SQLFluff, OpenAPI, ESLint, and Prettier all clean on the delta.

One inline note: a `edit`-only initiative grant can't reach PATCH because the view-gated load runs first.

bickelj (approved):
I see nothing objectionable on a very brief skim. Please take copilot and GLM-5.2's feedback into consideration.

I should also say that in a local GLM-5.2 session, GLM verified that the code here matches the permissions documentation for initiatives.

Issue #2554, opened 2026-08-18
Permissions `edit | changemaker` doc/code mismatch


bickelj-agent:
## Summary

`docs/PERMISSIONS.md` documents `edit | changemaker` as enabling "Create or update changemaker field values," but the code requires scope `changemakerFieldValue`, not `changemaker`. A grant of `edit | changemaker` (or `manage | changemaker`) does **not** allow creating changemaker field values. I verified this with a live HTTP probe against a real Keycloak-issued JWT using the `compose-ci.yml` stack.

## Steps to reproduce (live)

1. Start the stack: `sudo docker compose -f compose-ci.yml up --remove-orphans` (add `127.0.0.1 pdc-api`, `pdc-auth`, `pdc-s3` to `/etc/hosts`).
2. Get an admin token (service account `pdc-dev-ingest` / `password`) and a non-admin token (`user@pdc.local` / `password` via the `pdc-openapi-docs` client).
3. As admin: create a changemaker, a source on it, and a changemaker-field-value batch.
4. As admin: grant the non-admin user `edit | changemaker` (scope `["changemaker"]`, verbs `["edit"]`) on that changemaker:
   ```json
   POST /permissionGrants
   { "granteeType": "user", "granteeUserKeycloakUserId": "<user sub>",
     "contextEntityType": "changemaker", "changemakerId": 1,
     "scope": ["changemaker"], "verbs": ["edit"] }
   ```
5. As the non-admin user: `POST /changemakerFieldValues`.

## Expected (per docs/PERMISSIONS.md, Changemaker Permissions)

`201 Created` — the doc row `edit | changemaker` says it enables "Create or update changemaker field values."

## Actual

`403 Forbidden`:
```json
{ "name": "ForbiddenError",
  "message": "Authenticated user does not have permission to create field values for the specified changemaker." }
```

## Root cause

`src/handlers/changemakerFieldValuesHandlers.ts` checks:
```ts
hasChangemakerPermission(db, req, { changemakerId, permission: 'create', scope: 'changemakerFieldValue' })
```
`scope_set_permits_scope(['changemaker'], 'changemakerFieldValue')` is false (`scope = ANY(scopes) OR 'any' = ANY(scopes)`). `manage` does not help because it expands verbs, not scopes. So a grant whose scope is `changemaker` cannot satisfy a `changemakerFieldValue` scope check.

[...]

PR #2556, opened 2026-08-19
Apply permissions to `InitiativeFieldValue`

### LINKED ISSUE #2531: Add `InititativeFieldValue` permissions
We need to incorporate `InitiativeFieldValue` into the permissions system once the entity has been created.

slifty:
This PR adds the new `InitiativeFieldValue` to the permission system.  Since this is a field value permission we also mirror the `ProposalFieldValue` conditional permissions / base field category conditions.

Resolves #2531


src/handlers/initiativeFieldValuesHandlers.ts
```diff
@@ -136,22 +173,40 @@ const postInitiativeFieldValue = async (
...
-		value,
-		isValid: fieldValueIsValid(value, baseField.dataType),
-		goodAsOf,
+	if (
+		!(await canCreateInitiativeFieldValue(db, req, {
+			initiativeId,
+			baseFieldCategory: baseField.category,
+		}))
+	) {
+		throw new UnauthorizedError();
```
bickelj-agent: This throws `UnauthorizedError` (401), but the request is authenticated — the caller simply lacks the `create | initiativeFieldValue` grant. The neighboring `changemakerFieldValuesHandlers` throws `ForbiddenError` (403) for the same situation, and 403 is the correct status for an authenticated-but-unauthorized user. The same applies to the `patchInitiativeFieldValue` handler (line 245). Consider `ForbiddenError` here for consistency and correct semantics. GLM-5.2

bickelj-agent (commented):
The InitiativeFieldValue permission wiring is solid: the `permitted_initiative_field_value_ids[_among]` functions, the `canCreateInitiativeFieldValue`/`hasInitiativeFieldValuePermission` operations, the scope-map additions (`INITIATIVE` and `CHANGEMAKER` native scopes now include `INITIATIVE_FIELD_VALUE`), the `baseFieldCategory` conditions, and the read-query filtering all match the `ProposalFieldValue` pattern. Two issues stand out: the create/edit handlers throw `UnauthorizedError` (401) for authenticated-but-unauthorized callers where `ForbiddenError` (403) is correct, and the Changemaker docs block still carries the inaccurate `edit | changemaker` → "Create or update changemaker field values" row (issue #2554) right next to the new, correct `initiativeFieldValue` rows.

<details><summary>Other notes (not blocking)</summary>

- The shared `assertSourceIsReferenceable` helper is a clean DRY; the load-before-throw ordering preserves the 404-vs-403 distinction for invisible sources.
- `permitted_initiative_field_value_ids` (non-`_among`) omits the `NOT user_is_admin` guard on its grant branch, consistent with `permitted_proposal_field_value_ids`; the admin branch already returns all non-forbidden values so the grant branch is redundant-but-harmless for admins.
- The PR is `mergeable_state: dirty` and needs a rebase onto current `main` (the base has moved since the PR was opened).
- GLM-5.2

</details>

docs/PERMISSIONS.md
```diff
@@ -279,19 +279,22 @@ context key).
...
-| view   | initiative         | View the changemaker's initiatives                                 |
-| edit   | initiative         | Create or update initiatives owned by the changemaker              |
+| Verb   | Scope                | What It Enables                                                    |
+| ------ | -------------------- | ------------------------------------------------------------------ |
+| view   | changemaker          | View changemaker field values for the changemaker                  |
+| view   | proposal             | View proposals associated with the changemaker                     |
+|        |                      | View proposal versions associated with the changemaker             |
+|        |                      | View changemaker-proposal associations for the changemaker         |
+| view   | proposalFieldValue   | View proposal field values for the changemaker's proposals         |
+| edit   | changemaker          | Create or update changemaker field values                          |
```
bickelj-agent: This `edit | changemaker` → "Create or update changemaker field values" row is the inaccuracy tracked in #2554: creating changemaker field values requires scope `changemakerFieldValue` (the handler checks `create | changemakerFieldValue`), not `changemaker`. This PR adds the correct `view|create|edit | initiativeFieldValue` rows immediately below, which highlights that the sibling `changemaker`-scope field-value rows are still wrong. Worth either correcting these `changemaker` rows in kind or noting they are out of scope for #2531. GLM-5.2

slifty:
@bickelj this is ready for re-review I think!

bickelj-agent (commented):
Re-review after the rebase: both prior comments are resolved — the create/edit handlers now throw `ForbiddenError` (403) instead of `UnauthorizedError` (401), and the Changemaker docs block now uses the `changemakerFieldValue` scope for field-value view/create with the inaccurate `edit | changemaker` row removed. The `initiativeFieldValue` rows this PR adds (changemaker and initiative blocks, plus the new InitiativeFieldValue section and the `baseFieldCategory` conditions) are consistent with the `ProposalFieldValue` pattern and the underlying `permitted_initiative_field_value_ids[_among]` / `canCreateInitiativeFieldValue` SQL. No new in-line comments; the remaining `view | changemaker` "non-public attributes" wording is from the base (#2558), not this PR's diff. GLM-5.2

bickelj (approved):
I tried adding this permission in the https://utilities.philanthropydatacommons.org/pdc-service-preview-2556-6lpqu environment using an admin account:

```json
{
	"granteeType": "user",
	"granteeUserKeycloakUserId": "[my unprivileged UUID]",
	"contextEntityType": "changemaker",
	"changemakerId": 4,
	"scope": ["initiative", "initiativeFieldValue"],
	"verbs": ["view","create","edit"]
}
```

That returned a 201.

Then I logged out, logged in as the unprivileged user, and was able to create an initiative using this:

```json
{
  "changemakerId": 4,
  "title": "Dan is the man"
}
```

That returned a 201.

I tried posting an initiative field value for a non-existent initiative and got the expected 404.

I then posted an initiative field value using the example text from OpenAPI docs, and got a 404 on the base field referenced. I suppose that's a 404 but I thought we did 422 in that situation (or used to).

I then tried posting initiative field values but lacked the permission to create a source.

In all, I find nothing objectionable having done the above after skimming the source code and looking at GLM-5.2's follow-up review.

Very nice!

PR #2562, opened 2026-08-20
Remove `PlatformProviderResponse` entity

### LINKED ISSUE #2561: Remove `/platformProviderResponses`
This is a very old endpoint that existed before we had the concept of `ChangemakerFieldValues`.

We should remove it and its related code, since it's dead.

slifty:
This PR removes the concept of `PlatformProviderResponse` and its related API endpoint + documentation.  This entity has been replaced by `ChangemakerFieldValue`.

Resolves #2561


slifty:
I received confirmation that Exchange is not using this endpoint

bickelj-agent (commented):
Reviewed locally on top of main (commit dab64685). The removal is clean: handler, router, type, queries, OpenAPI schema/paths, and the integration test are all removed, and the new migration 0122 correctly calls `deaudit_table` before `DROP TABLE`; no source, docs, or ERD references remain (historical migrations 0055/0012 are appropriately left untouched).

— GLM-5.2

CHANGELOG.md
bickelj, line 8: Even if nobody's using it, I think we should bump the version to 1.0.0 or 0.42.0, because it is a removal of an endpoint. That will kick the new SDK build off which should help find any stragglers (I expect it will echo the find from the data-scripts repo).

bickelj (approved):
Let's do it, but bump the version here while doing so.

Thanks for getting rid of the cruft!

PR #2571, opened 2026-09-01
Pin `eslint-config-love`

slifty:
This PR pins eslint-config-love to a version that continues support for the import plugin.


bickelj-agent (commented):
Lockfile already resolves to 154.1.0, so this only tightens the spec to an exact pin and adds the matching semver-major ignore entry, consistent with the adjacent eslint entry. No high-confidence or high-impact issues found.

— GLM-5.2

bickelj (approved):
I don't know enough about ESLint to do a great review, but I do know I love exact version pins!

Issue #2455, opened 2026-06-16
[Enhancement]: Default permissions sets / permissions configurations.


kfogel:
### Summary

We need some mechanism for remembering and providing default permissions sets for different kinds of entities -- that is, defaults based on the grantee of a permissions grant (perhaps based on the type of entity the potential grantee is), and defaults based on the target object of the permission (looking at its owner and/or at its type).  How exactly these two orthogonal potential default sets interact is still to be determined.

TL;DR from @jmergy: "Permissioning grants are really assigned to records.  They're not assigned to records that don't exist yet.  [This issue is about both.] ... [Default permissions and permission inheritance.]"

One thing to keep in mind is that we want to be thinking in terms of organizations and groups, not in terms of individual users, as much as possible.  See [this comment](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/OpsTeamUpdates/near/262805) from @jmergy on 2026-06-16:

> Org and user dynamic: We know we have this in the platform and app, but everyone is aligned and understands functionality, for now, is really focus on organizations. Users are agents of orgs so we are going to try and keep it at org and group levels. This is more relevant for Exchange but also probably how we'll want to roll with permissions if we can. allAuths or Group permissionGrants as more the default with really staying away from user perms if we can avoid it.

Some specific questions:

* When a new Changemaker/Funder/DPP/whoever joins the PDC, what can they see and do by default?  What information about them can be seen (and/or updated) by others?
* What permissions should PDC Staff have by default, to facilitate onboarding?  (See [Zulip discussion](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/permissions/near/262844).)
* What defaults should a Changemaker have regarding being affected by data platform provider updates that contain information about that Changemaker?  (See [#2416](https://github.com/PhilanthropyDataCommons/service/issues/2416) for more context.)

There's also the question of how such defaults would be implemented.  Does the PDC Framework "Just Do It" on entity creation?  Or do we provide some kind of convenience wrapper (presumably via an API endpoint) that allows the caller to invoke the defaults, but it's still the caller's responsibility to do so?

This needs more design & thinking.  This description is just a starting point.

jmergy:
@kfogel @slifty 

I promised some user stories on default permissioning, I hope this is a good starting point. 

As a PDC admin, I want to be able to collect a set of permissionGrants that can be applied with variables so that I can more easily accomodate repeatable provisioning of access for organization and users. 

[...]

PR #2577, opened 2026-09-08
Add `DefaultPermissionGrant` entity and endpoints

slifty:
This PR adds a new `DefaultPermissionGrant` entity which will eventually be used to determine new permissions that get automatically added when a given entity is created.

Related to #2455


src/openapi/components/schemas/DefaultPermissionGrant.json
```diff
@@ -0,0 +1,17 @@
...
+	"allOf": [
+		{ "$ref": "./PermissionGrantBase.json" },
+		{
+			"type": "object",
+			"properties": {
+				"contextEntityType": {
+					"$ref": "./PermissionGrantEntityType.json",
+					"description": "The type of entity a grant made from this default applies to. `any` is not supported.",
+					"example": "changemaker"
+				}
```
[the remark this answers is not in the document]
slifty: I thiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiink this is just { one of those things } that we had to accept because the permission type is so complex.

src/database/migrations/0125-create-default_permission_grants.sql
```diff
@@ -0,0 +1,39 @@
+CREATE TABLE default_permission_grants (
+	id int PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+	grantee_type permission_grant_grantee_type_t NOT NULL,
+	grantee_user_keycloak_user_id
+	uuid REFERENCES users (keycloak_user_id) ON DELETE CASCADE,
```
bickelj-agent, line 5: This re-adds a `grantee_user_keycloak_user_id` FK that migration 0094 deliberately dropped from `permission_grants` to allow granting to Keycloak users before they exist in the local `users` table (commit cd38536) — the same pre-provisioning goal these defaults serve. Since `permission_grants` (the row a default expands into) has no such FK, a template is stricter than its result: creating a default for a not-yet-provisioned user fails with an FK violation (500) instead of being allowed. Is the constraint intentional, or should this mirror `permission_grants`?

— GLM-5.2

bickelj-agent (commented):
The end-to-end scaffolding (migration, operations, handlers, router, OpenAPI, tests, docs) follows the existing `permission_grants` patterns, and the shared-validation refactor in `PermissionGrant.ts` / `assertions.ts` is behavior-preserving. One concern inlined below: the new table re-adds a `grantee_user_keycloak_user_id` FK that migration 0094 deliberately removed from `permission_grants`.

— GLM-5.2

bickelj:
@slifty If I understand correctly, the justification for **not** having these default permissions apply directly/immediately (thus allowing the default to be revoked for everybody they apply to at any time via these endpoints) is to allow for fine-grained revocation?

In other words, it would be cool if these default permissions were live and directly effective. That way I could add a default permission and if I later found I needed to change the default, I wouldn't have to go through the N users or entities that they applied to, I could simply edit or remove the default. But the big downside of what I'm suggesting is that we could no longer decide to selectively remove/revoke the default from only one user or one row of an entity, thus non-fine-grained.

Am I understanding correctly?

I think it would be an acceptable compromise position to say that "if you want fine-grained control, you need to remove your defaults and do the fine-grained thing; if you want broad defaults, you can use defaults." Therefore I think these should not be something that a trigger reads and generates fine-grained permissions from, I think these should literally be fallback values after processing fine-grained permissions. And if they grant too much, the defaults should be revoked, you shouldn't have to go back in and look at every entity to revoke all those that happened in the meantime.

slifty:
@bickelj I think you're describing an expansion of the `universal permission grants` -- right now we can give a permission to all users in a single grant, we can give a permission to all scopes in a single grant, but we cannot currently give a permission to all entities of a given context (or all contexts) in a single grant.  I see the value for allowing that and support doing it.

That said, the use case for default permissions is different, and I think you honed in on why.  For example, we know we need to be able to (A) allow a changemaker to manage access to their data and be in the driver's seat, but (B) we also want to grant certain permissions by default when those changemakers are created.

Specifically: we know *today* that we want an admin to be able to give a data provider write access on all future changemakers while still allowing those changemakers to revoke that access.  I don't think your proposed solution meets that need.

---

One could say that should be a client-based feature -- e.g. this should be a feature of the admin UX -- but the downside of that is the defaults are rooted in data (basically, *some* system will need to be able to codify the grantee IDs that we want the default permission to be granted to.)

bickelj:
@slifty and I talked through this a bit, and the ability to have an auto-generated yet revocable permission grant makes the most sense. Whether that's a "default" is a naming concern. So I'm OK with this, basically.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
```diff
@@ -273,6 +285,7 @@ erDiagram
   PermissionGrant }o--o| Initiative : "references"
   PermissionGrant }o--o| InitiativeFieldValue : "references"
   PermissionGrant }o--|| User : "is created by"
+  DefaultPermissionGrant }o--|| User : "is created by"
```
bickelj, line 288: It would be nifty (not required) to have a "DefaultPermissionGrant generates PermissionGrant" somewhere in here or 2579.

docs/PERMISSIONS.md
```diff
@@ -253,6 +253,19 @@ manage. `PUT /permissionGrants/:id` additionally requires `manage` on the
...
+entities of a given type. It carries everything a permission grant carries
+except the key naming a specific context entity, so a default permission grant
+plus a newly created entity of its `contextEntityType` describes a permission
+grant.
+
+CRUD operations on default permission grants, at `/defaultPermissionGrants`,
+require the `pdc-admin` role. Default permission grants sit outside the
+permission system rather than within it: holding `manage` on an entity does
+not confer any access to them.
+
```
bickelj, line 268: I would like a bit more text to clearly show the effects of the `defaultPermissionGrant` through passage of time in a changing world. Here the changing world is the changes to other entities.

So something like step 1, a funder is created. Step 2, a default permission grant is created. Step 3, another funder is created. Step 4, the default permission grant is removed. What permissions remain and why? That kind of thing.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added a `defaultPermissionGrant` entity, managed via the `/defaultPermissionGrants` endpoints, that describes a permission grant to make against entities of a given type.
```
bickelj, line 12: Perhaps `auto-generate` in place of `make`?

src/routers/defaultPermissionGrantsRouter.ts
```diff
@@ -0,0 +1,42 @@
...
+	'/',
+	requireAuthentication,
+	requireAdministratorRole,
+	defaultPermissionGrantsHandlers.postDefaultPermissionGrant,
+);
+
+defaultPermissionGrantsRouter.get(
+	'/:defaultPermissionGrantId',
+	requireAuthentication,
+	requireAdministratorRole,
```
bickelj, line 24: I was trying to imagine a provenance of permissions where I could tie a generated permission back to this permission and in a meeting chatted briefly with @slifty about this. We can imagine other scenarios too, such as "what permissions would this action create and/or block those and not these permissions." But we agreed these are all potential future things to consider, not here. In other words, we could see a potential that non-administrators could view these defaults, but there's no need to start with that today.

bickelj (commented):
I see no bad things here.

I tried it in the preview environment and this (generated-by-openapi) post got rejected, neither here nor there if intended:

```json
{
  "granteeType": "user",
  "granteeUserKeycloakUserId": "550e8400-e29b-41d4-a716-446655440000",
  "scope": [
    "funder"
  ],
  "verbs": [
    "view"
  ],
  "conditions": {
    "proposalFieldValue": {
      "property": "baseFieldCategory",
      "operator": "in",
      "value": [
        "project"
      ]
    },
    "initiativeFieldValue": {
      "property": "baseFieldCategory",
      "operator": "in",
      "value": [
        "project"
      ]
    }
  },
  "contextEntityType": "funder"
}'
```

Error:

```json
{
  "message": "Condition keys must be present in the grant scope. Invalid keys: proposalFieldValue, initiativeFieldValue",
  "details": []
}
```

Once the FK is relaxed this is good to go. I left some musings while skimming the code.

bickelj (approved):
Approving - the above stuff is for the author's consideration.

docs/ENTITY_RELATIONSHIP_DIAGRAM.md
slifty: I hear that, but from an ERD perspective that would be inaccurate because there is no FK relationship between permission grant and default permission grant.

We don't use an ERD to talk about the business-logic relationships between entities (e.g. "users access proposals" or etc).

I do think your instinct here speaks to potential value in a nullable field to convey that a grant came from a default rule, but since that isn't happening today I won't update this doc.

docs/PERMISSIONS.md
slifty: I'll add this in the followup PR (in this PR that isn't actually accurate yet)

PR #2579, opened 2026-09-09
Use default permissions

slifty:
This PR actually applies default permissions when a permissionable entity is created.

Resolves #2455

### DIFF
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -261,6 +261,13 @@ except the key naming a specific context entity, so a default permission grant
 plus a newly created entity of its `contextEntityType` describes a permission
 grant.
 
+Whenever an entity is created, every default permission grant registered for
+that entity's type is turned into a permission grant against the new entity,
+alongside the `manage` grant its creator always receives. The grants are made
+in the same transaction as the entity, so an entity either exists with its
+default grants or does not exist at all. Editing or deleting a default
+permission grant leaves the grants it has already produced untouched.
+
 CRUD operations on default permission grants, at `/defaultPermissionGrants`,
 require the `pdc-admin` role. Default permission grants sit outside the
 permission system rather than within it: holding `manage` on an entity does
--- a/docs/PERMISSIONS.md
+++ b/docs/PERMISSIONS.md
@@ -253,6 +253,19 @@ manage. `PUT /permissionGrants/:id` additionally requires `manage` on the
 proposed context entity when the caller re-points a grant at a different
 entity.
 
+### Default Permission Grants
+
+A `defaultPermissionGrant` describes a permission grant to make against
+entities of a given type. It carries everything a permission grant carries
+except the key naming a specific context entity, so a default permission grant
+plus a newly created entity of its `contextEntityType` describes a permission
+grant.
+
+CRUD operations on default permission grants, at `/defaultPermissionGrants`,
+require the `pdc-admin` role. Default permission grants sit outside the
+permission system rather than within it: holding `manage` on an entity does
+not confer any access to them.
+
 ## Implemented Permissions
 
 The following sections describe permissions that are actually enforced by the
--- a/src/__tests__/applyDefaultPermissionGrants.int.test.ts
+++ b/src/__tests__/applyDefaultPermissionGrants.int.test.ts
@@ -1,0 +1,230 @@
+import request from 'supertest';
+import { app } from '../app';
+import {
+	getDatabase,
+	loadPermissionGrantBundle,
+	loadSystemUser,
+} from '../database';
+import {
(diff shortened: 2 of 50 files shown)


docs/PERMISSIONS.md
```diff
@@ -261,6 +261,13 @@ except the key naming a specific context entity, so a default permission grant
 plus a newly created entity of its `contextEntityType` describes a permission
 grant.
 
+Whenever an entity is created, every default permission grant registered for
+that entity's type is turned into a permission grant against the new entity,
```
bickelj, line 265: I would say "generates a new matching permission grant" rather than "is turned into a permission grant". The default remains and it's not clear that anything is generated.

src/__tests__/applyDefaultPermissionGrants.int.test.ts
```diff
@@ -0,0 +1,230 @@
...
+				taxId: '11-1111111',
+				name: 'Default Grant Co.',
+				keycloakOrganizationId: null,
+			})
+			.expect(201);
+
+		const grants = await loadAllPermissionGrants();
+		const testUser = await loadTestUser(db);
+		expect(grants.entries).toEqual([
+			expectObjectContaining({
```
bickelj, line 165: Could be stricter.

src/__tests__/applyDefaultPermissionGrants.int.test.ts
```diff
@@ -0,0 +1,230 @@
...
+			.put('/funders/default_grant_funder')
+			.type('application/json')
+			.set(adminUserAuthHeader)
+			.send({ name: 'Default Grant Funder', isCollaborative: false })
+			.expect(201);
+
+		const grants = await loadAllPermissionGrants();
+		expect(grants.entries).toEqual(
+			expectArrayContaining([
+				expectObjectContaining({
```
bickelj, line 195: Could be stricter.

bickelj-agent (commented):
Review from GLM-5.2. The apply path is sound: `DefaultPermissionGrant` is a key-less template and `createPermissionGrant` whitelists only writable fields, so the spread cannot leak `id`/`createdBy` (the latter comes from the acting user); insert-only gating is correct for the upsert paths (funder/dataProvider PUTs and `loadOrCreateChangemaker` gate on `wasInserted`, validated by the 'does not apply on PUT update' test), and every entity-creation path that makes a self-manage grant now also applies defaults within the same transaction. One consideration: because `applyDefaultPermissionGrants` propagates errors inside the entity's transaction, a single default that fails at apply time would roll back creation of every entity of that type, so the default-creation assertions are the sole guard against that blast radius.

bickelj (approved):
As an admin, using the preview environment at https://utilities.philanthropydatacommons.org/pdc-service-preview-2579-rmh4m/ I was able to create a defaultPermissionGrant using this body:

```json
{
	"granteeType": "user",
	"granteeUserKeycloakUserId": "my uuid is super private y'all",
	"contextEntityType": "funder",
	"scope": ["proposalFieldValue"],
	"verbs": ["manage"]
}
```

Then when I created a new funder named `neufunder` and then got the permission grants filtering by `neufunder`, I saw two permission grants as expected, the first being the usual "you made a thing so you get permissions on the thing" and the second being the above defaultPermissionGrant:

```json

  "entries": [
    {
      "id": 97,
      "scope": [
        "proposalFieldValue"
      ],
      "verbs": [
        "manage"
      ],
      "createdAt": "2026-09-14T17:33:52.508765+00:00",
      "createdBy": "redacted",
      "conditions": null,
      "granteeType": "user",
      "createdByUser": {
        "createdAt": "2025-09-25T14:23:21.794033+00:00",
        "keycloakUserId": "redacted",
        "keycloakUserName": "Jesse Bickel Admin"
      },
      "funderShortCode": "neufunder",
      "contextEntityType": "funder",
      "granteeUserKeycloakUserId": "redacted",
      "granteeKeycloakOrganizationId": null
    },
    {
      "id": 96,
      "scope": [
        "any"
      ],
      "verbs": [
        "manage"
      ],
      "createdAt": "2026-09-14T17:33:52.508765+00:00",
      "createdBy": "f0a9e005-fd50-4ab7-8148-d8f46c1f3f1d",
      "conditions": null,
      "granteeType": "user",
      "createdByUser": {
        "createdAt": "2025-09-25T14:23:21.794033+00:00",
        "keycloakUserId": "redacted",
        "keycloakUserName": "Jesse Bickel Admin"
      },
      "funderShortCode": "neufunder",
      "contextEntityType": "funder",
      "granteeUserKeycloakUserId": "redacted",
      "granteeKeycloakOrganizationId": null
    }
  ],
  "total": 2
}
```

It's looking pretty good here.

